PR #35, opened 2018-02-16
Firmware uploads to S3

mattlorey:
This isn't working yet, can't get the signed urls to be valid :/   If anyone wants to take a look, I have two version in `Beamware.Firmware.Upload.S3`

I'll take another look later after I clear my head


fhunleth:
Superseded by #86

PR #70, opened 2018-05-30
Document client server communcation

### LINKED ISSUE #64: Document device->nerves_hub channel API
Minimum requirements for the device to report to Nerves Hub:
* Firmware version - so that the server knows if new firmware needs to be sent down and can record that an upgrade completed.

Minimum needed to send back down to the device:
* Whether the device should update its firmware
* A URL to the new firmware

Nice to have notifications from the device:
* An error message if the firmware failed to download or apply - useful for debug

Other items to consider:
* If devices report more of their metadata (see https://github.com/nerves-project/nerves_runtime#nerves-system-and-firmware-metadata), the server can detect misconfigurations
* We've had requests for lightweight device management features like remote reboot, retrieving recent logs (like the most recent 16KB or so of log messages from RingLogger), and setting key/value pairs to modify the device configuration.
* Would be good to know when devices have not connected for some amount of time - this may not require anything from the protocol - just that the system tracks when a device is connected.

mjs2600:
Why
---
It's important to have a well thought out and consistent communication
procedure.

How
---
A `/docs/` directory was added and the documentation was put in there as a
markdown file.

Closes #64

### DIFF
--- a/docs/client-server.md
+++ b/docs/client-server.md
@@ -1,0 +1,53 @@
+# Client Server Communication
+
+This document outlines the basic communication protocol between a Nerves device
+and NervesHub. At a high level, the device connects to NervesHub over a
+websocket connection, authenticates with a certificate, and is able to receive
+commands and send updates about its current state.
+
+## Joining the Device Channel
+
+After booting, the client connects to NervesHub with its certificate. The
+certificate is used to determine the unique identity of the device, which is
+then stored on the socket. The device joins `device:#{device_identifier}` with a
+payload of `%{"version" => version}`. The server first verifies that the device
+is joining its own channel, by comparing the id in the channel name to the id on
+the certificate. Next, the server compares the running version of the firmware
+to the target version for that device on `NervesHub`. If those versions are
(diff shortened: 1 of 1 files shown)


docs/client-server.md
```diff
@@ -0,0 +1,51 @@
...
+
+This document outlines the basic communication protocol between a Nerves device
+and NervesHub. At a high level, the device connects to NervesHub over a
+websocket connection, authenticates with a token, and is able to receive
+commands and send updates about its current state.
+
+## Joining the Device Channel
+
+After booting, the client connects to
+`device:#{tenant}:#{product}:#{device_identifier}` with the token, specific to
```
fhunleth: We currently don't store `tenant` on the devices. We have `author` which might be close. (Here's the [list](https://github.com/nerves-project/nerves_runtime#nerves-system-and-firmware-metadata)). The UUID embedded in the client-side cert can be used to look up the tenant on the server, though. The device will know its UUID so it could provide it, but it's probably better for the server to use the one supplied in the cert.

PR #74, opened 2018-05-30
Document client server communcation

### LINKED ISSUE #64: Document device->nerves_hub channel API

mjs2600:
Why
---
It's important to have a well thought out and consistent communication
procedure.

How
---
A `/docs/` directory was added and the documentation was put in there as a
markdown file.

Closes #64

### DIFF
--- a/docs/client-server.md
+++ b/docs/client-server.md
@@ -1,0 +1,58 @@
+# Client Server Communication
+
+This document outlines the basic communication protocol between a Nerves device
+and NervesHub. At a high level, the device connects to NervesHub over a
+websocket connection, authenticates with a certificate, and is able to receive
+commands and send updates about its current state.
+
+## Joining the Device Channel
+
+After booting, the client connects to NervesHub with its certificate. The
+certificate is used to determine the unique identity of the device, which is
+then stored on the socket. The device joins `device:#{device_identifier}` with a
+payload of `%{"version" => version}`. The server first verifies that the device
+is joining its own channel, by comparing the id in the channel name to the id on
+the certificate. Next, the server compares the running version of the firmware
+to the target version for that device on `NervesHub`. If those versions are
+different, then the channel queues a firmware update message to be sent as soon
+as the channel finishes the join process. The final step of the join process is
+to respond to the device with a list of the group channels that it should join.
+
+## Joining Group Channels
+
+If the client receives a group channel from the device channel, it will attempt
+to join it. Authorization to join the group channel for that tenant and product
+are verified by checking the device's associations in the database. These
+channels provide the same interface as the individual device channels, but allow
+bulk management of devices. Group channels are named using the format
+`group:#{tenant}:#{product}:#{group_name}`.
+
+## Joining and Leaving Groups
+
+If a user adds or removes a device from a group while the device is connected,
+the server sends a message to join or leave the group channel. Messages to join
+and leave a channel look like `{"join", channel_name}` and `{"leave",
+channel_name}` respectively. If a user deletes a group, all members will receive
+a `{"leave", channel_name}` message.
+
+## Updating Firmware
+
+If the target version changes for a device or group of devices, the channel will
+issue an `{"firmware_update", %{"url" => update_url, "version" => version}`
+message to either the individual device channel, or the group channel. This will
+tell the firmware to start streaming the `.fw` file to the
+`Nerves.Firmware.Fwup` client. If the device is already on the version that it
+was instructed to update the firmware to, which may happen in groups, the device
+ignores the update command.
+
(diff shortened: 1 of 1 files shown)


docs/client-server.md
```diff
@@ -0,0 +1,53 @@
...
+bulk management of devices. Group channels are named using the format
+`group:#{tenant}:#{product}:#{group_name}`.
+
+## Joining and Leaving Groups
+
+If a user adds or removes a device from a group while the device is connected,
+the server sends a message to join or leave the group channel. Messages to join
+and leave a channel look like `{"join", channel_name}` and `{"leave",
+channel_name}` respectively. If a user deletes a group, all members will receive
+a `{"leave", channel_name}` message. If the device is already on the version
```
jlmcgehee21: > If the device is already on the version that it was instructed to upgrade to, which may happen in groups, it ignores the upgrade command.

Should this go under the "Upgrading Firmware" section?

docs/client-server.md
```diff
@@ -0,0 +1,53 @@
...
+that it was instructed to upgrade to, which may happen in groups, it ignores the
+upgrade command.
+
+## Upgrading Firmware
+
+If the target version changes for a device or group of devices, the channel will
+issue an `{"upgrade", %{"url" => upgrade_url, "version" => version}` message.
+This will tell the firmware to start streaming the `.fw` file to the
+`Nerves.Firmware.Fwup` client. It will also send updates to the device channel
+when it begins and finishes applying the update and before it reboots. If there
```
jlmcgehee21: > ~It~ The client will also send updates to the device channel when it begins and finishes applying the update and before it reboots.

fhunleth: Were you thinking of having `{"update_status", %{status="started", "message" => message}}` or something like that? I'm not sure we need much in terms of status. `fwup` will tell you the percentage complete, but I'm not sure we need to go that far.

docs/client-server.md
```diff
@@ -0,0 +1,53 @@
...
+are any issues downloading or applying the update, those are sent back as
+`{"update_error", %{"errors" => [errors]}}`.
```
fhunleth: Minor nit: Should it be `update*` or `upgrade*` for the firmware update/upgrade messages? I have a slight preference to `update` since you can downgrade firmware too. Mostly I just want to be consistent.

docs/client-server.md
mjs2600: @fhunleth I don't think we need much, but we've had bad updates brick devices before and it's nice now to know what state the device was last in. We also needed to be able to delay the reboot while we cleaned up sessions, so it's helpful to know if that's done, but hasn't been allowed to reboot yet.

docs/client-server.md
mjs2600: Yep, I totally agree. I'll change it.

docs/client-server.md
fhunleth: Ok - totally bummed that you've had bad updates that bricked devices. I think that we can make that recoverable. I'm good with whatever status you want to put in the protocol. We can always turn it off or selectively enable it in the future if it's an issue.

jlmcgehee21: @fhunleth, it wasn't a true "bricking" where the firmware update/nerves failed. But a device was functionally bricked because we pushed bad software that prevented the application from starting or server communication or w/e.  Only ever happened in staging type environments, but of course, it could happen elsewhere.

fhunleth: Good to know. At Smartrent, we have a `validate_fw` call that we make if the firmware can contact the network after an update. The plan is to enable a 5 minute timer to reboot if the server can't be contacted. If the firmware isn't validated, then it rolls back to the previous version. I haven't had time to push this out to the main Nerves systems (it's a little tricky and changes a lot of docs), but it might be a nice thing to eventually implement for nerves-hub.

mjs2600:
@fhunleth can you take another look at this when you get a chance?

PR #75, opened 2018-05-31
rename nerveshub to nerves_hub

stwf:


stwf:
@mobileoverlord The one place i couldn't change nerveshub to nerves_hub without breaking the tests  was:
https://github.com/nerves-hub/nerveshub/blob/96c34fda76e15490993f652c13415bed6c472cf0/test/nerves_hub/integration/websocket_test.exs#L14

Would that require a certificate change too? Isit a necessary change?

mobileoverlord:
The certificate would need to be regenerated again with the common name change for it to pass if you change this to nerves-hub. There is a readme you can update/follow to make the change and check in the new certificates.

PR #83, opened 2018-06-05
Device,Deployment tests and fixtures

jlmcgehee21:
Why:
* I will feel more confident developing upcoming features if there are
tests for the underlying functionality.

* `deployments_test.exs`, minor refactoring of `deployments.ex` and
`deployment.ex`, and updating `deployment_controller.ex` to match.
* `devices_test.exs`, minor refactoring of `devices.ex` and
`device.ex`, and updating `devices_controller.ex` to match.
* fixtures in `test/support/fixtures.ex`

Other Notes:
* I only added test coverage for functions that would directly impact my
upcoming feature(s).  In order to improve overall test coverage without
sacrifing velocity entirely, I will continue to follow this pattern in
the future.

### DIFF
--- a/lib/nerves_hub/deployments/deployment.ex
+++ b/lib/nerves_hub/deployments/deployment.ex
@@ -8,6 +8,8 @@ defmodule NervesHub.Deployments.Deployment do
   alias __MODULE__
 
   @type t :: %__MODULE__{}
+  @required_fields [:tenant_id, :firmware_id, :name, :conditions, :is_active]
+  @optional_fields []
 
   schema "deployments" do
     belongs_to(:tenant, Tenant)
@@ -21,16 +23,9 @@ defmodule NervesHub.Deployments.Deployment do
   end
 
   def changeset(%Deployment{} = deployment, params) do
-    fields = [
-      :name,
-      :conditions,
-      :is_active,
-      :firmware_id
-    ]
-
     deployment
-    |> cast(params, fields)
-    |> validate_required(fields)
+    |> cast(params, @required_fields ++ @optional_fields)
+    |> validate_required(@required_fields)
     |> validate_change(:conditions, fn :conditions, conditions ->
       types = %{tags: {:array, :string}, version: :string}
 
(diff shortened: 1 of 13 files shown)


lib/nerves_hub/deployments/deployment.ex
```diff
@@ -8,6 +8,8 @@ defmodule NervesHub.Deployments.Deployment do
   alias __MODULE__
 
   @type t :: %__MODULE__{}
+  @required_fields [:tenant_id, :firmware_id, :name, :conditions, :is_active]
```
mattlorey: I like this pattern, good seeing how other people do things like this

lib/nerves_hub/deployments/deployments.ex
```diff
@@ -34,16 +34,28 @@ defmodule NervesHub.Deployments do
...
+  def create_deployment(params) do
+    %Deployment{}
     |> Deployment.changeset(params)
     |> Repo.insert()
   end
 
+  @spec create_deployment_with_tenant(Tenant.t(), map) ::
+          {:ok, Deployment.t()} | {:error, Changeset.t()}
+  def create_deployment_with_tenant(tenant, params) do
+    params = Map.put(params, :is_active, false)
```
mattlorey: The incoming map with have string since it is coming from a form post, using an atom here will fail due to mixed string/atom keys

jlmcgehee21: Ah, right. Good point.  Great chance to add a controller test!  Will do that to address this.

lib/nerves_hub/devices/devices.ex
```diff
@@ -41,16 +41,29 @@ defmodule NervesHub.Devices do
...
+          {:ok, Device.t()}
+          | {:error, Changeset.t()}
+  def create_device_with_tenant(%Tenant{} = tenant, params) do
+    device =
+      tenant
+      |> Ecto.build_assoc(:devices)
+
+    device
+    |> Map.from_struct()
+    |> Map.merge(params)
```
mattlorey: The `params` here are coming from the controller, so we'll have mixed string and atoms again

jlmcgehee21: aaah. I got confused -- I meant `deployment_controller`.  I'll re-work this as well.

PR #84, opened 2018-06-07
Device Channel checks for firmware update on join

jlmcgehee21:
Why:
* We want to know if a device needs new firmware ASAP.
* https://github.com/nerves-hub/nerves_hub/issues/79

* Checking if `%{"version" => version}` in the join payload matches the
target version in the database, and acting accordingly.
* Also updates the `current_version` in the database to the version that
is reported in the join payload.

Caveats:
* Cannot send out a message with instructions for updating firmware
because that is yet to be implemented.

### DIFF
--- a/lib/nerves_hub/devices/device.ex
+++ b/lib/nerves_hub/devices/device.ex
@@ -33,18 +33,11 @@ defmodule NervesHub.Devices.Device do
     timestamps()
   end
 
-  def creation_changeset(%Device{} = device, params) do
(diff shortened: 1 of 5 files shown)


mattlorey:
@jlmcgehee21 I don't think our discussion this morning changes anything for this PR as it is - I know there will be changes once some decisions are made, but was going to merge this as is unless you want to hold off?

jlmcgehee21:
:shipit:

PR #85, opened 2018-06-08
Edit deployments/delete deployments

stwf:
addresses tickets:
https://github.com/nerves-hub/nerves_hub/issues/40
https://github.com/nerves-hub/nerves_hub/issues/41

### DIFF
--- a/lib/nerves_hub/deployments/deployment.ex
+++ b/lib/nerves_hub/deployments/deployment.ex
@@ -22,11 +22,26 @@ defmodule NervesHub.Deployments.Deployment do
     timestamps()
   end
 
+  def edit_changeset(%Deployment{} = deployment, params) do
+    fields = [
+      :name,
+      :conditions,
+    ]
+    deployment
+    |> cast(params, fields)
+    |> validate_required(fields)
+    |> validate_conditions()
+  end
+
   def changeset(%Deployment{} = deployment, params) do
     deployment
     |> cast(params, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
-    |> validate_change(:conditions, fn :conditions, conditions ->
+    |> validate_conditions()
+  end
+
+  defp validate_conditions(changeset, _options \\ []) do
+    validate_change(changeset, :conditions, fn :conditions, conditions ->
       types = %{tags: {:array, :string}, version: :string}
 
       version =
--- a/lib/nerves_hub/deployments/deployments.ex
+++ b/lib/nerves_hub/deployments/deployments.ex
@@ -34,6 +34,25 @@ defmodule NervesHub.Deployments do
     end
   end
 
+  @spec delete_deployment(Tenant.t(), Deployment.t()) ::
+          {:ok, Deployment.t()} | {:error, :not_found}
+  def delete_deployment(%Tenant{id: tenant_id}, %Deployment{id: deployment_id}) do
+    from(
+      d in Deployment,
+      where: d.tenant_id == ^tenant_id,
+      where: d.id == ^deployment_id
+    )
+    |> Repo.one!()
+    |> Repo.delete()
+    |> case do
+      {:error, _changeset} ->
+        {:error, :not_found}
+
+      {:ok, deployment} ->
+        {:ok, deployment}
+    end
+  end
+
   @spec create_deployment(map) :: {:ok, Deployment.t()} | {:error, Changeset.t()}
   def create_deployment(params) do
     %Deployment{}
--- a/lib/nerves_hub/deployments/deployments.ex
+++ b/lib/nerves_hub/deployments/deployments.ex
@@ -53,6 +53,13 @@ defmodule NervesHub.Deployments do
     end
   end
 
+  @spec update_deployment(Deployment.t(), map) :: {:ok, Deployment.t()} | {:error, Changeset.t()}
+  def update_deployment(deployment, params) do
+    deployment
+    |> Deployment.edit_changeset(params)
+    |> Repo.update()
+  end
+
   @spec create_deployment(map) :: {:ok, Deployment.t()} | {:error, Changeset.t()}
   def create_deployment(params) do
     %Deployment{}
--- a/lib/nerves_hub_web/controllers/deployment_controller.ex
+++ b/lib/nerves_hub_web/controllers/deployment_controller.ex
@@ -6,7 +6,7 @@ defmodule NervesHubWeb.DeploymentController do
   alias NervesHub.Deployments.Deployment
   alias Ecto.Changeset
 
-  plug(NervesHubWeb.Plugs.FetchDeployment when action in [:show, :toggle_is_active, :delete])
+  plug(NervesHubWeb.Plugs.FetchDeployment when action in [:show, :toggle_is_active, :delete, :edit, :update])
 
   def index(%{assigns: %{tenant: %{id: tenant_id}}} = conn, _params) do
     deployments = Deployments.get_deployments_by_tenant(tenant_id)
@@ -67,14 +67,7 @@ defmodule NervesHubWeb.DeploymentController do
           params
           |> Map.put("tenant_id", tenant.id)
           |> Map.put("is_active", false)
-          |> Map.put("conditions", %{
-            "version" => params["version"],
-            "tags" =>
-              params["tags"]
-              |> tags_as_list()
-              |> MapSet.new()
-              |> MapSet.to_list()
-          })
+          |> inject_conditions_map()
 
         result = Deployments.create_deployment(params)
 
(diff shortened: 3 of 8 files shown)


lib/nerves_hub_web/controllers/deployment_controller.ex
```diff
@@ -112,6 +108,71 @@ defmodule NervesHubWeb.DeploymentController do
...
+    end
+  end
+
+  def update(%{assigns: %{tenant: tenant, deployment: deployment}} = conn, %{
+        "deployment" => deployment_params
+      }) do
+    params = inject_conditions_map(deployment_params)
+
+    tenant
+    |> Firmwares.get_firmware(params["firmware_id"])
```
paulstatezny: Should we load the firmware via a plug?

I'm not sold either way... just throwing out the question.

lib/nerves_hub_web/templates/deployment/edit.html.eex
```diff
@@ -0,0 +1,71 @@
+<h1>Edit deployment</h1>
+
+<%= form_for @changeset, deployment_path(@conn, :update, @deployment), [as: :deployment], fn f -> %>
+  <h3 class="h3">Firmware Details</h3>
+
+  <table class="table" style="width: auto">
+    <tbody>
+      <tr>
+        <th>Product</th>
+        <td><%= @firmware.product %>
```
paulstatezny: We're missing a `</td>` here.

test/nerves_hub_web/controllers/deployment_controller_test.exs
```diff
@@ -54,4 +55,51 @@ defmodule NervesHubWeb.DeploymentControllerTest do
       assert html_response(conn, 200) =~ "Inactive"
     end
   end
+
+  describe "edit deployment" do
+    test "edits the chosen resource", %{conn: conn, current_tenant: tenant} do
```
paulstatezny: These tests are great! Loving them. 🙂

Super pedantic comment, but I feel like the description for this test (based on what's being asserted) should be more like, "displays an edit page".

lib/nerves_hub/deployments/deployment.ex
```diff
@@ -22,11 +22,27 @@ defmodule NervesHub.Deployments.Deployment do
     timestamps()
   end
 
+  def edit_changeset(%Deployment{} = deployment, params) do
```
jlmcgehee21: I don't think we need a separate `edit_changeset`.  Unless we are implying that the only things that should be editable are `[:name, :conditions]`.  If the intent is to have these be the only editable fields in the browser, I suggest this logic get moved to the controller.

lib/nerves_hub/deployments/deployments.ex
```diff
@@ -34,6 +34,32 @@ defmodule NervesHub.Deployments do
     end
   end
 
+  @spec delete_deployment(Tenant.t(), Deployment.t()) ::
+          {:ok, Deployment.t()} | {:error, :not_found}
+  def delete_deployment(%Tenant{id: tenant_id}, %Deployment{id: deployment_id}) do
```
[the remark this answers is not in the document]
jlmcgehee21: I agree here

lib/nerves_hub_web/controllers/deployment_controller.ex
```diff
@@ -112,6 +108,71 @@ defmodule NervesHubWeb.DeploymentController do
...
+    tenant
+    |> Firmwares.get_firmware(deployment.firmware_id)
+    |> case do
+      {:ok, firmware} ->
+        conn
+        |> render(
+          "edit.html",
+          deployment: deployment,
+          firmware: firmware,
+          changeset: Deployment.edit_changeset(deployment, %{}) |> tags_to_string()
```
jlmcgehee21: See comment on the `edit_changeset`

test/nerves_hub_web/controllers/deployment_controller_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule NervesHubWeb.DeploymentControllerTest do
   use NervesHubWeb.ConnCase.Browser
```
jlmcgehee21: Really love the tests that back up the new functionality!

lib/nerves_hub/deployments/deployments.ex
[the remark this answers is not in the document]
jlmcgehee21: If the session is created appropriately, and the tenant is put on `conn.assigns`, won't that prevent any tampering you're mentioning @stwf?

lib/nerves_hub/deployments/deployment.ex
stwf: ok, I like using separate changesets for everything. Why give someone an opportunity to inject data when you can remove it so easily. And it keeps the controllers clean. But of course either way works.

jlmcgehee21: @stwf, I'm definitely willing to discuss further.  My reasoning is as follows:

If this is the _only_ edit that will be allowed throughout the entire application logic, then I think this is a good move.  Otherwise, I would prefer to push this logic to the controller, which is the only place it will be used.

I'm totally willing to be convinced otherwise.  This is just the pattern that I am used to following.

lib/nerves_hub/deployments/deployments.ex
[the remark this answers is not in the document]
paulstatezny: Hmmm, I feel you bring up a great, broader question there @stwf. Namely, how do we want to handle access control? (Preventing users from accessing resources that they shouldn't.)

At SmartRent we've generally been doing that in the `Fetch*` plugs to pull ID's out of the URL and ensure that the user has access before allowing to proceed to the controller. Not that that's the "right" or only way.

I guess my question is: can we pick 1 way/place to perform all (or at least most) of our access control?

lib/nerves_hub/deployments/deployment.ex
paulstatezny: I share @stwf's perspective.

I think code is more easily understood (and therefore less bug prone) when written in "layers", where each layer is responsible for distinct, clearly defined things. So I think we should be consistent, that all input validation happen in one place. We already have the pattern of using changeset functions in the scheme modules.

I hear you about this comment, @jlmcgehee21:

> Unless we are implying that the only things that should be editable (even by our application code) are [:name, :conditions]

I think that is what we want, but I could be wrong. Thinking if they want to make a deployment for a different firmware, they should make a new deployment. Makes it more "formal", and require more steps so they have to slow down and think about what they're doing. Which we at SmartRent agreed (at one point :) is good to help the user make less mistakes. (Since creating a deployment means devices in the field start to update themselves!)

lib/nerves_hub/deployments/deployments.ex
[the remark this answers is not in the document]
jlmcgehee21: +1 @paulstatezny, I'm used to handling all access control and authorization with plugs as well.  But I definitely prefer that all of it happens via the same pattern throughout the application.

lib/nerves_hub/deployments/deployment.ex
jlmcgehee21: It seems we're at a standoff 😄 .  @mattlorey was on my side.  TBH, it might make sense to have this changeset, if it is the intention that it should be impossible to edit fields not listed within.  @mattlorey and I are nervous about the number of changesets exploding due to developer whims, but if it is a core part of every `update`, then it may belong here.  

Again, my number one concern is that we have a consistent pattern as the application grows.

mattlorey: This may have been an oversight in this `edit_changeset` specifically, but it does not allow updating `is_active` which will need to be changed.

So if from a edit page we can only change `name` and `conditions` and would use this changeset, would we have a different changeset for toggling `is_active`?

paulstatezny: Good point @mattlorey about `is_active`, I think it should be allowed for update.

I can see a future where changeset functions explode in number, (we have some crusty areas of code with that problem,) but I think that's related to a couple avoidable root causes: ("developer whims" not being a root cause)

1. Not architecting code well
2. Reviewers not being concerned about architecting code well or taking the time to consider good architecture

I think it could be said of most code patterns that developer laziness (not thinking carefully) will make a mess. So I don't think that's a helpful criticism.

jlmcgehee21: I won't withhold my approval due to this.  Both sides have valid points.  I don't want my opinions to hold up the project either. Would like to see the other change requests that all the reviewers agreed upon addressed though.

jlmcgehee21: I won't withhold my approval due to this.  Both sides have valid points.  I don't want my opinions to hold up the project either. Would like to see the other change requests that all the reviewers agreed upon addressed though.

lib/nerves_hub/deployments/deployments.ex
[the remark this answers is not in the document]
stwf: OK, done and it does look like the Plug is guaranteeing ownership correctly.

lib/nerves_hub/deployments/deployment.ex
stwf: OK, sorry, so I did change some more things around!
I don't really like the toggle behavior of the endpoint. It leaves a very real chance that pressing that button has the opposite effect that it says its going to have. So I just made it go through the regular update endpoint, using the edit_changeset.
To walk back a bit what I said before, I don't think stuff in nerves_hub should ever need info about whats going on in nerves_hub_web. But if its part of the domain logic that firmware_id can't be changed it seems ok to try and ensure that inside nerves_hub. Also, we can't publish an endpoint that lets this be changed, it has to be handled somewhere and this seems like a nice way to do it.
But I also agree just deciding and moving on quickly is the best path so whatever the decision is really fine.

Issue #91, opened 2018-06-13
Better handling of  firmware metadata


jlmcgehee21:
Currently, firmware `metadata` is being parsed for a few key fields, and then the entire `metadata` string is being stored in a `metadata` string field on a `firmware` record.  

The current issue, is that it is very common in the real world for the `metadata` field to be greater than the database imposed limit of 255 characters.

The proposed fix:
* A migration that removes the `metadata` string field, and adds the following relevant fields supplied by @fhunleth:
```
meta-product
meta-description
meta-version
meta-author
meta-platform
meta-architecture
meta-vcs-identifier
meta-misc
meta-uuid
```
* Refactor the `firmware_controller` and the `firmware` schema to handle these changes as well.

fhunleth:
Regarding missing fields:
1. Once you have `fwup` 1.2.0 installed, the only guaranteed field is `meta-uuid`. It's possible for people to hack their .fw files to remove all of the other fields.
2. Nerves makes it very difficult to remove `meta-platform`, `meta-architecture`, and `meta-version` and keep things working. If those don't exist, I'd reject the .fw file since it probably won't work anyway.
3. All of the other fields can be set to empty strings. Empty strings aren't printed in `fwup`'s metadata dump. There's no concept of null if `fwup` so if a field isn't present, writing an empty string to the database would be consistent.

danielspofford:
Do we want a dedicated column for `meta-creation-date` as well? That is read out of `metadata` [currently](https://github.com/nerves-hub/nerves_hub/blob/master/lib/nerves_hub/firmwares/firmware.ex#L56).

danielspofford:
It already exists :), just asked because it wasn't listed above and other duplicates were.

jlmcgehee21:
IMO, we should have columns for any metadata fields we store.

fhunleth:
@danielspofford Thanks for bringing up `meta-creation-date`. I actually regret adding that field to `fwup` several years ago and I've been trying to get rid of it ever since. It's hard to get rid of fields, though. My vote would be to remove from Nerves Hub. I also left out `meta-fwup-version` above. Both fields seemed like a good idea at the time, but have the effect of making reproduceable .fw files difficult to make. By that, I mean that if you feed the same inputs to `fwup`, you'd think that you'd get an identical `.fw` file. Those fields mess that up. I don't think that I know of a project that has actually used them for something useful.

danielspofford:
[...]

PR #95, opened 2018-06-17
Split firmware metadata into distinct fields

danielspofford:
Addresses #91

* Create migration and adjust schema to drop `timestamp` and `metadata`,
and add `author`, `description`, `misc`, `uuid`, `vcs_identifier`
* Adjust required and nullable fields
* Split `Firmware.metadata_item/2` into `fetch_metadata_item/2` and
`get_metadata_item/{2,3}` mirroring `Map.{fetch,get}` semantics
* Update templates and views to respect changes


jlmcgehee21:
I saw the change to the fixtures and that does address current tests, but would love it if there were some more tests here.  Coverage is rather low, so it would be awesome to add any other tests that may be relevant on any PR.

danielspofford:
Yeah, #90 will facilitate some good error path tests in the firmware controller (fixtures for corrupt and signed `.fw`s), but I can take another pass over this and see where I can buff up some coverage.

jlmcgehee21:
Thanks @danielspofford. Just anything that sticks out to you. Doesn't have to be exhaustive.

lib/nerves_hub/firmwares/firmware.ex
```diff
@@ -8,74 +8,65 @@ defmodule NervesHub.Firmwares.Firmware do
   alias __MODULE__
 
   @type t :: %__MODULE__{}
+  @optional_params [
+    :author,
+    :description,
+    :misc,
+    :product,
```
jlmcgehee21: After looking closer at this, we will need `product` + `version` _or_ just `uuid` to uniquely identify a firmware image (when a device reports this info from the outside world).  I think the right approach will be to make `product` required, and leave `uuid` optional.  This will allow us to look up based on `uuid` if it is available, but because it is new to `fwup`, it might not make it into some firmware builds, etc.

fhunleth: Is the question, "how do we verify that the firmware running on the device is the same as the firmware contained in a particular `.fw` file?"

jlmcgehee21: It's a statement that we can only guarantee being able to verify the firmware running on a device matches a particular `.fw` file if we have (`product` _and_ `version`) _or_ `uuid`.

Since we may not be able to count on `uuid` in `fwup` (since it's new).  I'm suggesting that we make `product` a required field, and leave `uuid` optional in order to allow for a firmware to be uploaded without a `uuid`.

jlmcgehee21: This is actually only true in an ideal world, because there are no database enforcements on a unique `product + version` combo of firmware.  Which is the motivation behind my comment here: https://github.com/nerves-hub/nerves_hub/issues/63#issuecomment-398167260

fhunleth: If `uuid` isn't available, how about comparing all of the metadata fields? That will catch cases where `vcs-identifer` changes, but `version` doesn't. It's hard to believe that differences in the other fields will be meaningful, but comparing all available fields seems preferable to making a couple fields special.

fhunleth:
@danielspofford Thanks for the PR! And @jlmcgehee21, thanks for reviewing and making the UUID change.

PR #96, opened 2018-06-19
current_firmware and target_deployment for device

jlmcgehee21:
Why:

* We want to make it easy to tell if a device needs an update.
* We want to be able to provide a URL when a device needs an update.

* Migration that adds `:target_deployment` and `:current_firmware` to
`Device`.
* Logic in the `device_channel` which finds the url for an update if it
is available.

Caveats:
* Web views/UI for the new fields need to be worked out.

### DIFF
--- a/lib/nerves_hub/firmwares/firmwares.ex
+++ b/lib/nerves_hub/firmwares/firmwares.ex
@@ -28,24 +28,12 @@ defmodule NervesHub.Firmwares do
     end
   end
 
-  @spec get_firmware(integer()) ::
+  @spec get_firmware_by_product_and_version(Tenant.t(), String.t(), String.t()) ::
           {:ok, Firmware.t()}
           | {:error, :not_found}
-  def get_firmware(id) do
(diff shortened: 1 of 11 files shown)


lib/nerves_hub/firmwares/firmwares.ex
```diff
@@ -28,6 +28,30 @@ defmodule NervesHub.Firmwares do
     end
   end
 
+  @spec get_firmware(integer()) ::
```
jlmcgehee21: @paulstatezny I wouldn't mind some specific feedback here.  In the socket, we don't have access to a tenant directly, since the device certificates are only tied to a `device_identifier`.  This renders the intent of `get_firmware/2` a bit useless since the only way to get the tenant is by looking it up _through_ the device record.

mattlorey: @jlmcgehee21 That function is still used in a plug to pull the firmware on the front end when we do know the tenant.  

So I think it's still valid - but I'm not sure if that answers your question/addresses your concern.

PR #105, opened 2018-07-12
Enable erlang distribution in production

mobileoverlord:
We are running at minimum two nodes in production in different availability zones and will need to cluster the group since we are using phoenix channels.

### DIFF
--- a/rel/config.exs
+++ b/rel/config.exs
@@ -30,13 +30,13 @@ environment :dev do
   # dev mode.
   set dev_mode: true
   set include_erts: false
-  set cookie: :"LcWJ?&&oKl41r:6VyC.kd`g[[wzlO[DT>2EXPKkQ>quN5XYx$=0}R}?w>jn9_Z!@"
+  set cookie: :""
 end
 
 environment :prod do
   set include_erts: true
   set include_src: false
-  set cookie: :"V0{_Lkp9^K6$}upoGQ=QWk[;kJY:!}G|`q;5}J[th}r_X@]3=/{AW)4O_YjMoFWV"
+  set cookie: :""
 end
 
 # You may define one or more releases in this file.
@@ -52,5 +52,6 @@ release :nerves_hub do
   set applications: [
     :runtime_tools
   ]
+  set vm_args: "rel/vm.args"
 end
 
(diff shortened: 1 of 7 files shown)


rel/scripts/ecs-cluster.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -e
+
+CLUSTER=nerves-hub
+SERVICE=nerves-hub
+
+METADATA=`curl http://169.254.170.2/v2/metadata`
```
fhunleth: Where does this IP address come from?

rel/config.exs
```diff
@@ -30,13 +30,13 @@ environment :dev do
   # dev mode.
   set dev_mode: true
   set include_erts: false
-  set cookie: :"LcWJ?&&oKl41r:6VyC.kd`g[[wzlO[DT>2EXPKkQ>quN5XYx$=0}R}?w>jn9_Z!@"
+  set cookie: :""
```
fhunleth: Did you blank out the cookie here since it doesn't get used? (I always get confused at the cookie that's put here since I always use the one in `vm.args`)

rel/scripts/ecs-cluster.sh
mobileoverlord: The IP address is being pulled from the internal metadata api. https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html

rel/config.exs
mobileoverlord: Correct, it’s unused now. The cookie is now the one in the vm.args which is set at container runtime from the env.

rel/vm.args
```diff
@@ -0,0 +1,24 @@
...
+##+A 5
+
+## Increase number of concurrent ports/sockets
+##-env ERL_MAX_PORTS 4096
+
+## Tweak GC to run more often
+##-env ERL_FULLSWEEP_AFTER 10
+
+# Enable SMP automatically based on availability
+-smp auto
```
fhunleth: This is ignored since there's no longer a non-SMP scheduler. There's a way to force Erlang to only use one core if that's desirable.

PR #107, opened 2018-07-13
Products Table

jlmcgehee21:
## Why:
* It is good to have a database schema that matches the structure of the
real world use cases.
* The `:product` string fields make it difficult to properly identify devices for deployments.

## This change addresses the need by:
* Products table (per the ERD)
* Scoping `devices`, `firmwares`, and `deployments` to a `product`.

## Questions this has brought up for me:
* There is no `delete` function in the device controller, is this intentional?
* Is there any reason the patterns established by the the phoenix generators weren't used on our existing controllers?
* I noticed plugs in the controller, but I typically have used them in the router with `pipe_through`.  This seems more clear to me, but I'm open to being told why it is wrong.

## Author's Note:
This got a bit out of hand when my changes broke all of the controllers.
I tried to implement my view of how this should work, but I may have
missed some things, and others may not like things I did (like scoping
various entities to a product).

### DIFF
--- a/lib/nerves_hub/deployments/deployment.ex
+++ b/lib/nerves_hub/deployments/deployment.ex
@@ -2,18 +2,21 @@ defmodule NervesHub.Deployments.Deployment do
   use Ecto.Schema
 
   import Ecto.Changeset
+  import Ecto.Query
 
   alias NervesHub.Accounts.Tenant
   alias NervesHub.Firmwares.Firmware
+  alias NervesHub.Products.Product
+
   alias __MODULE__
 
   @type t :: %__MODULE__{}
-  @required_fields [:tenant_id, :firmware_id, :name, :conditions, :is_active]
+  @required_fields [:product_id, :firmware_id, :name, :conditions, :is_active]
   @optional_fields []
 
   schema "deployments" do
-    belongs_to(:tenant, Tenant)
     belongs_to(:firmware, Firmware)
+    belongs_to(:product, Product)
 
     field(:name, :string)
     field(:conditions, :map)
@@ -40,6 +43,16 @@ defmodule NervesHub.Deployments.Deployment do
     |> cast(params, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
     |> validate_conditions()
+  end
+
+  def with_firmware(deployment_query) do
+    deployment_query
+    |> preload(:firmware)
+  end
+
+  def with_product(deployment_query) do
+    deployment_query
+    |> preload(:product)
   end
 
   defp validate_conditions(changeset, _options \\ []) do
--- a/lib/nerves_hub_web/controllers/firmware_controller.ex
+++ b/lib/nerves_hub_web/controllers/firmware_controller.ex
@@ -6,8 +6,8 @@ defmodule NervesHubWeb.FirmwareController do
   alias NervesHub.Firmwares.Firmware
   alias NervesHub.Accounts.TenantKey
 
-  def index(%{assigns: %{tenant: %{id: tenant_id}}} = conn, _params) do
-    firmwares = Firmwares.get_firmware_by_tenant(tenant_id)
+  def index(%{assigns: %{tenant: tenant, product: %{id: product_id}}} = conn, _params) do
+    firmwares = Firmwares.get_firmwares_by_product(product_id)
 
     render(conn, "index.html", firmwares: firmwares)
   end
@@ -17,14 +17,14 @@ defmodule NervesHubWeb.FirmwareController do
     |> render("upload.html", changeset: %Changeset{data: %Firmware{}})
   end
 
-  def do_upload(%{assigns: %{tenant: tenant}} = conn, %{
+  def do_upload(%{assigns: %{tenant: tenant, product: product}} = conn, %{
         "firmware" => %{"file" => %{filename: filename, path: path}}
       }) do
     with {:ok, tenant_key_id} <- verify_signature(path, tenant.tenant_keys),
          {:ok, metadata} <- Firmwares.extract_metadata(path),
          {:ok, architecture} <- Firmware.fetch_metadata_item(metadata, "meta-architecture"),
          {:ok, platform} <- Firmware.fetch_metadata_item(metadata, "meta-platform"),
-         {:ok, product} <- Firmware.fetch_metadata_item(metadata, "meta-product"),
+         {:ok, product_name} <- Firmware.fetch_metadata_item(metadata, "meta-product"),
          {:ok, version} <- Firmware.fetch_metadata_item(metadata, "meta-version"),
          author <- Firmware.get_metadata_item(metadata, "meta-author"),
          description <- Firmware.get_metadata_item(metadata, "meta-description"),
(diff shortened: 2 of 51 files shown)


mobileoverlord:
We should probably move to merge this one before https://github.com/nerves-hub/nerves_hub/pull/108. I can rebase the umbrella switch afterwards.

mattlorey:
Some answers to your questions from my perspective/memory:
 - It hadn't been decided how deletes would be handled - soft deletes, hard deletes, should you even be able to delete a device or just 'disable' it in some way.  Seems like this is something that could differ from organization to organization.
 - No particular reason, I for one don't like using `resources` as it isn't as clear as having the routes listed individually, but I see the benefit of naming consistency
 - We've discussed this a few times internally for our projects and for me personally having the plug in the controller is easier to follow, since if you are debugging/working on a controller function and don't know where something is getting assigned from, having the plug in the router, and possibly having to follow multiple `pipe_through`s is less convenient/more confusing than being able to see them in the controller where you are already at.  But this is just a matter of preference and as long as we're consistent, it's better than having some in the controller and some in the router.


Going through the changes now, 2 things that I noticed so far:
 - There is not link to the deployments page(s) from a product - is this intentional
 - The product pages aren't styled - is this intentional

I'll continue reviewing

lib/nerves_hub_web/controllers/firmware_controller.ex
```diff
@@ -17,14 +17,14 @@ defmodule NervesHubWeb.FirmwareController do
...
-  def do_upload(%{assigns: %{tenant: tenant}} = conn, %{
+  def do_upload(%{assigns: %{tenant: tenant, product: product}} = conn, %{
         "firmware" => %{"file" => %{filename: filename, path: path}}
       }) do
     with {:ok, tenant_key_id} <- verify_signature(path, tenant.tenant_keys),
          {:ok, metadata} <- Firmwares.extract_metadata(path),
          {:ok, architecture} <- Firmware.fetch_metadata_item(metadata, "meta-architecture"),
          {:ok, platform} <- Firmware.fetch_metadata_item(metadata, "meta-platform"),
-         {:ok, product} <- Firmware.fetch_metadata_item(metadata, "meta-product"),
+         {:ok, product_name} <- Firmware.fetch_metadata_item(metadata, "meta-product"),
```
mattlorey: I think this should be product_id

jlmcgehee21:
@mattlorey, none of those things that are missing were intentional, just side effects of big changes implemented without a pair.  Will be glad to fix those and any other problems.

WRT to the plugs, I have similar reasons for my viewpoint (I think it's less confusing if they are present in the router because `/parent/parent_id/child/child_id` scoping occurs there), but consistency is key above all, as you said.  I'm fine with going with the decision of the majority.

fwiw, I'm at a conference today and traveling tomorrow, so I may not get to this until monday.  So anyone should feel free to pull this down and make commits needed to get this through.

lib/nerves_hub/deployments/deployment.ex
```diff
@@ -2,18 +2,21 @@ defmodule NervesHub.Deployments.Deployment do
...
 
   alias NervesHub.Accounts.Tenant
   alias NervesHub.Firmwares.Firmware
+  alias NervesHub.Products.Product
+
   alias __MODULE__
 
   @type t :: %__MODULE__{}
-  @required_fields [:tenant_id, :firmware_id, :name, :conditions, :is_active]
+  @required_fields [:product_id, :firmware_id, :name, :conditions, :is_active]
```
paulstatezny: With the current associations we can have a deployment that has a firmware tied to one tenant, and that same deployment is tied to the product of a different tenant. This should be an illegal state, I think we'd all agree. We'd have to rely on the application code to not make the mistake, which seems too fragile to me.

> make illegal states unrepresentable

In the spirit of that quote, I would suggest:

1. Get rid of `firmwares.tenant_id` (but keep `firmwares.product_id`)
2. Get rid of `devices.tenant_id` (but keep `devices.product_id`)
3. Get rid of `deployments.product_id` (and preload/join the product through `firmwares`)

**Cost:** Slightly more database load for joining or extra queries (arguing this is a micro-optimization IMO)
**Benefit:** Rock solid guarantee that we don't have indirect associations that represent illegal states. A class of potential bugs go away, leaving more mental bandwidth for more important things.

It's a value system question. Maybe I'm wrong. 😄

paulstatezny: Another cost: "Less convenient" to not have immediate association with the product in these other entities.

mobileoverlord: I talked this through with Paul to catch up on things. From what I understand, the concern is about letting a tenant create a deployment with a product from another tenant. While this is _currently_ something that is not allowed, I believe that it is something that we are going to allow in the future. A dream use case for Nerves is to eventually allow tenants to make firmware / products "public" and let other tenants deploy the firmware to their device. The primary example of this would be that Nerves project examples could be deployed to new user boards so they can try out projects.

Paul and I agree that the state of this PR is good to go now. :)

lib/nerves_hub_web/controllers/firmware_controller.ex
mobileoverlord: @mattlorey This appears to be fetching the metadata from the fw file which i believe should contain the name and not the id.

mattlorey: Ah, I left that comment on the wrong line

lib/nerves_hub_web/controllers/firmware_controller.ex
```diff
@@ -38,7 +38,7 @@ defmodule NervesHubWeb.FirmwareController do
         description: description,
         misc: misc,
         platform: platform,
-        product: product,
+        product_name: product_name,
```
mattlorey: This is the one that should be `product_id`

mattlorey: https://github.com/nerves-hub/nerves_hub/pull/107/files#diff-7fc89b9b87bf2ab4ceeb424062343a34R18

mattlorey: I can fix this and push up the changes, I know you want to get the umbrella PR rebased on top of this and merged @mobileoverlord 

Just changing it to `product_id: product.id,` will allow the user to upload firmware again, but that isn't doing any validation that the firmware being uploaded matched the product and I don't know if that is desired?

I could compare the product name being pulled from the firmware file `product_name` to the name of the product (`product.name`) but I don't know if that is all of the extent we want to check that it matches.

jlmcgehee21: @mattlorey, I modified `create_firmware` to lookup the `product_id` if a `product_name` is passed: https://github.com/nerves-hub/nerves_hub/pull/107/files#diff-00c77d24057000afe97ba3c85f088d1cR65

mattlorey: Ah okay, I see that now.  The firmware I was testing with didn't match the name of the product I was trying to upload it to so it wasn't working.

I think we'll want to add some feedback to the end-user, but I'm good merging this for now so we can keep moving forward and I'll add a ticket to provide feedback to the user.

PR #113, opened 2018-07-16
Configure distinct db in test

danielspofford:
Why
---

* Previously dev and test were sharing the `db` database causing tests
to fail surprisingly

How
---

* Set test-specific `DATABASE_URL` env var in Makefile when running
tests
* Remove vestigial (even prior to this commit) `url: ...` and
`database: ...` mix repo configurations

PR
--

This addresses the issue of dev records being propagated "into" test and causing
tests to fail. There were a few misleading things at play here.

* Test config was configuring both `:url` and `:database`, potentially
implying some sort of merged functionality, or that the `:database` was taking
precedence. This is not the case, when `:url` is specified no other config is
aknowledged
* `:url` and friends like `:database` in mix configs were having no impact on
reality because `NervesHubCore.Repo` pulls in `DATABASE_URL` from the env at
runtime in its init callback overriding the `:url` option. This is what caused
dev and test to create and connect to the same database

This could be solved many different ways but I think using an env var across the
board provides consistency. In lieu of maintaining a `test.env` I opted to
override `DATABASE_URL` for test in the Makefile like we do for `MIX_ENV`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,10 +36,14 @@ rebuild-db:
 
 test: .env
 	. ./.env && \
-	    MIX_ENV=test mix test
+	DATABASE_URL=postgres://db:db@localhost:2345/nerves_hub_test \
+	MIX_ENV=test \
+	mix test
 
 test-watch: .env
 	. ./.env && \
-	    MIX_ENV=test mix test.watch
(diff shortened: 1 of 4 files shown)


mobileoverlord:
Going to merge this and manually kill the deployment since this only impacts test.

PR #120, opened 2018-07-16
Remove core prod ecto url config

danielspofford:
This config has no effect since the `NervesHubCore.Repo` module's init
callback reads from runtime and overrides `:url` anyway.


danielspofford:
`MIX_ENV=prod mix compile` works I checked 😅

PR #130, opened 2018-07-18
Fix user update form

jlmcgehee21:
Why:

* The password fields were email input fields

* Fixing the template.
* Testing


mobileoverlord:
Tested this locally, looks good to me

PR #134, opened 2018-07-19
create_deployment updates relevant devices

jlmcgehee21:
Why:

*  We want the creation of a deployment to update relevant devices that
should have this as a target version.

* Change `create_deployment` to find relevant devices and update their
`:target_deployment` asyncronously


jlmcgehee21:
@mobileoverlord anything preventing this one from going through?

PR #143, opened 2018-07-26
Feature/firmware is truth

jlmcgehee21:
Why:

* We want to avoid storing state on the server which could be stale

* Migrating database to match ERD.
  * Firmware is the single source of truth for info about a device.
* Removing url scoping of `product/product_id/device`.
* Move some files around in the test dirs to mimic the application
structure.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
@@ -48,12 +48,17 @@ defmodule NervesHubCore.Firmwares do
     end
   end
 
-  @spec get_firmware_by_uuid(String.t()) ::
+  @spec get_firmware_by_uuid(Tenant.t(), String.t()) ::
           {:ok, Firmware.t()}
           | {:error, :not_found}
-  def get_firmware_by_uuid(uuid) do
-    Firmware
-    |> Repo.get_by(uuid: uuid)
+  def get_firmware_by_uuid(%Tenant{id: t_id}, uuid) do
+    from(
+      f in Firmware,
+      where: f.uuid == ^uuid,
+      join: p in assoc(f, :product),
+      where: p.tenant_id == ^t_id
+    )
+    |> Repo.one()
     |> case do
       nil -> {:error, :not_found}
       firmware -> {:ok, firmware}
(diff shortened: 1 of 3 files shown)


apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
```diff
@@ -47,12 +48,12 @@ defmodule NervesHubCore.Firmwares do
     end
   end
 
-  @spec get_firmware_by_uuid(Tenant.t(), String.t()) ::
+  @spec get_firmware_by_uuid(String.t()) ::
           {:ok, Firmware.t()}
           | {:error, :not_found}
-  def get_firmware_by_uuid(%Tenant{} = tenant, uuid) do
+  def get_firmware_by_uuid(uuid) do
```
mobileoverlord: We _might_ want to still narrow the scope a little more here. It is possible that two tenants upload the same fw file.

mobileoverlord:
I took a look at the ERD that you included. Thanks for putting that in there 👍

apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
jlmcgehee21: yeah, was on the fence here. I'll fix this.

jlmcgehee21: I don't think there's a unique index in the db either...

mobileoverlord: Hurm, you raise a valid point. I didn't consider that it wouldn't really matter if the uuids are the same.

mobileoverlord: I think I am actually okay with your original implementation considering what @fhunleth said

mobileoverlord: A follow up from what we discussed in another channel. It’s best to keep tenant in here because if two tenants upload the same firmware with different signing keys they need to select the right one.

apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
```diff
@@ -50,9 +51,14 @@ defmodule NervesHubCore.Firmwares do
   @spec get_firmware_by_uuid(Tenant.t(), String.t()) ::
           {:ok, Firmware.t()}
           | {:error, :not_found}
-  def get_firmware_by_uuid(%Tenant{} = tenant, uuid) do
-    Firmware
-    |> Repo.get_by(tenant_id: tenant.id, uuid: uuid)
+  def get_firmware_by_uuid(%Tenant{id: t_id}, uuid) do
```
jlmcgehee21: @mobileoverlord I think this is fixed now

PR #150, opened 2018-07-30
Eager Firmware Update Message

jlmcgehee21:
Why:

*  We want to tell devices when they need new firmware

* On an update to a deployment, asyncronously send a message to the
`device:identifier` topic for each relevant device.

Other changes:
* Don't allow for the creation of a deployment with `%{is_active: true}`


mobileoverlord:
Looks good 👍

PR #163, opened 2018-08-06
Allow user cert deletion

jlmcgehee21:
Why:

* We want to delete certs that are bad/stale/etc.

* adding a delete route and controller.
* cleaning up the template.
* minor refactors to controller/router for testing of other controller
paths.

Other changes:
* Added a `.tool-versions` file for asdf users (like me :D)


mobileoverlord:
Tested this locally, looks good to me.

mobileoverlord:
Is there a race in the test. It passes locally for me.

jlmcgehee21:
I'm not sure. Don't think I did anything weird. It's passing locally for me too. I'm looking through it now.

PR #165, opened 2018-08-07
Device Status via Phx Presence

jlmcgehee21:
Why:

* We want to know what's happening with devices.

* Presence module in `NervesHubDeviceWeb` and adding some test
assertions in the device channel test.
* Device status via `Presence.list` in `DeviceChannel`


apps/nerves_hub_device/config/config.exs
```diff
@@ -10,11 +10,11 @@ config :nerves_hub_device,
   ecto_repos: [NervesHubCore.Repo],
   namespace: NervesHubDevice,
   websocket_auth_methods: [:ssl]
-  
+
 # Configures the endpoint
 config :nerves_hub_device, NervesHubDeviceWeb.Endpoint,
   render_errors: [view: NervesHubWWWWeb.ErrorView, accepts: ~w(html json)],
-  pubsub: [name: NervesHubWeb.PubSub, adapter: Phoenix.PubSub.PG2]
+  pubsub: [name: NervesHubDeviceWeb.PubSub, adapter: Phoenix.PubSub.PG2]
```
mobileoverlord: This should actually be `NervesHubWeb.PubSub`. The reason is that all the apps in the `Web` domain need to declare the same PubSub name, therefore instead of choosing a single one, we declare them all as connecting to the more generic name.

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/presence.ex
```diff
@@ -0,0 +1,74 @@
...
+  The function above fetches all users from the database who
+  have registered presences for the given topic. The fetched
+  information is then extended with a `:user` key of the user's
+  information, while maintaining the required `:metas` field from the
+  original presence data.
+  """
+  use Phoenix.Presence,
+    otp_app: :nerves_hub_device,
+    pubsub_server: NervesHubDeviceWeb.PubSub
+end
```
mobileoverlord: It seems that this module can be removed since the main supervisor is setting up the supervisor with the module `NervesHubDevice.Presence`. When I added that module, I was following the guide located at https://hexdocs.pm/phoenix/Phoenix.Presence.html

jlmcgehee21:
@mobileoverlord take a look at the latest changes when you get a chance

PR #167, opened 2018-08-07
cookie confirmation bar from https://cookieconsent.insites.com/download/

jlmcgehee21:
Looks like: 

image


mobileoverlord:
It would only appear the first time and then upon acceptance it writes to the cookie right?

jlmcgehee21:
yup @mobileoverlord, that was my experience locally

PR #173, opened 2018-08-09
Add device_certificates

mobileoverlord:
I guess in all my rebasing I snuck the migration in to a different PR. Sigh, oh well. Its in there. And here is the rest.

* Add device_certificates table.
* Update Device to has_many DeviceCertificates

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
@@ -1,7 +1,7 @@
 defmodule NervesHubCore.Devices do
   import Ecto.Query
 
-  alias NervesHubCore.Devices.Device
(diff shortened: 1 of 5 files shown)


jlmcgehee21:
`serial` in certificate land is `identifier` in database land.

mobileoverlord:
Actually there are two serial numbers that we are tracking here:

1. Device.identifier - This is a serial number that is set by the manufacturer or project owner. Used as a human readable / friendly serial, like `device-1234`
2. Certificate.serial - This is the serial number of the ssl certificate. 

We used this separation to address a concern that some users would replace the hardware's identifier if it were to undergo a repair that replaced major components where the identifier is derived from. Since this is on the device record we can allow this to happen without invalidating existing certificates.

Issue #168, opened 2018-08-08
Github Org Model


jlmcgehee21:
First:
- [x] change tenants to orgs
- [x]  make an org with the same name of a user and make that the users default org

Second:
- [x] allow users to be part of many orgs
- [x] allow orgs to be created
- [ ] set the user who created the org to be the org owner

jlmcgehee21:
- [x] change tenants to orgs

https://github.com/nerves-hub/nerves_hub_web/pull/170

jlmcgehee21:
@mobileoverlord -- that last checkbox is holding this up, but may be better addressed as part of a larger permissions system.  Right now all users have full permissions on an org because we have no permissions system at all.  Perhaps we close this and create an issue for per-org permissions?

PR #179, opened 2018-08-10
Create org with same name as user on account creation

jlmcgehee21:
Why:

* https://github.com/nerves-hub/nerves_hub_web/issues/168
* This is the next step to moving to the github org.

* Cleaning up some of the user changeset logic.
* Removing the `org_name` field from the account creation form.
* Injecting `org_name` into the user params in the controller.

Other changes:

* There was a "bug" where the password reset token was not expired when
it was used, so that was fixed and tested.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
@@ -1,6 +1,7 @@
 defmodule NervesHubCore.Accounts do
   import Ecto.Query
   alias Ecto.Changeset
+  alias Ecto.UUID
   alias NervesHubCore.Accounts.{Org, User, UserCertificate, Invite, OrgKey}
   alias NervesHubCore.Repo
   alias Comeonin.Bcrypt
@@ -12,6 +13,16 @@ defmodule NervesHubCore.Accounts do
     %Org{}
     |> Org.changeset(params)
     |> Repo.insert()
+  end
+
+  def change_user(user, params \\ %{})
+
+  def change_user(%User{id: nil} = user, params) do
+    User.creation_changeset(user, params)
+  end
+
+  def change_user(%User{} = org_key, params) do
+    User.update_changeset(org_key, params)
   end
 
   @doc """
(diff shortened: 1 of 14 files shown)


mobileoverlord:
One thing we have to consider as part of this is how to migrate the data that is already in production. Currently there is only one org `nerves-hub` but it has a couple users including the default user. We should probably delete the users and re create them. This should probably be done as a migration so it only happens once. Then we need to update the `seeds.exs`

jlmcgehee21:
hmm yeah. I can add that to the migration.

jlmcgehee21:
@mobileoverlord I'm definitely not an expert at data migrations in Ecto, and am not super in love with what I threw together, and feedback or total re-work is welcome.

apps/nerves_hub_core/priv/repo/migrations/20180810151416_default_user_org.exs
```diff
@@ -0,0 +1,34 @@
...
+        o in Org,
+        where: o.id not in ^new_org_ids,
+        preload: :users
+      )
+      |> Repo.all()
+      |> Enum.filter(fn o -> Enum.empty?(o.users) end)
+      |> Enum.map(fn o -> Repo.delete(o) end)
+  end
+
+  def down do
```
mobileoverlord: There is no going back :)

jlmcgehee21: Lol, yeah. Don't think there's a way to make that happen

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -89,7 +100,7 @@ defmodule NervesHubCore.Accounts do
   def create_user(%Org{} = org, params) do
```
mobileoverlord: We should make it that create_user essentially does create_user_with_org with the params set that it tries to create the default org. You should never be able to create a new user without creating an org by the same name

mobileoverlord:
I cleaned up a warning and rebased this on master. Make sure to pull

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
jlmcgehee21: @mobileoverlord I agree, but was going to save this (and some other desirable refactors) for when I added the `many-many` relationship between users/orgs.

mobileoverlord:
🚀 Lets do this!

PR #188, opened 2018-08-14
Add API for managing devices and device certs

mobileoverlord:
This PR adds the following endpoints

* post `/devices` - create a new device
* get `/devices/:identifier/certificates` - list all device certificates
* post `/devices/identifier/certificates` - create a new device certificate by supplying a certificate signing request.

While working through this PR it came to my attention that we would have issues with device topics. Before this PR, topic names were designated using `device:#{device.identifier}`. The problem with doing it with _just_ the identifier is that the identifier is only unique to the org and not to the database. Therefore, it would've been possible for devices with the same identifier from different orgs to end up on the same channel. Furthermore, I've identified that there was a dangling index from the `tenant_id` days that was allowing duplicate devices with the same identifier to be added for the same org. I've added a migration to fix this issue and re wrote the channel and socket logic to use `firmware_uuid` as a way to group the connections. In prep for allowing devices to auto configure, I've added `last_known_firmware_uuid` to the device model. In order to send messages to individual devices, we use socket ids. You can broadcast to `device_socket:#{device.id}` for direct messages.


apps/nerves_hub_core/priv/repo/migrations/20180813191326_fix_unique_index_devices_identifier.exs
```diff
@@ -0,0 +1,10 @@
+defmodule NervesHubCore.Repo.Migrations.FixUniqueIndexDevicesIdentifier do
+  use Ecto.Migration
+
+  def change do
+    rename(table(:devices), :org_id, to: :tenant_id)
+    drop(unique_index(:devices, [:tenant_id, :identifier], name: :devices_tenant_id_identifier_index))
+    rename(table(:devices), :tenant_id, to: :org_id)
```
jlmcgehee21: This happened in a previous migration: https://github.com/nerves-hub/nerves_hub_web/blob/6f27ba9139b910d3d252e26b630a9a06c107fc29/apps/nerves_hub_core/priv/repo/migrations/20180808143752_tenant_to_org.exs#L18

apps/nerves_hub_device/test/nerves_hub_device_web/websocket_test.exs
```diff
@@ -194,42 +159,45 @@ defmodule NervesHubDeviceWeb.WebsocketTest do
       )
     end
 
-    test "Cannot connect and authenticate to channel with non-existing serial" do
-      opts =
-        @ssl_socket_config
-        |> Keyword.put(:caller, self())
+    #   test "Cannot connect and authenticate to channel with non-existing serial" do
```
jlmcgehee21: Is this test supposed to be deleted or replaced with a test using a non-existing `firmware_uuid`?

apps/nerves_hub_core/priv/repo/migrations/20180814014545_add_last_known_firmware_uuid_to_device.exs
```diff
@@ -0,0 +1,10 @@
+defmodule NervesHubCore.Repo.Migrations.AddLastKnownFirmwareUuidToDevice do
+  use Ecto.Migration
+
+  def change do
+    alter table(:devices) do
+      add(:last_known_firmware_uuid, :binary_id, null: true)
```
jlmcgehee21: I don't really see the advantage here.  Why not just preload the `last_known_firmware` and use its `uuid` instead of duplicating the data (which allows someone to screw it up in the future).

mobileoverlord: This was to allow devices to connect with unknown firmware versions. There is other protection in place that would not allow this to happen today. It should be handled in another PR though, I'll go the route you mentioned of accessing it through the last_known_firmware.uuid

apps/nerves_hub_device/test/nerves_hub_device_web/websocket_test.exs
mobileoverlord: This test doesn't really work much anymore. I removed the secondary level of authorization in the channel because it is all happening at the socket now. This case covered when they would show up with a uuid that has no firmware record. Which I believe there is a test for.

mobileoverlord:
@jlmcgehee21 I pushed fixes that should address the notes above.

PR #209, opened 2018-08-17
Many to many for orgs and users

jlmcgehee21:
Why:

* This is a very standard use case for core contributors, and probably
others.

* Migration to handle the many to many relationship.
* Moving some of the org <> user creation constraints into the user schema.
* Adding some composable user query helpers.
* Creating an org with the same name as the user on account creation.
* Treating the earliest `inserted_at` org as the "default" org for a user.

### DIFF
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/user_controller.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/user_controller.ex
@@ -6,6 +6,11 @@ defmodule NervesHubAPIWeb.UserController do
 
   action_fallback(NervesHubAPIWeb.FallbackController)
 
+  defp whitelist(params, keys) do
+    keys
+    |> Enum.into(%{}, fn x -> {x, params[to_string(x)]} end)
+  end
+
   def me(%{assigns: %{user: user}} = conn, _params) do
     render(conn, "show.json", user: user)
   end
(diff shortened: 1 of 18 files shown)


jlmcgehee21:
I may do a bit more polishing, but I think this is functionally complete.

apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/user_controller.ex
```diff
@@ -6,17 +6,22 @@ defmodule NervesHubAPIWeb.UserController do
 
   action_fallback(NervesHubAPIWeb.FallbackController)
 
+  defp whitelist(params, keys) do
```
jlmcgehee21: My goal here is to avoid dealing with string keys in `NervesHubCore`.  I like the idea that external applications should be responsible with sanitizing their own data before they send it to `NervesHubCore`.

jlmcgehee21:
@mobileoverlord I'm pretty good with where this is sitting now

apps/nerves_hub_core/priv/repo/migrations/20180813204454_many_orgs_many_users.exs
```diff
@@ -0,0 +1,16 @@
+defmodule NervesHubCore.Repo.Migrations.ManyOrgsManyUsers do
+  use Ecto.Migration
+
+  def change do
+    create table(:users_orgs) do
+      add(:user_id, references(:users, on_delete: :delete_all))
+      add(:org_id, references(:orgs, on_delete: :delete_all))
```
mobileoverlord: I am concerned over the `delete_all` option here. Does this mean that if a user is removed from an org by deleting the record that links the user to that org, the delete will cascade and delete the org too?

PR #220, opened 2018-08-23
add font-awesome and haml rendering

stwf:
Added the haml renderer which was discussed, and I took the liberty of adding font-awesome, which provides lots of useful icons and is open source. No UI fixes but this will pave the way for the rest.


apps/nerves_hub_www/assets/package-lock.json
```diff
@@ -20,10 +20,10 @@
       "integrity": "sha512-hOs7GfvI6tUI1LfZddH82ky6mOMyTuY0mk7kE2pWpmhhUSkumzaTO5vbVwij39MdwPQWCV4Zv57Eo06NtL/GVA==",
       "dev": true,
       "requires": {
-        "fast-deep-equal": "2.0.1",
-        "fast-json-stable-stringify": "2.0.0",
-        "json-schema-traverse": "0.4.1",
-        "uri-js": "4.2.2"
+        "fast-deep-equal": "^2.0.1",
```
stwf: I couldn't get npm to not change the versions to the general "^2.0.1" from the specific "2.0.1" for example. ??

apps/nerves_hub_www/assets/webpack.config.js
```diff
@@ -69,7 +69,7 @@ var common = {
       },
       {
         test: /\.(png|jpg|gif|svg)$/,
-        loader: "file-loader?name=../priv/static/images/[name].[ext]"
+        loader: "file-loader?name=/images/[name].[ext]"
```
fhunleth: Doesn't removing `../priv/static` break this?

stwf: no I think its broken the way it is. Without the change images get put inside priv/priv/static/images, where they won't get picked up by the browser.

PR #221, opened 2018-08-23
Add clarity around addition/subtraction of user-orgs

jlmcgehee21:
Why:

* Calling "update" on a user while passing `%{orgs: [org1, org2]}` is
confusing and requires a lot of work _outside_ of the `nerves_hub_core`
application in order to ensure the desired effect occurs.

* I think there were some bugs (of no consequence elsewhere in the outside application yet) around updates

* adding `add_user_to_org` and `remove_user_from_org` functions which
clarify what's going on.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
@@ -341,6 +341,24 @@ defmodule NervesHubCore.Accounts do
     |> Repo.update()
   end
 
+  def add_user_to_org(%User{} = user, %Org{} = org) do
+    all_orgs = user |> User.with_all_orgs() |> Map.get(:orgs, [])
+    params = %{orgs: all_orgs ++ [org]}
+
+    user
+    |> User.update_orgs_changeset(params)
+    |> Repo.update()
+  end
+
+  def remove_user_from_org(%User{} = user, %Org{} = org) do
+    all_orgs = user |> User.with_all_orgs() |> Map.get(:orgs, [])
+    params = %{orgs: all_orgs -- [org]}
+
+    user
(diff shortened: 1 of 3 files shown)


apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -341,6 +341,24 @@ defmodule NervesHubCore.Accounts do
     |> Repo.update()
   end
 
+  def add_user_to_org(%User{} = user, %Org{} = org) do
+    all_orgs = user |> User.with_all_orgs() |> Map.get(:orgs, [])
+    params = %{orgs: all_orgs ++ [org]}
```
mobileoverlord: I don't think ordering matters, we could add it to the head of the list. `[org | all_orgs]`

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -341,6 +341,24 @@ defmodule NervesHubCore.Accounts do
...
+
+    user
+    |> User.update_orgs_changeset(params)
+    |> Repo.update()
+  end
+
+  def remove_user_from_org(%User{} = user, %Org{} = org) do
+    all_orgs = user |> User.with_all_orgs() |> Map.get(:orgs, [])
```
mobileoverlord: You could use `{_, all_orgs} = Enum.split_with/2` here to pop the org out.

mobileoverlord:
I am being nit picky about style, but I try to stay reserve the use of `++` / `--` for lists if possible. If a user does have a lot of orgs it could have a long list to walk.

jlmcgehee21:
That's good feedback. I'm not super familiar with a lot of the proper idioms for these types of operations in Elixir 👍

PR #223, opened 2018-08-24
add font-awesome

stwf:


apps/nerves_hub_www/assets/webpack.config.js
```diff
@@ -69,7 +69,7 @@ var common = {
       },
       {
         test: /\.(png|jpg|gif|svg)$/,
-        loader: "file-loader?name=../priv/static/images/[name].[ext]"
```
stwf: I think this is the right path. If you clear out the /priv/static folder and restart the app images are copied into priv/priv/static/images. This change puts them into the right folder.

Issue #217, opened 2018-08-22
Limit number of provisioned devices per org


mobileoverlord:
We should limit this to 5 devices per org. The setting should be configurable. As part of implementing this, we need to allow devices to be deleted.

jlmcgehee21:
@mobileoverlord, would you want this config to be at the umbrella level or at the `nerves_hub_core` level?

mobileoverlord:
Yeah we should probably handle it there for the API's sake since it does not have www as a dependency

PR #227, opened 2018-08-24
Delete devices via browser and API

jlmcgehee21:
Why:

* If we want to limit devices per org/product, we need to allow users to
delete devices: https://github.com/nerves-hub/nerves_hub_web/issues/217

* routes, controller, and context functions.
* tests to verify behavior


mobileoverlord:
Can we verify that when we delete a device that it also deletes that device's `device_certificates`

jlmcgehee21:
Yeah. Will add a test for that

jlmcgehee21:
@mobileoverlord that was a good catch, due to the initial setup of the foreign key constraint, it was actually impossible to delete a device that had any certs associated.

PR #233, opened 2018-08-26
Update UI for Product index page

stwf:
Also upgraded font awesome and fix ed issues with the footer not sticking to the page bottom


stwf:
screen shot 2018-08-25 at 9 43 10 pm

fhunleth:
Looks nice.

Let me make sure that I understand the icons.
The + icon is to create a new product.
The trashcan is to delete a product.
Are the other two "show" and "edit"?

stwf:
They are for firmware, and deployments. There is a tooltip hover thing explaining that, but I know its not the best. I can see the case for little text buttons for that too. Just figured I'd start here, so opinions welcome.

Also different icon choices:
https://fontawesome.com/icons?d=gallery&m=free

stwf:
Clicking on the name, it's a link is the show, and you said to get rid if the edit.

fhunleth:
Sounds good to me. I don't have any better ideas right now for the icons.

Issue #208, opened 2018-08-17
Add missing fields to deployments detail page


fhunleth:
Here's the current state:

![image](https://user-images.githubusercontent.com/64669/44276520-b33e0800-a215-11e8-96e5-941f814ee7b1.png)

It's missing the version and conditions. It could also use a little tidying up.

mobileoverlord:
merged in master

PR #246, opened 2018-09-05
Fix deployment show page

jlmcgehee21:
Why:

* Wasn't showing all the info, and it looked weird
* https://github.com/nerves-hub/nerves_hub_web/issues/208

* Making it show info and not look weird


jlmcgehee21:
screen shot 2018-09-04 at 6 57 20 pm

Issue #242, opened 2018-09-03
Convert Accounts.User :name to :username


mobileoverlord:
`name` is being used to also define the users default `org` `name` and therefore it has to be unique. It would be a little weird to ask someone to enter there name only to find out that it has been taken. This kind of behavior is more likely when it comes to usernames. We would like to do the following

* Change `name` to `username`.
* Validate that the chosen username contains only URL safe characters.
* Enforce that the username is unique.

PR #253, opened 2018-09-06
Rename name on user to username

danielspofford:
Why
---

* Addresses #242

How
---

* User schema name is now username, is unique, and can only contain url
safe characters
* Invites no longer dictate username, it can be choosen on invite
acceptance
* Update fixtures, seeds, controllers, views, and tests to respect
changes

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
@@ -194,12 +194,12 @@ defmodule NervesHubCore.Accounts do
 
   @spec get_org(integer()) ::
           {:ok, Org.t()}
-          | {:error, :not_found}
+          | {:error, :org_not_found}
   def get_org(id) do
     Org
     |> Repo.get(id)
     |> case do
-      nil -> {:error, :not_found}
+      nil -> {:error, :org_not_found}
       org -> {:ok, org}
     end
   end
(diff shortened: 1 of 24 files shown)


danielspofford:
Just pushing this up, I need to take a final look at it.

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -309,25 +309,24 @@ defmodule NervesHubCore.Accounts do
     query
     |> Repo.one()
     |> case do
-      nil -> {:error, :not_found}
+      nil -> {:error, :invite_not_found}
```
jlmcgehee21: 👍

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -309,25 +309,24 @@ defmodule NervesHubCore.Accounts do
...
+    user_org = %{name: user_params.username}
 
     Repo.transaction(fn ->
-      with {:ok, user} <- create_user(%{orgs: [org]} |> Enum.into(user_params)),
+      with {:ok, user} <- create_user(%{orgs: [user_org, org]} |> Enum.into(user_params)),
            {:ok, _invite} <- set_invite_accepted(invite) do
         {:ok, user}
       else
-        _ -> {:error}
+        {:error, error} -> Repo.rollback(error)
```
jlmcgehee21: Good catch

danielspofford:
The migration included in this PR renames a `%User{}`s name to username and enforces a uniqueness constraint. This will fail on migration in deploy if prod has data that violates this.

There is also a changeset validation introduced to ensure usernames only contain url safe characters. It is possible that usernames exist currently that don't respect this, however, this won't crop up in deploy, only when updating those users.

mobileoverlord:
❤️ Thanks @danielspofford

PR #256, opened 2018-09-06
Add styles for the devices table.

squaresurf:
Why:

* The table did not match other pages.

* Copy the styles from the deployments table.
* Commit .python-version file for those using pyenv (python version is
  needed for node-sass).


mobileoverlord:
Thanks :bowtie:

PR #259, opened 2018-09-07
Eliminate Static Firmware Fixtures

jlmcgehee21:
Why:

* The small set of static firmware fixtures limited our ability to test
well.

* `NervesHubCore.Support.Fwup` which wraps `fwup` and gives elixir-y
programmatic access to `fwup` for development and testing purposes.
* Converting all tests away from using the static files.

Other notes:

* There is still a lot of room for improving the way firmware is being handled, but this is a nice step in the right direction.


mobileoverlord:
I am fiddling with rebasing `eliminate_static_firmware_fixtures` and I am getting a handful of errors when running. Looks like when we rolled the test certs it caused issues? not sure. Also, Since a lot of the sub projects redefine the `FWUP` modules, can you follow the pattern that we defined in https://github.com/nerves-hub/nerves_hub_web/blob/master/test/support/fixtures.ex#L1-L8

That will silence the warnings for 
```
warning: redefining module NervesHubCore.Support.Fwup (current version loaded from /Users/jschneck/Developer/nerves-hub/nerves_hub_web/_build/test/lib/nerves_hub_core/ebin/Elixir.NervesHubCore.Support.Fwup.beam)
  /Users/jschneck/Developer/nerves-hub/nerves_hub_web/test/support/fwup.ex:1
```

jlmcgehee21:
I can definitely make the fix for the warnings @mobileoverlord -- I'm not sure what's up with the errors your seeing. The fwup wrapper is pretty dumb and the async tests may be messing it up if you already got that in.

mobileoverlord:
@jlmcgehee21 Yup, async got merged down

jlmcgehee21:
K, @mobileoverlord, I'll get my local sync'ed up and see if I can repro locally.  If I can, then I should be able to handle the fix myself.

jlmcgehee21:
@mobileoverlord rebased and addressed failures. Again... 😜

mobileoverlord:
Looks good, and nobody else submit anything between an hour ago and now, so, I guess its time to merge it. 😄

PR #261, opened 2018-09-14
Enable real time device status updates

danielspofford:
Why
---

- So there is a way to hook into an organization's devices' status
lifecycles

How
---

- When a device connects to a `"firmware:#{fw_uuid}"` topic, it
`Presence.track/4`s itself on a `"devices:#{org_id}"` topic
- When a browser connects to a `devices:#{org_id}` topic, it will
receive presence state and diffs regarding all devices within that org

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
@@ -400,4 +400,17 @@ defmodule NervesHubCore.Accounts do
         {:error, :not_found}
     end
   end
+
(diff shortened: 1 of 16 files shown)


mobileoverlord:
I just tested this locally and its working. We may want to trim down the timeout times from either the socket disconnecting, or the presence sync. It appears `online` much faster then it will appear `offline`. I had to power the unit down and wait because it didn't seem to timeout faster then it took the device to reboot. Which isn't a terrible problem to have, but, we can do better :). Lets tackle this in another issue.

mobileoverlord:
❤️ Thanks @danielspofford

PR #266, opened 2018-09-19
Create org_limits

mobileoverlord:
Currently, firmware upload size is limited by the number defined for `Plug.Parsers` multipart upload byte limit. This is set currently to roughly 160 MB. Some organizations are going to require the ability to upload firmware that is larger than this limit. To accommodate for this I had to increase the overall limit that is enforced in the endpoint by `Plug.Parsers`. I set this value to 1 GB. The second enforcement now happens in the controller when reading the body. It will look up the limits from the `org_limits` table, if there are none, it will pass back the defaults which are set to 160 MB.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
@@ -2,7 +2,7 @@ defmodule NervesHubCore.Accounts do
   import Ecto.Query
   alias Ecto.Changeset
   alias Ecto.UUID
-  alias NervesHubCore.Accounts.{Org, User, UserCertificate, Invite, OrgKey}
+  alias NervesHubCore.Accounts.{Org, User, UserCertificate, Invite, OrgKey, OrgLimit}
   alias NervesHubCore.Repo
   alias Comeonin.Bcrypt
 
@@ -13,6 +13,34 @@ defmodule NervesHubCore.Accounts do
     %Org{}
     |> Org.creation_changeset(params)
     |> Repo.insert()
+  end
+
+  def create_org_limit(params) do
+    %OrgLimit{}
+    |> OrgLimit.changeset(params)
+    |> Repo.insert()
+  end
+
+  def delete_org_limit(%OrgLimit{} = org_limit) do
+    org_limit
+    |> Repo.delete()
+  end
+
+  def update_org_limit(org_limit, params) do
+    org_limit
+    |> OrgLimit.update_changeset(params)
+    |> Repo.update()
+  end
+
+  def get_org_limit_by_org_id(org_id) do
+    query = from(ol in OrgLimit, where: ol.org_id == ^org_id)
+
+    query
+    |> Repo.one()
+    |> case do
+      nil -> {:ok, %OrgLimit{}}
+      org_limit -> {:ok, org_limit}
+    end
   end
 
   def change_user(user, params \\ %{})
--- a/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
@@ -1,6 +1,7 @@
 defmodule NervesHubCore.Firmwares do
   import Ecto.Query
 
+  alias NervesHubCore.Accounts
   alias NervesHubCore.Accounts.{OrgKey, Org}
   alias NervesHubCore.Firmwares.Firmware
   alias NervesHubCore.Products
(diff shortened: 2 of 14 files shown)


jlmcgehee21:
I like this general direction, but won't be able to look over it entirely until this evening.

mobileoverlord:
I view this as a patch to let us customize the org limits until we introduce proper billing and plans.

fhunleth:
I'm :+1: on this feature. The backstory is that Le Tote needs this for their x86 kiosk systems. It would be great to shrink the kiosk size down and I suspect some significant savings can be had, but I can't help out right now. This gets them over the hump.

I'm going to approve so the fix isn't held up, but hopefully Jeff can get a chance to look at it too.

mobileoverlord:
Yeah, I think this is flexible enough to also accommodate overriding the number of firmware and number of device restrictions we globally have now as well.

apps/nerves_hub_core/lib/nerves_hub_core/accounts.ex
```diff
@@ -15,6 +15,34 @@ defmodule NervesHubCore.Accounts do
     |> Repo.insert()
   end
```
jlmcgehee21: This `accounts` context is getting a bit long in the tooth. Might be worth considering breaking it into a few different pieces in the near future.

apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
```diff
@@ -70,35 +71,13 @@ defmodule NervesHubCore.Firmwares do
           | {:error, :invalid_signature}
           | {:error, any}
   def prepare_firmware_params(%Org{} = org, filepath) do
-    org = NervesHubCore.Repo.preload(org, :org_keys)
+    byte_size = :filelib.file_size(filepath)
+    {:ok, %{firmware_size: firmware_size_limit}} = Accounts.get_org_limit_by_org_id(org.id)
 
-    with {:ok, %{id: org_key_id}} <- verify_signature(filepath, org.org_keys),
```
jlmcgehee21: Got excited this giantness was gone....

apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
```diff
@@ -181,4 +160,37 @@ defmodule NervesHubCore.Firmwares do
       {:error}
     end
   end
+
```
jlmcgehee21: Then saw this... lol

apps/nerves_hub_core/priv/repo/migrations/20180919161402_create_org_limits_table.exs
```diff
@@ -0,0 +1,14 @@
...
+
+  def change do
+    create table(:org_limits) do
+      add(:org_id, references(:orgs, null: false))
+      add(:firmware_size, :integer, null: false)
+
+      timestamps()
+    end
+
+    create(index(:org_limits, [:org_id]))
```
jlmcgehee21: Does this need to be a `unique_index`? I see the `has_one` and `unique_constraint` in the schema.

jlmcgehee21: fwiw this is truly a question on my part, I'm not totally sure on the dynamics of enforcing uniqueness in the schema vs at the db level with ecto, but it does seem to me like you would want both unless I'm missing something.

mobileoverlord: I was actually curious about this myself, and at first, I had a unique index in there, but it complained that there was already one inserted. I guess the relationship definition takes care of adding on for you.

apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
mobileoverlord: 😄

apps/nerves_hub_core/priv/repo/migrations/20180919161402_create_org_limits_table.exs
jlmcgehee21: :+1:

danielspofford: @mobileoverlord @jlmcgehee21

If you create the unique index in the DB, and do nothing regarding that in the schema's changesets, then when you violate the uniqueness an exception will be raised with no opportunity to pattern match on an `{:error, changeset}`. If you `unique_constraint` in the schema's changeset but do not create a unique index in the DB, then uniqueness will not be enforced. If you do both, uniqueness will be enforced and rather than raising on a violation, you will get a changeset error back. I would assume Justin's issues stemmed from the DB being in an unexpected state.

**More importantly**: If you want to enforce that column being unique, you definitely would have wanted that line to be `unique_index`. Right now you should be able to insert 2 rows into `:org_limits` which have the same `org_id` value.

jlmcgehee21: That was my hunch...

Issue #10, opened 2018-02-06
Deployment View


mattlorey:
Detail view of a deployment

Acceptance Criteria
- Logged in user should only be able to see deployments for their own tenant
- Show name, criteria and status of deployment
- Ability to toggle the status of a deployment between inactive and active

Technical Acceptance Criteria
- Change `status` to a boolean `is_active` field

PR #272, opened 2018-09-21
Add richer seeds for :dev env

jlmcgehee21:
Why:

* It's useful to see data in the web views to get a feel for how the app
will look/feel.
* I was trying to determine if
https://github.com/nerves-hub/nerves_hub_web/issues/10 was able to be
marked as closed, and realized that I couldn't tell without manually
creating a bunch of stuff.

* Adding a user with two orgs where each org has multiple devices,
firmwares and deployments in `seeds.exs` when `Mix.env() == :dev`


mobileoverlord:
Yup, I was just in need of this myself

PR #273, opened 2018-09-22
Fix device status on load is always "online"

mobileoverlord:
When the device page loads all device statuses load initially as "online" and then update quickly to "offline". This is due to presence not handling the nil case for when a device is missing from presence completely. This adds the missing case. I am unsure how to add a test without making a decision on a front end testing library.


danielspofford:
I think this could be tested service side, just test rendering the view, so no front end decision making would be required. That being said the fix LGTM.

mobileoverlord:
@danielspofford You test shamed me into at least checking. I validated that this test failed under the non patched conditions.

PR #274, opened 2018-09-22
Transactional firmware creation and deletion

### LINKED ISSUE #258: Firmware Zombies
Firmware files are persisted to the file store _before_ the record is added to the database.  If we don't have a transactional guarantee and rollback strategy in place, a failed DB insert will yield a zombie firmware file.

`prepare_firmware_params` (https://github.com/nerves-hub/nerves_hub_web/blob/master/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex#L72) is called in the controllers _before_ `create_firmware`.

danielspofford:
Resolves #258

What
----

- When creating or deleting firmware files always work inside a
transaction and act on the database record before the file persistance
to ensure a persisted file (remote or local) will not be dereferenced
but still exist

How
---

- `NervesHubCore.Firmwares`
  - Remove `prepare_firmware_params/2` and roll functionality into
  `create_firmware/3` to publicly ensure an atomic operation
  - Add `metadata/2` to `Upload.S3` and `Upload.File` modules so that
  information about about where a firmware file will be persisted to is
  available before actually persisting it
- `NervesHubCore.Fixtures`
  - Split firmware file creation into its own fixture apart from
  firmware creation within the application

---

Not overjoyed with https://github.com/nerves-hub/nerves_hub_web/pull/274/files#diff-d43b46de1cbc8975ae36bc46322877beR37. It is using dependency injection to allow me to ensure the DB insertion succeeds but the file persistence fails. In similar cases previously I have used https://github.com/plataformatec/mox. That way we could have an `Upload` behavior and use mocks in test to force it to behave how we wish. That being said, I decided to go the easy route and not force a dependency decision in front of this PR unless discussion justifies it.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
@@ -1,11 +1,15 @@
 defmodule NervesHubCore.Firmwares do
   import Ecto.Query
-
   alias NervesHubCore.Accounts
+  alias Ecto.Changeset
   alias NervesHubCore.Accounts.{OrgKey, Org}
   alias NervesHubCore.Firmwares.Firmware
   alias NervesHubCore.Products
   alias NervesHubCore.Repo
+
+  @type upload_file_2 :: (filepath :: String.t(), filename :: String.t() -> :ok | {:error, any()})
+
+  @uploader Application.fetch_env!(:nerves_hub_core, :firmware_upload)
 
   @spec get_firmwares_by_product(integer()) :: [Firmware.t()]
   def get_firmwares_by_product(product_id) do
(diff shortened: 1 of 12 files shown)


apps/nerves_hub_core/lib/nerves_hub_core/firmwares.ex
```diff
@@ -153,18 +140,18 @@ defmodule NervesHubCore.Firmwares do
...
   end
 
-  def upload_firmware(filepath, filename, org_id) do
-    if uploader = Application.get_env(:nerves_hub_core, :firmware_upload) do
-      uploader.upload_file(filepath, filename, org_id)
-    else
-      {:error}
-    end
+  @spec insert_firmware(map()) :: {:ok, Firmware.t()} | {:error, Changeset.t()}
+  def insert_firmware(params) do
```
jlmcgehee21: Really feel like this should be a private function. Can't think of any reason to insert a firmware if the conditions in create_firware are not satisfied.

danielspofford: Agreed

apps/nerves_hub_core/test/nerves_hub_core/firmwares/firmwares_test.exs
```diff
@@ -51,7 +50,7 @@ defmodule NervesHubCore.FirmwaresTest do
       }
 
       assert {:error, %Ecto.Changeset{errors: [uuid: {"has already been taken", []}]}} =
-               Firmwares.create_firmware(new_params)
+               %Firmware{} |> Firmware.changeset(new_params) |> Repo.insert()
```
jlmcgehee21: I'm not totally against this, but why couldn't you test this side effect of create_firmware?

danielspofford: Certainly could, and now we do!

PR #275, opened 2018-09-23
Only open socket on devices page

### LINKED ISSUE #268: Do not connect to channels on logged out pages
When visiting pages that do not require a log in, the channel connect is still attempted resulting in a lot of errors in the console. We should limit the channel connection to only the pages that require it. For now I think thats just the device pages.

danielspofford:
Resolves #268 

Why
---

- Connecting to this socket will fail on non-authed / non-devices pages flooding
client and server logs with failures

How
---

- Only `require` the `socket.js` code when on the devices page

---

This is a somewhat naive approach but gets the job done now while dodging
heavier front end efforts. Open to discussion about alternate approaches.

### DIFF
--- a/apps/nerves_hub_www/assets/js/app.js
+++ b/apps/nerves_hub_www/assets/js/app.js
@@ -1,3 +1,6 @@
 import "phoenix_html"
 import 'bootstrap';
-import socket from "./socket"
+
+if(window.location.pathname === '/devices') {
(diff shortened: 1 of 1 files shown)


danielspofford:
Tests failed to deadlock that I believe is resolved in #274

mobileoverlord:
I am good with this for now.

PR #281, opened 2018-09-27
Fix firmware delete on S3

mobileoverlord:
Deleting firmware from S3 is currently raising. It turns out that in the past, it was just preparing a transaction and not applying the request. This fixes both the return pattern match as well as actually deleting the files.


mobileoverlord:
I totally agree.

PR #283, opened 2018-09-27
Add button to delete firmware

ConnorRigby:
Only being able to delete firmware from the command-line was sort of annoying

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/firmware_controller.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/firmware_controller.ex
@@ -87,4 +87,13 @@ defmodule NervesHubWWWWeb.FirmwareController do
       {:error}
     end
   end
+
+  def delete(%{assigns: %{current_org: org, product: product}} = conn, %{"id" => id}) do
+    with {:ok, firmware} <- Firmwares.get_firmware(org, id),
+         :ok <- Firmwares.delete_firmware(firmware) do
+      conn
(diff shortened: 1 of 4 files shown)


ConnorRigby:
@mobileoverlord I've added some error checking to the controller. I only used a simple `put_flash` to display a generic error.

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/firmware_controller.ex
```diff
@@ -87,4 +87,13 @@ defmodule NervesHubWWWWeb.FirmwareController do
       {:error}
     end
   end
+
+  def delete(%{assigns: %{current_org: org, product: product}} = conn, %{"id" => id}) do
+    {:ok, firmware} = Firmwares.get_firmware(org, id)
+    :ok = Firmwares.delete_firmware(firmware)
```
mobileoverlord: The assumption of `:ok` will raise if things are not okay. When I run this locally, any firmware that has a constraint has a constraint error that is raised. We need to add the proper validation to the ecto changeset for the invalidating constraints so they are returned as ecto errors and we need this to capture those errors and put them in the flash when they happen.

fhunleth:
@ConnorRigby Would you be able to get back to this? This seems useful and I'm hoping that the conflicts aren't too bad.

ConnorRigby:
Yeah i think there were some changes that will make this code much easier to implement. Will take a look today.

ConnorRigby:
I've rebased and updated this feature. I think it should be ready now;

CC: @fhunleth @mobileoverlord

PR #284, opened 2018-10-01
Firmware TTL

mobileoverlord:
Uploading test firmware from CI can quickly result in clutter and should be garbage collected. This PR introduces a "time to live" setting that is applied to firmware records upon creation. TTL is stored in seconds and represents the time to keep firmware _after_ all its associations have been removed. This is tracked by setting the `tty_until` `utc_datetime` stamp on the firmware record whenever is has been removed from all associations. If an association is made again, the timestamp is removed and it will not get garbage collected. The default `ttl` is set for 7 days and can be configured per org using `org_limits`.

As a side effect, the API needed to change how it accepts the firmware upload from the CLI in order to accept both a file and additional params like `ttl`. This way, CI can configure the test firmware to get garbage collected sooner than other kinds of firmware. Merging this will break existing `nerves_hub_cli` users who are not using the `firmware-ttl` branch.

### DIFF
--- a/apps/nerves_hub_core/lib/nerves_hub_core/accounts/org_limit.ex
+++ b/apps/nerves_hub_core/lib/nerves_hub_core/accounts/org_limit.ex
@@ -9,18 +9,44 @@ defmodule NervesHubCore.Accounts.OrgLimit do
   @type t :: %__MODULE__{}
 
   @required_params [:org_id]
-  @optional_params [:firmware_size, :firmware_per_product, :devices]
+  @optional_params [
+    :devices,
+    :firmware_per_product,
+    :firmware_size,
+    :firmware_ttl_seconds_default,
+    :firmware_ttl_seconds
+  ]
+
+  @defaults [
+    # Max number of devices per org
+    devices: 5,
+    # Max number of firmwares per product
+    firmware_per_product: 5,
+    # Max firmware size (160 Mb)
+    firmware_size: 167_772_160,
+    # Default firmwre ttl seconds (7 days)
+    firmware_ttl_seconds_default: 604_800,
+    # Max firmwre ttl seconds (7 days)
+    firmware_ttl_seconds: 604_800
+  ]
 
   schema "org_limits" do
     belongs_to(:org, Org)
 
-    # 160 Mb
-    field(:firmware_size, :integer, default: 167_772_160)
-    field(:firmware_per_product, :integer, default: 5)
-    field(:devices, :integer, default: 5)
+    field(:devices, :integer, default: @defaults[:devices])
+    field(:firmware_per_product, :integer, default: @defaults[:firmware_per_product])
+    field(:firmware_size, :integer, default: @defaults[:firmware_size])
+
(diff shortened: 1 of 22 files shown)


fhunleth:
I'm :+1: on this feature. This will be very useful.

apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
```diff
@@ -130,10 +130,26 @@ defmodule NervesHubCore.Devices do
...
+      error ->
+        error
+    end
   end
 
   def delete_device(%Device{} = device) do
     Repo.delete(device)
+    |> case do
+      {:ok, device} ->
+        Firmwares.update_firmware_ttl(device.last_known_firmware_id)
```
jlmcgehee21: There are a lot of separated calls to `Firmwares.update_firmware_ttl`, and each placement is critical for proper functionality.  I _think_ they are all covered by tests, so it's good to ship for now IMO... However, for ease of understanding for new contributors (as well as our future selves), a slight change in this implementation may be in order in the future. Maybe move to the changeset of each schema, or maybe some larger reaching filter that updates ttl conditionally, and doesn't spread out the logic controlling which records require a ttl update.  I'm saying all this with just having spend the last 10 minutes thinking about the implementation of this feature, so take it with a grain of salt.

apps/nerves_hub_core/lib/nerves_hub_core/accounts/org_limit.ex
```diff
@@ -9,19 +9,44 @@ defmodule NervesHubCore.Accounts.OrgLimit do
...
+    :firmware_ttl_seconds_default,
+    :firmware_ttl_seconds,
+    :firmware_ttl_seconds_max
+  ]
+
+  @defaults [
+    devices: 5,
+    firmware_per_product: 5,
+    firmware_size: 167_772_160,
+    firmware_ttl_seconds_default: 604_800,
```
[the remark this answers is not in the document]
mobileoverlord: Thanks for pointing this out, `firmware_ttl_seconds_max` was unused and unnecessary.

apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
mobileoverlord: I initially attempted this through the schema validations, but it brought on problems that you'd expect from using changesets as after insert callbacks. It was riddled with conditions where it was hard to explicitly unset the ttl. I was thinking that there might be some benefit derived through the use of `Ecto.Multi` but I was unable to produce anything notable.

jlmcgehee21: Yeah, I don't think it should hold anything up -- but as the application and contributor list grows, it will be good to keep an eye on this as a place for potential optimization of code friendliness.  Should I make an issue so we don't lose track of this?

mobileoverlord: I opened an issue [here](https://github.com/nerves-hub/nerves_hub_web/issues/285) to track

PR #287, opened 2018-10-02
Add endpoints for a device to consume.

ConnorRigby:
This adds the ability to ask an HTTP endpoint if an update is available.


mobileoverlord:
Looks good, I like that you are using the same join reply from the channel infrastructure.

PR #288, opened 2018-10-03
Ugly code snippet on front page

gemantzu:
You don't have to accept this, it is just a suggestion, do your thing on this, I just thought it might help. Be aware it might affect `templates/firmware/index.html.eex` which also has a `pre` tag.

### DIFF
--- a/apps/nerves_hub_www/assets/css/app.scss
+++ b/apps/nerves_hub_www/assets/css/app.scss
@@ -125,3 +125,6 @@ main {
   }
 }
 
+pre {
+  overflow: hidden;
+}


jlmcgehee21:
@gemantzu -- thanks, I'll give it look to check for side effects.  On the surface, if you don't mind, it might be better to make a new css class for the code snippet(s)

gemantzu:
I would do it that way, my only point is that it should be fixed, that's all :) I just couldn't test it at the time as I was @ work.

danielspofford:
Closing in favor of #343.

Thanks @gemantzu for calling this out 💖!

PR #290, opened 2018-10-04
Capture additional certificate information required for revocation

mobileoverlord:
In preparation for providing our own certificate revocation lists, we need to begin to track the `authority_key_id` from the certificate in addition to the `serial_number`.

### DIFF
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/device_certificate_controller.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/device_certificate_controller.ex
@@ -16,8 +16,15 @@ defmodule NervesHubAPIWeb.DeviceCertificateController do
     with {:ok, device} <- Devices.get_device_by_identifier(org, identifier),
          {:ok, %{"cert" => cert}} <- CertificateAuthority.sign_device_csr(csr),
          {:ok, serial} <- Certificate.get_serial_number(cert),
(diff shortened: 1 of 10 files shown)


fhunleth:
I approve (not surprisingly).

Out of curiosity, why don't we save the cert to the db rather than pluck fields out of it?

mobileoverlord:
I’m trying to split revocation into digestible chunks. In a follow up PR, the CA will begin to track all certificates in its own DB, we just need these two keys to reference it.

apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/device_certificate_controller.ex
```diff
@@ -16,8 +16,15 @@ defmodule NervesHubAPIWeb.DeviceCertificateController do
     with {:ok, device} <- Devices.get_device_by_identifier(org, identifier),
          {:ok, %{"cert" => cert}} <- CertificateAuthority.sign_device_csr(csr),
          {:ok, serial} <- Certificate.get_serial_number(cert),
-         {not_before, not_after} <- Certificate.get_validity(cert),
-         params <- %{serial: serial, not_before: not_before, not_after: not_after},
+         {:ok, authority_key_id} <- Certificate.get_authority_key_id(cert),
```
jlmcgehee21: On the surface, it looks like you could move this and the same logic from the api `user_controller` to a single function in `NervesHubCore.Certificate`.

PR #293, opened 2018-10-12
Fix wrong path for local firmware downloads

### LINKED ISSUE #292: Local Firmware downloading not working?
```
%NervesHubCore.Firmwares.Firmware{
    __meta__: #Ecto.Schema.Metadata<:loaded, "firmwares">,
    architecture: "arm",
    author: "The Farmbot Team",
    deployments: #Ecto.Association.NotLoaded<association :deployments is not loaded>,
    description: "The Brains of the Farmbot Project",
    id: 20,
    inserted_at: ~N[2018-10-12 17:46:05.998800],
    misc: nil,
    org_id: nil,
    org_key: #Ecto.Association.NotLoaded<association :org_key is not loaded>,
    org_key_id: 7,
    platform: "rpi3",
    product: #Ecto.Association.NotLoaded<association :product is not loaded>,
    product_id: 5,
    size: nil,
    ttl: 604800,
    ttl_until: nil,
    updated_at: ~N[2018-10-12 17:46:06.072889],
    upload_metadata: %{
      "local_path" => "/tmp/3/7c8588d2-3e02-5406-a7af-7947da326f66.fw",
      "public_path" => "http://nerves-hub.org:4000/firmware/3/7c8588d2-3e02-5406-a7af-7947da326f66.fw"
    },
    uuid: "7c8588d2-3e02-5406-a7af-7947da326f66",
    vcs_identifier: nil,
    version: "6.4.12"
  }
```

My `public_path` is all sorts of weird. I cant find a `route` in any of the routers that would match that. This may be a local configuration error, in which case this step is undocumented

ConnorRigby:
This was a typo from a [previous
commit](https://github.com/nerves-hub/nerves_hub_web/commit/8669c534c5bb2d880a20e8ca5ffad7f95b096f48#diff-3ee9958556f69d386641ab916f438305R6)

Fixes  #292

### DIFF
--- a/apps/nerves_hub_core/config/dev.exs
+++ b/apps/nerves_hub_core/config/dev.exs
@@ -3,7 +3,7 @@ use Mix.Config
 config :nerves_hub_core, firmware_upload: NervesHubCore.Firmwares.Upload.File
 
 config :nerves_hub_core, NervesHubCore.Firmwares.Upload.File,
-  local_path: System.tmp_dir(),
+  local_path: Path.join(System.tmp_dir(), "firmware"),
   public_path: "/firmware"
 
 # config :nerves_hub_core, NervesHubCore.Firmwares.Upload.S3, bucket: System.get_env("S3_BUCKET_NAME")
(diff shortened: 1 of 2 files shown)


jlmcgehee21:
I feel pretty strongly that changing this config value should _not_ break the code.  I think the best solution is to change https://github.com/nerves-hub/nerves_hub_web/blame/master/apps/nerves_hub_www/lib/nerves_hub_www_web/endpoint.ex#L24 to load the paths from the config instead.

jlmcgehee21:
I'm willing to be convinced otherwise, and/or let someone else approve this change though.

ConnorRigby:
I think it's a good idea also. Updated so one file cant break the other

PR #303, opened 2018-11-30
Org supplied ca signing certs and auto-registration

mobileoverlord:
This PR begins to add support for orgs to supply their own ca signing certs for device certificate validation. Currently, all device certificates are validated against the NervesHub CA path. With support for custom certs, organizations can supply their own ca certs and NervesHub will use them for ssl peer validation.

This also adds the capability to auto register the first certificate a device presents. For example, if you create a new device but do not generate any certificates for it, NervesHub will capture and trust the first path validated certificate presented.


apps/nerves_hub_device/lib/nerves_hub_device_web/channels/user_socket.ex
```diff
@@ -1,6 +1,5 @@
 defmodule NervesHubDeviceWeb.UserSocket do
```
ConnorRigby: Maybe not in the scope of this PR, but i get tripped up every time i read the name of this file/module. 
In the future can we rename this to `DeviceSocket`?

apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
```diff
@@ -212,6 +212,44 @@ defmodule NervesHubCore.Devices do
...
+  @spec get_ca_certificate_by_aki(binary) :: CACertificate.t() | nil
+  def get_ca_certificate_by_aki(aki) do
+    query =
+      from(
+        c in CACertificate,
+        where: c.aki == ^aki
+      )
+
+    query
+    |> Repo.one()
```
ConnorRigby: I think lines 226-234 can all be replaced with:
```
Repo.get_by(CACertificate, aki: aki)
```

https://hexdocs.pm/ecto/Ecto.Repo.html#c:get_by!/3

apps/nerves_hub_core/lib/nerves_hub_core/devices.ex
```diff
@@ -212,6 +212,44 @@ defmodule NervesHubCore.Devices do
...
+  @spec get_ca_certificate_by_serial(binary) :: CACertificate.t() | nil
+  def get_ca_certificate_by_serial(serial) do
+    query =
+      from(
+        c in CACertificate,
+        where: c.serial == ^serial
+      )
+
+    query
+    |> Repo.one()
```
ConnorRigby: Same. Lines 243-250 can be replaced with:
```
Repo.get_by(CACertificate, serial: serial)
```

jlmcgehee21:
@mobileoverlord -- rather than trusting the first cert automatically, I've seen some services offer a "devices pending authentication" view, which forces you to manually approve a device before the cert is trusted.  This may be a safer path if it doesn't block whatever use-case you were imagining for trusting by default.

mobileoverlord:
The auto registration mechanism only applies to certificates that were signed by trusted, path validated CA certs. So if an organization tells NervesHub to trust devices signed by their own CA, it seems reasonable to trust new devices signed by their CA cert. I think I am okay with this proposition as the new default. In the future we could build a way to customize it when we are presented with an org where their domain would benefit from a human intervention step.

fhunleth:
@jlmcgehee21 Auto-registration is off by default. If you want auto-registration, you need to upload a CA certificate. Then when a device connects with an unknown certificate, Nerves Hub will see if that cert has been signed by a CA cert in your organization. If yes, then the device is added to Nerves Hub. 

I'm assuming there's a misunderstand here, but if not, could you let us know of a use case where you'd supply a CA cert to Nerves Hub and want a manual step to activate the device? I was thinking that anyone wanting to manually activate devices would upload device certs to Nerves Hub and not deal with an auto-registration system.

mobileoverlord:
> I was thinking that anyone wanting to manually activate devices would upload device certs to Nerves Hub and not deal with an auto-registration system.

In this case, as soon as they upload the device certificate, auto-registration is off for that device.

jlmcgehee21:
@fhunleth @mobileoverlord, I missed that it was contingent on a CA certificate already being added to the org.

PR #304, opened 2018-12-01
Add API for interacting with CA certificates

mobileoverlord:
This PR adds resource endpoints for managing `ca_certificates`. 

The following endpoints are added:

* POST `/:org_name/ca_certificates` - create a new certificate from a pem.
* GET `/:org_name/ca_certificates` - list all ca_certificates
* GET `/:org_name/ca_certificates/:serial` - show information about a single certificate
* DELETE `/:org_name/ca_certificates/:serial` - delete the certificate with the specified serial

### DIFF
--- a/apps/nerves_hub_api/test/nerves_hub_api_web/controllers/ca_certificate_controller_test.exs
+++ b/apps/nerves_hub_api/test/nerves_hub_api_web/controllers/ca_certificate_controller_test.exs
@@ -1,0 +1,64 @@
+defmodule NervesHubAPIWeb.CACertificateControllerTest do
+  use NervesHubAPIWeb.ConnCase, async: true
+
+  alias NervesHubCore.{Devices, Certificate}
+
+  describe "index" do
+    test "lists all ca certificates", %{conn: conn, org: org} do
+      conn = get(conn, ca_certificate_path(conn, :index, org.name))
+      assert json_response(conn, 200)["data"] == []
+    end
+  end
+
+  describe "create ca certificate" do
+    test "renders key when data is valid", %{conn: conn, org: org} do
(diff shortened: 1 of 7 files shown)


apps/nerves_hub_api/test/nerves_hub_api_web/controllers/ca_certificate_controller_test.exs
```diff
@@ -0,0 +1,64 @@
...
+
+      conn = post(conn, ca_certificate_path(conn, :create, org.name), params)
+      resp_data = json_response(conn, 201)["data"]
+      assert %{"serial" => ^serial} = resp_data
+    end
+
+    test "renders errors when data is invalid", %{conn: conn, org: org} do
+      conn = post(conn, ca_certificate_path(conn, :create, org.name), cert: "")
+
+      assert json_response(conn, 500)["errors"] != %{}
```
pdgonzalez872: Question: Should you be specific on the errors? Maybe this is a good way to show what the errors could be.

mobileoverlord: I certainly like the spirit of your comment, however, I believe that if we were to list all the types of errors that could happen here, we would be essentially copying each column in the schema that is required and checking when we omit it. My plan to handle this kind of event is in an integration test suite between `nerves_hub_web` and `nerves_hub_cli`. I am trying to figure out how to do this without making all the tests sequential.

pdgonzalez872: sorry, I missed this somehow @mobileoverlord . 

Sometimes listing everything doesn't yield that much value in the end. Nice, an integration test between both apps would be an awesome approach! :+1:

PR #305, opened 2018-12-03
Rename nerves_hub_core to nerves_hub_web_core

mobileoverlord:
Why?

Because it's been a bit since we had a PR that touches a ridiculous number of files. 

Actually, this is so we can free up the module name `nerves_hub_core` so it could be used as a hex package on the client side. The problem this is trying to solve: We need to include the API modules from `nerves_hub_cli` at runtime of a phoenix app, but without `mix` which the CLI heavily relies on. Therefore, the goal is to split out the API interface bits from the CLI and publish them as `nerves_hub_core`. This is also how hex names their server API interface package.


apps/nerves_hub_core/README.md
```diff
@@ -1,21 +1,21 @@
-# NervesHubCore
+# NervesHubWebCore
 
 **TODO: Add description**
 
 ## Installation
 
 If [available in Hex](https://hex.pm/docs/publish), the package can be installed
-by adding `nerves_hub_core` to your list of dependencies in `mix.exs`:
+by adding `nerves_hub_web_core` to your list of dependencies in `mix.exs`:
```
ConnorRigby: I like the consistency, but should we delete the sub-readmes from each of the projects? They don't really offer anything useful lol

mobileoverlord: Agreed, it would be a good follow up

PR #321, opened 2018-12-19
Add table to Org panel to display limits

ConnorRigby:
screenshot from 2018-12-19 08-12-48

Doesn't look the greatest in my opinion but i think this is useful. It might even be useful to display the current usage for these things in a followup.


mobileoverlord:
Surfacing this data is important. Thanks @ConnorRigby ❤️

PR #323, opened 2019-01-03
Set last known firmware in device plug

### LINKED ISSUE #322: `NervesHub.update()` behaves unexpectedly with a never-seen-before-device
I think it is because

https://github.com/nerves-hub/nerves_hub_web/blob/8b09a433ede2f6cc7bb1417bb2c5814b96f448c4/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex#L16

sets the last known firmware whereas

https://github.com/nerves-hub/nerves_hub_web/blob/master/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex#L10

does not. And

https://github.com/nerves-hub/nerves_hub_web/blob/8b09a433ede2f6cc7bb1417bb2c5814b96f448c4/apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex#L82

returns `[]` if there is no last known firmware.

danielspofford:
Why
---

* A device with no last known firmware that tries to use device apis to
fetch an update would previously fail to resolve new firmware because
eligible deployments can only be resolved if a last known firmware
exists for a device

PR
--

* Resolves #322 
* Is powered by: https://github.com/nerves-hub/nerves_hub/pull/50 but is written to be backwards compatible without it

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex
@@ -14,7 +14,9 @@ defmodule NervesHubDeviceWeb.Plugs.Device do
          {:ok, cert} <- X509.Certificate.from_der(cert),
          serial <- Certificate.get_serial_number(cert),
          {:ok, cert} <- Devices.get_device_certificate_by_serial(serial),
-         {:ok, device} <- Devices.get_device_by_certificate(cert) do
+         {:ok, device} <- Devices.get_device_by_certificate(cert),
+         uuid_header <- get_req_header(conn, "x-nerveshub-uuid"),
+         {:ok, device} <- update_last_known_firmware(uuid_header, device) do
       assign(conn, :device, device)
     else
       _err ->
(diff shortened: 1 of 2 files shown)


apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex
```diff
@@ -14,7 +14,14 @@ defmodule NervesHubDeviceWeb.Plugs.Device do
          {:ok, cert} <- X509.Certificate.from_der(cert),
          serial <- Certificate.get_serial_number(cert),
          {:ok, cert} <- Devices.get_device_certificate_by_serial(serial),
-         {:ok, device} <- Devices.get_device_by_certificate(cert) do
+         {:ok, device} <- Devices.get_device_by_certificate(cert),
+         fw_uuid_header <- get_req_header(conn, "x-nerveshub-uuid") do
+      device =
+        case fw_uuid_header do
+          [fw_uuid] -> Devices.update_last_known_firmware(device, fw_uuid)
```
danielspofford: Oop. This returns a tagged tuple and that needs to be handled.

PR #325, opened 2019-01-04
Update devices when tags are updated.

### LINKED ISSUE #309: Updating a devices tags should send a update_available message
Changing a device's tags from the frontend doesn't send a device a message telling it to update to the correct firmware.

ConnorRigby:
Fixes #309 

This will dispatch eligible update_available channel messages when a `Device` is updated.

### DIFF
--- a/apps/nerves_hub_api/test/nerves_hub_api_web/controllers/device_controller_test.exs
+++ b/apps/nerves_hub_api/test/nerves_hub_api_web/controllers/device_controller_test.exs
@@ -1,5 +1,5 @@
 defmodule NervesHubAPIWeb.DeviceControllerTest do
-  use NervesHubAPIWeb.ConnCase, async: true
+  use NervesHubAPIWeb.ConnCase, async: false
   alias NervesHubWebCore.{Devices, Fixtures}
 
   describe "create devices" do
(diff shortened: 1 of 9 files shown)


ConnorRigby:
I was able to fix tests failing, but at the expense of any test that calls `update_device` can no longer be async. I know there must be a better way of doing that. Leaving it here for review.

ConnorRigby:
the issue i had was with shared DbConnection. Ecto coudn't determine who owns the connection in a test because of the spawned task in `update_device`

danielspofford:
This may eventually be resolved by Ecto levaraging this stuff:
https://github.com/elixir-lang/elixir/issues/7995
https://github.com/elixir-lang/elixir/pull/8510

Until then:
- modify the app code to allow other pids to use your DB connection (gross)
- modify the app code to not make the db call from another process (ew)
- do not run the test async

ConnorRigby:
Right now i opted for `do not run the test async` because it required the least modification. The tests currently pass right now, but having them be non async is sort of annoying.

ConnorRigby:
Right now i opted for `do not run the test async` because it required the least modification. The tests currently pass right now, but having them be non async is sort of annoying.

PR #328, opened 2019-01-10
Support just-in-time registration in device api

### LINKED ISSUE #327: Support Just-in-Time registration over Device API
When connecting via web socket, `NervesHubDevice.SSL.verify_device/1` is called here:

https://github.com/nerves-hub/nerves_hub_web/blob/9786178040e9b880e20580de86e74f41bc35b16b/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/user_socket.ex#L23

And is defined here:

https://github.com/nerves-hub/nerves_hub_web/blob/master/apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex#L41

That is the code that facilitates JITR. Whereas here:

https://github.com/nerves-hub/nerves_hub_web/blob/master/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex#L16

It will fail to find a device cert and return a `403`.

I think this could be remedied by calling `NervesHubDevice.SSL.verify_device/1` instead of `Devices.get_device_certificate_by_serial(serial)` in that last link. I will open a PR shortly.

danielspofford:
PR
--

* Resolves #327

Why
---

* Some devices poll for updates via `NervesHub.update()` which prior to
this commit would not cause the device to have a device certificate
generated for them

How
---

* Leverage `NervesHubDevice.SSL.verify_device/1` like the device socket
does

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/plugs/device.ex
@@ -1,7 +1,7 @@
 defmodule NervesHubDeviceWeb.Plugs.Device do
   import Plug.Conn
 
-  alias NervesHubWebCore.{Devices, Certificate}
+  alias NervesHubWebCore.Devices
 
   def init(opts) do
     opts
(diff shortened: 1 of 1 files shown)


fhunleth:
I looked at the change and it looks right to me. Just to double check that I'm not missing something subtle. In the "why" section, it says "would not cause the device to have a device certificate generated for them". Should that be "would not cause the device's device certificate to be authenticated and registered."

danielspofford:
Yep I meant it how you re-articulated it, should be on the same page 👍

Issue #329, opened 2019-01-16
Password reset e-mail contains invalid url


NickNeck:
The current deployed nerves-hub version sends an e-mail for the password reset with an URL like `http://password-reset/a14ec49e-0598-4a92-8101-xxxxxxxxxxx`.

fhunleth:
@NickNeck Thanks for the report. It's fixed now if you'd like to retry.

PR #330, opened 2019-01-16
Fix forgotten password email link (#329)

fhunleth:


jlmcgehee21:
We don't have any tests for the correctness of that url (or any email templates in genera).  It's pretty easy to add with Bamboo: https://hexdocs.pm/bamboo/readme.html#testing

Not sure that's a reason to explicitly reject this PR though.  I suppose if we have an issue documenting the need for email tests, I'd be down to ship this as is.

fhunleth:
I added an issue: https://github.com/nerves-hub/nerves_hub_web/issues/333.

jlmcgehee21:
❤️ -- thanks @fhunleth

PR #332, opened 2019-01-16
Readme setup updates

fhunleth:


fhunleth:
Yeah, hang on... I'll do that. It needs rebasing anyway due to the yarn change.

fhunleth:
Done. Rebased too.

Issue #31, opened 2018-02-15
Device status


fhunleth:
It's useful to know a high level status of the device (like red, yellow, green or failed, warnings, ok). The idea is that once you get enough devices, it's convenient to be able to see at a glance what's happening rather than clicking through to a device detail view. 

It seems difficult to combine all interesting device status into one field, so here's a high level idea that has two columns:

* Overall status
  * Red/failed -> offline for > 1 day
  * Yellow/warning -> connected in the past day; an error has been submitted
  * Green/ok -> connected in the past day; no error logs

* Firmware update status
  * Red/failed -> firmware update needed, but errors or retries detected
  * Yellow/warning -> firmware out-of-date. An update will be requested on next connect
  * Green/ok - firmware up-to-date

The firmware update status field would replace the "update pending" column

jlmcgehee21:
@fhunleth, since we're using channels, what about using presence here?

fhunleth:
Yeah, I had been thinking that a while ago, but I haven't thought about it recently.

mobileoverlord:
I am absolutely on board using Presence. Production already supports it. The api, device, and www servers are already clustered.

jlmcgehee21:
Yeah, we used in on our firmware update server. I wasn't on the project during the time we did it, but it seemed to work super well.  Of course we'll have to track a bit more info to have `red, yellow, green` based on the above definitions, but I think Presence can be at the heart of it.

I'd like this to be my next major task after I get a few of these cleanup type issues out of the way.

jjcarstens:
We have lots of statuses now. New issues for more fine tuned changes if desired

PR #335, opened 2019-01-18
Update last_communication info for devices, and render in frontend.

jlmcgehee21:
(related to: https://github.com/nerves-hub/nerves_hub_web/issues/31)

Why:

* It's useful, especially for users whose devices are not maintaining a
constant connection.

* Updating `last_communication` in `plug` for http requests and on
  `join` function for channels.

Other Notes:
* Though the DB field is named `last_communication`, it is currently only being updated on _connection_, so the language in the UI reflects that.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/device/index.html.eex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/device/index.html.eex
@@ -14,6 +14,7 @@
           <th class="col-3">Identifier</th>
           <th class="col-2">Version</th>
           <th class="col-3">Status</th>
+          <th class="col-3">Last Connection (local time)</th>
           <th class="col-2">Tags</th>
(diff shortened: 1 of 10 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/templates/device/index.html.eex
```diff
@@ -14,6 +14,7 @@
           <th class="col-3">Identifier</th>
           <th class="col-2">Version</th>
           <th class="col-3">Status</th>
+          <th class="col-3">Last Connection (local time)</th>
```
fhunleth: I'm mostly on board with printing out local time. I wonder if there will be confusion between the times being browser local time vs. device local time. I.e., my devices are spread across several time zones. "Browser local time" seems verbose. If it's easy to print out the timezone, then that certainly would clarify.

PR #339, opened 2019-01-20
Fix elixir 1.8 deprecation warning for :seconds time unit

mobileoverlord:


mobileoverlord:
I looked and this is compatible with elixir 1.7.x as well

PR #342, opened 2019-01-28
Fix last_communication bug and test js in CI

danielspofford:
Why
---

* Fixes a bug where if a device has no `last_communication` it was
  errantly parsed into `Invalid Date`.

How
---

* Update date formatting logic to account for `never` value.
* Test date formatting logic.
* Add an eslint config.
* Add a prettier config.
* Test JS, verify formatting, and lint JS in circleci.
* Remove comments from and format socket js file. Also remove
  unreferenced `resp` variables.


apps/nerves_hub_www/rel/Dockerfile.build
```diff
@@ -4,7 +4,7 @@ ADD . /build
 WORKDIR /build
 RUN mix deps.clean --all && mix deps.get
 
-FROM node:9.11.2 as assets
+FROM node:8.11.3 as assets
```
mobileoverlord: Was this an intentional change to use node 8 instead of 9?

mobileoverlord:
Wow, this is great. Thanks @danielspofford 🤗

Issue #347, opened 2019-01-29
List more information on account certificates so that they're easier to manage


fhunleth:
**Describe the bug**
From running various experiments, I've managed to create 5 "Account Certificates". The description for all of them is the same and I'm unsure which ones are safe to delete.

**To Reproduce**
Run `mix nerves_hub.user auth` repeatedly.

**Expected behavior**
The certificates are labeled with something that can help me figure out if they're old. I'm not sure what's easier - maybe the creation date or the date they were last used?

**Screenshots**

![image](https://user-images.githubusercontent.com/64669/51940743-d41fa180-23e0-11e9-85f4-673f9f8defd5.png)

PR #351, opened 2019-02-02
Add last_used to Device, CA, and User Certificates

mobileoverlord:
This PR adds `last_used` to all certificates and updates the values anytime those certificates are used to auth to the server. This is the first step towards fixing https://github.com/nerves-hub/nerves_hub_web/issues/347


mobileoverlord:
@ConnorRigby I agree, we can plug the real one in here for testing. I think direct connections with websockets were used here to both illustrate it can be tested agains any websocket client, and because there were complications with running some of `nerves_hub`'s dependencies on CI. This should be solved now though.

Issue #320, opened 2018-12-18
Allow a device to have a "fallback" firmware


ConnorRigby:
I had all intentions of implementing this today, but ran out of time.
Related to: #319 

If a device connects with an unknown firmware, it should have the ability to apply a known good firmware.

I've tried this a few ways to no avail so far:

* adding a `fallback_firwmare_id` to `Device`
* adding a `default` Deployment per `product`

I didn't like either of them but am willing to revisit if someone else likes those ideas.
I think this feature would also easily lead into forcing a device onto a particular firmware which i also tried to implement naively. the `Deployment` association gets in the way here a bit.

ConnorRigby:
I like this idea to some extent. I thought of doing it this way also. As far as i know we don't use any connect/join params in the client. It could be added there. Seems a little strange to implement it this way, but i can't exactly pin point _why_

jlmcgehee21:
The concern might be a "cold start" issue, where you don't know a valid fallback firmware uuid to bake into the firmware at the time of building it.

However, this problem still exists from the opposite side.  If a device connects with an unknown firmware, how can we safely decide what it should fall back to in NervesHub. If the firmware is unknown to NervesHub, how can NervesHub make that decision?

This complexity is why we initially built it to only work with no firmware versions. I know that's sidestepping some real-world issues, but the solution is tricky, especially if you were to have _many_ devices come online with unknown firmwares.  You would have no guarantees regarding which firmwares these devices could accept (or fall back to), which would make moving them all onto a known version a potentially nightmarish endeavor.

mobileoverlord:
@fhunleth and I had an offline conversation about this. Here is my understanding of the result of that conversation.

There are a few issues that are related, here are the issues that we identified:
* Allowing the update of
  * Connecting with unknown firmware and determining deployment eligibility.  
  * Connecting with firmware that has been garbage collected. This would be a symptom of plugging in a device thats been in the closet and offline for a long period of time. 
* Preventing the update of:
  * Connecting with firmware that has the same semantic version but a different uuid. This scenario is more typical when trying to debug a pipeline issue or test devices.
  * Connecting with firmware that was updated out of band to a higher version.

This is the strategy we came up with

* Unknown firmware will supply additional metadata to the server either through connect params or an after join request. 
* The supplied product metadata can be used to find deployments for that org for that product.
* The supplied version metadata is then used in a requirement match of `supplied_version < deployment_firmware_version`. 
[...]

PR #357, opened 2019-02-11
Embed firmware_metadata into Device

mobileoverlord:
As detailed in https://github.com/nerves-hub/nerves_hub_web/issues/320 our desired functionality is to use the firmware version number supplied with the device connection parameters to determine deployment eligibility. After making that change, it made sense to _always_ use the metadata supplied by the device on the server side. This PR removes the linked `last_know_firmware_id` and adds an embedded schema `firmware_metadata` to the device model.

To get this to fully work, the client will need to be updated to pass the extra metadata for both PhoenixChannel and HTTP requests. The code is structured to allow the current client to remain functional. If _only_ the firmware uuid is provided (the current client), the firmware_metadata will be constructed by first looking up the firmware by uuid. When updating the device.firmware_metadata, if any part fails and returns nil, the device record will not be updated. This behavior will likely change as we migrate to the newer client where every request can include all the metadata by default.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
@@ -1,0 +1,40 @@
+defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  @type t :: %__MODULE__{}
+
+  @required_params [
+    :uuid,
+    :product,
+    :architecture,
+    :version,
+    :platform
+  ]
+
+  @optional_params [
+    :author,
+    :description,
+    :vcs_identifier,
+    :misc
+  ]
+
+  embedded_schema do
+    field(:uuid)
+    field(:product)
+    field(:version)
+    field(:architecture)
+    field(:platform)
+    field(:author)
+    field(:description)
+    field(:vcs_identifier)
+    field(:misc)
+  end
+
+  def changeset(%__MODULE__{} = metadata, params) do
+    metadata
+    |> cast(params, @required_params ++ @optional_params)
+    |> validate_required(@required_params)
+  end
+end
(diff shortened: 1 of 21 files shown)


apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
```diff
@@ -0,0 +1,173 @@
+defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
```
jlmcgehee21: Based on the way many of these functions are being used in the new code, I feel like they being up a level in NervesHubWebCore.Firmwares.  Will be glad to list the specific offending functions in the AM, but on mobile now, so keeping this short.

mobileoverlord: I believe you might be referring to the ones that pull params from `Firmware`. There are cases that are allowed in this code that I would want to remove in the future. Such as, from `Firmware` and the case where it tries to get it from a firmware uuid alone. I think the client should have to _always_ pass up information about itself.

jlmcgehee21: Just realized how unclear my comment was -- 

I think all the `params_from_XXXX` functions belong one level up in `NervesHubWebCore.Firmwares`, as these are operations that need to be performed in the generic `Firmwares` scope as apposed to the specific `FirmwareMetadata` schema scope.

I'm willing to be convinced otherwise, and tbh, it's not worth blocking this work.  So just lmk if you don't think that's a good change, and I'll give my 👍

mobileoverlord: Ah I see what you are saying. Yeah, I think that’s a good idea. I’ll shuffle it around. Thanks :)

mobileoverlord:
@jlmcgehee21 I reworked the layout to address your comments above.

PR #359, opened 2019-02-12
Add quantum and schedule firmware job

mobileoverlord:
This brings in `quantum` and schedules the firmware garbage collection job on production to run every 15 minutes. It is running quantum as global since production is clustered and we only want these jobs to run on a single node.


apps/nerves_hub_web_core/config/prod.exs
```diff
@@ -3,6 +3,15 @@ use Mix.Config
 config :ex_aws,
   region: "${AWS_REGION}"
 
+config :nerves_hub_web_core, NervesHubWebCore.Scheduler,
+  global: true,
+  jobs: [
+    garbage_collect_firmware: [
+      schedule: "*/15 * * * *",
```
fhunleth: omg. It's 2019 and cron tab format is the best we got?

PR #361, opened 2019-02-12
Remove last_known_firmware from Device

mobileoverlord:


mobileoverlord:
The data migration is complete. Since there was a brief time where the `firmware_metadata` was not populated for the devices, some devices may appear as `offline` with `unknown` firmware until the server restarts or the device reconnects.

PR #362, opened 2019-02-12
Multiple firmware

mobileoverlord:
This PR fixes a few edges that were seen in the production logs.

When looking up the firmware in the db, `get_by_firmware_uuid` was expecting there to only be one result. This updates it to more then 1


fhunleth:
How is it possible to have more than one firmware with the same uuid? They should be the same firmware image.

mobileoverlord:
The same firmware is uploaded to two different orgs

mobileoverlord:
and they are signed with different fwup keys

fhunleth:
ok, that's legit.

PR #363, opened 2019-02-13
Connect devices to device topic

mobileoverlord:
This transitions devices from connecting to the topic `firmware:#{firmware.uuid}` to `device`. The existing `firmware:*` are internally routed to the `device` join logic.


mobileoverlord:
I reworked this after thinking through the device side a little more. The intention now is that the device will not have to supply any information and can only connect to the topic `device`. The server will internally subscribe it to a unique channel upon join of the topic of `device:#{device.id}`. Thus simplifies the client code so it does not need to know where to look for its identifier. Instead,= the identity of the device is solely placed in the validity of the certificate it presents.

PR #366, opened 2019-02-20
Firmware tranfers

mobileoverlord:
This PR does the following
  * Add `firmware_transfers` table
  * Digest server access logs from S3 and create transfer records.

The intention of this is to eventually allow orgs to see how much data they are sending to clients for deployments.


test/support/fixtures.ex
```diff
@@ -58,6 +58,17 @@ defmodule NervesHubWebCore.Fixtures do
     }
   end
 
+  def firmware_transfer_params(org_id, firmware_uuid) do
+    %{
+      org_id: org_id,
+      firmware_uuid: firmware_uuid,
+      remote_ip: "111.222.333.444",
```
fhunleth: It doesn’t look like this matters, but it’s an invalid IP address.

PR #371, opened 2019-03-05
Force device certificates to match on serial, aki, and validity

mobileoverlord:
This PR updates the logic which queries the database for existing trusted device certificates to match on serial, authority key id, and validity. This fixes a case where a signer or multiple signers would create certificates with the same certificate serial number.


fhunleth:
I just realized that matching on device cert's validity is superfluous, since the device cert's serial number is a hash of the device's public key and validity fields. This still feels like a good thing to do long term especially if we ever stop computing the device cert's serial number from those fields.

PR #372, opened 2019-03-05
Revert "Force device certificates to match on serial, aki, and validity"

mobileoverlord:
Older versions of the NervesHub client are having issues connecting. Lets revert the commit until we have a test in place for that.

This reverts commit 42fae0eae51a441591e9ddc2fbe1b65b44424189.


mobileoverlord:
This may not be necessary, the issue could be a data issue. Still investigating. keeping this here just in case.

mobileoverlord:
Turns out this was a data issue that only affected clients with certificates that were generated before the `aki` field was tracked on the `device_certificate` records. This was a time that predated support for custom ca certs, therefore, the data was fixed by adding the `aki` of the NervesHub CA device signing root cert.

PR #376, opened 2019-03-07
Expose device description through API

fhunleth:


fhunleth:
I wasn't sure whether description was left out on purpose?

mobileoverlord:
It wasn’t left out on purpose, I’m sure that the other templates could use a double check.

PR #382, opened 2019-03-18
Add Product User roles

mobileoverlord:
This PR is intended to be merged after https://github.com/nerves-hub/nerves_hub_web/pull/377

Similar to Org User roles, Product User roles grant permissions to users for individual products inside an organization. 

To keep things as small as possible, this PR only transitions the current structure to go through `ProductUsers` when loading products but does not yet implement filtering for those roles. After applying this PR, a manual data migration will be required add existing users for an org to all the products in that org.


fhunleth:
Do the roles work something like this:

* `:read` - user can see information on the product, deployments, and devices, but can't change anything
* `:write` - user can modify product metadata, create deployments and modify devices
* `:admin` - user can delete the product? Anything else?

mobileoverlord:
Yes, in addition, the `:admin` role will be able to add / remove other product user roles.

This PR only modifies product creation to take a `user` and automatically give that user an `:admin` role. Implementing scoping for `:read` and `:write` will be added in another PR as well as the API for allowing admins to give other users a product role.

mobileoverlord:
This is being pulled apart and merged in stages

PR #394, opened 2019-03-26
Update README.md

jjcarstens:
I'm currently going through local dev setup and these are some of the things I noticed that were either outdated or could use a little clarity.


jjcarstens:
I'm kind of at a lost here on why this is failing. Nothing changed should be affecting tests 🤔

jjcarstens:
K, I'm officially not crazy. Looks like a bad tz_data

https://twitter.com/laut/status/1110960785117888513

PR #395, opened 2019-03-27
Update README.md

jjcarstens:
I'm currently going through local dev setup and these are some of the things I noticed that were either outdated or could use a little clarity.

Duplicate of #394 to see if I can get the CI to pass. Last ditch effort incase its a problem with the cache

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,15 +41,13 @@ rebuild-db:
 	mix run apps/nerves_hub_web_core/priv/repo/seeds.exs
 
 test: .env
+	MIX_ENV=test && \
 	. ./.env && \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -41,15 +41,13 @@ rebuild-db:
 	mix run apps/nerves_hub_web_core/priv/repo/seeds.exs
 
 test: .env
-	. ./.env && \
-	DATABASE_URL=postgres://db:db@localhost:2345/nerves_hub_test \
 	MIX_ENV=test \
+	. ./.env && \
```
fhunleth: This was noticed on Slack by iautom8things: Lines 44 and 45 need to be swapped or the `MIX_ENV=test` set gets lost. Same thing with lines 49 and 50 below.

PR #396, opened 2019-03-28
Update README.md (and allow forks to run CI)

jjcarstens:
Copy of #394 & #395 (Third times the charm... 🤞)

Updated the readme with some points I went through while setting this up locally. Also made change to CI config to hopefully allow PRs from forks to run the CI in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ help:
 	@echo "server - start the server"
 	@echo "iex-server - start the server with the interactive shell"
 	@echo "reset-db - reinitialize the database"
+	@echo "reset-test-db - reinitialize the test database"
 	@echo "test - run the unit tests"
 
 .env:
@@ -34,6 +35,11 @@ reset-db: .env
 	. ./.env && \
 	make rebuild-db
 
+reset-test-db: .env
+	DB=db_test \
+	MIX_ENV=test \
+	make reset-db
+
 rebuild-db:
 	mix ecto.drop && \
 	mix ecto.create && \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -41,15 +41,13 @@ rebuild-db:
 	mix run apps/nerves_hub_web_core/priv/repo/seeds.exs
 
 test: .env
+	MIX_ENV=test && \
```
fhunleth: Lines 44 and 45 are swapped. Same with 49 and 50. This makes the `MIX_ENV=test` line a no-op. `mix test` shouldn't need it. I don't know if `mix test.watch` sets the mix environment automatically.

fhunleth: Just to be explicit, if setting `MIX_ENV` is required, this should be:
```
    . ./.env && \
   MIX_ENV=test \
   mix test
```
However, I'd prefer
```
    . ./.env && \
   MIX_ENV=test mix test
```
since that makes the environment variable part more explicit.

But having said all that, I don't know why `MIX_ENV=test` needs to be set for `mix test` anyway.

jjcarstens: so `MIX_ENV=test` is not required for `mix test`. However, I have it this way going _first_ for the `.env` file which was changed to do a little deduction on the env and adjust the `DATABASE_URL` to use a database of `db_test` when running tests.

So likewise I also documented that you can use `MIX_ENV=test make reset-db` to setup that same db for running tests.

I went back and forth on trying to do that fancy deduction in `.env`. I do think separating out the database for development and testing is beneficial, but I'd be okay expecting to use the same for each case if that is preferred.

If we want to move the fancy change our of `.env`, then I would  just need to hardcode a `DATABASE_URL` in the makefile and also adjust the steps to setup for being able to run `mix test` locally

Makefile
```diff
@@ -34,22 +35,25 @@ reset-db: .env
 	. ./.env && \
 	make rebuild-db
 
+reset-test-db: .env
+	DB=db_test \
+	MIX_ENV=test \
```
jjcarstens: Put this here so the next `rebuild-db` step will use the compiled `test` build

PR #400, opened 2019-04-01
add Phoenix LiveView structure

jjcarstens:
* Adds the structure to support Phoenix LiveView according to its docs
* Changes `apps/nerves_hub_www/mix.exs` to use released versions of phoenix and phoenix_live_reload instead of the rc
* Updates to Phoenix v1.4.3 as required (see https://github.com/phoenixframework/phoenix_live_view/issues/141)


ConnorRigby:
It looks like the circle config for `test_javascript` needs to run `mix deps.get` now unfortunately.

ConnorRigby:
adding this snippet to the config.yml right [here](https://github.com/nerves-hub/nerves_hub_web/blob/1c68448e722bfe0ba4b1feb8b822a34503af90b1/.circleci/config.yml#L161) might fix it.
```yml
      - restore_cache:
          keys:
            - v1-mix-cache-{{ .Branch }}-{{ checksum "mix.lock" }}
            - v1-mix-cache-{{ .Branch }}
- v1-mix-cache
```

jjcarstens:
@ConnorRigby Just adjusted to pull the `phoenix_live_view` js from the git repo instead of the file path. Should be good enough until officially released since that's also how we're getting the elixir live_view stuff

Issue #317, opened 2018-12-18
Invitations should lookup an existing account or register one


ConnorRigby:
**Describe the bug**
Accepting an invite prompts a user to make an account. If you are being added to a second 
org, you already have an account and just need to sign in to accept the invite.

**To Reproduce**
1) Create account
2) Create 2 orgs
3) Invite second user to join org 1
4) Create second account
5) invite second user to join org 2
6) ???
7) Profit

**Expected behavior**
Expected to be able to login if my email is already associated with an Account. 

**Additional context**
This is important for having multiple users with a `staging` and `production` environment.

`admin@some.org` - Root account
`deploy@some.org` - CI account so certs can be revoked if it gets pwnd

`deploy` needs to be part of the same nerves-hub orgs as `admin`

jjcarstens:
I think it would be nice to have a "notifications" part of the nerves hub UI where an existing user could see this request and just accept there once logged in. But I think that's a much bigger cut.

For now, going to just adjust this to find or create the user and function from there (unless there are any objections)

PR #402, opened 2019-04-01
fix invitations to support existing users

jjcarstens:
Fixes https://github.com/nerves-hub/nerves_hub_web/issues/317

This will now attempt a user lookup when you try to invite a user to an org.
* If user exists, adds user to the org and sends an email to user to notify
* If user does not exist, functions the same as before (sending an invitation)
* If user exists _and_ already belongs to the org, reports back to org#edit that the user is already a member

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www/accounts/email.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www/accounts/email.ex
@@ -23,4 +23,13 @@ defmodule NervesHubWWW.Accounts.Email do
     |> put_layout({NervesHubWWWWeb.LayoutView, :email})
     |> render("forgot_password.html", user: user)
   end
+
+  def org_user_created(email, %Org{} = org) do
+    new_email()
+    |> from(@from)
+    |> to(email)
+    |> subject("New NervesHub Organization Added")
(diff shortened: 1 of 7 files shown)


apps/nerves_hub_www/lib/nerves_hub_www/accounts/email.ex
```diff
@@ -23,4 +23,13 @@ defmodule NervesHubWWW.Accounts.Email do
     |> put_layout({NervesHubWWWWeb.LayoutView, :email})
     |> render("forgot_password.html", user: user)
   end
+
+  def org_user_created(email, %Org{} = org) do
+    new_email()
+    |> from(@from)
+    |> to(email)
+    |> subject("New NervesHub Organization Added")
```
jlmcgehee21: This shouldn't block, but maybe "You have been added to a new NervesHub organization"?

jjcarstens: I'm good with whatever popular opinion is here. I am definitely not a copy expert. I can change it if needed

fhunleth: Mind if we make a second pass on the email text? I'll make a proposal in a separate issue since it affects more than just this email.

PR #407, opened 2019-04-03
Add structure for Audit Logs

jjcarstens:
Initial structure to start saving audit logs (no implementation). Heavily inspired by [Hexpm audit log](https://github.com/hexpm/hexpm/blob/master/lib/hexpm/accounts/audit_log.ex)
* adds `NervesHubWebCore.Accounts.AuditLog` schema
* adds `NervesHubWebCore.Types.Resource` to handle the resource type of AuditLog
* migration to create table and fields for AuditLog

Few things to note:
* The `resource_type` is stored in the DB as a string, but when loaded it represented as an atom of an existing module. I went back and forth on whether to just have it as a string when loaded in the structs and changesets so if you have an opinion, speak up. My guess is we might want to do more with the atom, but maybe not? It'd also be a really easy change later on without affecting the DB
* `resource_type` is only valid if its a module that uses Ecto.Schema and is a runtime check. I'd actually prefer this to be compile time, but that's a bigger change so I'd rather do that separately next. (see `NervesHubWebCore.Types.Resource.allowed_resources/0`)
* ~~`User` and `Org` associations are included more as a convenience but this should still support instances where user or org may not be involved (like device reboot report, etc)~~

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/audit_log.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/audit_log.ex
@@ -1,0 +1,31 @@
+defmodule NervesHubWebCore.Accounts.AuditLog do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+  import EctoEnum
+
+  alias NervesHubWebCore.Accounts.{Org, User}
+  alias NervesHubWebCore.Types.Resource
+
+  @primary_key {:id, :binary_id, autogenerate: true}
+  @required_params [:action, :actor_id, :actor_type, :params, :resource_id, :resource_type]
+
+  defenum(Action, :action, [:create, :update, :delete])
+
+  schema "audit_logs" do
+    field(:action, Action)
+    field(:actor_id, :id)
+    field(:actor_type, Resource)
+    field(:params, :map)
+    field(:resource_id, :id)
+    field(:resource_type, Resource)
+
+    timestamps(updated_at: false)
+  end
+
+  def changeset(%__MODULE__{} = audit_log, params) do
+    audit_log
+    |> cast(params, @required_params)
+    |> validate_required(@required_params)
+  end
+end
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/types.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/types.ex
@@ -25,4 +25,48 @@ defmodule NervesHubWebCore.Types do
     def load(tags), do: Ecto.Type.load(type(), tags)
     def dump(tags), do: Ecto.Type.dump(type(), tags)
   end
+
+  defmodule Resource do
+    @behaviour Ecto.Type
+
+    def type, do: :string
+
+    def cast(resource) when is_atom(resource) do
+      resource
+      |> to_string()
+      |> cast()
+    end
+
+    def cast(resource) when is_bitstring(resource) do
+      if resource in allowed_resources() do
+        {:ok, String.to_existing_atom(resource)}
+      else
+        :error
+      end
+    end
+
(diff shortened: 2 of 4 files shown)


jjcarstens:
There are a few initial use cases to start which are logging potentially problematic events (uploading new firmware/deployment) and logging when a user triggers a reboot (feature to come soon😉), etc. 

We could go with an enum for typical CRUD actions and I think that would be fine. Having as a string allows use to be more flexible in the future to report scoped actions that might not be as apparent within simple CRUD definitions (I.e. `device.reboot`, `user.update.new_permission`, etc). That also allows querying a little more specifically on a single field.

On the flip, it’s not as defined and makes it much easier to dump in bad data. We could make it a custom type like resource to handle validating a key string structure which might allow the string flexibility and a bit more structure like the enum. 

What do you think?

jlmcgehee21:
It's easier to start with an enum and migrate to a string than it is to go from a string to an enum.

My choice would be to make it an enum with the currently known actions that would be logged, then expand that list as needed in the future.

danielspofford:
It is a good idea to maintain a well-defined/documented list of actions. Having them as strings vs atoms at the application level shouldn't matter in this case WRT performance, only ergonomics: `:"device.reboot"` vs `"device.reboot`, which makes me lean string.

The choice at the database level comes down to: do you want to limit the value of that field via the application or the DB. Some downsides of a Postgres enum: https://github.com/gjaldon/ecto_enum#important-notesgotchas. A dedicated actions table + FK relationship is a possible heavier solution here as well that still has DB level constraints.

I'm not convinced DB level protections are going to buy us anything here, but it seems a relatively painless buy-in 🤷🏻‍♂️.

jlmcgehee21:
:+1: This could go either way. I suppose as long as the devs are in control of the strings that go into the DB, then I don't have a strong opinion.

mobileoverlord:
I’ve recently debated the same thing and ended up adding `ecto_enum` for the roles. I also believe hat it is easier to go back to strings if we choose to. I would say we should be consistent and therefore would suggest using an enum type.

jjcarstens:
changed `:action` to be an enum. Were there opinions about leaving `resource_type` as is and letting it validate according to what schemas are in the app? Or also change that to an enum as well?

mobileoverlord:
This is looking good structurally, but I am concerned in its flexibility for creating audit log entries caused by a device. For example, when a device connects to the server, it modifies its own record to update the last connection timestamp. I could see how this would be handled by inserting an `update` action into the `AuditLog`, but the schema seems to expect that audit log entries would be made solely by users in an org. Should we reference `user` and `device` ad make it that at least one is required?

jjcarstens:
Chatted with @mobileoverlord offline and we decided to change the structure a bit:
* Removed `org` and `user` from the schema
* Instead, added concept of `actor` which is identifying who is performing the action on the resource
  * `actor_id`: id of the actor resource
  * `actor_type`: the resource type of the actor
* All fields required (except `:id` and `:inserted_at`) - If this becomes a pain later on, we might try defaulting `actor` data to be the same as the resource if `nil` (assuming the resource acted for itself).

apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/audit_log.ex
```diff
@@ -0,0 +1,32 @@
...
+    field(:resource_type, Resource)
+
+    timestamps(updated_at: false)
+  end
+
+  def changeset(%__MODULE__{} = audit_log, params) do
+    required_params = __MODULE__.__schema__(:fields) -- [:id, :inserted_at]
+
+    audit_log
+    |> cast(params, __MODULE__.__schema__(:fields))
```
mobileoverlord: Wouldn't doing it this way allow the `inserted_at` / `id` to be forced?

Issue #417, opened 2019-04-08
Update all device fields without a reload


fhunleth:
Currently only the device online/offline field updates without a page reload. When upgrading firmware, the firmware version changes, but you need to reload the page to see it.

Issue #401, opened 2019-04-01
Support remote device reboot


fhunleth:
Sometimes remotely rebooting a device can fix weird issues in the field. This is never a satisfying solution, but it's nice to have the ability when you're desperate. 

The idea with this issue is to add an API and a button in the web UI to the device. In the UI, it would be nice to have an "are you sure" check to prevent accidents.

fhunleth:
For my purposes, this is an immediate or nothing event. I.e., if the device is connected to NervesHub, then the reboot message is sent. If it's not connected, then nothing.

I could imagine that a deferred reboot to the next connection attempt could be both useful and very confusing. I'm not ready to address the way that that should be handled.

PR #420, opened 2019-04-09
Support device reboot

jjcarstens:
Resolves https://github.com/nerves-hub/nerves_hub_web/issues/401

* Few conditions must be met:
  * User must belong to Org of device _and_ be an admin
  * Device must be online to attempt to trigger a reboot
* Audits when a user attempts a reboot
* Changes `device#show` to use LiveView
  * Updates automagically when presence data changes for a device
  * resolves https://github.com/nerves-hub/nerves_hub_web/issues/417
* supports displaying flash message in LiveView pages

4 quadrants of the gif are:

`device#show` page | Chrome dev console (show changing elements)
--- | ---
logs for server | device logs (to show closed session on reboot)

device_show_reboot


mobileoverlord:
@jjcarstens @fhunleth and I discussed this over a call. The question was regarding the update to add `:reboot` to  `Action` at https://github.com/nerves-hub/nerves_hub_web/pull/420/files#diff-ee862541304c58fa7b0d9420a344f5aeR12

Since `Action` is an enum type, I'd prefer to prevent many changes to it. Adding `:reboot` gave me the impression of volatility which equals pain on enum types. The go fwd change we decided on was to allow `AuditLog` to not only represent "who changed what persistent state" but to include logs regarding requests to change. Therefore, asking a device to reboot would be an `:update` `Action` with some searchable metadata in the `params` describing that it was asked to reboot.

jjcarstens:
Updated with the @mobileoverlord suggestion. Ready for another 👀

PR #422, opened 2019-04-10
Humanize display of organization info

### LINKED ISSUE #415: Display organization limits in human-readable units
**Describe the bug**
Some mental math is needed to understand the org limits right now:

image

**Expected behavior**
Display unreferenced firmware gc time in days and firmware size in MB or GB.

pcmarks:
Fixes #415
Show Firmware size in MiB, GiB
Show TTL in days, etc. depending on value.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_view.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_view.ex
@@ -1,3 +1,5 @@
 defmodule NervesHubWWWWeb.OrgView do
   use NervesHubWWWWeb, :view
+
+  import NervesHubWWWWeb.LayoutView, only: [humanize_size: 1, humanize_seconds: 1]
 end
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org/edit.html.eex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org/edit.html.eex
@@ -18,15 +18,15 @@
         <th>Device Limit</th>
         <th>Firmware per Product</th>
         <th>Firmware Size</th>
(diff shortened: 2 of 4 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_view.ex
```diff
@@ -1,3 +1,29 @@
...
+  @gib :math.pow(2, 30)
+  @mib :math.pow(2, 20)
+  @kib :math.pow(2, 10)
+  @precision 3
+
+  def humanize_ttl_seconds(seconds) do
+    seconds
+    |> Timex.Duration.from_seconds()
+    |> Timex.Format.Duration.Formatter.format(:humanized)
+  end
```
jjcarstens: I think these helper functions will useful in quite a few views. I think it'd probably be better to put them in `LayoutView` or as a helper module that we can import in specific views that may want them. What do you think?

fhunleth: I like that idea since it will make the code easier to unit test. I think `humanize_size` could use a couple unit tests on second look.

jjcarstens: Also just throwing this out there but there is a `Size` library with a `humanize!` macro that could also handle this:
https://github.com/jfcalvo/size

Doesn't really matter to me either way. Just an FYI incase that was easier to use

pcmarks: I guess I would lean towards putting them in LayoutView for now.

fhunleth:
Looks good!

Issue #365, opened 2019-02-13
Record the time a connection was lost


fhunleth:
NervesHub currently records the time of the most recent connection. This is good if the device is connected, but if the connection is lost, it would be useful to know when it was lost.

PR #424, opened 2019-04-10
Record time on device disconnect

jjcarstens:
Fixes https://github.com/nerves-hub/nerves_hub_web/issues/365

This assumes that a device socket being terminated is a "last connection"
and piggy backs on that field. So looking at the status of `offline`, you
would assume the `last_connection` time being displayed is when it disconnected.

I debated adding a new field for this so open to that if others would rather that be implemented.


fhunleth:
HTTP endpoint users - I assume that you're not looking at the online/offline field since it's not relevant for you. Not in this issue, but later on should be combine the "online" and "last connection" columns? I.e., one "last connected" column that says "connected" if the device is connected (Phoenix channel case) or the last connected timestamp.

PR #431, opened 2019-04-13
Add remote IEx feature

jjcarstens:
This is the initial support for connecting to an IEx server on a remote device from the browser and being able to submit commands and receive back the IO. It is all done through a LiveView session to update lines as they come.

(Big thanks to @ConnorRigby as most of this work is due to his initial POC and help understanding the I/O transport.)

Requires https://github.com/nerves-hub/nerves_hub/pull/86 on a device

I'm adding this as a draft PR for now cause there is probably more polishing to do, but I wanted to get some more 👀 on it to help think of cases I might be missing. So happy to hear any changes you might want to see.

I also went the route of only allowing _one_ IEx server to run on a device at once but allowing multiple users to to view that session (and interact if an `admin`). After sleeping on it a bit, I'm not sure that's the best. Maybe better to allow spawning multiple IEx servers for different users by limit the number of servers that can run on a device so it doesn't get bogged? Thoughts?

### When user is _not_ an `:admin`
not_allowed_remote

### Viewing IEx session from multiple users in other browsers
ezgif com-video-to-gif-2


apps/nerves_hub_www/mix.exs
```diff
@@ -67,7 +67,8 @@ defmodule NervesHubWWW.MixProject do
       {:distillery, "~> 2.0"},
       {:floki, "~> 0.20", only: [:test, :dev]},
       {:nerves_hub_web_core, in_umbrella: true},
-      {:nerves_hub_device, in_umbrella: true}
+      {:nerves_hub_device, in_umbrella: true},
+      {:ansi_to_html, git: "https://github.com/jjcarstens/ansi_to_html"}
```
jjcarstens: I made a pull request to this library for the needed changes, but its kind of stale (last activity 2 years ago). So use my fork for now and if nothing budges after a few days then I'll be wrapping this into a new library

fhunleth:
This is really cool!

fhunleth:
FYI - This isn't a draft any more. I didn't get an email, so not sure if others may have missed it as well.

ConnorRigby:
I'm stoked this is finally happening!

jjcarstens:
One more time - This now uses `presence` to decide if the devices is connected and has remote console enabled vs just letting the device connect on the ConsoleChannel no matter what.

Ready for 👀

PR #434, opened 2019-04-17
Check for equal or higher Org role before Product role

mobileoverlord:
This PR allows a user with an equal or higher `OrgUser` role to have implicit permission when a `ProductUser` role is required. For example. Viewing a product requires a `product: :read` role but users with an `org: :read` role or higher such as [`:admin, :delete, :write`] will also work.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/role_validate_helpers.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/role_validate_helpers.ex
@@ -7,11 +7,29 @@ defmodule NervesHubWebCore.RoleValidateHelpers do
     validate_org_user_role(conn, org, user, role)
   end
 
-  def validate_role(%{assigns: %{user: user, product: product}} = conn, product: role) do
-    if NervesHubWebCore.Products.has_product_role?(product, user, role) do
-      conn
-    else
-      halt_role(conn, "product " <> to_string(role))
+  def validate_role(
+        %{assigns: %{product: %{org: %Ecto.Association.NotLoaded{}} = product}} = conn,
+        product: role
+      ) do
+    product = NervesHubWebCore.Repo.preload(product, :org)
+
+    conn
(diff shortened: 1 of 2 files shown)


apps/nerves_hub_web_core/lib/nerves_hub_web_core/role_validate_helpers.ex
```diff
@@ -7,11 +7,29 @@ defmodule NervesHubWebCore.RoleValidateHelpers do
     validate_org_user_role(conn, org, user, role)
   end
 
-  def validate_role(%{assigns: %{user: user, product: product}} = conn, product: role) do
-    if NervesHubWebCore.Products.has_product_role?(product, user, role) do
-      conn
-    else
-      halt_role(conn, "product " <> to_string(role))
+  def validate_role(
+        %{assigns: %{product: %{org: %Ecto.Association.NotLoaded{}} = product}} = conn,
```
ConnorRigby: Couldn't we add the preload to the `NervesHubWebCore.Accounts` module? I guess it doesn't matter where that happens.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/role_validate_helpers.ex
```diff
@@ -27,6 +45,14 @@ defmodule NervesHubWebCore.RoleValidateHelpers do
     end
   end
 
+  def validate_product_user_role(conn, org, user, role) do
+    if NervesHubWebCore.Accounts.has_org_role?(org, user, role) do
```
jjcarstens: Since this is `validate_product_user_role`, shouldn't this be `NervesHubWebCore.Products.has_product_role?/3`?

mobileoverlord: Ah, you're right, actually this whole function was added as a mistake.

Issue #348, opened 2019-01-29
Display CA certificates somewhere in the web app


fhunleth:
**Describe the bug**
CA certificates are important for device auto-registration, but they currently only show up in the CLI

**To Reproduce**
Add a bunch of CA certificates with `mix nerves_hub.ca_certificate register ...`. Go to the web UI.

**Expected behavior**
It's possible to list out CA certificates. It would be really useful to see how many times they were used and their expiration dates.

pcmarks:
Just want to make sure: the database stores three types of certificates: ca_certificates, device_certificates and user_certificates. This issue is referring to **ca_certificates**, right?

fhunleth:
Yes. 

Also, the Web UI already displays user certificates. Device certificates might be interesting to have, but they'd naturally go under the device informational screen.

pcmarks:
On a question of naming. 

1. Since ca certificates are tied to organizations, it would seem that the table `ca_certificates` could be renamed `org_certificates`. This is similar to the `user` and `user_certificate` tables. This would require adding some Ecto migrations, etc., however.

2. Leave the ca certificate table name the same but the Ecto module would be called `org_certificate` and in the code implement `org_certificate_controller`, `org_certificate_view`, etc.

I favor 2. 

Thoughts?

fhunleth:
I feel uneasy changing the name. `user_certificate`s are used to authenticate users. `ca_certificates` are not used to authenticate organizations. `ca_certificates` are used the first time devices connect to authenticate their device certificates.

fhunleth:
I might suggest `device_ca_certificate`, but that's getting long to type.

pcmarks:
Thanks for the clarification.

It turns out that ca certificates are already accounted for under devices, e.g., `NervesHubWebCore.Devices.CACertificates` So that can stay the same but there will be a ca_certificate controller, etc. that is reached via the organization settings page - if that makes sense. 

If you want, I can sketch out how the UI will behave and look like tomorrow.

I've already made the changes that allow one to see the ca certificates associated with an organization. It looks and works like the `User Certificates` button on the user settings page.

fhunleth:
That makes sense.  It sounds like you're putting the certificate list in the right place, so I don't think it's necessary for us to review how the UI behaves. If you're unsure, feel free to let us know. However, we did recently get an offer for help with the UI. I'm quite happy to postpone UI perfection for their help.

pcmarks:
[...]

Issue #315, opened 2018-12-17
CI / API User


danielspofford:
Right now you can use the local keys from a user to authenticate within CI or whatever. It is unfortunate that you either have to use _your_ keys, or go through the process of registering a user with an email, a website password, etc. just to use the local keys.

It would be nice if I could generate a CI / API "user" specific to an org that didn't require anything else like email, but provided me with a key and cert.

Note that by using _your_ user's cert and key that you are forced to give the CI more access than you might want (to other orgs for example).

ConnorRigby:
This is related to #317 i think. I was just planning on making a dummy email and stuff, but having an actual CI "User" is a valid option also.,

mobileoverlord:
I think I like the simplicity of using the User certificate infrastructure to grant access to servers, especially with the addition of roles on products and orgs. To make it easier to generate the user certificates, we might want to add the ability to create a User cert from the web site. This was brought up in #435

PR #435, opened 2019-04-18
Add ability to list Device (CA) Certificates

pcmarks:
Issue #348

Added button to Organization Settings page to take you to a list of 
Device (CA) cerftificates. Shown in the same format as User 
Certificates.
Took the liberty to change the User and Organization Settings page to 
display what type of page one is on - organization or user. For example 
the User Settings page now says: "<username> User Settings". Updated 
relevant unit tests.

Note issue will be created to make the "New" and "Delete" buttons on the Device (CA) Certificates page operational.

Also, removed new button from User Certificates page. New user certificates should only be created from the CLI.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/account_certificate/index.html.eex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/account_certificate/index.html.eex
@@ -1,4 +1,4 @@
-<h2><%= @user.username %> User Certificates <span><%= button "", class: "btn btn-success fa fa-plus", title: "Create New Account Certificate", to: account_certificate_path(@conn, :new), method: :get %></span></h2>
+<h2><%= @user.username %> User Certificates </h2>
 <div class="row">
   <div class="w-100 shadow nhw_list">
     <div class="card">
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_certificate/index.html.eex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_certificate/index.html.eex
@@ -1,8 +1,8 @@
-<h2><%= @user.username %> User Certificates <span><%= button "", class: "btn btn-success fa fa-plus", title: "Create New Account Certificate", to: account_certificate_path(@conn, :new), method: :get %></span></h2>
+<h2><%= @current_org.name %> Device (CA) Certificates <span><%= button "", class: "btn btn-success fa fa-plus", title: "Create New Account Certificate", to: account_certificate_path(@conn, :new), method: :get %></span></h2>
 <div class="row">
   <div class="w-100 shadow nhw_list">
     <div class="card">
-      <table id="account_certificates" class="table">
+      <table id="organization_certificates" class="table">
         <thead>
           <tr class="d-flex">
             <th class="col-2">Description</th>
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_certificate_controller.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_certificate_controller.ex
@@ -1,0 +1,13 @@
+defmodule NervesHubWWWWeb.OrgCertificateController do
+  use NervesHubWWWWeb, :controller
+
+  alias NervesHubWebCore.Devices
+
+  def index(%{assigns: %{current_org: org}} = conn, _params) do
+    conn
+    |> render(
+      "index.html",
+      certificates: Devices.get_ca_certificates(org)
+    )
+  end
+end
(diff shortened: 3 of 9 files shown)


fhunleth:
I spoke with @pcmarks and he'll be adding a short unit test for the device CA certs soon.

fhunleth:
Also, I'm not sure if there's any controversy around the user cert button removal change or the settings screen title change. If there is, let's split the PR so that the device CA list can get in there.

jjcarstens:
Whats the reasoning to restrict User Cert creation to the CLI?

fhunleth:
We may eventually want user cert creation in the Web UI, but right now, 1) it doesn't work (you'll get an internal server error), and 2) it doesn't feel like a priority since people using user certs are already pretty comfortable with the CLI.

mobileoverlord:
User cert creation via the web would be convenient when your org has a user that is used for CI, and you needed a simple way of generating a certificate for your servers. Otherwise you need to switch users on the CLI, which creates a new User cert each time.

fhunleth:
Is user cert creation on the web UI important enough for someone to create an issue for it?

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_certificate/index.html.eex
```diff
@@ -0,0 +1,41 @@
+<h2><%= @current_org.name %> Device (CA) Certificates <span><%= button "", class: "btn btn-success fa fa-plus", title: "Create New Account Certificate", to: account_certificate_path(@conn, :new), method: :get %></span></h2>
```
mobileoverlord: Should this `Create New Account Certificate` be Create New Organization Certificate

and `account_certificate_path` `org_certificate_path`?

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/account_certificate/index.html.eex
```diff
@@ -1,4 +1,4 @@
-<h2>Account Certificates <span><%= button "", class: "btn btn-success fa fa-plus", title: "Create New Account Certificate", to: account_certificate_path(@conn, :new), method: :get %></span></h2>
+<h2><%= @user.username %> User Certificates </h2>
```
jjcarstens: We might want to have it be `#{username} User Account Certificates` here to match the route path and what its trying to portray.

mobileoverlord:
@fhunleth 

> Is user cert creation on the web UI important enough for someone to create an issue for it?

The root issue I would be trying to solve is covered in this issue: 
https://github.com/nerves-hub/nerves_hub_web/issues/315

I'll update the comments there.

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_certificate/index.html.eex
fhunleth: Wouldn't that be confusing since device CA certificates don't authenticate organizations?

fhunleth:
I'm going to close the PR. @pcmarks - could you please break the PR up so that we can review one change at a time?

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_certificate/index.html.eex
mobileoverlord: Oh I get it, this is to take you to the user certificate creation for the org.

mobileoverlord: I take back my comments

pcmarks: Yes, it should be.

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/account_certificate/index.html.eex
pcmarks: Right now there is no `org#show` action, only `org#edit`. Similarly, there is no `account#show` action, only an `account#edit` action. 

Is the consensus that I add a `org#show` action along with its template that would show the list of certificates? But, I think that begs the question: should the `org#edit` page also show the list of certificates **and** allow for the creation of a device (ca) certificate? My guess is that that would be a different issue.

fhunleth: I don't think that I care on what file the code is put in, but registering a device CA certificate is a big ordeal that is much more than filling out some database fields. My opinion is that we wrap this up as @pcmarks wrote it so that the device CA certs can actually be listed. The rest can be done in followup PRs, but I'd recommend we review the impact of properly creating/registering device CA certs together first.

PR #440, opened 2019-04-18
Record audit for deployment changes

jjcarstens:
* Records audit for deployment `update` and `create` events
  * little cleanup around the controllers. No functionality change except for `audit`
* Records audit log for a device if a deployment change triggers an update
* Creates `NervesHubWebCore.AuditLogs` context
  * moves `AuditLog` from `Accounts` to `AuditLogs` context
  * adds `audit/4` and `audit!/4` functions to serialize an audit log
* Adds `changes` field to `AuditLog` to store what changes happened on an `update` action

With this, I can start displaying audit events in a data feed on `device#show` so one can quickly see what's been happening with the device (and maybe who did it).

`AuditLog` started to feel like it didn't fit within the `NervesHubWebCore.Accounts` context, so I moved it into its own `AuditLogs` context under the idea that its been designed more polymorphic and may not always be a log within an account space. What do you think?


mobileoverlord:
❤️ This looks great!

PR #441, opened 2019-04-19
Display Device (CA) Certificates

pcmarks:
Issue #348 

Device (CA) certificates are displayed on their own page. The page can be reached from the Organization Settings page via the button "Device (CA) Certificates.

No adding or deleting of these certificates are available from this page

Unit test added (only :index)


mobileoverlord:
Looks good, thanks!

Issue #427, opened 2019-04-11
Add the ability to get a list of users associated with an organization.


pcmarks:
There is currently no way to get a list of the users that are associated  with an organization. A good place might be the organization settings page: Settings > Organization Settings.

PR #443, opened 2019-04-19
Display organization's users in web interface

pcmarks:
Issue: #427

From the Organizations Setttings page select the "Users" button. A list 
of the organization's users is shown on a separate page.

One unit test added.


apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org_user/index0.html.eex
```diff
@@ -0,0 +1,41 @@
+<h2>Account Certificates</h2>
```
fhunleth: This looks like it was accidentally added to the commit.

pcmarks: Should be gone now.

PR #444, opened 2019-04-19
Add audit log feed to device#show

jjcarstens:
* Adds `_audit_log_feed.html.eex` template to be reusable in other views
  * Adds `AuditLogView` with helpers used in `_audit_log_feed.html.eex`
* Adds pagination to the audit log feed in `device#show` view
* few sprinkles of CSS (as much as my sanity can handle)

I'm clearly not a designer, but I think its a good first step. Basically the lines just show `actor -> resource` relationship. If you click on it, you can see the `params` and `changes` of the audit log as well. Also, if the actor/resource is not the displayed item, then it is hyperlinked to that resource (well...sort of. Deployments and resources are scoped to a product and I don't have that in the `AuditLog` record. Decided to let that sit for now and decide if we want to do a bunch of query lookups for it later on)

As a bonus, its paginated here. So for the device#show, its inclusive and you can see every audit log for it.

device_audit_feed


jjcarstens:
Updated for builds to use elixir 1.8.1 and all cylinders are firing now 😎

fhunleth:
I realize that this is just getting the feature in and that there's certainly more to be done. Did you want me to comment on the format of the log messages or hold off?

jjcarstens:
Comment away (or make issue). It took me a good amount of time just to get this so I opted for getting it in and adjusting format gradually

fhunleth:
Sounds good. This is totally awesome. I'm so glad that you pushed this since the current state is definitely usable. I'll save a note for next week. I kind of want to savor the current progress for a bit longer before adding any new issues.

Issue #411, opened 2019-04-05
Unreferenced firmware not being garbage collected


fhunleth:
**Describe the bug**
Old firmware images that aren't referenced by a deployment are not being garbage collected after months.

**Expected behavior**
Unreferenced firmware images eventually get garbage collected.

mobileoverlord:
This was fixed in https://github.com/nerves-hub/nerves_hub_web/pull/447#issuecomment-485516343

mobileoverlord:
The firmware ttl dates were set starting today. It will take up until the default ttl time (7 days) for the firmware to be deleted.

PR #447, opened 2019-04-22
Use Firmwares.delete_firmware/1 when garbage collecting

mobileoverlord:
Fixes an issue where the database record would be cleaned up but the files would not be deleted.


jjcarstens:
Does this also resolve https://github.com/nerves-hub/nerves_hub_web/issues/411?

Issue #403, opened 2019-04-02
Email all users of an organization when a user is added or removed


fhunleth:
Updates to organizational settings should notify all users of that organization as a precaution against accidental or malicious changes.

I expect there to be more places, but the intention for this issue to only apply to users being added to or removed from organizations.

pcmarks:
For adding a user there are two routes and controller support:

`POST /invite  => AccountController.accept_invite()`
`POST /register => AccountController.create()`

However, there are no routes nor any controller support for removing a user. 

Should the creation of remove user routes and controllers be in a separate issue? If so, then this issue should be limited to adding a user - not removing one.

fhunleth:
Take a look at `/orgs/:org_name/users/:username`. See https://github.com/nerves-hub/nerves_hub_web/blob/master/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/org_user_controller.ex.

At the commandline, you can run `mix nerves_hub.org user add|remove USERNAME` to exercise that endpoint. It would be nice to remove users from organizations from the web UI as well and I'd agree that that's a separate issue.

pcmarks:
OK. I was focused on the UI interface and forgot about the API.

jjcarstens:
This seems to have been implemented. I am getting notifications when new user joins org

PR #452, opened 2019-04-26
Email all on add remove user

pcmarks:
Issue #403 

When a user is added or removed from an organization, either by an API call or (in the future) via the the web UI, an email is sent to all other users in that organization notifying them of this action. The subject line includes the username of the initiator of the action.

Unit tests added.

### DIFF
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/org_user_controller.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/org_user_controller.ex
@@ -2,6 +2,8 @@ defmodule NervesHubAPIWeb.OrgUserController do
   use NervesHubAPIWeb, :controller
 
   alias NervesHubWebCore.Accounts
+  alias NervesHubWWW.Accounts.Email
+  alias NervesHubWWW.Mailer
 
   action_fallback(NervesHubAPIWeb.FallbackController)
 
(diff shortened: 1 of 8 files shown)


apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/org_user_controller.ex
```diff
@@ -35,6 +44,13 @@ defmodule NervesHubAPIWeb.OrgUserController do
     with {:ok, user} <- Accounts.get_user_by_username(username),
          {:ok, _org_user} <- Accounts.get_org_user(org, user),
          :ok <- Accounts.remove_org_user(org, user) do
+      # Now let everyone in the organization - except the one being removed -
+      # that this user has been removed from the organization.
```
fhunleth: I think it's nice to have the person being removed get an email too in case the removal was unintentional or malicious.

pcmarks: You're right. Let me make that change.

apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/org_user_controller.ex
```diff
@@ -35,6 +44,13 @@ defmodule NervesHubAPIWeb.OrgUserController do
     with {:ok, user} <- Accounts.get_user_by_username(username),
          {:ok, _org_user} <- Accounts.get_org_user(org, user),
          :ok <- Accounts.remove_org_user(org, user) do
+      # Now let everyone in the organization - except the one being removed -
```
mobileoverlord: Everything looks good, can we update this comment now that it emails everyone?

pcmarks: Yes, let me fix that.

PR #454, opened 2019-04-29
Return device firmware metadata

pcmarks:
Return a device's firmware metadata on an API request.

This change is necessary to support https://github.com/nerves-hub/nerves_hub_cli/issues/102


fhunleth:
Could you rebase on `master`?

pcmarks:
I forgot to remove that one commit: "tell org users...". I will do it shortly.

Issue #419, opened 2019-04-08
Put deployment into a "needs attention" state when two many errors are reported


fhunleth:
If a firmware update fails on too many devices, it should be possible for NervesHub to stop the deployment on its own. This would need to be configurable based on the use case since some organizations could be more sensitive to stopping deployments automatically than others. 

This depends on #418.

Issue #418, opened 2019-04-08
Put devices into a "needs attention" state when too many failures


fhunleth:
Currently, it's possible for a device to get into a loop where it repeatedly gets told to update its firmware. NervesHub assists in making sure this doesn't happen by checking firmware signatures, metadata, etc. for compatibility, but in the end, the device has the final say. There could also be other reasons why firmware images don't update successfully like really flaky network connections or something buggy happening on the device that reboots it mid-update. Whatever the reason, if a device fails n times in m minutes, then it should be put into a "needs attention" state where NervesHub ignores it for a while (maybe an hour or a day) or until the operator intervenes and clears the condition.

fhunleth:
I assume that it will be best to break out failure detection to other issues. The failure that most interests me is when a device is total to update to the same firmware uuid more than twice in 10 minutes. (The threshold is arbitrary.) This would catch most of the errors I'm concerned about even though I know we can do better if we update the client to report more detail on errors.

PR #455, opened 2019-04-30
Mark deployments and devices that need attention

jjcarstens:
First attempt at https://github.com/nerves-hub/nerves_hub_web/issues/418 and https://github.com/nerves-hub/nerves_hub_web/issues/419

* Adds new fields to `Deployment` schema
  * :device_failure_threshold
  * :device_failure_rate_seconds
  * :device_failure_rate_amount
  * :failure_threshold
  * :failure_rate_seconds
  * :failure_rate_amount
  * :healthy
* Adds `healthy` bool to `Device` schema

The new `deployment` fields are to calculate a failure rate and failure threshold at both device and deployment levels.
The `failure rate` == X number of device update failures within Y seconds
The `failure threshold` is total number of devices in a `healthy == false` state that are permitted. Devices get in a `healthy == false` state when they attempt to update from the same deployment and exceed the `device_failure_threshold` (total number of times its allowed to attempt to update) or `device_failure_rate` (X attempted updates in Y seconds)

This is a good start, but not perfect. It will help cover some cases to get the ball rolling. But I would consider it a little inaccurate since to make some of these assertions we're relying on looking up device updates in the audit log with `send_update_message: true` in order to say "that device is attempting the same up multiple times and is probably failing 🤔". I went with this route because we mostly tell devices to update via a broadcast with no feedback from the device, but it can probably be approved.

Next step will be to get some feedback from devices about its attempts or connections to try some stronger asserts _first_ on what is breaking before checking loose audit events to see what might be going on.

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
@@ -6,7 +6,16 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
   """
 
   use NervesHubDeviceWeb, :channel
-  alias NervesHubWebCore.{Accounts.Org, Devices, Devices.Device, Firmwares}
+
+  alias NervesHubWebCore.{
+    Accounts.Org,
+    AuditLogs,
+    Deployments.Deployment,
+    Devices,
+    Devices.Device,
+    Firmwares
+  }
+
   alias NervesHubDevice.Presence
 
   intercept(["presence_diff"])
@@ -24,7 +33,15 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
          {:ok, device} <- Devices.received_communication(device) do
       Phoenix.PubSub.subscribe(NervesHubWeb.PubSub, "device:#{device.id}")
       deployments = Devices.get_eligible_deployments(device)
-      join_reply = Devices.resolve_update(device.org, deployments)
+      join_reply = Devices.resolve_update(device, deployments)
+
+      if join_reply.update_available do
+        AuditLogs.audit!(hd(deployments), device, :update, %{
+          from: "channel_join",
+          send_update_message: true
+        })
+      end
+
       send(self(), {:after_join, device, join_reply.update_available})
       {:ok, join_reply, socket}
     end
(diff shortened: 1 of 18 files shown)


apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex
```diff
@@ -81,27 +80,33 @@ defmodule NervesHubWebCore.Devices do
...
-      _error ->
-        []
-    end
+    from(
+      d in Deployment,
+      where: d.is_active,
+      where: d.needs_attention == false or is_nil(d.needs_attention),
+      join: f in assoc(d, :firmware),
+      join: p in assoc(d, :product),
+      where: p.org_id == ^device.org_id and p.name == ^meta.product,
```
jjcarstens: I adjusted this query here to match up the product in the same query and avoid the extra Product lookup we were doing.

I figure with all this AuditLog/rate lookups now, we might as well save a query when we can ¯\\_(ツ)_/¯

apps/nerves_hub_web_core/priv/repo/migrations/20190429171105_add_rate_and_thresholds_to_deployments.exs
```diff
@@ -0,0 +1,17 @@
...
+
+  def change do
+    alter table(:deployments) do
+      add(:device_failure_threshold, :integer, default: 3)
+      add(:device_failure_rate_seconds, :integer, default: 180)
+      add(:device_failure_rate_amount, :integer, default: 5)
+
+      add(:failure_threshold, :integer, default: 50)
+      add(:failure_rate_seconds, :integer, default: 300)
+      add(:failure_rate_amount, :integer, default: 5)
```
jjcarstens: These defaults are totally arbitrary. Totally open to ideas here.

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
```diff
@@ -25,6 +25,11 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
       Phoenix.PubSub.subscribe(NervesHubWeb.PubSub, "device:#{device.id}")
       deployments = Devices.get_eligible_deployments(device)
       join_reply = Devices.resolve_update(device.org, deployments)
+
+      if join_reply.update_available do
+        AuditLogs.audit!(hd(deployments), device, :update, %{send_update_message: true})
```
jjcarstens: Audit the update message here to match how we audit when a device or deployment gets updated and triggers a firmware update on the device

jjcarstens:
FYI, I've updated a few things here and its ready to go! 🎉 Appreciate any 👀

fhunleth:
I like the health status icon by the deployments and devices a lot! I feel like I understand what's happening and a really high level and it seems good to me.

PR #461, opened 2019-05-01
Fix devices#index status weirdness

jjcarstens:
Turns out when I moved the devices#index page to LiveView, I forgot to remote the js
code connecting to a socket and watching for presence change events. So every now and
then there was a race condition between the JS updating devices statuses on the page
and LiveView.

So to fix, :boom: the socket javascript!

(Thanks to @mobileoverlord and @chrismccord for helping find this craziness)


jjcarstens:
So previously I updated `phoenix_live_view` but turns out there is a new bug there (see https://github.com/phoenixframework/phoenix_live_view/issues/201) which breaks all of our specs.

So I removed those updates and will work on cleaning that up later. For now, this is good to go

PR #463, opened 2019-05-03
Add rollbar crash report support

mobileoverlord:
Add `rollbax` for Rollbar support. This will capture crashes and send reports to rollbar.


fhunleth:
Is there another way to enable this? It doesn't feel right forcing users running their own NervesHub instances to subscribe to Rollbar.

mobileoverlord:
We should revisit this when we update to elixir 1.9. Runtime config in 1.9 will make it a lot easier for us to conditionally include / disable this.

PR #464, opened 2019-05-06
Add fwup progress when device is updating

jjcarstens:
Requires https://github.com/nerves-hub/nerves_hub/pull/97

Displays a fwup progress based on messages from `nerves_hub` while the device is updating.
Shows progress in `device#show` and `device#index` views.

A good visual check that your deployment is doing what you want and devices are reacting

fwup_progress


mobileoverlord:
🚢 🇮🇹

PR #469, opened 2019-05-09
Update LivView and change session handling

jjcarstens:
A recent update of Phoenix.LiveView changed how tests were getting session data
and subsequently broke all of our LV tests because our sessions were so yyuuge
(which is bad LV practice, I now know).

So this updates LV and adjusts all their routes and session handling to be minimal
viable data to recreate the session vs what we _were_ doing which was passing full
structs and tons of data into the session which was being stored as a huge hashed
hunk of info on a LV page.

The only real implementation addition is in the `NervesHubWWWWeb.Router` to pipe
live paths through role validation so we can fail the connection _before_ attempting
and LV mounts and lookups like the controllers do.

Also, removes unused controller actions for the web.

Sorry it's so big. I could try to break it up more, but thought it might be best to fix all the LV stuff now in one big swoop


jjcarstens:
I honestly have no idea why `build-api` and `build-www` would be failing suddenly. I'm 100% positive I didn't change anything with fwup.

mobileoverlord:
Looks like fwup is 1.3.1 now, I’m unsure why older versions are no longer available after updates.

mobileoverlord:
🐑 🇮🇹

PR #476, opened 2019-05-20
Scope failure queries to firmware UUID

### LINKED ISSUE #474: Make device threshold lookup firmware instead of deployment
So looks like a common behavior for firmware updates may be to just update an existing deployment with a new firmware (thanks to @fhunleth for pointing that out..). But the initial checks for threshold assume a new firmware == new deployment.

The way we check device failure threshold is by the deployment id in an audit log, so using the same deployment n times > threshold will fail and put device in an unhealthy state.

We should instead link an update message to a specific firmware uuid so device/deployment threshold/rate is calculated for that specific fw, no matter what deployment the fw gets applied to

jjcarstens:
Fixes #474

Adds `firmware_uuid` to an Audit Log event created when a device is told to update.
Then when we use the AuditLog to check failure rates and thresholds, we check for that UUID
as well to ensure we're correctly matching failures. This allows reusuing deployments with
different firmwares and adds little more accuracy to device failure checking.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/audit_logs/audit_log.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/audit_logs/audit_log.ex
@@ -4,7 +4,11 @@ defmodule NervesHubWebCore.AuditLogs.AuditLog do
   import Ecto.Changeset
   import EctoEnum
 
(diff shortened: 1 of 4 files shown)


jjcarstens:
O right..totally forgot about the tests. Let me get that in quick...

jjcarstens:
Updated

PR #477, opened 2019-05-21
Use human friendly descriptions for audit logs

jjcarstens:
Changes the audit log feed to use a more easily understandable phrase about what event occurred
that was audited instead of icons with a click-to-expand behavior. This description is generated
when creating the audit log and stored in the record.

Also changes it to look more like a feed (i guess?)

Includes a script in `apps/nerves_hub_web_core/priv/repo/migrations/20190520202940_add_description_to_audit_logs.exs` to be run after the new field is added via migration.

Screen Shot 2019-05-21 at 3 01 43 PM


mobileoverlord:
Looks good.

PR #478, opened 2019-05-22
Add instructions on installing asdf

aforward:
Taken from the nerves training PDF.  The documentation refers to `asdf` so I thought it might be useful to link in the instructions to getting that up and running.


aforward:
Done.

fhunleth:
Thanks!

PR #479, opened 2019-05-23
Support remote iex lines that are lists

jjcarstens:
A recent change in https://github.com/nerves-hub/nerves_hub/pull/102 added support to handle
io_requests requiring a function to be run to format the data. In some cases, that ends up
being a list of characters or numbers (representing a character) that would break the console
session because to regex is used to generate the ANSI HTML and it couldn't handle the list.

So this just calls ~~`to_string/1`~~ `IO.iodata_to_binary/1` on data that comes in from the device before attempting to convert
to HTML and display.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/console.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/console.ex
@@ -84,7 +84,13 @@ defmodule NervesHubWWWWeb.DeviceLive.Console do
         %Broadcast{event: "put_chars", payload: %{"data" => line}},
         %{assigns: %{lines: lines}} = socket
       ) do
-    line = AnsiToHTML.generate_phoenix_html(line, @theme)
+    # Lines may come in as a list of characters and/or numbers to represent
+    # a string. So call to_string here just to convert before attempting
+    # any transposing to HTML which relies on strings.
+    line =
+      IO.iodata_to_binary(line)
+      |> AnsiToHTML.generate_phoenix_html(@theme)
+
     new_lines = List.insert_at(lines, -1, line)
     {:noreply, assign(socket, :lines, new_lines)}
   end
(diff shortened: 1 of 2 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/console.ex
```diff
@@ -84,7 +84,13 @@ defmodule NervesHubWWWWeb.DeviceLive.Console do
         %Broadcast{event: "put_chars", payload: %{"data" => line}},
         %{assigns: %{lines: lines}} = socket
       ) do
-    line = AnsiToHTML.generate_phoenix_html(line, @theme)
+    # Lines may come in as a list of characters and/or numbers to represent
+    # a string. So call to_string here just to convert before attempting
+    # any transposing to HTML which relies on strings.
+    line =
+      to_string(line)
```
fhunleth: So I didn't expect `to_string/1` to work, but it definitely seems like it does. The route I'm familiar with is calling `IO.iodata_to_binary/1` to flatten iodata. The list implementation of `to_string/1` calls into Erlang's unicode module... probably ok, but not sure. Would you mind switching to `IO.iodata_to_binary/1`?

jjcarstens: No problem. Updated.

jjcarstens: For record, it looks like the difference is that `IO.iodata_to_binary` does not attempt any encoding  and treats integers in a list as raw bytes.

https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/io.ex#L572-L573

While `to_string` and `List.to_string` call unicode which attempts to strict convert a chartist to a utf-8 encoding string.

fhunleth: Thanks for looking that up.

PR #480, opened 2019-05-23
Adjust line spacing for IEx console

jjcarstens:
I don't know what I'm doing, but this condenses lines closer to better match output
you would see in the terminal. Should make things a little easier to read for now

**Before**
before

**After**
after


jjcarstens:
Hold on this for one second..I have a better fix..

jjcarstens:
K, adjusted one more thing to clean up extra unneeded newlines that are coming in.

PR #486, opened 2019-05-28
Update Coveralls token again

fhunleth:


fhunleth:
FYI - While debugging this, I changed the CircleCI token so the one listed above isn't right any longer.

jjcarstens:
So this was [initially added](https://github.com/nerves-hub/nerves_hub_web/pull/396/files#diff-1d37e48f9ceff6d8030570cd36286a61R92) so that people making PR's from a fork would be able to have tests complete. It seems like maybe this setting in Circle gets precedence, but from a fork that secret token gets omitted from the job and we'd want this in

fhunleth:
:scream:

fhunleth:
Alrighty - I deleted the token from CircleCI's environment so that there is one and only one place for it!

PR #488, opened 2019-06-03
Remove dashboard and default to product index

mobileoverlord:
This PR removes the dashboard page and changes the initial page to the product index. The jumbotron from the dashboard will display on the product index if there are no products. Once a product is created, the index shows the product list and you are no longer presented with the "lets get started by creating a new product" screen.


mobileoverlord:
/products with no products

Screen Shot 2019-06-03 at 11 54 30

/products after creating a product
Screen Shot 2019-06-03 at 11 55 44

PR #499, opened 2019-06-20
Move devices under products

mobileoverlord:
Sorry for the massive PR, but I'm unsure of how to make this change any smaller.

This PR does the following

  * Add `product` association to `device`
  * Move devices under products

The biggest impact that this has is that the api routes would change and require changes to `nerves_hub_user_api`

Device management endpoints moved from
```
/orgs/≤org_name>/devices
```
to
```
/orgs/≤org_name>/products/<product_name>/devices
```


apps/nerves_hub_device/lib/nerves_hub_device/presence.ex
```diff
@@ -25,10 +25,11 @@ defmodule NervesHubDevice.Presence do
 
   def find(device, default \\ nil)
 
-  def find(%Device{id: device_id, org_id: org_id}, default) do
-    "devices:#{org_id}"
+  def find(%Device{id: device_id, product_id: product_id}, default) do
+    "product:#{product_id}:devices"
     |> Presence.list()
     |> Map.get("#{device_id}", default)
+    |> merge_metas
```
jjcarstens: I don’t think you need this `merge_metas` because the `Presence.list/1` above will call fetch, which does the meta merging

PR #500, opened 2019-06-20
Add example gettext

Theuns-Botha:


mobileoverlord:
❤️ 💚 💙

fhunleth:
@Theuns-Botha Thanks for contributing this! I suspect that we'd merge this regardless, but if you have a regression test for the language selection piece, that would be helpful.

mobileoverlord:
This is looking really good. It appears that we are going to need to update our CI automation to run `mix gettext.merge priv/gettext` when building the production container?

Theuns-Botha:
That is debatable, I suspect a developer that is working on the UI should actually run this and verify his/her own work. Which would negate the usefulness of running it separately.

mobileoverlord:
Ah I understand, so the po files are checked into git then.

Theuns-Botha:
Correct

mobileoverlord:
Ah! I edited the gitignore only to realize it already had a `/priv/static` entry. can you please force push and I'll merge :)

Theuns-Botha:
I will push when I am back home again(I am at work and the code is on my personal PC)

fhunleth:
No need to do anything. I fixed and merged it manually. Thanks!

PR #505, opened 2019-06-21
Add Crowdin configuration

fhunleth:
[skip ci]


fhunleth:
Not sure on this at all, but it seems like I have to have the config file in `master` for Crowdin to work. :(

PR #514, opened 2019-06-22
Fix navigation on drop down when switching products between orgs

mobileoverlord:


mobileoverlord:
Jumping between products in different orgs does not update the current org. Here is an example of switching from `letote | kiosk` to `nerves-project | system_test`

Screen Shot 2019-06-22 at 06 29 51
Screen Shot 2019-06-22 at 06 30 02

Then if you click on a button like "View Devices" you get an internal server error.

PR #516, opened 2019-06-24
Fix device channel topics

jjcarstens:
With recent change to scope devices to product, some device channel topics were changed and a few were missed, so this just fixes that.


fhunleth:
This is all internal to the web server, right? I.e., no device API changes?

jjcarstens:
Correct. Internal topics for NervesHub and Presence.

PR #523, opened 2019-06-25
Add catch-all for LiveView mounts

jjcarstens:
When we change anything about the session structure of a LiveView, it causes tons of rollbar errors after deployment because any existing browser tabs/windows are trying to reuse the old session structure which no longer has a matching `mount/2` clause for it.

So, this adds a catch-all to redirect to the home path if that apps.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/live/deployment_live/show.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/live/deployment_live/show.ex
@@ -27,6 +27,19 @@ defmodule NervesHubWWWWeb.DeploymentLive.Show do
     {:ok, socket}
   end
 
+  # Catch-all to handle when LV sessions change.
+  # Typically this is after a deploy when the
+  # session structure in the module has changed
+  # for mount/2
+  def mount(_, socket) do
+    socket =
+      socket
+      |> put_flash(:info, "The software running on NervesHub was updated to the latest version")
+      |> redirect(to: Routes.home_path(socket, :index))
+
+    {:stop, socket}
+  end
+
   def handle_event(
         "delete",
         _val,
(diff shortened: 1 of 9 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/live/deployment_live/show.ex
```diff
@@ -27,6 +27,18 @@ defmodule NervesHubWWWWeb.DeploymentLive.Show do
     {:ok, socket}
   end
 
+  # Catch-all to handle when LV sessions change.
+  # Typically this is after code changes to the session
+  # in the a LiveView module is deployed
+  def mount(_, socket) do
+    socket =
+      socket
+      |> put_flash(:info, "LiveView session was changed")
```
fhunleth: What's the nice phrase that websites use for this? "Updated session to latest version" or something like that? Also, does it have to redirect to the home path or can it reload the current page?

jjcarstens: I don't know of any standard nice phrase that websites use here. Definitely open to suggestions.

For the redirect, the home path is the safest. The socket has _very_ little available to it and if we get to this point, we have no idea what the session looks like or what we can grab from it. Maybe more can be done to get some sort of path or current page from the conn? Not sure..would need more research

fhunleth: Ok, then I'm fine with redirected to home, then.

For the messaging, maybe "The software running on NervesHub was updated to the latest version".

jjcarstens: Changed

PR #539, opened 2019-07-21
Handle live view mount exceptions

mobileoverlord:
If a product, device, or any other database record that a live view required to mount is deleted, and a page remains open, live view will raise and continue to attempt to mount. This will redirect any exceptions back to the home.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web.ex
@@ -58,6 +58,25 @@ defmodule NervesHubWWWWeb do
       }
 
       alias Phoenix.Socket.Broadcast
+
+      defp socket_error(socket, error, opts \\ []) do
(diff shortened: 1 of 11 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web.ex
```diff
@@ -58,6 +58,25 @@ defmodule NervesHubWWWWeb do
...
+        socket =
+          socket
+          |> put_flash(:info, error)
+          |> redirect(to: redirect)
+
+        {:stop, socket}
+      end
+
+      def live_view_error(:update) do
+        "The software running on NervesHub was updated to the latest version"
```
fhunleth: Should this say something like "Reload the page to continue" or something like that?

fhunleth: Oh, I see. It redirects. So you don't need to do anything. In that case, my only comment is to add a period to end the sentence.

PR #544, opened 2019-08-01
Staging environment

mobileoverlord:
This PR adds the ability to deploy to the staging environment and adds an approval phase to production deployments.


mobileoverlord:
Quantum was causing issues with scaling the cluster up, especially from a cold start. I would like to replace Quantum with another job scheduling solution. Until then, I'll be running the firmware clean up job manually.

PR #545, opened 2019-08-09
Move email functionality into core

jjcarstens:
This is to help decouple api from www.

No functionality change. Just moving files around for now


apps/nerves_hub_device/config/release.exs
```diff
@@ -30,7 +30,7 @@ config :ex_aws, region: System.fetch_env!("AWS_REGION")
 
 config :nerves_hub_device, NervesHubDeviceWeb.Endpoint, server: true
 
-config :nerves_hub_www, NervesHubWWW.Mailer,
+config :nerves_hub_www, NervesHubWebCore.Mailer,
```
mobileoverlord: `nerves_hub_www` -> `nerves_hub_web_core`

apps/nerves_hub_web_core/lib/nerves_hub_web_core/views/email_view.ex
```diff
@@ -1,7 +1,12 @@
...
-  use NervesHubWWWWeb, :view
+defmodule NervesHubWebCore.EmailView do
+  use Phoenix.View,
+    root: "lib/nerves_hub_web_core/templates",
+    namespace: NervesHubWebCore
+
+  import Phoenix.HTML
 
   def base_url do
+    # TODO: We probably should not tie this to NervesHubWWWWeb?
```
mobileoverlord: This part would have to get fixed. The configuration for `nerves_hub_www` is not available for `nerves_hub_api` and causes an enumeration on `nil`. This breaks any api requests that triggers the delivery of en email.

apps/nerves_hub_device/config/release.exs
jjcarstens: Fixed

apps/nerves_hub_web_core/lib/nerves_hub_web_core/views/email_view.ex
jjcarstens: Changed the configs for `nerves_hub_web_core` up a bit to account for this. Right now, its hard coded so it could get out of sync with the nerves_hub_www endpoint, but once I get #546 figured out then we will have one spot to make shared values for these 2 apps at compile time.

mobileoverlord:
❤️ Thanks!

PR #547, opened 2019-08-13
test using quantum random strategy

jjcarstens:


jjcarstens:
Fun experiment, didn't work.

PR #553, opened 2019-08-16
Use Oban for job scheduling

jjcarstens:
* Adds `:oban` as dep for jobs
* Includes migrations required to use oban
* ~~repurposes `NervesHubWebCore.Scheduler` to handle queuing jobs based on cron schedules~~
* creates `NervesHubWebCore.Worker` behaviour to use with workers
* Moves previous workers into `NerveHubWebCore.Workers` scope

~~The scheduling is basic - It parses the cron text, gets the next datetime it should be run, creates a job record so oban will pick it up and run at the determined time, then reschedules a message to the process to schedule the next job. Scheduling is judicious so no job is prevented by another failing to be scheduled.~~

Totally adjusted this PR to follow some of the ideas from sorentwo here:
* https://sorentwo.com/2019/08/02/oban-recipes-part-3-reliable-scheduling.html

TL;DR - the first time a job runs, it will schedule the next job to run in the future creating a continuous loop

It gets a little heavy on the meta so any new workers can simply use the `NervesHubWebCore.Worker` behavior with an implementation and not have to think too hard about the scheduling.


config/test.exs
```diff
@@ -67,6 +67,8 @@ config :nerves_hub_web_core, NervesHubWebCore.CertificateAuthority,
 
 config :nerves_hub_web_core, NervesHubWebCore.Mailer, adapter: Bamboo.TestAdapter
 
+config :nerves_hub_web_core, Oban, queues: false, prune: :disabled
```
mobileoverlord: We might want to prune these records. I could see them piling up pretty quickly. Lets keep ~ 2 weeks

apps/nerves_hub_web_core/lib/nerves_hub_web_core/application.ex
```diff
@@ -11,13 +13,65 @@ defmodule NervesHubWebCore.Application do
...
+        worker_queues =
+          for worker <- worker_modules() do
+            {worker.config[:queue], worker.config[:concurrent_jobs]}
+          end
+
+        Keyword.put(config, :queues, Keyword.merge(config_queues, worker_queues))
+    end
+  end
+
+  defp worker_modules() do
```
mobileoverlord: 🙈 i've thought about using this pattern so many times

config/test.exs
jjcarstens: Done

mobileoverlord:
Lookin great, thanks 🙌

PR #565, opened 2019-10-02
Add org_id to AuditLog

mobileoverlord:
This PR adds `org_id` to `firmwares`, `device_certificates`, `deployments`, and `audit_logs`. This reference is being added so it is easy to get to when creating audit logs on resources of these types. This PR also adds an audit log entry any time a device connects to the server.


apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/show.ex
```diff
@@ -1,6 +1,6 @@
 defmodule NervesHubWWWWeb.DeviceLive.Show do
   use NervesHubWWWWeb, :live_view
-
+  require Logger
```
jjcarstens: Is this leftover from some testing or do we need logger here?

apps/nerves_hub_web_core/mix.exs
```diff
@@ -50,7 +50,7 @@ defmodule NervesHubWebCore.MixProject do
       {:plug, "~> 1.7"},
       {:plug_cowboy, "~> 2.0"},
       {:ecto_sql, "~> 3.0"},
-      {:ecto_enum, "~> 1.2"},
+      {:ecto_enum, github: "dnsbty/ecto_enum"},
```
jjcarstens: Is this intentional or a different enum library?

apps/nerves_hub_web_core/priv/repo/migrations/20191002193424_add_oban_beats_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule NervesHubWebCore.Repo.Migrations.AddObanBeatsTable do
```
jjcarstens: Just for records sake, what is this “oban beats” table?

apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/show.ex
mobileoverlord: I think its safe that this can be removed.

apps/nerves_hub_web_core/mix.exs
mobileoverlord: This is intentional. Updating Ecto requires that `Ecto.Type`s all define these two additional callbacks. The main project for `ecto_enum` has not been updated yet. I guess it would be safer for one of us to maintain a fork of this fork.

apps/nerves_hub_web_core/priv/repo/migrations/20191002193424_add_oban_beats_table.exs
mobileoverlord: whiskey and headphones.

mobileoverlord:
I split this apart into 3 PRs

PR #566, opened 2019-10-03
Update deps and Ecto.Types

mobileoverlord:
Update deps and Ecto. Also adds migration for `oban` tables and adds `embed_as /1` and `equal?/2` to Ecto.Type(s)


apps/nerves_hub_web_core/mix.exs
```diff
@@ -50,7 +50,7 @@ defmodule NervesHubWebCore.MixProject do
       {:plug, "~> 1.7"},
       {:plug_cowboy, "~> 2.0"},
       {:ecto_sql, "~> 3.0"},
-      {:ecto_enum, "~> 1.2"},
+      {:ecto_enum, github: "mobileoverlord/ecto_enum"},
```
fhunleth: 😱 

What happened with `:ecto_enum`?

mobileoverlord: Ecto was updated and requires that `Ecto.Type`s define two additional callback functions. The main library has not merged the PRs yet that fix this. I forked the fork that fixes the issue for now.

PR #569, opened 2019-10-10
Allow authenticating with username and email

### LINKED ISSUE #473: Allow username or email for login
When a user registers for NervesHub, they create a username and an email address with their account.

The username basically never gets used, mentioned, or displayed again in the web experience and login is based on email address. In the CLI, username is used a little more frequently, but not much.

If we're requiring both, we should allow the user to login with either username or email address

hassanshaikley:
fixes #473 

💚

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/session/new.html.eex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/session/new.html.eex
@@ -5,9 +5,9 @@
     </div>
     <div class="modal-body">
       <div class="form-group">
-        <label for="email_input">Email</label>
-        <%= text_input f, :email, class: "form-control", id: "email_input" %>
-        <%= error_tag f, :email %>
+        <label for="email_or_username">Email/Username</label>
+        <%= text_input f, :email_or_username, class: "form-control", id: "email_or_username" %>
+        <%= error_tag f, :email_or_username %>
       </div>
 
       <div class="form-group">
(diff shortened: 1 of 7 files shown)


hassanshaikley:
If we look for username or email in `nerves_hub_web_core/accounts.ex` then we shouldn't need to adjust the form, right?

I'll get on these changes!

jjcarstens:
I think the form will need to be adjusted somewhat. Because right now it uses `email_tag` which attempts to validate the field before submitting:

image

And we'll want to change text to indicate `username` or `email` could be used prob.

hassanshaikley:
I see! Awesome.

hassanshaikley:
Alright, just pushed.

Some things I am not sure about:

1. Do I still match on username for the CLI? Or should I remove that code as it will work if you refer to the username parameter as "email" anyways.
2. Is it acceptable to keep calling it email on the front end or should I change it to `email_or_username`?

Thanks!

jjcarstens:
1. I think its fine to leave that match in for the CLI. Then they have the option to be more explicit or not
2. `email_or_username` might be a little more correct, but I'm good either way

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/session/new.html.eex
```diff
@@ -6,7 +6,7 @@
     <div class="modal-body">
       <div class="form-group">
         <label for="email_input">Email</label>
```
jjcarstens: I think we should change this label to `Email/Username` (or something of the sort)

hassanshaikley:
How's that!

hassanshaikley:
😄

mobileoverlord:
This has been merged into master manually, Thanks!

PR #572, opened 2019-10-13
allow spaces to delimit in tag field

supersimple:
Hi @jjcarstens et al,
I noticed tags were not allowed to be delimited by spaces. It seems they were also not allowed to contain spaces, so I whipped up this PR (i hope you dont mind not discussing it with you first) to support space and/or comma delimited lists in tags. 
If there is some other reason not to support this or if you just dont like it, feel free to close it.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/types.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/types.ex
@@ -10,18 +10,14 @@ defmodule NervesHubWebCore.Types do
 
     def cast(tags) when is_bitstring(tags) do
       tags
-      |> String.split(",", trim: true)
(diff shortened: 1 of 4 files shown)


supersimple:
I see. After reading that, it seems the key is to be consistent and to display them (show routes) with commas.
Would this PR be worthwhile if the CLI were updated to also support spaces?

fhunleth:
If we were to enable spaces in tags in the UI, we'd definitely want the CLI to support it as well.

We (maybe mostly me, though) got burned by spaces in tags several months ago. It seems like spaces in tags work well on other websites. I'm not super excited about supporting them in NervesHub, but I'll try to keep an open mind if someone really wants them.

supersimple:
grandpa

I wasnt aware of the backstory. I'll go ahead and close this.

Issue #557, opened 2019-09-11
Some passwords don't work


fhunleth:
**Describe the bug**
It has been reported twice that some passwords don't work with NervesHub.

It has been hypothesized that it's a certain character in the password. One person
used 1Password to generate word-based passwords like "gape-ubiquity-darn-icebound" and had trouble. Another person had a 37 character long password with `$` or `%` and `:`. 

**To Reproduce**
1. Create a new user from the commandline
2. Try to log into nerves-hub.org

**Expected behavior**
All passwords work or you get an error when specifying the password that something is invalid.

**Screenshots**
N/A

**Additional context**

PR #574, opened 2019-10-14
500 on invite

### LINKED ISSUE #472: Internal Server Error on creating account from invite
When creating an account from an emailed invite, I cause an Internal Server Error if I do not enter an email address in the user name field in the registration page.

Note that the registration was nonetheless successful with my email address as user name.

**To Reproduce**
1. Get an email invite
2. Click on link in invite
3. Enter a non-email in the user name field (e.g. jfcloutier)
4. Enter a password in the password field
5. Press "Register" button

**Expected behavior**
I would expect the user name field to be pre-filled with the email address the invite was sent to (it should probably also be read-only)
I should only need to enter a password and press "Register"

 - OS: Ubuntu 19.04
 - Browser: Chrome
 - Version 73.0.3683.86

supersimple:
Fixes #472 

I noticed this issue as well today. Turns out the conn.assigns does not hold the user info at this point, so it was raising. The username is readily available from the params though so it was an easy fix. i didnt have time to research when this might have started happening (or find the commit.) But I suspect if you have Honeybadger or Rollbar, etc hooked up it would be easier to track down by looking for when the `(KeyError) key :user not found in: %{}` errors started coming in.

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/account_controller.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/account_controller.ex
@@ -76,8 +76,7 @@ defmodule NervesHubWWWWeb.AccountController do
   end
 
   def accept_invite(conn, %{"user" => user_params, "token" => token} = _) do
-    clean_params =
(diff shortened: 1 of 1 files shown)


jjcarstens:
Closes #557

fhunleth:
Yay!!!! Thanks!

PR #583, opened 2019-12-11
Issue #582 - add user invite link to org users page

### LINKED ISSUE #582: Add link to invite user in /settings/{org}/users
Right now you can invite a user from `/settings/{org}/profile`, but not from `/settings/{org}/users`. Sometimes that can cause confusing since `users` is where many might look to first when trying to add a user to their org.

chazwatkins:
* Add "Invite user" link to an organization's users page when the organization is not the current user

closes #582


mobileoverlord:
Thanks for adding this. ❤️ This was merged manually to master.

PR #584, opened 2019-12-11
move firmware deletion error message into context

chazwatkins:
* Fix duplication of error message when deleting firmware that has an associated deployment
  * Move error message from `NervesHubWWWWeb.FirmwareController` and `NervesHubAPIWeb.FirmwareController` into `NervesHubWebCore.Firmwares.Firmware`
context to remove the need to update the error message in multiple places.


mobileoverlord:
Thanks! ❤️ Merged into master

PR #588, opened 2019-12-20
Fix connections for devices with trusted certificates

mobileoverlord:
This fixes the SSL verify_fun to support allowing certificates from devices with expired ca's. The CA cert needs to remain in the database even after it has expired for this to work.


mobileoverlord:
A little more information, `verify_fun` is called for each certificate in the chain, this was only anticipating it to be the ca certificate getting handled through this funtion. It turns out that in this scenario it is called twice, one for each cert in the chain.

PR #592, opened 2020-02-07
Restrict organization names

LostKobrakai:
It seems the nerves_hub cli cannot deal with spaces, so I added a bit of validation for that. It's way more restricting than just spaces though `~r/^[A-Za-z0-9-_]+$`. If that's to restricted feel free to adjust the regex.


fhunleth:
Wouldn't it be better to fix whatever is breaking due to spaces?

LostKobrakai:
Not sure, but given the experience i have with other tools (e.g. recently with dockerhub) those restrictions are to be found often. Especially for cli usage I would expect spaces to cause trouble. In my specific case it seemed like the space was url encoded, making it fail to be correctly detected.

mobileoverlord:
I am leaning more towards handling this problem by eliminating spaces, but I would need to check to see if there are any existing orgs that have spaces.

mobileoverlord:
I migrated the production data to replace ` ` with `-` in anoy org names. There weren't too many.

PR #596, opened 2020-03-12
Proper sort order for datetime column

LostKobrakai:


LostKobrakai:
And just noticed this would need to be able to care about "no connection" as well, which by now it doesn't.

PR #600, opened 2020-03-25
Send firmware metadata to device in update message

mobileoverlord:
When sending an update to a device, pass all firmware metadata. This information is useful to present when determining if the client should choose to apply the update.


fhunleth:
Could you update the docs with this change? https://docs.nerves-hub.org/device-api/phoenix-channel. Also, does the REST endpoint need to be updated for consistency?

Also, how are we handling API versioning on the Phoenix channel side?

jjcarstens:
That’s a good point. We don’t do any versioning with firmware channel response. Maybe that’s a good field to add as well

mobileoverlord:
Can we get a re review now that docs have been published to https://docs.nerves-hub.org/device-api/overview

Issue #604, opened 2020-04-14
Deployment active / inactive button does not function


mobileoverlord:
**Describe the bug**
Activating a deployment / deactivating a deployment does not work from the web site.

**To Reproduce**
1. Deployment Show
2. Make inactive / Make active
3. nothing happens

mobileoverlord:
Merged in master and deployed

PR #605, opened 2020-04-14
Fix live view deployment activate button

mobileoverlord:
Phoenix live view was updated and the behavior of `phx-value` changed to `phx-value-<var>`. This updates the show page on deployments to fix the activate / deactivate button.


mobileoverlord:
Fixes https://github.com/nerves-hub/nerves_hub_web/issues/604

PR #615, opened 2020-07-07
Support full terminal emulation in remote IEx console

jjcarstens:
This adds support for using xterm.js as the web terminal emulation for the remote IEx console. It simplifies things a bit and allows the fun terminal features we all know and love like tab completion, history search, arrow movement, etc.

To use this new web console, a device will need to update `nerves_hub_link` with these changes: https://github.com/nerves-hub/nerves_hub_link/pull/35

For now, both the LiveView terminal and this will be supported and routing happens automatically based on what the device supports. If the device is still using the API for LiveView, a big deprecation warning is shown:
deprecation


apps/nerves_hub_web_core/lib/nerves_hub_web_core/audit_logs/audit_log.ex
```diff
@@ -98,7 +98,7 @@ defmodule NervesHubWebCore.AuditLogs.AuditLog do
   end
 
   def create_description(%{description: description} = audit_log, _, _)
-      when not is_nil(description) or not description == "" do
+      when is_binary(description) and byte_size(description) > 0 do
```
ConnorRigby, line 101: You shouldn't need `is_binary` here. `byte_size` enforces binary

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/device/console_live.html.leex
```diff
@@ -1,3 +1,13 @@
+<div class="alert alert-danger">
+  <h2>Deprecation Notice</h2>
+  <p>
+    This device is using an older version of the remote IEx console API which will soon be removed from NervesHub.
```
fhunleth, line 4: Do we know what "soon" is? I'm thinking that this might live for a while since there are some production devices that might not be easy to upgrade. Unless it's a maintenance burden on us, I think I'd drop the word "soon".

Also, I'm super glad that you put this here since it serves to inform people of the awesome console support that they can get by upgrading.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/audit_logs/audit_log.ex
jjcarstens: Ah, good catch. Totally thought I had to type this

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/device/console_live.html.leex
jjcarstens: I don't have a timeframe. "soon" was the add precedence, but not attached to it

jjcarstens:
Changed the names:
* `tty_data` => `up`
* `tty_text` => `down`

This should be ready for another look

PR #622, opened 2020-08-04
Basic styling for UX refresh

coryjewell:
* Set up new base styles 
* Add new icons
* Add new scss files
* Add phoenix active link for use in the dropdown menu: https://github.com/danhper/phoenix-active-link
* Style navbar, tabs, login, password reset
* Add some no result handling for pages
* Started updates to tables

**Note** the changes in `apps/nerves_hub_www/lib/nerves_hub_www_web.ex`. Was trying to set it up for liveviews according to official documentation.


apps/nerves_hub_www/lib/nerves_hub_www_web/views/session_view.ex
```diff
@@ -1,3 +1,7 @@
 defmodule NervesHubWWWWeb.SessionView do
   use NervesHubWWWWeb, :view
+
+  def permit_uninvited_signups do
```
jjcarstens, line 4: This is duplicated from `NervesHubWWWWeb.LayoutView` and probably an easy place to DRY a bit.

Maybe just an import?
```elixir
import NervesHubWWWWeb.LayoutView, only: [permit_uninvited_signups: 0]
```

coryjewell:
@jjcarstens this was meant to be on a private fork for now! expect the changes to be coming soon!

mobileoverlord:
We want to capture feedback from the community while we work towards applying the redesign without having the site be in a half and half state of transition. @jjcarstens We should figure out a way to collaborate so we can prevent NervesHub from being half broken half useable until its all been applied.

apps/nerves_hub_www/lib/nerves_hub_www_web/views/session_view.ex
coryjewell: thanks for this -- will update

fhunleth:
How about we start a branch for reviewing/merging these new changes?

PR #625, opened 2020-08-25
Add support for uploading trusted device certs

mobileoverlord:
When using pre-signed crypto auth chips such as the ATECC608A-TNGTLS, trusted device certificates are often provided from the supplier. These need to be uploaded to the server to allow them to connect and authenticate.

This adds the following endpoints to the API:

/org/product/device/certificate/create
/org/product/device/certificate/show
/org/product/device/certificate/delete


fhunleth:
Nice. I've been wanting that even for ones that are programmed at manufacturing since it seems like if we have to register the serial number with NervesHub, we might as well register the cert while we're at it. That would also make it possible to reduce the signing cert expiration to just cover manufacture rather than shelf time.

PR #627, opened 2020-08-26
Delta Updates

Mrjaco12:
Why:

* It more efficient on bandwidth to send a delta between old and new firmware files

* Adding a FirwarePatch db table
* Adding functions for getting patches by id and by source/target firmware ids
* Adding a Patcher behaviour module
* Adding a default implementation of the Patcher behaviour
* Adding xdelta/zip/unzip as part of docker build
* Using local filesystem and xdelta to generate firmware patches
* When a device checks for new firmware or a deployment is updated, send back a patch
* Create a patch on request if it doesn't exist
* Add a FirmwarePatch fixture
* Add mox for testing the patcher behaviour
* Implement an Upload behavior for ease of testing with mox
* Stub the upload mock with the file implementation of the upload behaviour
* Add tests around creating patches and checking for updates
* Update readme with xdelta dependency
* Increase timeouts when potentially creating a patch as upload to S3 is slow


Mrjaco12:
@mobileoverlord @jlmcgehee21 here is the first pass at delta updates. This PR should not be merged as it is missing the feature of checking the source/target firmwares to confirm that they both support patching before creating/sending patch files. However I think it's far enough along to be worth a review to make sure I'm on the right track and I was also curious how circle would run.

Some areas where I'm not 100% confident and would appreciate extra scrutiny:

1. I've simply increased timeouts [here](https://github.com/nerves-hub/nerves_hub_web/pull/627/files#diff-e125badcc4b381adec5a2defdb966e8bR233) and [here](https://github.com/nerves-hub/nerves_hub_web/pull/627/files#diff-ecb324861b32008d62bd5b0873c760e9R376) to deal with S3 taking a while to upload. There's probably a better way to deal with this.
2. I'm unsure if I'm using the best approach to [download firmware files](https://github.com/nerves-hub/nerves_hub_web/pull/627/files#diff-0f247e5f6d59e4ac206e9955f378a8a6R31). I've seen this same method used in other places with much more attention paid to configuration and organizing processes around starting and stopping the client. That didn't feel necessary here but I could be overlooking something.
3. In general I'm not sure how well the pattern of download files -> process/create new files -> upload files -> clean up files, will work with our AWS setup. Do we have enough extra disk space on our instances to support this? Other gotchas?

Thanks for the 👀 from anyone who has a chance to look.

fhunleth:
I looked at the delta patch creation logic. Seems fine to me - definitely good enough to exercise this feature. When the dust settles, I'd like to move this functionality over to a shell script that's bundled with `fwup`. There's low hanging fruit to trim the `.fw` files further (not as big as the `rootfs.img` one, but they'll be legit), and keeping regression tests on delta update generation seems easier to maintain in `fwup` than here.

Mrjaco12:
I like that approach. Testing the actual delta creation stuff is pretty painful with all the system calls going on in there. That would seem to indicate to me that there might be a better place to manage that specific logic.

Mrjaco12:
I've updated this to include adding a patchable flag to firmware/metadata. This patchable flag is determined by checking the `meta.conf` of a firmware zip for the presence of a `delta-source-raw-offset` or `delta-source-raw-count` key. If there are other considerations, they should be pretty easy to add. This does not include verification of the FWUP version that is running on the device as there is currently not a way to check that data, that I know of.

fhunleth:
I think that the only practical way of getting the device's `fwup` version is to add that to the metadata that `nerves_hub_link`, etc., are already reporting up when they connect. (I write this after I typed a couple ways of deriving the version of `fwup` in the source `.fw` file, but those routes feel hard to implement and brittle.)

Mrjaco12:
@mobileoverlord just an udpate, I think this PR should be good to go as a first release of the delta updates. We've got checking the fwup version and firmware metadata. Just wanted to make sure you knew this one was ready for you to look at now that Elixir conf is wrapped.

oantila:
@mobileoverlord @fhunleth Any objections to getting this merged to Prod? We'd like to get this live this week if possible.

fhunleth:
I skimmed this again. As far as I can tell, if you're not using delta updates, nothing will change for you. Therefore, I'm on board. Very excited about this feature!

fhunleth (approved):

mobileoverlord:
❤️

PR #628, opened 2020-08-26
Make from_email customizable

Mrjaco12:
Why:

* Private/Custom instances may need a custom from email that is validatable

* Getting from_email from env vars
* Provide a default value if not from_email is set


Mrjaco12:
I'm not sure if there is a good/reasonable way to test this customization. It's probably small enough to not require it but I would be interested if there were ideas on how to do so.

Mrjaco12:
@mobileoverlord Anything else we need to do on this one?

mobileoverlord (approved):

PR #629, opened 2020-08-27
Fixes ISE when the default org name is invalid

brianberlin:
I tried to sign up as `brian.berlin` and I got a 500 error. My fix applies the same validation on orgs name that is being done to the users username. I wonder though if there are usernames in other production environments that will end up having trouble editing their account. I guess that depends on how long the bug has been around.

Other options might be to remove characters not allowed by the org name from the username before saving. Or maybe use the orgs name validation on users username.


jjcarstens:
I'm not sure I understand the underlying issue here? I feel like `Brian.berlin` formatted username should be acceptable (especially if we're allowing `brian_berlin`)

I could just be missing some past discussion on why we made this decision previously. I need to dig in a bit more in the morning

PR #630, opened 2020-09-09
UI Refresh

coryjewell:
### Why
* NervesHub UI is quite dated and in need of a refresh! This is a wide reaching theme overhaul to update the site to be more usable and give it a more modern look.
* Few new features are being added here. This is largely focused on rearranging and restyling existing features.

### How
* _Lots_ of new css, icons, and changes to templates
* Some vocabulary updates which are reflected only on the frontend: organization -> workspace, tag -> group
* Better tooltips to describe different form fields
* Added ability to reboot devices from the device index screen
* Now using either moment.js or Timex for date/time formatting

### Notes
* This is obviously an incredibly large update. I'd recommend pulling down and poking around locally, or @mobileoverlord mentioned getting it on staging to review as well.
* There are still some items on the TODO list for styling updates. Notably, the deployments pages will likely be added here this week. All pages are at least as usable as they were previously with this update alone.
* Also untouched yet is the homepage (but it does look better with a dark theme)


fhunleth:
Is it possible to rebase away all of the cleanup and merge commit to make this easier to review? Or are they all intermediate steps, so everything should be squashed down to 1 commit?

Also, since this feels way too large to expect a thorough review, I expect this will come down to us spot checking a few places and skimming. And then we squash/merge so that if something goes wrong in production, we have an quick revert to avoid any downtime. Does that make sense?

jjcarstens:
I've approved this to be deployed to staging so we can tinker with it.

jjcarstens:
NVM. That seems borked...We might need @mobileoverlord magic here to get staging deployed

mobileoverlord:
We have to enable circle ci to use secrets in PRs from forks.

ConnorRigby:
Just loaded this up on my machine. It looks good!

Is mobile a focus right now or is that for a later date? I bring this up because while it is clunky, the mobile version of the current site is usable, but with the new UI nothing seems to scale appropriately and it's very hard to use. 

I also like the rename of `tags` to `groups` in the UI, but i have concerns about it. Some places refer to it as `tags` still (deployment create, deployment view, deployment edit, device list). 
deployment create
deployment view
deployment edit
device list

The other concern i have about this is there will be a difference between the UI and the CLI after this change, so we need to decide if that's something that needs to be addressed. 
Forcing users to learn a new term for an existing concept seems less than ideal to me. I have a similar concern about
renaming Organizations to Workspaces. 

The next thing might be a feature request for a future update, but on the Org index, there is a lot of white space around each org 
This seems like there could be some useful at a glance info. Number of deployments, devices online etc. Leaving here for now, but will likely open a feature request for this when this PR is merged. 
org dashboard empty space

Overall it looks really good. Thanks for putting this work in!

coryjewell:
Thanks for the feedback!

As mentioned above, I'm currently working on updating the deployment pages so tags haven't received their treatment there just yet. Expect this work to be done (and here) by the end of Friday though!

The whitespace you pointed out on the products page will be addressed _soon_ as well. The plan is to show total devices in each project at the bottom of its square.

Updating mobile styling is definitely still on my plate as well. It has mostly gone untouched as of yet. 

We still have lots of work coming but wanted to get this update out to make the overall process of merging and reviewing a bit smoother.

I think your concerns on the vocabulary updates are valid. We thought that these words provide a bit more understanding to users on the frontend but it does leave the code in a conflicting state at this point. @oantila can give some further explanation as well.

jjcarstens:
I'm okay considering vocabulary updates, but I'm thinking we shouldn't throw it into this big PR. I think we should try to keep that separate and grouped (frontend/backend/cli) as much as possible. Can that be separated out into the next PR?

coryjewell:
^ I'll update to revert the vocabulary changes

jjcarstens:
I notice a few more commits have been added to this ❤️ 

I'm deploying to staging now and going to test. Can we hold off on adding any more to this PR for now and queue them up after this is merged? Pending any _major_ issues, I will prob opt to just getting this in and dealing with nits/small issues as they come

coryjewell:
That works! 

The last thing I have right now waiting is updates to the deployment show page and forms. Those changes are sitting in a PR on our fork but I can hold off on getting anything else in here if you'd prefer.

jjcarstens (approved):
Ok, I've had my fill of playing with this in staging and think its good to go!

Any objections?

fhunleth:
Same - I'm good with squashing the commits and merging.

mobileoverlord:
YOLO!

PR #634, opened 2020-09-16
Don't capitalize org and product in navigation

jjcarstens:
New UI asserts capitalization on orgs and products in the navigation menu. However, if I name my org lowercased with other characters, I'd like to see that in the nav as well instead of having my names reformatted for me in views.

Super nit - what do you think?

/cc @Mrjaco12 @coryjewell 

### Old
image

### New
image


mobileoverlord:
Hummmmm, the things that you can't unsee I believe that I am on board with this pet named peeve that you've got @jjcarstens

mobileoverlord (approved):

PR #639, opened 2020-09-17
Mobile styling and updates to deployment pages

coryjewell:
### Why
- App had hardly any styles to handle mobile devices
- Deployment show page and forms need to be updated to match the rest of the UI

### How
- Added many new mobile css rules (most notably to tables and main navigation)
- Style updates to the deployment show page
- Reworked deployment creation user flow: 
  - Firmware version is now selected on the same screen that you create the deployment. No longer a two step process. Still checks to see if firmware has actually been uploaded before going to this form.
  - Firmware version can now be changed from the edit screen on the deployment edit screen
  - Threw threshold settings under an "advanced options" toggle on the deployment edit screen
- Also fixed an annoying clipping issue with the main menu showing small parts of text underneath of it

### Notes / Questions
- Is there a larger reason I should know about (and consequences to) why a deployment couldn't change its firmware version previously?


jjcarstens:
> _Is there a larger reason I should know about (and consequences to) why a deployment couldn't change its firmware version previously?_

No. In fact, one can change the firmware associated with a deployment via the CLI. We do this frequently at Smartrent. I think the inability to change firmware for a deployment in web is just a "yet to be implemented" feature

apps/nerves_hub_www/lib/nerves_hub_www_web/views/deployment_view.ex
```diff
@@ -62,8 +62,8 @@ defmodule NervesHubWWWWeb.DeploymentView do
   def active(%Deployment{is_active: true}), do: "Yes"
   def active(%Deployment{is_active: false}), do: "No"
 
-  def opposite_status(%Deployment{is_active: true}), do: "Inactive"
-  def opposite_status(%Deployment{is_active: false}), do: "Active"
+  def opposite_status(%Deployment{is_active: true}), do: "Off"
+  def opposite_status(%Deployment{is_active: false}), do: "On"
```
jjcarstens, line 66: Right now the CLI can get a report of deployment being `is_active` so I might scope this under the umbrella of needing to be done with renames all at once. What do you think?

I also created #641 to help track some of the proposed renames and link when making the changes..

jjcarstens (commented):
Looking good! 💟 🎉 

My only nit is `Version Requirement` as an "Advanced Option" - I consider that pretty typical for a deployment and prob can just be shown with he other normal settings. Thoughts?

image

apps/nerves_hub_www/lib/nerves_hub_www_web/views/deployment_view.ex
mobileoverlord: On / Off are already out in prod, I think we should make a move to update the docs with the new terms.

coryjewell:
Taking version requirement out of advanced options sounds great if it's more commonly used. I'll make an update

apps/nerves_hub_www/lib/nerves_hub_www_web/views/deployment_view.ex
jjcarstens: 🤦 Thats what I get for not checking prod.

jjcarstens (approved):

mobileoverlord:
I merged this into main. Thanks!

PR #640, opened 2020-09-18
Org User management

Mrjaco12:
Why:

* Admins may need to remove users or change user roles

* Added an edit template
* Added actions/routes for edit, update, delete
* Add a view helper to get role options
* Only show action menu for users that aren't the current user
* Add tests for each action and the new OrgUserView function

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_user_controller.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_user_controller.ex
@@ -2,6 +2,10 @@ defmodule NervesHubWWWWeb.OrgUserController do
   use NervesHubWWWWeb, :controller
 
   alias NervesHubWebCore.Accounts
+  alias NervesHubWebCore.Accounts.{Email, Org}
+  alias NervesHubWebCore.Mailer
+
+  plug(:validate_role, org: :admin)
 
   def index(%{assigns: %{org: org}} = conn, _params) do
     conn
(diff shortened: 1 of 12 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_user_controller.ex
```diff
@@ -11,4 +15,58 @@ defmodule NervesHubWWWWeb.OrgUserController do
       org: org
     )
   end
+
+  def edit(%{assigns: %{org: org}} = conn, %{"user_id" => user_id}) do
+    {:ok, user} = Accounts.get_user(user_id)
+    {:ok, org_user} = Accounts.get_org_user(org, user)
```
brianberlin, line 21: Maybe you could use `with` here and let the fallback controller catch the `:not_found`s. But I think you'll need to add a function clause with a nice message.

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org/invite.html.eex
```diff
@@ -18,11 +18,13 @@
       <span class="tooltip-info"></span>
       <span class="tooltip-text">User role can be changed from the command line</span>
     </label>
-    <select class="form-control" disabled id="role_input">
-      <option value="Admin">Admin</option>
-      <option value="Basic">Basic</option>
-    </select>
-    <div class="select-icon"></div>
+    <div class="pos-rel">
+      <select class="form-control" disabled id="role_input">
```
jjcarstens, line 22:
```suggestion
      <select class="form-control" id="role_input">
```
We probably don't want this disabled by default anymore since there are ways to edit?

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_user_controller.ex
Mrjaco12: Reviewed this a bit more and I think, for consistency, the best approach would be to add a plug for getting the user/org_user and handling 404s in that plug. Does that sound good?

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/org/invite.html.eex
Mrjaco12: This field doesn't actually do anything right now since the invite action doesn't have any way to add/update a role so I think it makes sense to keep it disabled for now until we figure out how to assign roles on invite.

jjcarstens (approved):

mobileoverlord (approved):

Issue #641, opened 2020-09-22
Renames


jjcarstens:
With the new UI, some new structure and renames have been proposed. It was asked to remove from the UI change and instead incrementally change so that we could adjust both web, API, and CLI at the same time to reflect.

This is an issue to track some of those proposed changes.

| Completed  | Old Name | New Name |
| --- | --- | --- |
|   | `Organization` | `Workspace` |
|   | `tags` | `groups` |
|  | Deployment `Active/Inactive` | Deployment `On/Off` |
|  | `Healthy`/`Unhealthy` | `Quarantined` (no replacement for healthy?) |

joshk:
As this has been open for three years, I'd suggest closing this issue as their have been other discussions floating around naming.

PR #642, opened 2020-09-23
Allow deployment state in API requests/response

jjcarstens:
Part of https://github.com/nerves-hub/nerves_hub_web/issues/641

From discussion in #639 

Deployments now display a `state` value of `on/off` string in the UI. This adds that to the API response and allows updating a deployment state as well so that changes can also be reflected in the CLI


jjcarstens:
cc/ @Mrjaco12 @coryjewell

mobileoverlord (approved):

mobileoverlord:
Thanks @jjcarstens ❤️

PR #643, opened 2020-09-23
Device filtering and pagination on index page

Mrjaco12:
Why:

* We want users to be able to filter for specific devices
* We don't want to show more than 25 devices on the screen
* We want fairly standard pagination interaction with filtering/sorting

* Updating the index template with logic for showing/hiding filters
* Adding handle_events to handle show/hide, reset, and update filters
* Add a function to receive pagination events
* Add a function to paginate based on pagination options
* Add default pagination options
* Add a function to receive an event to update the per_page count
* Update functions to filter after receiving new device data
* Update functions to paginate after receiving new device data/sorting or filtering
* Add tests


jjcarstens (approved):
LGTM 🎉

mobileoverlord:
🎉

PR #644, opened 2020-09-24
Fixes deployment redirect when the name changes

brianberlin:
Also removed unused import


jjcarstens (approved):
❤️thanks @brianberlin !

PR #648, opened 2020-09-24
Fix delta updates failing

Mrjaco12:
What:

* Delta updates were failing for a number or reasons:

1. The `inets` application wasn't available
2. The `xdelta3` executable wasn't available

How:

* Add `inets` to extra_applications
* Make sure delta update executables are installed so they'll be accessible


jjcarstens (approved):

fhunleth:
What needs inets?

Mrjaco12:
We're using `:httpc` to fetch the firmwares. It seemed like the easiest way.
https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/patcher/default.ex#L18

fhunleth (approved):

PR #658, opened 2020-10-05
Update deps to latest

mobileoverlord:
This will update the deps to their latest versions and Elixir to 1.10.4-otp-23


apps/nerves_hub_www/lib/nerves_hub_www_web/templates/layout/app.html.eex
```diff
@@ -27,7 +27,7 @@
       "content": {
         "message": "NervesHub uses cookies to ensure you get the best experience we can provide."
       }
-    })});
+    })});x`
```
jjcarstens, line 30:
```suggestion
    })});
```
Looks like accidental keyboard slam here...

jjcarstens (approved):

Issue #660, opened 2020-10-06
Supper registering device CA certificate with a description via the API


jjcarstens:
[`NervesHubAPI.CACertificateController` doesn't support creating the cert with a description](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/ca_certificate_controller.ex#L23)

That option should be added so it can be supported with the CLI like it is in web.

PR #663, opened 2020-10-07
Adds description to ca_certificate.json render function

brianberlin:
connects to #660 
connects to nerves-hub/nerves_hub_user_api#22 
Seems like this was in place already just not a part of the response. 🤷‍♂️

### DIFF
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/views/ca_certificate_view.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/views/ca_certificate_view.ex
@@ -14,7 +14,8 @@ defmodule NervesHubAPIWeb.CACertificateView do
     %{
       serial: ca_certificate.serial,
       not_before: ca_certificate.not_before,
-      not_after: ca_certificate.not_after
+      not_after: ca_certificate.not_after,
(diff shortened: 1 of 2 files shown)


jjcarstens (approved):
Awesome!

But this is only half the battle. This will return the description in the response. But we still need to adjust the create endpoint to allow the incoming `description` field

brianberlin:
@jjcarstens I maybe confused but i think that's already happening [here](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/ca_certificate_controller.ex#L37).

From here what I gather needs to be done... Need to an update to `NervesHubUserAPI.CACertificate.create` as well as a few edits to `Mix.Tasks.NervesHub.CaCertificate`. I was unable to get all the things running together last night to test locally.

jjcarstens:
@brianberlin You're totally right! I skimmed straight over what was in the function definition and missed that it attempts to pull the description from the params 🤦

You're right in that only `NervesHubUserAPI` and the mix tasks in `:nerves_hub_cli` need to be updated

Thats is totally good to go 🎉

mobileoverlord (approved):

PR #667, opened 2020-10-13
pool_size & prepare Repo options dynamic configuration added

alde103:
Why:
- It seems that pool_size parameter (NervesHubWebCore.Repo) is hardcoded.
- It is not possible to disable prepared statements in Ecto (runtime). 

How:
- init/2 function (at repo.ex), handles the enviroment variables just as url.


fhunleth:
@mobileoverlord Could you make a determination on this one?

jjcarstens:
@alde103 I believe these can be adjusted by adding or changing the options in the [`prod.exs` config](https://github.com/nerves-hub/nerves_hub_web/blob/main/config/prod.exs#L36) (Or similar env)

You should also be able to change config options in any of the apps `release.exs`, if I'm understanding the use-case here correctly

Issue #665, opened 2020-10-10
DB failure on deployment update won't broadcast fw updates


jjcarstens:
We recently had a case where a deployment was updated with a new firmware via `nerves_hub_cli, but target devices failed to receive the request to update.

Updating a deployment with a new firmware triggers various DB lookups to identify devices that may be expected to receive the fw update. In this case, those lookups and subsequent broadcast to devices failed (about [here](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/deployments.ex#L91-L93)) because there was network latency for compute modules in AWS.

I'm unsure if the DB packets dropped, or just timeout in NH, etc, but its prob something to consider hardening by moving pieces of this update check to monitored, recoverable tasks with an aggressive enough retry scheme to help ensure devices will be told to update (i.e. have some incrementing backoff time after each error instead of blasting through retries within a couple seconds)

jjcarstens:
Discovered a rollbar that helps showcase this issue. Triggered from manually turning deployment off/on again (which is same behavior as using the CLI)

https://rollbar.com/nerves-hub/nerves_hub_www/items/31/occurrences/138359048600/

<img width="1201" alt="image" src="https://user-images.githubusercontent.com/11321326/96038918-1b53fd00-0e25-11eb-9f46-b5156731f33f.png">

jjcarstens:
The task is running [`Devices.send_update_message/2`](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex#L130-L167) and that has a lot of queries in there. Its prob worth optimizing as well to maybe group to a single transaction.

Also prob better for fire off each task asynchronously and supervised (with retry scheme) without waiting on the task completions. Thoughts?

jjcarstens:
Also, the [`Devices.send_update_message/2`](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex#L130-L167) doesn't check device availability. Meaning we're doing that lookup for every device that the deployment potentially targets, even if it is not online. Its prob worth checking that the device is online first, and skipping offline devices.

In our case, only ~20 devices are online, but the deployment could potentially target ~50

jjcarstens:
Got another report of timeout when a single device checked in

[...]

PR #675, opened 2020-10-14
Temporarily disable delta updates

jjcarstens:
Addresses https://github.com/nerves-hub/nerves_hub_web/issues/665 by temporarily disabled delta updates to prevent timeouts and failing updates until a better solution can be developed.

cc/ @Mrjaco12


Mrjaco12 (approved):

Mrjaco12:
Seems good enough to me.

Mrjaco12:
Actually tests will probably fail with this...

fhunleth (approved):
I'm good with this or any variations that do the same basic thing

Issue #421, opened 2019-04-10
Support deletion of accounts


fhunleth:
Currently it's possible to create an account, but once you're done with NervesHub, there's no way to delete the account.

chazwatkins:
Would we prefer a soft or hard delete?

Should the user's **owned** orgs, devices, firmware, deployments, etc be removed when the user is removed?

fhunleth:
That's a good question.

I curious on others opinions, but I'd want a soft delete. I'd like to preserve data so that connections from orphaned devices could be traced to their original owners and handled specially should they become disruptive. Hard deleting orphaned firmware and deployments seems less problematic. I'd be fine with soft deleting those too with the assumption that orphaned firmware files would eventually be garbage collected.

brianberlin:
Excuse me while i fumble around with this draft PR. I'm still getting the hang of all these relationships and constraints. I was able to get some code together for removing a basic account without org users. Once a second org is created and users are invited it looks to become a little more challenging. I've also not gotten into the soft delete part which will also take some cycles to get right.

More rambles... It looks like if we want devices to be soft deleted, we'll also need to do the same for orgs and products. I think because of the way tags work we should be able to delete everything else. In order to tackle orgs with org users, I should maybe shift focus to deleting orgs next. Should any admin be able to delete an org? 🤔

jjcarstens:
I would assume an admin level should probably be able to delete an org since we're dealing with soft delete.

ConnorRigby:
Closed via #421

PR #676, opened 2020-10-21
Ability to delete accounts

brianberlin:
connect #421 

This pull request allows users to delete their accounts when they're done using NervesHub. When an account is deleted, most things can be "hard deleted". However, in order to handle orphaned devices that may become disruptive we elect to "soft delete". I've identified users, orgs, org_users, products, and devices to become soft deleted.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/remove_account.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/remove_account.ex
@@ -1,0 +1,133 @@
+defmodule NervesHubWebCore.Accounts.RemoveAccount do
+  import Ecto.{Query, Changeset}
+  alias Ecto.Multi
+
+  alias NervesHubWebCore.{
+    Accounts,
+    Firmwares,
+    Deployments.Deployment,
+    Products,
+    Repo,
+    OrgKey,
+    Devices
+  }
+
+  alias Firmwares.{Firmware, FirmwareDelta, FirmwareTransfer}
+  alias Accounts.{Org, OrgUser, OrgKey, Invite, User, UserCertificate, OrgLimit, OrgMetric}
+  alias Devices.{DeviceCertificate, Device, CACertificate}
+  alias Products.{Product, ProductUser}
+
+  def remove_account(user_id) do
+    Multi.new()
+    |> Multi.run(:user_id, fn _, _ -> {:ok, user_id} end)
+    |> Multi.run(:org_ids, &get_org_ids/2)
+    |> Multi.run(:firmware_ids, &get_firmware_ids/2)
+    |> Multi.delete_all(:invites, &query_by_org_id(Invite, &1))
+    |> Multi.delete_all(:device_certificates, &query_by_org_id(DeviceCertificate, &1))
+    |> Multi.delete_all(:ca_certificates, &query_by_org_id(CACertificate, &1))
+    |> Multi.delete_all(:deployments, &query_by_org_id(Deployment, &1))
+    |> Multi.delete_all(:firmware_deltas, &query_firmware_deltas/1)
+    |> Multi.delete_all(:firmware_transfers, &query_by_org_id(FirmwareTransfer, &1))
+    |> Multi.merge(&delete_firmwares/1)
+    |> Multi.delete_all(:product_users, &query_product_users/1)
+    |> Multi.delete_all(:org_keys, &query_by_org_id(OrgKey, &1))
+    |> Multi.delete_all(:org_limits, &query_by_org_id(OrgLimit, &1))
+    |> Multi.delete_all(:org_metrics, &query_by_org_id(OrgMetric, &1))
+    |> Multi.update_all(:soft_delete_products, &soft_delete_by_org_id(Product, &1), [])
+    |> Multi.update_all(:soft_delete_devices, &soft_delete_by_org_id(Device, &1), [])
+    |> Multi.update_all(:soft_delete_org_users, &soft_delete_by_org_id(OrgUser, &1), [])
+    |> Multi.update_all(:soft_delete_orgs, &soft_delete_orgs(&1), [])
+    |> Multi.delete_all(:user_certificates, &query_user_certificates/1)
+    |> Multi.update(:soft_delete_user, &soft_delete_user/1)
+    |> Repo.transaction()
+  end
+
+  defp query_org_users do
+    from(
+      org_user in OrgUser,
+      join: user in assoc(org_user, :user),
+      where: is_nil(user.deleted_at),
+      select: org_user.org_id
+    )
+  end
+
+  defp get_org_ids(repo, %{user_id: user_id}) do
+    where_org_has_users = where(query_org_users(), [ou], ou.user_id != ^user_id)
+
+    query =
+      query_org_users()
+      |> where([ou], ou.user_id == ^user_id)
+      |> except(^where_org_has_users)
+
+    {:ok, repo.all(query)}
+  end
+
+  defp get_firmware_ids(repo, %{org_ids: org_ids}) do
+    query =
+      from(
+        firmware in Firmware,
+        where: firmware.org_id in ^org_ids,
+        select: firmware.id
+      )
+
+    {:ok, repo.all(query)}
+  end
+
+  defp delete_firmwares(%{firmware_ids: firmware_ids}) do
+    Enum.reduce(firmware_ids, Multi.new(), fn firmware_id, multi ->
+      multi_key = "delete_firmware_#{firmware_id}"
+
+      Multi.run(multi, multi_key, fn _, _ ->
+        Firmware
+        |> Repo.get(firmware_id)
+        |> Firmwares.delete_firmware()
+      end)
+    end)
+  end
+
+  defp truncated_utc_now do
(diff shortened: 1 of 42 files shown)


fhunleth:
@brianberlin Thanks for sending this PR and congrats on getting past draft! Could you rebase and squash out the intermediate work? I'm not sure what the conflict is. If you need assistance, let us know.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/remove_account.ex
```diff
@@ -0,0 +1,126 @@
...
+    |> Multi.update(:soft_delete_user, &soft_delete_user/1)
+    |> Repo.transaction()
+  end
+
+  defp get_org_ids(repo, %{user_id: user_id}) do
+    query =
+      from(
+        org_user in OrgUser,
+        join: org in assoc(org_user, :org),
+        where: org_user.user_id == ^user_id,
```
jjcarstens, line 51: Should we also check here if the org has any other users so we don't delete an Org this user is a part of but still has active users?

Its also been a minute since I was heavy into DB queries - Maybe that is already considered here?

jjcarstens (commented):
The soft delete functionality seems okay to me. But I might be a little loss on the Account removal. It seems like if we're removing an account, we'll also soft delete the org that user/account was a part of even if that org is still active with other users?

apps/nerves_hub_web_core/priv/repo/migrations/20200822003346_firmware_patch_table.exs
```diff
@@ -1,4 +1,4 @@
-defmodule NervesHubWebCore.Repo.Migrations.FirmwarePatchTable do
+defmodule NervesHubWebCore.Repo.Migrations.FirmwareDeltaTable do
```
jjcarstens, line 1: I think we also need to remove this change and just rebase from main to get the needed firmware patch/delta pieces that were changed later on

fhunleth:
@brianberlin Do you have time to fix the tests this week? @jjcarstens and I are looking at merging the PR and it seems really close. We just want to make sure the tests pass and then do a quick pass on trying it out in staging.

brianberlin:
@fhunleth Sorry I haven't had much time to spare lately. I'll get it fixed up soon.

brianberlin:
@fhunleth I think all the tests should pass now. I still need to address @jjcarstens concerns.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/accounts/remove_account.ex
```diff
@@ -0,0 +1,133 @@
...
+    |> Multi.run(:user_id, fn _, _ -> {:ok, user_id} end)
+    |> Multi.run(:org_ids, &get_org_ids/2)
+    |> Multi.run(:firmware_ids, &get_firmware_ids/2)
+    |> Multi.delete_all(:invites, &query_by_org_id(Invite, &1))
+    |> Multi.delete_all(:device_certificates, &query_by_org_id(DeviceCertificate, &1))
+    |> Multi.delete_all(:ca_certificates, &query_by_org_id(CACertificate, &1))
+    |> Multi.delete_all(:deployments, &query_by_org_id(Deployment, &1))
+    |> Multi.delete_all(:firmware_deltas, &query_firmware_deltas/1)
+    |> Multi.delete_all(:firmware_transfers, &query_by_org_id(FirmwareTransfer, &1))
+    |> Multi.merge(&delete_firmwares/1)
```
brianberlin, line 31: This part of the multi could be an issue. It will make an s3 request for each firmware object that needs to be deleted. If it were to get halfway through and hit an error and roll back. That would cause the account to fail again and again. Thoughts on how to handle?

brianberlin: It looks like oban is available. Maybe adding a job for each firmware in the `Firmwares.delete_firmware/1` function?

brianberlin:
@fhunleth I think this maybe ready. 🤞

apps/nerves_hub_web_core/lib/nerves_hub_web_core/deployments.ex
```diff
@@ -5,6 +5,7 @@ defmodule NervesHubWebCore.Deployments do
   alias NervesHubWebCore.Deployments.Deployment
   alias NervesHubWebCore.Firmwares
   alias NervesHubWebCore.Devices
+  alias NervesHubWebCore.Devices.Device
```
jjcarstens, line 8: Was this intended to be used? It seems like this is an unused alias

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex
```diff
@@ -122,18 +122,26 @@ defmodule NervesHubWebCore.Firmwares do
   end
 
   def delete_firmware(%Firmware{} = firmware) do
+    do_delete = fn ->
```
jjcarstens, line 125:
```suggestion
    do_delete_firmware = fn ->
```

The with statement below doesn't match here and instead references `do_delete_firmware.()`. I like that better, so prob just change here

apps/nerves_hub_web_core/lib/nerves_hub_web_core/repo.ex
```diff
@@ -27,4 +29,20 @@ defmodule NervesHubWebCore.Repo do
...
+    |> update()
+  end
+
+  def soft_delete_changeset(struct_or_changeset) do
+    deleted_at = DateTime.truncate(DateTime.utc_now(), :second)
+
+    Ecto.Changeset.change(struct_or_changeset, deleted_at: deleted_at)
+  end
+
+  def with_undeleted(query) do
```
jjcarstens, line 45:
```suggestion
  def exclude_deleted(query) do
```
This is a fairly big nit, but "undeleted" makes me think think records were once deleted, but now are not, and we're doing complex query to deduce that.

I don't know if my suggestion is the best function name alternative, but I think something a little simpler for what it is doing would be nice.

apps/nerves_hub_web_core/test/nerves_hub_web_core/accounts/accounts_test.exs
```diff
@@ -201,7 +194,7 @@ defmodule NervesHubWebCore.AccountsTest do
 
     assert user.email == expected_email
 
-    assert {:ok, %User{email: expected_email, orgs: [%Org{}]}} =
+    assert {:ok, %User{email: _expected_email, orgs: [%Org{}]}} =
```
jjcarstens, line 197:
```suggestion
    assert {:ok, %User{email: ^expected_email, orgs: [%Org{}]}} =
```
Does this fail? I would probably say if we have an "expected" email, we should pin it to assert we get whats expected

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/account_controller.ex
```diff
@@ -38,6 +38,26 @@ defmodule NervesHubWWWWeb.AccountController do
...
+         {:ok, _} <- Accounts.remove_account(user.id) do
+      conn
+      |> put_flash(:info, "Success")
+      |> redirect(to: "/login")
+    end
+  end
+
+  def delete(conn, %{"user_name" => username}) do
+    conn
+    |> put_flash(:error, "Failed")
```
jjcarstens, line 57: Do we know why it is "Failed" here? Is it because the user did not confirm username?

brianberlin: I'm not sure much thought went into this initially. I updated it with a better message if the username doesn't match and left any unexpected errors from the remove_account function unaddressed. I'm not really sure what should be handled at this point. Maybe a rollbar occurrence is what we want.

apps/nerves_hub_web_core/test/nerves_hub_web_core/accounts/accounts_test.exs
brianberlin: ah yes, I was going through and silencing warning without looking too hard at the tests. pinned it. test passed.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/repo.ex
brianberlin: yep, i like that better. the thing i worried about with this was where it was placed. It's not really a repo thing it's a query thing. I thought about putting the same function in each module that would be soft deleted but then settled on this. 🤷

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex
brianberlin: guess i didn't run test before my last push. whoops.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/deployments.ex
brianberlin: ⚡

jjcarstens (approved):
🎉 Thanks @brianberlin 

This LGTM and really sorry for the delay. There has been quite a push on some other things that we've been behind on and are more immediately needed. Once those are in, we'll get this in as well to test around with.

PR #677, opened 2020-10-21
Fix device index status reset bug

### LINKED ISSUE #670: Device view loses all data when one device changes state
**Describe the bug**
When I view all devices, if one changes state, pretty much everything else on the screen loses its state. It's easier to see in the video:

https://youtu.be/zW3RSTe81Ag

**To Reproduce**

Go to the device's view. I usually sort on "Last handshake", but I don't think that it has to do with that. Then wait for a device to change state. Firmware updates are good since lots of devices reboot and that results in a flurry of state changes.

Mrjaco12:
Why:

* We were not applying presence statuses to new devices from DB
* Fixes #670 

* Apply statuses from presence after a presence_diff event and devices are fetched

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/index.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/live/device_live/index.ex
@@ -222,12 +222,14 @@ defmodule NervesHubWWWWeb.DeviceLive.Index do
   end
 
   def handle_info(
-        %Broadcast{event: "presence_diff", payload: payload},
+        %Broadcast{event: "presence_diff", payload: %{leaves: leaves}},
         %{assigns: %{org: org, product: product}} = socket
       ) do
     devices = Devices.get_devices_by_org_id_and_product_id(org.id, product.id)
+    joins = Presence.list("product:#{product.id}:devices")
 
-    socket = assign_display_devices(socket, sync_devices(devices, payload))
+    socket =
+      assign_display_devices(socket, sync_devices(devices, %{joins: joins, leaves: leaves}))
 
     {:noreply, socket}
   end


jjcarstens (commented):
I'm not quite following the fix here? `payload` should already be loaded with any necessary devices in the `:joins` key. This now adds another lookup to the presence data when the purpose of getting `presence_diff` messages is to avoid lookups. In theory, if the device was not just joined or left, then it should be ignored?

Mrjaco12:
Yeah I agree this is pretty ugly. The catch here is that we are refetching all the devices when this event comes in, we do that because the join or leave that is associated with the `presence_diff` might not be in our current device data set.

Now that I've typed that all out, maybe the fix is to just not fetch new data on `presence_diff`. I'll take a look at that approach and see if there are any gotchas.

Mrjaco12:
Ok, after testing this out the down side to this approach is that if you have a filter set and a device that didn't match the filter, changes to match the filter then it won't appear in the list e.g. if you were filtering by connected devices and one of you offline devices connected, we wouldn't automatically add that device in unless we got back and add all the devices from the DB and then refilter, sort, etc.

jjcarstens:
so if join/leave is not in device data set, does that cause a crash which messes things up?

Is the "device data set" the limited devices in current view?

Mrjaco12:
Yes, the "device data set" is the list of devices that are currently being displayed so it's limited to what has been filtered and paginated. Everytime a user updates filter/pagination/sorting we refetch all the devices and re-filter, re-paginate, re-sort. This gets kind tricky when the `presence_diff` event comes through. It doesn't crash and mess things up if the join/leave isn't in the device data set, but it won't properly update the device status and reflect those changes on the front end.

fhunleth:
@Mrjaco12 @jjcarstens @mobileoverlord What's the status of this PR? It would really be nice to get this fixed.

Mrjaco12:
I don't know of a better/simpler way to fix this bug unless we're happy to make the tradeoff of:

> if the join/leave isn't in the device data set, but it won't properly update the device status and reflect those changes on the front end

Or if @jjcarstens has a better idea.

jjcarstens:
I'm thinking maybe the better move here in the long run is to store all devices in state on the server. Any filter or presence updates server state and changes the list of devices to assign to client state for display.

jjcarstens:
Actually, thats a half baked idea. I'm out of date with LiveView, so I don't know if its outside the realm of possibility to have session/server state that doesn't get hashed and applied to the HTML view...but I _assume_ is possible?

If not, then this is probably the route to go for now as its simpler to do the ets lookup that store all devices in the client side state

oantila:
Checking in on this since I'm not sure what the consensus is based on the last comments. Can we merge this or is there additional work to be done to address the original issue? @Mrjaco12 @jjcarstens

jjcarstens (approved):
:shipit:

oantila:
Nice, thanks!

PR #681, opened 2020-10-30
Append timestamp to audit_log help-text message

### LINKED ISSUE #680: Timestamps on device activity logs
**Describe the bug**
Currently the activity logs look like this:

image

For very recent activities, knowing that something happened seconds or minutes ago is convenient, but beyond that, it would be great if the timestamps were shown. This would make it possible to correlate NervesHub events to events going on elsewhere. For example, in the picture above, knowing the event was "1 day ago" is not super helpful.

chazwatkins:
closes #680 

Appends timestamp to the audit log to determine the exact time an event occurred

image


jjcarstens (approved):
🙌 ❤️ 🎉 

Thanks @chazwatkins!

mobileoverlord (approved):

PR #683, opened 2020-11-02
Make products patchable

### LINKED ISSUE #674: Allow delta updates to be opt-in
[Firmwares get labeled as "patchable" if they have a few fwup fields in their `fwup.conf`](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex#L439), but that's basically all it takes to opt-in to firmware patches. Past that, everything happens autonomously.

We prob should allow opt-in settings on the org/firmware/deployment/whatever so that firmware patches can be avoided if needed without a whole new firmware deployed just to change some fields in `fwup.conf`

Mrjaco12:
This fixes #674 

Why:

* We want users to be able to configure patchability of firmware per product

* Add a patchable flag in the DB
* Update the router to allow updating products
* Update the product form to use the correct endpoint
* Update product form with new fields
* Update CSS for patchable field
* Add a new condition to the fetch_firmware_url function for patchable product
* Add tests to ensure that patches are created for patchable products
* Add test for updating product

Note: The `get_firmware_url` function is getting a bit wild with parameters. If anyone has ideas about how to reorganize this to make check the check a little nicer, would be appreciated.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
@@ -8,7 +8,7 @@ defmodule NervesHubWebCore.Products.Product do
   alias NervesHubWebCore.Products.ProductUser
 
   @required_params [:name, :org_id]
-  @optional_params []
+  @optional_params [:delta_updatable]
 
   schema "products" do
     has_many(:devices, Device)
@@ -19,6 +19,7 @@ defmodule NervesHubWebCore.Products.Product do
     belongs_to(:org, Org)
 
     field(:name, :string)
+    field(:delta_updatable, :boolean, default: true)
 
     timestamps()
   end
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
@@ -19,7 +19,7 @@ defmodule NervesHubWebCore.Products.Product do
     belongs_to(:org, Org)
 
     field(:name, :string)
-    field(:delta_updatable, :boolean, default: true)
+    field(:delta_updatable, :boolean, default: false)
 
     timestamps()
   end
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/views/product_view.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/views/product_view.ex
@@ -12,7 +12,8 @@ defmodule NervesHubAPIWeb.ProductView do
 
   def render("product.json", %{product: product}) do
     %{
-      name: product.name
+      name: product.name,
+      delta_updatable: product.delta_updatable
     }
   end
 end
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex
@@ -143,7 +143,7 @@ defmodule NervesHubWebCore.Devices do
           {:error, :not_found} -> nil
         end
 
-      {:ok, url} = Firmwares.get_firmware_url(source, target, fwup_version)
+      {:ok, url} = Firmwares.get_firmware_url(source, target, fwup_version, product)
       {:ok, meta} = Firmwares.metadata_from_firmware(target)
 
       Phoenix.PubSub.broadcast(
@@ -419,7 +419,7 @@ defmodule NervesHubWebCore.Devices do
           {:error, :not_found} -> nil
         end
 
-      {:ok, url} = Firmwares.get_firmware_url(source, target, fwup_version)
+      {:ok, url} = Firmwares.get_firmware_url(source, target, fwup_version, product)
       {:ok, meta} = Firmwares.metadata_from_firmware(target)
 
       %{update_available: true, firmware_url: url, firmware_meta: meta}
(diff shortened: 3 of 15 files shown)


Mrjaco12:
@coryjewell I hacked some stuff together for the patchable checkbox but you probably have some better ideas if you wanted to add anything here that would be welcome.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
```diff
@@ -19,6 +19,7 @@ defmodule NervesHubWebCore.Products.Product do
     belongs_to(:org, Org)
 
     field(:name, :string)
+    field(:patchable, :boolean, default: true)
```
mobileoverlord, line 22: Maybe we should make this false for now until we reach stability.

mobileoverlord (commented):
I'll have to check out the styling, but the code looks like it should do it.

Mrjaco12:
Screen Shot 2020-11-03 at 1 01 15 PM
Here's what the styling looks like right now. For reference.

fhunleth:
Could the text be updated to change "patchable" to "Generate delta firmware updates"? Or something along those lines that's consistent with the documentation at https://docs.nerves-hub.org/nerves-hub/setup/firmware#delta-updates. We'll also need the docs updated to tell users to make sure they enable this setting.

I think that it would also be nice to have one of those circle question mark icons by the option and to use it to point users at the delta firmware update docs.

Mrjaco12:
Updating the label is no problem. I'll also make sure to add a PR for the docs. I'll have to take a look at the tooltip stuff. IDK off the top of my head how much extra that is.

Mrjaco12:
I've added the updated label and tooltip. Right now our tooltips disappear after you lose hover on the label so it's kinda hard to get a link working well inside the tooltip.
Screen Shot 2020-11-04 at 9 58 35 AM

fhunleth:
Thanks! I'm sure that there are a dozen ways of improving it, but it seems plenty good enough to me. I'm approving. Someone else should approve on the code side, though.

fhunleth (approved):

Mrjaco12:
Screen Shot 2020-11-04 at 11 04 46 AM
Updated UI from Cory.

fhunleth:
This is a nit, but that checkbox doesn't "Enable delta firmware updates". It enables the chunk of code that creates the delta firmware updates on NH. There are a few steps to enable delta firmware updates. That's why I like the link to the documentation.

fhunleth:
Before this gets merged, could we change the word `patchable` in the db to another word? Perhaps, `generate_firmware_deltas` or something similar? The word also seems inaccurate in the CSS, but that seems easier to change in the future.

Mrjaco12:
You're referring to the DB column? I would push back on changing that because the [term patch](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_patch.ex) is [already used](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware.ex#L49) in [several places](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/patcher/default.ex) including the DB and consistency there seems worthwhile to maintain.

Mrjaco12:
Also to be clear we still have the tooltip and it works much better now:
Screen Shot 2020-11-05 at 12 36 10 PM happy to keep tweaking the label text to make it more accurate.

fhunleth:
> You're referring to the DB column? I would push back on changing that because the [term patch](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_patch.ex) is [already used](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware.ex#L49) in [several places](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/patcher/default.ex) including the DB and consistency there seems worthwhile to maintain.

Could you update the term everywhere, then? I agree that consistency is important. I missed the commits that added the word "patchable", but that should have been changed or discussed since it's different than how many of us have been referring to the "delta updates" publicly. I know that it would have been better to have this changed earlier, but it seems like it only gets more painful in the future.

Mrjaco12:
Yeah we can definitely update that term everywhere. It seems like it might be outside of the scope of what we want to do here but I can take a look at that if we think that's gonna be a blocker to this PR.

fhunleth:
Yes. It's up to you. I'd probably update the term everywhere in a separate PR. Then get that PR merged and then update this one. I don't have strong opinions on the how. If you'd prefer to do it all in this PR, that seems fine to me too. I'd just prefer not to have a transient column added to the db and then renamed soon after.

oantila:
Wanted to check on this one. Any chance we can get it merged soon? We got one client that's itching to use Deltas. :)

Mrjaco12:
@oantila I'm going to do another PR to update language and then we'll get this one merged in. Hopefully I'll get some time to update the patch <=> delta update language soon.

fhunleth:
@Mrjaco12 Could you rebase this on top of #693 and update the naming? It sounds like this is ready to merge once that happens.

fhunleth:
@Mrjaco12 One question - are the generated firmware updates being garbage collected?

Mrjaco12:
The way it works right now is deltas will cascade with firmwares. So when a firmwares is cleaned up or removed the corresponding delta will be removed as well. We don't clean up deltas individually.

fhunleth:
@mobileoverlord @jjcarstens This one is ready to merge, right? I believe it was on the list to merge on patch Tuesday, but got dropped when the meeting was cancelled.

PR #685, opened 2020-11-03
Add progressbar to device details page

coryjewell:
### Why
- The progress bar showing a device updating should be displayed on the device details page

### How
- Added the progress bar to the device `_header` 
- Slightly updated the styles for the progress bar to be more inline with the rest of the styles.

### DIFF
--- a/apps/nerves_hub_www/assets/css/app.scss
+++ b/apps/nerves_hub_www/assets/css/app.scss
@@ -143,3 +143,25 @@ html body {
     background-color: var(--warning);
   }
 }
+
+.progress {
+  background: rgba(255, 255, 255, 0.1);
+
+  .progress-bar {
+    background-color: var(--primary);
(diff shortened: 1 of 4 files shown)


jjcarstens (approved):
Tested in staging.

LGTM 🎉 

If I can inject a nit, I'm not a huge fan of the `X` right next to `Updating` status on the device page. It would be nice if that could also be a change in this PR as well, but I'm good going through without it also for now
image

image

coryjewell:
I agree with you about the X next to updating, but I'll add that in a separate PR

fhunleth:
@coryjewell Do you have any progress on fixing the "X" by the "Updating" text? If it's an easy fix, that would be really nice to get in since it's a little disturbing to devices start updating and immediate go offline.

mobileoverlord:
merged into main

PR #693, opened 2020-11-10
Rename firmware_patch to firmware_delta

Mrjaco12:
Why:

* We want standard language
* Unblock #683 

* Updating occurrences of "patch" to some format of "delta"
* Added a migration to update all DB references from patch to delta


fhunleth (approved):
I'm approving that the language is consistent and more accurate in some places. I noticed a couple places where I probably wouldn't have minded patch vs. delta, but it seems like the right thing to use the same term everywhere.

Could someone else review it for anything that might accidentally break? I'm ready to merge this. I assume that #683 will need a rebase with name changes, but then I think that I'm ready to merge it as well.

jjcarstens (approved):

PR #697, opened 2020-11-19
Restructure Console page

### LINKED ISSUE #654: Adjust the remote console size
The remote console's width and height are hardcoded. You can see this by watching that text wraps before the edge of the black rectangle like below:

image

The device-side code supports setting the width and height of the console. If the web side can somehow send this information back to the device, the device will wrap text at the right place.

### LINKED ISSUE #689: Console page shows devices as offline when they're online
It looks like devices only show up as offline on the console page. I haven't see a device show up as online yet, so I suspect that the logic either isn't right or it changed since #666 was implemented. Here's a screenshot:

image

This seems like it should be easy to reproduce, but if it turns out not to be, I'll try to get more details.

### LINKED ISSUE #696: Fix console scrolling
Moved from [this comment](https://github.com/nerves-hub/nerves_hub_web/issues/664#issuecomment-707362699)

There is some weird scrolling behavior in the console page that causes the whole page to scroll when you're scrolling in the console window. It should just scroll in the console window.

I don't know what is involved to fix it, but it sure would be nice 😉

jjcarstens:
Based on some discussion from slack.

Resolves #689
Resolves #696
Resolves #654

* `console` button now opens console to a new window or tab
* console is the full window with minimal details at the top
* console is resizable
* device status _should_ update with presence changes...I'm not that good at javascript, so YMMV

image

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/user_console_channel.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/user_console_channel.ex
@@ -51,39 +51,4 @@ defmodule NervesHubWWWWeb.UserConsoleChannel do
   defp console_topic(device_id) do
     "console:#{device_id}"
   end
-
-  # defp sync_device(%Device{id: id} = device, payload) when is_map(payload) do
-  #   id = to_string(id)
-  #   joins = Map.get(payload, :joins, %{})
-  #   leaves = Map.get(payload, :leaves, %{})
-
-  #   cond do
-  #     meta = joins[id] ->
-  #       updates =
-  #         Map.take(meta, [
-  #           :console_available,
-  #           :firmware_metadata,
-  #           :fwup_progress,
-  #           :last_communication,
-  #           :status
-  #         ])
-
-  #       Map.merge(device, updates)
-
-  #     leaves[id] ->
-  #       # We're counting a device leaving as its last_communication. This is
-  #       # slightly inaccurate to set here, but only by a minuscule amount
-  #       # and saves DB calls and broadcasts
-  #       disconnect_time = DateTime.truncate(DateTime.utc_now(), :second)
-
-  #       device
-  #       |> Map.put(:console_available, false)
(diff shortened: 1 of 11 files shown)


jjcarstens:
The scrolling "works" - disabling page scrolling and letting only the terminal scroll exist was inconsistent (sometimes the prompt would show out of view, especially when resizing). Someone smarter with CSS will need to figure that out

fhunleth:
This looks great!

What's the deal with the quotes around "works", though? I don't think I understand what compromise that we're going to have to live with until the CSS is fixed. Like is it a big deal? Please say no.

jjcarstens:
It's totally usable and scrolling works. When you scroll, the buttons at the top bounce around a bit and when you get to the bottom, the top buttons will hide (halfway or fully) and requires just a bit of upward scroll to show the buttons/info again.

It's just annoying and I don't know how to fix it without messing up the prompt line, so opted to just deal with a bit of annoying bouncing.

I'll fix tests so we can deploy to staging and you can seee

jjcarstens:
Here's a preview of the "bouncing"

scroll_bounce

ConnorRigby:
Screenshot from 2020-11-19 11-03-49

fhunleth (approved):
I've been playing around with this and it's really nice! Thank you @jjcarstens and @ConnorRigby!

apps/nerves_hub_www/lib/nerves_hub_www_web/user_console_channel.ex
```diff
@@ -24,4 +51,39 @@ defmodule NervesHubWWWWeb.UserConsoleChannel do
   defp console_topic(device_id) do
     "console:#{device_id}"
   end
+
+  # defp sync_device(%Device{id: id} = device, payload) when is_map(payload) do
```
fhunleth, line 55: Should `sync_device/2` be deleted? It looks like it's not needed any more.

mobileoverlord (approved):
Looks cool!

Issue #702, opened 2020-11-24
DeviceCert with no Common Name fails to verify


ConnorRigby:
While debugging a different issue, we found a device in the logs spamming the server with reconnects because the cert has no common_name

jjcarstens:
I'm assuming we want verification to still fail when CommonName in the cert does not match the device identifier, but with a better error here instead?

fhunleth:
Yes, the problem was that the lack of the common name would turn into an exception which would fill the logs. The first desire was for it to return error rather than crash. Then, and I'm not sure if this is possible, it would be nice if the error back to the user could tell them to add a common name to their certs. It's a weird error in that the cert being presented passes authentication, but it was made without the required contents for use with NervesHub.

jjcarstens:
This was fixed with #715

PR #703, opened 2020-11-24
Hex cert

### LINKED ISSUE #701: Show certificate serial number in hex to match openssl
**Describe the bug**

When inspecting certificates in NH, the serial number looks like this:

image

When using OpenSSL to dump certificates, they look like this:

```
$ openssl x509 -in cert.pem -text
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            4b:7d:f4:cc:2d:21:e8:58:63:de:e5:fc:c7:6f:a1:b2
...
```

It would be nice if the serial number were printed in hex to make it easier to compare. A side benefit would be that it might not wrap any more in the UI.

ConnorRigby:
Fixes #701

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
@@ -10,4 +10,9 @@ defmodule NervesHubWWWWeb.OrgCertificateView do
     |> Enum.chunk_every(2)
     |> Enum.join(":")
   end
+
+  def format_serial(serial_str) when is_bitstring(serial_str) do
+    String.to_integer(serial_str)
+    |> format_serial()
+  end
 end
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
@@ -2,4 +2,12 @@ defmodule NervesHubWWWWeb.OrgCertificateView do
   use NervesHubWWWWeb, :view
 
   alias NervesHubWWWWeb.LayoutView.DateTimeFormat, as: DateTimeFormat
+
+  def format_serial(big_long_integer) when is_integer(big_long_integer) do
+    big_long_integer
+    |> Integer.to_string(16)
+    |> to_charlist()
+    |> Enum.chunk_every(2)
+    |> Enum.join(":")
+  end
 end
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
@@ -11,7 +11,7 @@ defmodule NervesHubWWWWeb.OrgCertificateView do
     |> Enum.join(":")
   end
 
-  def format_serial(serial_str) when is_bitstring(serial_str) do
+  def format_serial(serial_str) when is_binary(serial_str) do
     String.to_integer(serial_str)
     |> format_serial()
   end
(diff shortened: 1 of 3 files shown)


jjcarstens:
What would you think about being able to click through hex and internet representation of the serial? I do have times where I want the integer value

fhunleth:
When do you want the integer value? Perhaps those places need to be updated too.

apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
```diff
@@ -2,4 +2,12 @@ defmodule NervesHubWWWWeb.OrgCertificateView do
   use NervesHubWWWWeb, :view
 
   alias NervesHubWWWWeb.LayoutView.DateTimeFormat, as: DateTimeFormat
+
+  def format_serial(big_long_integer) when is_integer(big_long_integer) do
```
fhunleth, line 6: I've learned that serial numbers are converted to strings throughout NervesHub. That means this will fail. However, I really think that serial numbers should be passed around as integers since that's how the X509 library natively uses them.

jjcarstens: So I'm all for this, but in order to do this we'd have to keep the serial stored in the DB as a string because some serials are too large to fit as a Postgres integer data type (which max is bigint, or int8). We'd run into something like this:

```
** (Postgrex.Error) ERROR 22003 (numeric_value_out_of_range) value "422154175989008335861639249273408275957749326847" is out of range for type integer
```

If we're okay storing as one way, but using as another, then we can do it. Though it will require that every DB query would need to be adapted to convert serials to strings on the lookup anyway. So...is it worth it?

jjcarstens (approved):

apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
fhunleth: I looked at the certificate handling code, and I'm more interested in catching the missing fields issue (this is #702). The "represent certificates as decimal number strings for Postgres" makes me sad, but I've lost motivation to address that.

apps/nerves_hub_www/lib/nerves_hub_www_web/views/org_certificate.ex
```diff
@@ -2,4 +2,17 @@ defmodule NervesHubWWWWeb.OrgCertificateView do
...
+    big_long_integer
+    |> Integer.to_string(16)
+    |> to_charlist()
+    |> Enum.chunk_every(2)
+    |> Enum.join(":")
+  end
+
+  def format_serial(serial_str) when is_bitstring(serial_str) do
```
fhunleth, line 14:
```suggestion
  def format_serial(serial_str) when is_binary(serial_str) do
```

apps/nerves_hub_www/test/nerves_hub_www_web/views/org_certificate_view_test.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHubWWWWeb.OrgCertificateViewTest do
+  use ExUnit.Case
+  alias NervesHubWWWWeb.OrgCertificateView
+
+  test "serial number is formatted in hex" do
+    OrgCertificateView.format_serial(112_346_101_875_805_641_052_401_911_002_393_715_100) ==
```
fhunleth, line 6:
```suggestion
    OrgCertificateView.format_serial("112346101875805641052401911002393715100") ==
```

fhunleth (approved):

mobileoverlord (approved):

PR #711, opened 2021-01-07
Device moving - Web

### LINKED ISSUE #687: Can't use the same firmware signing key in two organizations
**Describe the bug**
Register a firmware signing key and then try to register the same one in another org. It fails with an error message that it already exists, but it's not obvious why when you're in a different org.

Why would you do this?
1. If you need to create multiple orgs to segregate products managed by different teams, you still might want to keep the same dev firmware signing keys so that it's less confusing which one you use for development.
2. You don't care about firmware signatures on a hobby device. While NervesHub forces signatures, it can be nice to ship around devices to others that they can hook up and update via NervesHub. It seems wrong to disable signatures entirely since we want to promote that as a best practice, but letting people use known firmware keys seems reasonable and then it's easy for them to swap them out when they want to lock down their device.

jjcarstens:
Add support for moving devices between products/orgs.

The context is in place and this also adds support to complete the function in the web. The next step will be to reproduce from the API . 

In the move, this will attempt to copy any required firmware keys from one org to another if it detects that a device uses it. Coincidentally, this required I resolve #687 via 7e59db8

EDIT:

UI updated

https://user-images.githubusercontent.com/11321326/105080779-8cbbd100-5a4e-11eb-9c9b-7d6b540ed75e.mov

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
@@ -64,7 +64,7 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
   def handle_in("fwup_progress", %{"value" => percent}, socket) do
     Presence.update(
       socket.channel_pid,
-      "product:#{socket.assigns.device.product_id}:devices",
+      tracking_topic(socket),
       socket.assigns.device.id,
       %{fwup_progress: percent}
     )
@@ -75,7 +75,7 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
   def handle_in("status_update", %{"status" => status}, socket) do
     Presence.update(
       socket.channel_pid,
-      "product:#{socket.assigns.device.product_id}:devices",
+      tracking_topic(socket),
       socket.assigns.device.id,
       %{status: status}
     )
@@ -87,7 +87,7 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
     # Device sends "rebooting" message back to signify ack of the request
     Presence.update(
       socket.channel_pid,
-      "product:#{socket.assigns.device.product_id}:devices",
+      tracking_topic(socket),
       socket.assigns.device.id,
       %{rebooting: true}
     )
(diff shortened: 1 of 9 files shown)


fhunleth:
That's pretty cool and the video is super helpful. I'll look more at this tomorrow.

coryjewell:
@jjcarstens I'll spend some time today to help fix up the styling on that "move" menu

coryjewell:
I've got this set up and working but I think there are a couple of issues with using this navigation component here:
- The user must select a product but has to hover over organizations to get to the product. In the main navigation, organizations are clickable but here they are not so there are some inconsistencies.
- Experience is messy on mobile -- the solution for the main nav was to go full screen with the toggle at the top showing your org:product -- that isn't as elegant here. I'd suggest disabling this on mobile if we want to use this component.
- We're putting these nav elements into a non-nav location (this can obviously be fixed with some additional work)

My suggestion would be to have a simple select dropdown displaying products with their org name like:
- Org1: Product 1
- Org1: Product 2 
- Org2: Product 1
- Org3: Product 1
- etc

This is low effort and easiest for the user to use and understand (downside here could be if someone has 100s of products)

Screenshot

jjcarstens:
Thanks @coryjewell - I'm not apt with UI so I'm going to opt for the simple list that might be a little repetitive. If the UI becomes bothersome, someone more efficient and experienced than me can change the look. Changes coming right up...

jjcarstens:
Actually @coryjewell, if you could help with the UI of this new dropdown list we want, that would be fantastic. I thought it was "easy", but I'm failing miserably and am timing out. If the design is in place, I can get the rest of the functionality added and it would be incredibly helpful to me 🙏

coryjewell:
@jjcarstens I'll try to set aside some time tomorrow to work on this! 

I'll get the UI set up and then if you can just properly hook up functionality I think we'll be solid (so basically what you just said).

coryjewell:
Sorry for the delay, I'll have something up by EOD tomorrow

tonnenpinguin:
This would be super useful :)

Is there any way I can help out to move this forward?

jjcarstens:
Sure! It needs to be rebased and then polished. I think there might be one more edge case to consider, but I can't remember at this point 😞 it might be we need to consider soft deletes devices before attempting a move and making sure the device certificates go as well?

tonnenpinguin:
I rebased on the current main branch, but can't push here, so I updated the respective branch on our fork. 
Can you pull in the changes from there?

https://github.com/sportalliance/ft-nerves-hub-web/tree/device-moving

jjcarstens:
@tonnenpinguin I don't really see any changes in your branch pertaining to this device moving. Although there are a lot of other changes around ecs and running NH server. I've rebased this branch against main - Can you rebase against this branch, remove the irrelevant commits to this change, and then push it up again?

tonnenpinguin:
Thinking about this the sportalliance repo is probably not the best one to work on upstream changes. I'll have some time in the upcoming days to take a look at this in my private fork :)

tonnenpinguin:
I rebased your PR on the next branch, fixed some minor issues and implemented the hiding of the checkbox on mobile feature that was requested :)

I don't think I can commit on this one, so you might have to pull from my branch :)

https://github.com/sportalliance/ft-nerves-hub-web/tree/device_moving

jjcarstens:
Closing in favor of #823

PR #712, opened 2021-01-15
Switch to select input for moving devices

coryjewell:
### Why
- Input for moving a device needs to be properly styled
- Went away from reusing the main navigation as a form input and instead am using a select organized with `optgroup`s
- This form still needs to be hooked up to perform the previous functionality (as discussed)


coryjewell:
@jjcarstens This is the work we discussed in https://github.com/nerves-hub/nerves_hub_web/pull/711

jjcarstens:
I see there is an `apply` button - Would this instead allow us to trigger the move when selecting a drop down item instead (with the similar `data-confirm` behavior to give a chance to back out?

coryjewell:
Yup that's correct

jjcarstens (approved):

PR #714, opened 2021-01-22
bump elixir, erlang, alpine versions

jjcarstens:
This will be useful to have Elixir 1.11 (which is blocking #711 )


mobileoverlord:
Looks like this is failing because 1.11 requires us to reboot after applying sys config so we can configure the kernel. This error is happening on migration

```
ERROR! Cannot configure :kernel because :reboot_system_after_config has been set to false and :kernel has already been loaded, meaning any further configuration won't have an effect.

```

jjcarstens:
Fixed. I think restarting after a migration should be fine to ensure release is reloaded.

mobileoverlord (approved):

PR #715, opened 2021-01-25
Use DeviceCertificate fingerprints for SSL verification 

jjcarstens:
This adds SHA certificate and public key fingerprints for device certificates to be used in SSL functions for determining the trusted state of a certificate in NH.

Once DERs for devices have been collected and the existing device state is caught up, the DB lookup will rely soley on pinned device certificate fingerprints.

See
[OWASP Certificate and Public Key Pinning](https://owasp.org/www-community/controls/Certificate_and_Public_Key_Pinning) for more details


jjcarstens:
This is probably ready enough to be looked at - Tests need a lot of changing, but the logic flow is there

jjcarstens:
This is now official ready with extensive unit tests around SSL verification

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex
```diff
@@ -258,6 +273,25 @@ defmodule NervesHubWebCore.Devices do
     |> Repo.insert()
   end
 
+  @spec create_ca_certificate_from_x509(Org.t(), X509.Certificate.t(), binary() | nil) ::
```
fhunleth, line 276: I feel like all of the `*ca_certificate*` functions really don't belong in `devices.ex`. This isn't really your problem, but I can't shake the feeling of wanting to move it to avoid further complicating an already complicated file. Perhaps in a future PR, we move all of the create_ca_certificate and get_ca_certificates functions out...

apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
```diff
@@ -1,111 +1,184 @@
...
+    now = DateTime.utc_now()
+    is_after? = DateTime.compare(now, db_cert.not_after) == :gt
+    is_before? = DateTime.compare(now, db_cert.not_before) != :gt
+
+    if is_before? or is_after? do
+      # Maybe should be :cert_expired ?
+      :invalid_issuer
     else
-      _e -> :error
+      {:ok, db_cert}
```
fhunleth, line 160:
```suggestion
      :ok
```
This changes some other code above. It looked like `db_cert` was going to be rebound above. To avoid any confusion, I figured it would be better to just return `:ok` here.

apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
```diff
@@ -1,111 +1,184 @@
...
+  # This happens when public_key is matched in the DB,
+  # but no DB certs matches the incoming OTP Cert der
+  defp maybe_register_from_existing_public_key(otp_cert, device) do
+    with {:ok, cn} <- check_common_name(otp_cert),
+         true <- cn == device.identifier || :mismatched_cert,
+         {:ok, db_ca} <- check_known_ca(otp_cert),
+         {:ok, db_ca} <- check_expiration(db_ca),
+         true <- db_ca.org_id == device.org_id || :mismatched_org,
+         der = Certificate.to_der(otp_cert),
+         {:ok, _} <- :public_key.pkix_path_validation(db_ca.der, [der], []),
```
fhunleth, line 122: Is this line tested by the unit tests? It returns `{:error, {:bad_cert, reason}}` and I don't remember that in `ssl_test.exs`.

jjcarstens: Yes. Its tested on lines 121 and 126 - When checking device cert being expired or having bad signature

also for the case where the public key isn't know, its test on lines 170 and 185

fhunleth (approved):

PR #724, opened 2021-02-05
Support device import from CSV

jjcarstens:
Adds pages for device importing. It attempts to help a bit with identifying bad lines or data coming in with the expectation the user will fix those lines in the CSV and try later when expected. It is idempotent to make this case easier as well.

This probably needs quite a bit of UI fixing as most is copy/pasta from areas I knew worked, but I'm tapped out in my ability and its "good enough" for me (/cc @coryjewell @Mrjaco12 or anyone else good at UI..)

image

https://user-images.githubusercontent.com/11321326/107067565-aeca8880-679c-11eb-9094-fd9c44ab059e.mov


fhunleth (approved):

jjcarstens:
Updated this PR:

* There is now an arbitrary `1000` line limit for any uploaded CSV
* The organizations device limit is not honored when attempting to import

fhunleth:
>     * The organizations device limit is not honored when attempting to import

not == now?

mobileoverlord:
> > ```
> > * The organizations device limit is not honored when attempting to import
> > ```
> 
> 
> not == now?

This appears to be the case, looks like the code is tracking the device_limit

jjcarstens:
Derp 🤦 
Yep - thanks auto-correct

mobileoverlord (approved):

PR #726, opened 2021-02-08
Begin fixing dialyzer

ConnorRigby:
While looking around to add a feature, I noticed that

1) dialyzer isn't run in CI
2) the existing specs were mostly wrong
3) there aren't very many existing specs

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
@@ -3,8 +3,6 @@ defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
 
   import Ecto.Changeset
 
-  @type t :: %__MODULE__{}
-
   @required_params [
     :uuid,
     :product,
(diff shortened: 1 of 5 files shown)


fhunleth:
Thank you for adding some docs and specs!!!!

I think this is a pretty involved endeavor and would rather have some changes than all at once. I can't tell if the test error is due to the changes here or not. However, I'd be super happy to even have small steps. For example, enable Dialyzer, but don't fail CI if it fails. Then merge all of your doc updates. Then merge the easy Dialyzer fixes, etc.

@jjcarstens @mobileoverlord Are you two ok with this?

@ConnorRigby I see that you marked it as draft, but would you be comfortable getting us started with the really low risk changes?

ConnorRigby:
>  I see that you marked it as draft, but would you be comfortable getting us started with the really low risk changes?
yep i'm fine with that. I was hoping to at least get the existing failures cleaned up, but the CI change can be reverted to allow for these incremental updates to be applied.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
```diff
@@ -21,6 +19,25 @@ defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
...
+  @typedoc """
+  fwup meta data about firmware
+
+  [read more here](https://github.com/fwup-home/fwup#global-scope)
+  """
+  @type t :: %__MODULE__{
+          architecture: String.t(),
+          author: String.t() | nil,
+          description: String.t() | nil,
+          fwup_version: Version.build(),
```
fhunleth, line 31:
```suggestion
          fwup_version: Version.build() | nil,
```

fhunleth (approved):

mobileoverlord (approved):
Looks good to me.

PR #728, opened 2021-02-10
Use OTP 23.0.4

jjcarstens:
TL;DR is OTP >= 23.1 is making lots of changes around SSL/TLS, especially to support TLS 1.3. There some to be some kinks to work out for NH use-case so lets use the version we know works until that new hotness is ready

In OTP 23.1+, Devices that are already registered, but still send the device signer CA in the `cacerts` option will fail to connect if the signer CA has expired. This has always been ignored since the device is already registered.

In OTP 23.2+, there are some big changes to support TLS 1.3 which seem to have changed expected behavior for our `verify_fun/3` callback. Specifically, it seems to not validate the peer even when the device cert and the signer CA cert is included which subsequently means that first device connect fails

23.0.4 is the last known working OTP version for our setup, so move back to there while everything else is figured out


fhunleth (approved):

jjcarstens:
Tested the conditions in staging and all is 👌

PR #731, opened 2021-02-10
Add specs for functions involved in the firmware update payload

ConnorRigby:
This adds some specs and types for the payload that gets sent to the device when an update is available.
The `NervesHubWebCore.Firmwares.FirmwareMetadata.metadata` type isn't as specific as i wanted, but Dialyzer refuses to play ball. without listing the payload as `optional` which isn't actually true.
Would love a second look

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex
@@ -99,8 +99,8 @@ defmodule NervesHubWebCore.Firmwares do
   end
 
   @spec create_firmware(
-          Org.t(),
-          String.t(),
+          org :: Org.t(),
+          filepath :: Path.t(),
           params :: map(),
           opts :: [{:upload_file_2, upload_file_2()}]
         ) ::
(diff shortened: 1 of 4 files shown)


apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
```diff
@@ -38,6 +38,19 @@ defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
           version: Version.build()
         }
 
+  @type metadata :: %{
```
jjcarstens, line 41: I would say we actually stop defaulting to this map and use the structure that has been defined. It seems silly to define struct, but then not really use it except when loading from DB

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
```diff
@@ -38,6 +38,19 @@ defmodule NervesHubWebCore.Firmwares.FirmwareMetadata do
...
+          optional(:author) => String.t() | nil,
+          optional(:description) => String.t() | nil,
+          optional(:fwup_version) => Version.build() | nil,
+          optional(:misc) => String.t() | nil,
+          optional(:platform) => String.t(),
+          optional(:product) => String.t(),
+          optional(:uuid) => Ecto.UUID.t(),
+          optional(:vcs_identifier) => String.t() | nil,
+          optional(:version) => Version.build()
+        }
```
jjcarstens, line 52:
```suggestion
  @type metadata :: %{
          :architecture => String.t(),
          :author => String.t() | nil,
          :description => String.t() | nil,
          optional(:fwup_version) => Version.build() | nil,
          :misc => String.t() | nil,
          :platform => String.t(),
          :product => String.t(),
          :uuid => Ecto.UUID.t(),
          :vcs_identifier => String.t() | nil,
          :version => Version.build()
        }
```

Also, for simplicity, this should work as well. `fwup_version` only comes in the device channel and not from the firmware DB struct or fwup, therefor it is "optional"

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares/firmware_metadata.ex
ConnorRigby: I really like this idea, but it seems like something that can/should be saved for a followup PR

apps/nerves_hub_web_core/lib/nerves_hub_web_core/firmwares.ex
```diff
@@ -515,32 +525,37 @@ defmodule NervesHubWebCore.Firmwares do
...
   end
 
+  @typep metadata_string() :: String.t()
+  @typep metadata_key() :: String.t()
+  @typep metadata_value() :: String.t() | nil
+
+  @spec get_fwup_metadata(Path.t()) :: {:ok, metadata_string()} | {:error, String.t()}
   defp get_fwup_metadata(filepath) do
-    case System.cmd("fwup", ["-m", "-i", filepath]) do
+    case System.cmd("fwup", ["-m", "-i", IO.iodata_to_binary(filepath)]) do
```
fhunleth, line 534:
```suggestion
    case System.cmd("fwup", ["-m", "-i", filepath]) do
```

If the spec is correct, then this should be fine.

fhunleth (approved):

PR #732, opened 2021-02-12
set `update_available: true` in when a deployment is updated

ConnorRigby:
this is a work around fix for an issue where the payload that gets sent to devices when an update is available is
different if the update happened on `join` or while a device is connected.

I don't believe this is a long term fix, but a work around. 

There are **two different** functions that calculate if a device should receive an update, the other being called `resolve_update` in the same module as is changed here. 
While trying to document and make sense of this payload on the [client side here](https://github.com/nerves-hub/nerves_hub_link_common/commit/db64d190f331184535c26eab253020cec24afb06#diff-3b8725c177a5e6ae0bf2f15149de38d16bf617b238617ecd7cfce5f94a503247) 
i only took that first function into account as i assumed that the other case would use the same function. I think ultimately what needs to happen here is a small refactor to get both of these systems to output the same data as i don't believe having two different flavors of this very important payload is sustainable long term.

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/views/device_view.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/views/device_view.ex
@@ -1,6 +1,7 @@
 defmodule NervesHubDeviceWeb.DeviceView do
   use NervesHubDeviceWeb, :view
   alias NervesHubDeviceWeb.DeviceView
+  alias NervesHubWebCore.Devices.UpdatePayload
 
   def render("show.json", %{device: device}) do
     %{data: render_one(device, DeviceView, "device.json")}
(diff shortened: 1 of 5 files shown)


jjcarstens (approved):

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
```diff
@@ -140,7 +140,9 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
     :ok
   end
 
-  def terminate(_reason, _state), do: :ok
+  def terminate(reason, _state) do
+    IO.inspect(reason, label: "????")
```
jjcarstens, line 144: Is this still needed? Maybe change to Logger statement?

ConnorRigby:
This has a side effect of fixing the documentation found [here](https://docs.nerves-hub.org/device-api/phoenix-channel#server-events)

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices/update_available.ex
```diff
@@ -0,0 +1,48 @@
+defmodule NervesHubWebCore.Devices.UpdateAvailable do
```
jjcarstens, line 1:
```suggestion
defmodule NervesHubWebCore.Devices.UpdatePayload do
```
I would probably call this `UpdatePayload` since it is more than the context of "update available"

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
ConnorRigby: whoops missed this one. I don't think it needs a log, this function never even got called while testing for me.

jjcarstens (approved):

fhunleth (approved):

PR #737, opened 2021-02-17
Adjust payload for Device Channel join reply

jjcarstens:
This fixes the breaks with the new UploadPayload.

Unfortunately, it negates even having an UploadPayload structure, but only on initial reply when a device joins the Phoenix channel. This is a fault in `:nerves_hub_link` that considers the existence of `firmware_url` key to mean an update is available. When update is not actually available, that would be nil and cause `:nerves_hub_link` to completely crash.

This is fixed in https://github.com/nerves-hub/nerves_hub_link/pull/63 on the client side.

For now, this cuts up our nice structure to prevent existing devices in the field from breaking. In the future, this can be removed once enough devices have updated.

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
@@ -20,6 +20,8 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
 
   alias Phoenix.Socket.Broadcast
 
+  require Logger
+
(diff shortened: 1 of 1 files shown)


fhunleth:
For better or worse, I think we need to discuss this fix before the merge since there is no removing the old way. Devices could be in a warehouse and turned on a year from now or worse. There are only two options that I can think of:

1. Keep the old way
2. Bump the API version (whatever that means) so that NervesHub knows whether the device needs the v1 message or the newer v2 message.

fhunleth (approved):

Issue #316, opened 2018-12-18
Require proof of private key when uploading signer certificates


fhunleth:
Since the signer/CA certificates enable devices to be auto-registered, it's really important that NervesHub gets the right ones and that they're legit. AWS IoT has the same issue. Their solution is to have you sign them to prove ownership of the private key. It seems like what they did would be reasonable for us. See this screenshot:

![ca_registration](https://user-images.githubusercontent.com/64669/50159738-680c3280-02a5-11e9-972b-5b64373f3199.png)

PR #741, opened 2021-02-26
Fix registrations using NervesHubCA provided signer cert

jjcarstens:
Device signer CA provided by NervesHubCA are added to the DB and used when user does not supply their own Signer CA. Adding to the DB prevents someone else from claiming ownership.

Eventually be removed by https://github.com/nerves-hub/nerves_hub_web/issues/316


jjcarstens:
Tested in staging 👍

ConnorRigby (approved):

fhunleth (approved):

PR #742, opened 2021-03-01
Don't record an update request to the audit log when not requesting one

ConnorRigby:


apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
```diff
@@ -121,12 +140,14 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
     # If we get here, the device is connected and high probability it receives
     # the update message so we can Audit and later assert on this audit event
     # as a loosely valid attempt to update
-    AuditLogs.audit!(deployment, socket.assigns.device, :update, %{
-      from: "broadcast",
-      send_update_message: true
-    })
+    if not socket.assigns.download_uuid do
```
jjcarstens, line 143:
```suggestion
    if is_nil(socket.assigns.download_uuid) do
```

`not` will require an explicit true/false condition for comparison. `nil` or a binary will fail it

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
```diff
@@ -42,15 +42,18 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
         Devices.resolve_update(device, deployments)
         |> build_join_reply()
 
-      if join_reply.update_available do
+      if join_reply.update_available and params[:download_uuid] == nil do
```
jjcarstens, line 45:
```suggestion
      if join_reply.update_available and params["download_uuid"] == nil do
```

I believe `params` will be a JSON decoded map here, so it will need to check binary keys

apps/nerves_hub_device/lib/nerves_hub_device_web/channels/device_channel.ex
```diff
@@ -183,4 +183,16 @@ defmodule NervesHubDeviceWeb.DeviceChannel do
   end
 
   defp build_join_reply(up), do: up
+
+  defp should_audit_log?(join_reply, params)
```
jjcarstens, line 187: There is no default, so this prob isn't needed?

jjcarstens (approved):

PR #750, opened 2021-04-13
Update dependencies

ConnorRigby:


jjcarstens (approved):

apps/nerves_hub_web_core/lib/nerves_hub_web_core/repo.ex
```diff
@@ -11,8 +11,6 @@ defmodule NervesHubWebCore.Repo do
     {:ok, Keyword.put(opts, :url, System.get_env("DATABASE_URL"))}
   end
 
-  def reload(%module{id: id}), do: get(module, id)
-
```
fhunleth, line 15: Should these lines be deleted?

jjcarstens: Yes. This is now a function of Ecto.Repo in newer versions. So it's clashing with the dependency

fhunleth (approved):

PR #757, opened 2021-05-12
Just In Time Provisioning

ConnorRigby:
This PR enables automatically registering Devices if enabled by the CA Cert


apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices.ex
```diff
@@ -327,14 +327,15 @@ defmodule NervesHubWebCore.Devices do
   def get_ca_certificates(%Org{id: org_id}) do
     from(ca in CACertificate, where: ca.org_id == ^org_id)
     |> Repo.all()
+    |> Repo.preload([:jitp])
```
jjcarstens, line 330:
```suggestion
    from(ca in CACertificate, where: ca.org_id == ^org_id, preload: :jitp)
    |> Repo.all()
```
`Repo.preload` called after the query will loop through the list and do more lookups to preload. Its prob better off to include the lookups as part of the query transaction

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices/ca_certificate.ex
```diff
@@ -22,8 +22,32 @@ defmodule NervesHubWebCore.Devices.CACertificate do
...
+    def changeset(jitp, params) do
+      jitp
+      |> cast(params, [:tags, :description])
+      |> validate_required([])
+    end
+  end
+
   schema "ca_certificates" do
     belongs_to(:org, Org, where: [deleted_at: nil])
+    belongs_to(:jitp, JITP)
```
jjcarstens, line 50:
```suggestion
    has_one(:jitp, JITP)
```

Should this be a `has_one`? Otherwise, but `ca_certificate` and `jitp` belong to each other with seems circular?

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices/ca_certificate.ex
```diff
@@ -22,8 +22,32 @@ defmodule NervesHubWebCore.Devices.CACertificate do
...
+      # Will be coppied to the device when it's created
+      field(:tags, NervesHubWebCore.Types.Tag)
+      field(:description, :string)
+      timestamps()
+    end
+
+    def changeset(jitp, params) do
+      jitp
+      |> cast(params, [:tags, :description])
+      |> validate_required([])
```
jjcarstens, line 44: Is this needed if it is not validating anything?

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_certificate_controller.ex
```diff
@@ -104,4 +129,25 @@ defmodule NervesHubWWWWeb.OrgCertificateController do
...
+  end
+
+  @spec validate_csr(Plug.Conn.t(), tuple()) :: :ok | {:error, :invalid_csr}
+  def validate_csr(conn, csr) do
+    session = get_session(conn)
+
+    # This line could probably raise. the only way to get here is by mucking around in Session Storage
+    code = session["registration_code"] || "nil check"
+
+    case Certificate.get_common_name(csr) do
```
jjcarstens, line 145: Do we need to do any path validations on this CSR to make sure its valid? Is there a way that the common name could be put into the CSR when its total garbage?

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_certificate_controller.ex
```diff
@@ -17,8 +17,18 @@ defmodule NervesHubWWWWeb.OrgCertificateController do
...
 
-  def new(conn, _params) do
-    render(conn, "new.html", changeset: %Changeset{data: %CACertificate{}})
+  def new(%{assigns: %{org: org}} = conn, _params) do
+    registration_code = generate_code()
+    # todo: put this somewhere more sensible
+    products = NervesHubWebCore.Repo.preload(org, :products).products
+
+    conn
+    |> put_session(:registration_code, registration_code)
```
jjcarstens, line 26: This is probably fine for now, but I dont think we'll have the session luxury when trying to support this from the CLI and might need to do it somewhere else. (Or disallow using the CLI for CA signer certificate registration

apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
```diff
@@ -18,6 +19,7 @@ defmodule NervesHubWebCore.Products.Product do
     has_many(:firmwares, Firmware)
     has_many(:product_users, ProductUser)
     has_many(:users, through: [:product_users, :user])
+    has_one(:jitp, CACertificate.JITP)
```
jjcarstens, line 22:
```suggestion
    has_one(:ca_certificate, CACertificate)
    has_one(:jitp, CACertificate.JITP, through: :ca_certificate)
```

Is this needed? If so, maybe we are better server using `through` or preloading?

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices/ca_certificate.ex
ConnorRigby: jitp doesn't `belong_to` jitp here. `has_one` is the opposite of `belongs_to` where `has_one` doesn't actually use any database records if i remember correctly. the `ca_certificates` table has a column called `jitp_id` but the `jitp` table does **not** have a `ca_certificate_id`. I'm pretty sure this is "normal" for this kind of thing, but i've also been known to get it backwards.

apps/nerves_hub_web_core/lib/nerves_hub_web_core/devices/ca_certificate.ex
ConnorRigby: I think @fhunleth wants both of these fields to be required. Just left em out while debugging

apps/nerves_hub_web_core/lib/nerves_hub_web_core/products/product.ex
ConnorRigby: This isn't actually used anywhere, i just wanted to be consistent with the relationships in case we need it in the future.

apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
```diff
@@ -162,13 +162,32 @@ defmodule NervesHubDevice.SSL do
...
+      _ ->
+        case Devices.create_device(%{
+               identifier: cn,
+               org_id: org_id,
+               product_id: jitp.product_id,
+               tags: jitp.tags,
+               description: jitp.description
+             }) do
+          {:ok, _d} = resp -> resp
+          _ -> :device_not_registered
```
jjcarstens, line 186:
```suggestion
          _ -> :device_registration_failed
```

It might be nice for more explicit error here

jjcarstens (approved):

apps/nerves_hub_www/lib/nerves_hub_www_web/controllers/org_certificate_controller.ex
```diff
@@ -52,24 +61,38 @@ defmodule NervesHubWWWWeb.OrgCertificateController do
...
            aki: aki,
            ski: ski,
-           not_before: not_before,
-           not_after: not_after,
+           not_before: cert_not_before,
+           not_after: cert_not_after,
            der: X509.Certificate.to_der(cert),
-           description: Map.get(params["ca_certificate"], "description")
+           description: Map.get(params["ca_certificate"], "description"),
+           jitp: Map.get(params, "jitp")
```
jjcarstens, line 95:
```suggestion
           jitp: Map.get(params["ca_certificate"], "jitp")
```

`jitp` is nested inside of `ca_certificate` fields

jjcarstens:
Looks like only one more minor change is needed.

And its worth noting that this could be affected by https://github.com/nerves-hub/nerves_hub_web/issues/762. Specifically, if you "delete" a device through Web UI, it is soft deleted and next connect will still work, but will not be visible anywhere in the UI.

One idea around this is that during the `verify_cert` check where we lookup a device, we can scope to exclude soft deleted devices. Then later when checking for JITP, we restore the soft deleted record if JITP exists

PR #764, opened 2021-06-01
use hexpm images for CI testing

jjcarstens:


fhunleth:
Any reason to stick with an old Erlang or is that what's in prod?

jjcarstens:
K, this is working now. We'll need it for #757 

It abandons javascript testing because something is going on with circleci cache that I just dont want to deal with right now. I dont feel like we're missing out on much there since it only tests one js function

fhunleth (approved):

PR #765, opened 2021-06-02
Ignore SSL auth when device has been soft deleted

jjcarstens:
Right now if a device is deleted, it is a soft delete. When it
attempts to connect again, SSL verification will succeed but the
device will not show anywhere in the UI.

This instead fixes that to prevent authentication for devices that are
deleted and require the user to explicitly restore it.

### DIFF
--- a/docs/device_authentication.md
+++ b/docs/device_authentication.md
@@ -26,7 +26,10 @@ device's certificate on file.
 Devices connect to NervesHub over TLS. NervesHub requests a certificate from the
 device. After the TLS stack receives the certificate, it does some processing
 and then passes it on to NervesHub code that does the following.  Note that
-nothing in the X.509 certificate from the device is trusted at the beginning.
+nothing in the X.509 certificate from the device is trusted at the beginning and
+if the device has been soft deleted, all certificate validation is skipped and
+authentication halted. The device record will need to be restored (or completely
+deleted) to authenticate again.
 
 1. Compute a SHA-1 hash on the device certificate in DER form. This is
    called the certificate fingerprint.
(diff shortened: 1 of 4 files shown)


mobileoverlord:
Looks good, I think we only match on `{:ok, _}` for the returns, just thinking maybe we should wrap it in an `{:error, :device_deleted}` or something.

mobileoverlord (approved):

jjcarstens:
I don't know if we'd want to wrap it since the device record still technically exists and sometimes we do want it from the DB, so we don't need to consider it an error.

Though I guess in these specific lookup functions we could do that? 🤔

ConnorRigby (approved):

docs/device_authentication.md
```diff
@@ -26,7 +26,8 @@ device's certificate on file.
 Devices connect to NervesHub over TLS. NervesHub requests a certificate from the
 device. After the TLS stack receives the certificate, it does some processing
 and then passes it on to NervesHub code that does the following.  Note that
-nothing in the X.509 certificate from the device is trusted at the beginning.
+nothing in the X.509 certificate from the device is trusted at the beginning and if the device has been soft deleted, all certificate validation is skiped and authentication halted. The device record will
+need to be restored (or completed deleted) to authenticate again.
```
tonnenpinguin, line 30:
```suggestion
nothing in the X.509 certificate from the device is trusted at the beginning and if the device has been soft deleted, all certificate validation is skipped and authentication halted. The device record will
need to be restored (or completely deleted) to authenticate again.
```

tonnenpinguin: :)

Issue #699, opened 2020-11-23
Deleting a device from the web works, but goes to the wrong view


fhunleth:
**Describe the bug**
Delete a device from the web UI. The web UI reports an error and takes you to the NervesHub home page. It does look like the device is deleted, though.

**To Reproduce**

1. Create a device on a product (in any way)
2. Log into the web UI
3. Find the device and delete it

brianberlin:
@fhunleth i wasn't able to reproduce this issue. I found something that was maybe introduced with 4e6ab4b and resolved by 9c7e1fb but nothing in recent commits. think maybe it's stale now.

jjcarstens:
Fixed with #769

PR #769, opened 2021-06-03
Device uniqueness fixes

jjcarstens:
099249f changes the uniqueness requirement for devices. Previously, a device identifier could be created in an org if that device was deleted. This causes lots of errors if it ever tries to connect again and leaves potential for lots of orphaned device records so we need to not exclude deleted devices in the index. Part of the migration will delete duplicate records before restoring the index.

The rest is to support being able to restore/destroy soft deleted device records in order to avoid unknown errors trying to recreate devices or connecting deleted ones


jjcarstens:
confirmed this does what I want it to in staging 🙌

ConnorRigby (approved):

jjcarstens:
This also fixes #699

PR #770, opened 2021-06-04
Fix allow_signups runtime config is not available

pojiro:
The reason why this issue occurs is due to the fact that
the value of allow_signups is assigned in the init function of plug.

This is mentioned in [the Plug documentation](https://github.com/elixir-plug/plug/blob/fa579592ebf53306e3a4b13e2414a40997add1f7/lib/plug.ex#L22-L24).
> Note that init/1 may be called during compilation and as such
> it must not return pids, ports or values that are specific to the runtime.

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/plugs/allow_uninvited_signups.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/plugs/allow_uninvited_signups.ex
@@ -6,30 +6,34 @@ defmodule NervesHubWebCore.Plugs.AllowUninvitedSignups do
 
   alias Phoenix.Controller
 
-  def init(opts) do
-    Keyword.put(opts, :allow_signups, allow_signups())
-  end
-
-  def call(conn, allow_signups: true), do: conn
+  def init(opts), do: opts
 
   def call(%{private: %{phoenix_router: APIRouter}} = conn, _opts) do
-    conn
-    |> put_resp_header("content-type", "application/json")
-    |> send_resp(
-      403,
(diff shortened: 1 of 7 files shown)


jjcarstens (approved):

apps/nerves_hub_web_core/lib/nerves_hub_web_core/plugs/allow_uninvited_signups.ex
```diff
@@ -6,30 +6,34 @@ defmodule NervesHubWebCore.Plugs.AllowUninvitedSignups do
...
+    else
+      conn
+      |> Controller.put_flash(:error, "Sorry signups are not enabled at this time.")
+      |> Controller.redirect(to: "/")
+      |> halt()
+    end
   end
 
-  defp allow_signups do
+  defp allow_signups() do
```
tonnenpinguin, line 36: How about renaming this to `allow_signups?`? This might make it a bit clearer that this is returns a boolean

pojiro: Agreed. Thank you.

pojiro:
My PR is failing the test.
I've also confirmed that running the test multiple times causes the error at local. I'm currently investigating.

pojiro:
The [last commit](https://github.com/nerves-hub/nerves_hub_web/pull/770/commits/bc9e57d990e527b1aaeebb6a68ea58f5d09020a6) fixes the problem that was breaking the test.

jjcarstens (approved):

PR #794, opened 2022-01-11
New Feature: Import microchip tpm file

ConnorRigby:
This implements a new uploader to process preprovisioned modules.
See the official docs for more information: https://ww1.microchip.com/downloads/en/DeviceDoc/TrustPlatformManifestFileFormat_2019-09-26_A.pdf

### DIFF
--- a/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products.ex
+++ b/apps/nerves_hub_web_core/lib/nerves_hub_web_core/products.ex
@@ -229,6 +229,25 @@ defmodule NervesHubWebCore.Products do
     Product.changeset(product, %{})
   end
 
+  def parse_json_data(%{"header" => %{"uniqueId" => serial}} = data) do
+    certificate =
+      data
+      |> Map.fetch!("payload")
+      |> Base.url_decode64!(ignore: :whitespace, padding: false)
+      |> Jason.decode!()
+      |> Map.fetch!("publicKeySet")
+      |> Map.fetch!("keys")
+      |> Enum.at(0)
+      |> Map.fetch!("x5c")
+      |> Enum.at(0)
+      |> Base.decode64!(ignore: :whitespace, padding: false)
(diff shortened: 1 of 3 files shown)


apps/nerves_hub_web_core/lib/nerves_hub_web_core/products.ex
```diff
@@ -229,6 +229,25 @@ defmodule NervesHubWebCore.Products do
     Product.changeset(product, %{})
   end
 
+  def parse_json_data(%{"header" => %{"uniqueId" => serial}} = data) do
```
fhunleth, line 232:
```suggestion
  def parse_jose_data(%{"header" => %{"uniqueId" => serial}} = data) do
```

I skimmed the PDF. It sound like the intention is that their file follow Javascript Object Signing and Encryption (JOSE) and in theory, this should work for anyone else who exports JOSE files. How about renaming this and the following references `jose`?

apps/nerves_hub_www/lib/nerves_hub_www_web/templates/product/import.html.leex
```diff
@@ -40,6 +41,17 @@
       <%= live_file_input @uploads.csv, class: "upload-select"%>
     </div>
   </form>
+
+  <form phx-change="validate" phx-drop-target="<%= @uploads.json.ref %>">
+    <h1> Upload a JSON file: </h1>
```
fhunleth, line 46:
```suggestion
    <h1> Upload a JOSE file (e.g., TrustPlatform Manifest File): </h1>
```

I'm not sure where to best document this. It just feels really useful and that people should know.

ConnorRigby: Updated!

fhunleth (approved):
Looks good to me.

fhunleth:
Thanks! 🚀

PR #813, opened 2022-04-07
Don't push "latest" tag to Docker Hub from next-main branch

dnrce:
The migrate and deploy tasks both use the versioned tag:

https://github.com/nerves-hub/nerves_hub_web/blob/f72afa9868c41519126dbe7b9ece5fb7d9688bbd/.circleci/config.yml#L57

https://github.com/nerves-hub/nerves_hub_web/blob/f72afa9868c41519126dbe7b9ece5fb7d9688bbd/.circleci/config.yml#L72

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,6 @@ docker_build_release: &docker_build_release
     command: |
       docker build \
         -t nerveshub/$APP_NAME:$DOCKER_TAG \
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):

fhunleth:
@mobileoverlord - Could you confirm that you no longer need this in `next-main`?

PR #821, opened 2022-05-30
Fix S3 config

tonnenpinguin:
This fixes the default s3 host, removes unnecessary config overrides and makes sure the presigned download url doesn't require an extra redirect

Before this the generated urls would look like `https://amazonaws.com/nerves-hub-production-web/firmware/3/[...]`
After they evaluate to `https://nerves-hub-production-web.s3.amazonaws.com/firmware/3/[...]`


jjcarstens (approved):

tonnenpinguin:
In order to get uploading to work I had to remove these lines:
`config :ex_aws, :s3, host: System.get_env("AWS_HOST", "s3.amazonaws.com")`

The reason is that the s3 host actually depends on the AWS region being used and is not static.
See this issue: https://github.com/ex-aws/ex_aws/issues/278

apps/nerves_hub_web_core/mix.exs
```diff
@@ -47,6 +47,7 @@ defmodule NervesHubWebCore.MixProject do
     [
       {:phoenix, "~> 1.6"},
       {:phoenix_html, "~> 3.1"},
+      {:phoenix_live_view, "~> 0.16"},
```
jjcarstens, line 50: More just curious - Why is live view needed in the core?

apps/nerves_hub_web_core/mix.exs
```diff
@@ -67,7 +68,8 @@ defmodule NervesHubWebCore.MixProject do
       {:jason, "~> 1.2", override: true},
       {:ecto, "~> 3.4", override: true},
       {:mox, "~> 1.0", only: [:test, :dev]},
-      {:nimble_csv, "~> 1.1"}
+      {:nimble_csv, "~> 1.1"},
+      {:base62, "~> 1.2"}
```
jjcarstens, line 72: This is fine. But typically you can just add `extra_applications: [:base62]` to the project since its a dep of the root umbrella.

jjcarstens (approved):

apps/nerves_hub_web_core/mix.exs
tonnenpinguin: Because some templates where updated to heex, which apparently requires live_view to be parsed

apps/nerves_hub_web_core/mix.exs
tonnenpinguin: Didn't know that. Cool, thanks! :)

PR #823, opened 2022-06-02
Device moving

jjcarstens:
Replacement for #711

Support moving devices between orgs.

/cc @tonnenpinguin


fhunleth:
@jjcarstens Could you rebase when you get a chance?

fhunleth (approved):

PR #834, opened 2022-06-10
Start tracking version history

jjcarstens:
While NervesHub has been running in many production environments,
it has not been versioned with records of changes. This will set
the initial version and start tracking from here.

This will also allow us tag docker builds with version numbers


.circleci/config.yml
```diff
@@ -18,14 +18,18 @@ docker_env: &docker_env
       else
         TAG=$CIRCLE_TAG
       fi
+      VERSION=$(cat VERSION | tr -d '\n')
```
dnrce, line 21: "uuoc"
```suggestion
      VERSION=$(tr -d '\n' < VERSION)
```

.circleci/config.yml
```diff
@@ -37,6 +41,9 @@ docker_save: &docker_save
       docker save \
         nerveshub/$APP_NAME:$DOCKER_TAG \
         -o /docker/$APP_NAME-$DOCKER_TAG.tar
+      docker save \
+        nerveshub/$APP_NAME:$DOCKER_VERSION_TAG \
+        -o /docker/$APP_NAME-$DOCKER_VERSION_TAG.tar
```
dnrce, line 46: This and `&docker_import` could be made slightly more efficient by only saving this image once, and adding any desired tags after loading but before pushing. But it probably isn't much time and I don't think we're paying for the storage or bandwidth so 🤷.

jjcarstens: Ya, I think the whole CI setup needs a refresh, but Its not something I have a lot of time to dedicate to right now - So I'm opting for the quickest win

dnrce (approved):

PR #835, opened 2022-06-20
Fix deployment toggling

tonnenpinguin:
For some reason LiveView kept sending an empty string instead of the boolean value. Maybe there's another fix for this?

### DIFF
--- a/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/deployment/show.html.heex
+++ b/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/deployment/show.html.heex
@@ -1,7 +1,7 @@
 <div class="action-row">
   <%= link "All Deployments", to: Routes.deployment_path(@socket, :index, @org.name, @product.name), class: "back-link"%>
   <div class="btn-group" role="group" aria-label="Deployment Actions">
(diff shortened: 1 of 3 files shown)


apps/nerves_hub_www/lib/nerves_hub_www_web/templates/deployment/show.html.heex
```diff
@@ -1,7 +1,7 @@
 <div class="action-row">
   <%= link "All Deployments", to: Routes.deployment_path(@socket, :index, @org.name, @product.name), class: "back-link"%>
   <div class="btn-group" role="group" aria-label="Deployment Actions">
-    <a class="btn btn-outline-light btn-action" aria-label={"Turn #{opposite_status(@deployment)} deployment"} phx-click="toggle_active" phx-value-isactive={!@deployment.is_active}>
+    <a class="btn btn-outline-light btn-action" aria-label={"Turn #{opposite_status(@deployment)} deployment"} phx-click={JS.push("toggle_active", value: %{is_active: !@deployment.is_active})}>
```
jjcarstens, line 4:
```suggestion
    <a class="btn btn-outline-light btn-action" aria-label={"Turn #{opposite_status(@deployment)} deployment"} phx-click="toggle_active" phx-value-is_active={"#{!@deployment.is_active}"}>
```
It seems LV now wants string literals and no longer wants to accept boolean values with `phx-value-*` - I just tested this and it works and seems less overkill that including the whole JS hook

PR #836, opened 2022-06-27
Run CI on tagged commits

dnrce:
CircleCI defaults to not running for tags, so any dependent jobs need an
explicit, permissive tag filter in order to run. See
https://circleci.com/docs/2.0/configuration-reference#tags.


jjcarstens (approved):

jjcarstens:
Thanks!

PR #837, opened 2022-06-29
Fix deployment#show test

jjcarstens:
Didn't look closely enough to the last deployment show page fix to see it didn't actually build and test the fork


tonnenpinguin:
Oh, sorry about that!
Aren't tests run for PRs?

jjcarstens:
Technically, all the settings are on. But it seems to only run one piece of the test suite for PR's from forks, and not the whole thing. ¯\_(ツ)_/¯ 
image

PR #839, opened 2022-07-01
Allow pinned device certs without a registered CA

### LINKED ISSUE #838: Known devices with expired CAs can no longer connect
**Describe the bug**
In our org we manually pin certs for each of our devices, rendering the device CA chain-of-trust essentially moot. We don't have CAs added to the org, because our understanding has been that org device CAs are only needed for trust during JITP. This has worked well for us in the past, but after an attempted upgrade to NervesHub v1.0.0, some of our devices were unable to reconnect to NervesHub due to a certificate error. The client logs:
```
13:59:14.466 [error] GenStateMachine #PID<0.21479.49> terminating
** (exit) {:socket_error, {:tls_alert, {:internal_error, 'TLS client: In state connection received SERVER ALERT: Fatal - Internal Error\n '}}}
    (stdlib 3.13) gen_statem.erl:1360: :gen_statem.loop_state_callback_result/11
    (stdlib 3.13) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
```
and the device server logs:
```
14:30:53.432 [notice] TLS :server: In state :wait_cert at ssl_handshake.erl:362 generated SERVER ALERT: Fatal - Internal Error
 - {:unexpected_error, {:case_clause, :cert_expired}}
```
The affected devices were provisioned with CAs that have since expired. Even though their device certs are known, they are now being rejected when attempting to connect. This seems to be related to OTP 24 and specifically to https://github.com/nerves-hub/nerves_hub_web/commit/e67f4a64aa14592f62cdd1204f5d2c44aefdfee9.

**To Reproduce**
1. Provision a device with a pinned cert (i.e. non-JITP) in an older version of NervesHub (tested with 8c1dc08).
2. Update NervesHub past e67f4a6.
3. The device can no longer connect.

**Expected behavior**
The device is able to connect despite having an expired CA.

jjcarstens:
Fixes #838

We were previously requiring that the signer CA to be registered even if the
device certificate was already pinned, which was incorrect. This fixes that to
skip the check if we're being presented with a signer CA so that validation of the
device certificate can happen.

If the device cert is pinned, the check will pass. If not, it will go through the
normal validation flow and fail due to an expired, unregistered signer CA anyway

### DIFF
--- a/apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
+++ b/apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
@@ -43,30 +43,20 @@ defmodule NervesHubDevice.SSL do
   # This can happen if the Signer CA is not included in the request
   # and only the device cert/key is. Or if some other unknown CA
   # was included.
-  def verify_fun(otp_cert, {:bad_cert, :unknown_ca}, state) do
+  def verify_fun(otp_cert, {:bad_cert, err}, state) when err in [:unknown_ca, :cert_expired] do
     aki = Certificate.get_aki(otp_cert)
     ski = Certificate.get_ski(otp_cert)
 
     if aki == ski do
       # Because Signer CAs are required to be registered first, we don't
-      # really care about it coming in here
+      # really care about it coming in here. Likewise, if this is an
+      # unregistered CA, we can just move on so that the device can
+      # still attempt to present it's certificate to check if it has been
+      # pinned or not. Veririfcation will fail there if the device cert
+      # and it's signer CA is unknown
       {:valid, state}
     else
       do_verify(otp_cert, state)
-    end
-  end
-
-  def verify_fun(otp_cert, {:bad_cert, :cert_expired}, state) do
-    # If the CA is expired but already registered then we should
-    # still allow it in the request. If this is a device attempting
-    # to register, the validation will fail later on due to the expired.
-    #
-    # If this is an existing device presenting an expired cert in the chain
-    # then allowing the expired CA cert prevents the request from being
-    # terminating prematurely
-    case check_known_ca(otp_cert) do
-      {:ok, _ca} -> {:valid, state}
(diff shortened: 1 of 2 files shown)


apps/nerves_hub_device/lib/nerves_hub_device/ssl.ex
```diff
@@ -43,33 +43,23 @@ defmodule NervesHubDevice.SSL do
   # This can happen if the Signer CA is not included in the request
   # and only the device cert/key is. Or if some other unknown CA
   # was included.
-  def verify_fun(otp_cert, {:bad_cert, :unknown_ca}, state) do
+  def verify_fun(otp_cert, {:bad_cert, err}, state) when err in [:unknown_ca, :cert_expired] do
```
fhunleth, line 46: Could you remind me why the device certificate is being checked for expiration if it's pinned?

jjcarstens: It is not. We are reacting here to the SSL libs detection of the device/ca cert being expired or unknown. In the `do_verify/2` below, if the cert is in our DB, then it is allowed (even if expired).

fhunleth: I'm slightly unsure of your answer. Let me try to restate to make sure that I understand.

The SSL library IS checking the device certificate's expiration BEFORE the check for whether the certificate is pinned.

My question is on why any code (no matter where it exists) is checking any field in the certificate before the hash of the certificate is checked.

fhunleth: I spoke with Jon offline. It looks like this is the best we can do without a decent amount of work since some path verification is done no matter what before we get a chance to add checks in the `verify_fun` callback.

fhunleth (approved):

dnrce:
Thanks, folks!!

PR #841, opened 2022-07-05
Add test for expired CA

LostKobrakai:
Related to #838. 

As out internal issue was caused by the CA being expired we wanted to add a test for that scenario as well to prevent eventual regressions.

### DIFF
--- a/apps/nerves_hub_device/test/nerves_hub_device_web/channels/websocket_test.exs
+++ b/apps/nerves_hub_device/test/nerves_hub_device_web/channels/websocket_test.exs
@@ -183,7 +183,9 @@ defmodule NervesHubDeviceWeb.WebsocketTest do
       refute_receive({"presence_diff", _})
     end
 
-    test "already registered certificate with expired signer CA can connect", %{user: user} do
+    test "already registered expired certificate with expired signer CA can connect", %{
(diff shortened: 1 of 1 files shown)


apps/nerves_hub_device/test/nerves_hub_device_web/channels/websocket_test.exs
```diff
@@ -182,6 +182,79 @@ defmodule NervesHubDeviceWeb.WebsocketTest do
       assert Presence.device_status(device) == "online"
       refute_receive({"presence_diff", _})
     end
+
+    test "already registered certificate with expired signer CA can connect", %{user: user} do
```
jjcarstens, line 186:
```suggestion
    test "already registered expired certificate with expired signer CA can connect", %{user: user} do
```
I believe the signer expiration is already tested  on [line 405](https://github.com/nerves-hub/nerves_hub_web/blob/main/apps/nerves_hub_device/test/nerves_hub_device_web/channels/websocket_test.exs#L405) with `valid certificate expired signer can connect`.

But this new test does check that both the pinned cert and signer are expired, so probably worth while to keep in

jjcarstens (approved):

Issue #843, opened 2022-07-25
v1.0.2, cannot upload firmware by GUI


pojiro:
**Describe the bug**

With v1.0.2, we cannot upload firmware by GUI, because HTML input tag type file doesn't work.
I think this is because joshjs's potential bug reveals in improvements up to v1.0.2.

The error message is following.

![image](https://user-images.githubusercontent.com/4096956/180681149-860eac0d-cdf8-440b-bbbf-563b0627295a.png)

The error prevents to execute following jQuery.
https://github.com/nerves-hub/nerves_hub_web/blob/87682ed36269309e7a779badb0999e8b6e0568a7/apps/nerves_hub_www/assets/js/app.js#L21-L32

**To Reproduce**

0. `mix assets.setup && make iex-server`
1. Go to `http://localhost:4000/org/NervesTeam/SmartRentHub/firmware/upload`
2. Click on 'Click to upload file'
3. Then select some firmware you have. This ends with no action.

**Expected behavior**

can upload firmware

**Desktop (please complete the following information):**
 - OS: [Linux]
 - Browser [chrome, firefox]
 - Version [nerves_hub_web: v1.0.2]

**Additional context**

I made a [PR](https://github.com/mamunhpath/josh.js/pull/10) to fix this joshjs's issue.

I also have an idea of what to do if the PR is not accepted. That is to stop using joshjs, currently joshjs is used only one place,

https://github.com/nerves-hub/nerves_hub_web/blob/87682ed36269309e7a779badb0999e8b6e0568a7/apps/nerves_hub_www/lib/nerves_hub_www_web/templates/home/index.html.heex#L17

so I think we can choose to stop using joshjs. I already tried to uninstall joshjs on my self-hosted nerves-hub and confirmed uploadable and deployable.

tonnenpinguin:
I noticed the josh exceptions a while ago already and ended up dropping josh from our fork as well because I didn't see the benefit of it being there in the first place :)

By the way: Another fix is to just wrap the `new Josh()` call in a try catch block. That would also allow the jquery to be executed, although it might be worthwhile to figure out a way around using jquery in the first place 🙃

pojiro:
> I already tried to uninstall joshjs on my self-hosted nerves-hub and confirmed uploadable and deployable.

#844 is the PR includes commit i tried.

fhunleth:
Thanks for looking into this and fixing!

PR #844, opened 2022-07-25
propose to stop using joshjs for issue

pojiro:
propose to stop using joshjs for issue #843

WHY: joshjs's bug prevents firmware upload by GUI

With v1.0.2, we cannot upload firmware by GUI.
This is because joshjs bug has appeared in the improvements made so far.
Fortunately, joshjs is now applied to only one location, so propose to stop using it.


jjcarstens (approved):

jjcarstens:
Works for me. I have no attachment to JoshJS

PR #849, opened 2022-07-30
proposal: make certs just in time we need

### LINKED ISSUE #753: make reset-db fails due to duplicated use of device certificates
**Describe the bug**
Resetting the Test DB currently fails as we try to seed multiple products (which in turn seed devices) with the same device certificate.

This is the console output
```
** (MatchError) no match of right hand side value: {:error, #Ecto.Changeset<action: :insert, changes: %{aki: <<43, 96, 184, 137, 70, 11, 241, 25, 18, 250, 138, 122, 110, 37, 254, 201, 20, 64, 142, 90>>, der: <<48, 130, 2, 31, 48, 130, 1, 197, 160, 3, 2, 1, 2, 2, 20, 73, 242, 12, 70, 160, 183, 5, 39, 165, 5, 12, 167, 211, 136, 226, 3, 250, 63, 35, 255, 48, 10, 6, 8, 42, 134, 72, 206, 61, ...>>, fingerprint: "4B600694CAFBB4B1E310F4BBC0CE673F94C3D0A2", not_after: ~U[2023-09-16 15:33:00Z], not_before: ~U[2018-09-17 15:33:00Z], org_id: 2, public_key_fingerprint: "09A16B27D38E8CD256A084A0479337F22D0D69EB", serial: "422154175989008335861639249273408275957749326847", ski: <<144, 44, 253, 69, 48, 161, 225, 92, 154, 123, 70, 207, 62, 229, 85, 58, 134, 210, 98, 15>>}, errors: [public_key_fingerprint: {"public key already associated with another device", []}], data: #NervesHubWebCore.Devices.DeviceCertificate<>, valid?: false>}
    (nerves_hub_api 0.1.0) /Users/bernhardhackl/src/nerves_hub_web/test/support/fixtures.ex:291: NervesHubWebCore.Fixtures.device_certificate_fixture/2
    (elixir 1.11.4) lib/enum.ex:1411: Enum."-map/2-lists^map/1-0-"/2
    (elixir 1.11.4) lib/enum.ex:1411: Enum."-map/2-lists^map/1-0-"/2
    apps/nerves_hub_web_core/priv/repo/seeds.exs:61: NervesHubWebCore.SeedHelpers.nerves_team_seed/1
    (elixir 1.11.4) lib/code.ex:931: Code.require_file/2
    (mix 1.11.4) lib/mix/tasks/run.ex:146: Mix.Tasks.Run.run/5
    (mix 1.11.4) lib/mix/tasks/run.ex:86: Mix.Tasks.Run.run/1
    (mix 1.11.4) lib/mix/task.ex:394: Mix.Task.run_task/3
    (mix 1.11.4) lib/mix/cli.ex:84: Mix.CLI.run_task/2
    (elixir 1.11.4) lib/code.ex:931: Code.require_file/2
make[1]: *** [rebuild-db] Error 1
make: *** [reset-db] Error 2
```

This also contains a good pointer to where the issue is: `NervesHubWebCore.SeedHelpers.nerves_team_seed/1` tries to seed multiple devices, but doesn't use a unique device certificate for each of them.

**To Reproduce**
1. Reset your test database
2. Run `docker compose up`
3. Run `make reset-db`

pojiro:
THIS PR IS NOT FOR FIX, FOR PROPOSAL, refs #850

This PR can fix #753 by creating device certificate just in time.
Currently nerves_hub_web depends real certs fixtures, https://github.com/nerves-hub/nerves_hub_web/tree/main/test/fixtures/ssl.

I'd like to propose that

1. test should not depends on real certs, cause we have to maintenance them for expiration.
    but the current implementation may be intended to reduce testing time, If so, it needs to be considered.
2. create like `nerves_hub_certificate` name library
    using nerves_hub_cli for this PR is not appropriate. why i use it is i do not want write same code as nerves_hub_cli. the features related to certificates are needed by both client and  server. so i think we may need the shared lib, like `nerves_hub_certificate`.


joshk:
@fhunleth @jjcarstens any thoughts or feedback on this PR?

jjcarstens:
I like this idea. I'm not sure yet where the shared certificate definition is best lived, but it's currently on our minds to get in. Since this is fairly far behind, going to close for now while we figure things out

PR #863, opened 2022-10-31
Fix Accounts.get_user_by_email_or_username/1

zolakeith:
Previously the function would generate a query with the OR grouped incorrectly. This commit changes the position of the `exclude_deleted/0`
call so that the query looks for `(username: X or email: X) and not deleted`
instead of `(username: X and not deleted) or email: X)`


zolakeith:
Stumbled on this by accident, results in a `Ecto.MultipleResultsError at POST /login` when attempting to login with a previously deleted email address.

jjcarstens (approved):

jjcarstens:
Thanks @zolakeith !

Issue #795, opened 2022-01-14
Add support for a data retention policy on the logs


fhunleth:
Currently clearing old data from the audit logs requires manual intervention. It would be nice to have a UI option to set the number of days logs should be retained and have a cron job clear out old entries. A per-org setting would suffice.

joshk:
I was hoping this could be a Product setting, but looking into the Audit Log schema and how Audit Logs are linked to Orgs, it might make more sense to have this be an Org setting, at least for now.

PR #872, opened 2022-12-09
Add worker for periodically cleaning up audit logs

LostKobrakai:
At sparkmeter we've had some very noicy devices, which generated so many logs, that it needed cleanup before the system became usable again. This adds a worker (which I hope I added correctly) to periodically cleanup the audit log, so things stay working. By default the worker is disabled, so we don't delete data for people expecting nothing to get deleted.

### DIFF
--- a/apps/nerves_hub_api/config/release.exs
+++ b/apps/nerves_hub_api/config/release.exs
@@ -79,3 +79,13 @@ config :nerves_hub_web_core, NervesHubWebCore.CertificateAuthority,
     certfile: "/etc/ssl/#{host}.pem",
     cacertfile: "/etc/ssl/ca.pem"
   ]
+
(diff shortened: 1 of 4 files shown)


mdwaud:
Relates to https://github.com/nerves-hub/nerves_hub_web/issues/795

jjcarstens (commented):
I think I'm curious the motivation to do count based here vs time based? I feel like the more typical use-case is time based (30 days, 4 month, etc) instead a hard number limit but maybe I'm missing something?

fhunleth:
@jjcarstens Both are needed. The one in this PR is needed to protect against device spamming. We also need a time-based one to support compliance with data retention policy.

fhunleth:
I think we should merge this one and then add a time-based check in a subsequent PR.

jjcarstens:
Gotcha. That makes sense and I agree.

jjcarstens (approved):

PR #875, opened 2022-12-21
Revert "Support TLS 1.3"

fhunleth:
This reverts commit 7bc942093dbee8666fe4f34d3f07e0b59110c2ed.

Devices that aren't running Erlang 25.2 won't be able to connect with
this change.


jjcarstens:
This has since been resolved

PR #880, opened 2023-01-20
Add API endpoint for fetching JITP

zolakeith:
Hello NervesHub team!

This commit adds an additional endpoint to retrieve JITP for a given CACertificate. The certificate is identified by base64 encoded SKI resulting in an url like:

`GET orgs/test/jitp/NfGQXczGpi24p2/zGmJiLI4ahs4=`

which returns:

```
{:ok,
 %{
   "data" => %{
     "description" => "jitp description",
     "product" => "test",
     "tags" => ["test"]
   }
 }}
```

I originally tried to fit this new endpoint underneath the CACertificates endpoint, but it clashes with the existing "show" for that resource.

The CACertificateController.create was also updated to support passing in JITP as base64 encoded json. The "product" is expected to passed as the name of an existing product. This update was added to support testing the JITP feature in the nerves_hub_user_api library.

The motivation for this feature is that I have a service which our device connects to, and in order to authenticate the certificate via devices endpoint, I need to know it's product name. Using NervesKey we are only able to set "/O=org/CN=<serial_number>" so when the device connects to us, we have the certificate, the serial number, and organisation.

I've tried to follow conventions as much as possible, but feel free to request changes if required.

### DIFF
--- a/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/ca_certificate_controller.ex
+++ b/apps/nerves_hub_api/lib/nerves_hub_api_web/controllers/ca_certificate_controller.ex
@@ -1,7 +1,7 @@
 defmodule NervesHubAPIWeb.CACertificateController do
   use NervesHubAPIWeb, :controller
 
-  alias NervesHubWebCore.{Devices, Certificate}
+  alias NervesHubWebCore.{Devices, Products, Certificate}
 
   action_fallback(NervesHubAPIWeb.FallbackController)
 
@@ -27,16 +27,19 @@ defmodule NervesHubAPIWeb.CACertificateController do
          aki <- Certificate.get_aki(cert),
          ski <- Certificate.get_ski(cert),
          {not_before, not_after} <- Certificate.get_validity(cert),
-         params <- %{
-           serial: serial,
-           aki: aki,
-           ski: ski,
-           not_before: not_before,
-           not_after: not_after,
-           der: X509.Certificate.to_der(cert),
-           description: Map.get(params, "description")
-         },
-         {:ok, ca_certificate} <- Devices.create_ca_certificate(org, params) do
+         params <-
+           %{
+             serial: serial,
+             aki: aki,
+             ski: ski,
+             not_before: not_before,
+             not_after: not_after,
+             der: X509.Certificate.to_der(cert),
+             description: Map.get(params, "description")
+           }
+           |> maybe_merge_jitp(org, params),
+         {:ok, ca_certificate} <-
+           Devices.create_ca_certificate(org, params) do
       conn
       |> put_status(:created)
       |> put_resp_header(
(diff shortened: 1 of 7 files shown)


jjcarstens (commented):
Sorry @zolakeith that its taken so long to get to this!

I see that you tried to include this with the CACertificate controller, but mentioned it clashed. Can you explain a bit more on that? I would think that we could just update the `NervesHubAPIWeb.CACertificateView` to include the JITP association (if any) there instead. This might actually be more preferred

zolakeith:
Thanks @jjcarstens - off the top of my head I don't remember the complications - however I'll be happy to investigate whether I can refactor this to use `NervesHubAPIWeb.CACertificateView` as that sounds like a more sensible place for this to live. I'll dig back into my code when I get a chance and either update this pr or report back.

zolakeith:
Sorry again for the delay here - I'm hoping I'll get a chance to investigate this week.

zolakeith:
Morning @jjcarstens I've finally had the time to get back to this today and I now am able to explain why I didn't implement it using `NervesHubAPIWeb.CACertificateView`

Just to back up a little:

1. the motivation is to access the JITP data for a given a device certificate
2. the service only has two pieces of information: the `DeviceCertificate` and the `org_name`

Inside the `DeviceCertificate` the `AKI` is the same as the `SKI` for the `CACertificate`.

That means we only have the `SKI` for the `CACertificate`, not the `":serial"` which is what the `"/show"` expects:

```
scope "/ca_certificates" do
  ...
  get("/:serial", CACertificateController, :show)
end
``` 

Hence my dilemma. 

My options were:

1. Add another route to `CACertificateController`:

```
scope "/ca_certificates" do
  ...
  get("/ski/:ski", CACertificateController, :show)
  get("/:serial", CACertificateController, :show)
end
``` 

2. Add another controller, eg, JITPController


Obviously at the time I opted for number 2. However if you have another suggestion how to solve this, I'll be happy to consider.

zolakeith:
Ok I've played around a little with this, and using the route:

```
scope "/ca_certificates" do
  ...
  get("/ski/:ski", CACertificateController, :show)
  get("/:serial", CACertificateController, :show)
end
```

means the controller will just need another `show/2`, something like this:

```
defmodule NervesHubAPIWeb.CACertificateController do
  ...
  def show(%{assigns: %{org: _org}} = conn, %{"ski" => ski} = params) do
    with {:ok, ca_certificate} <- Devices.get_ca_certificate_by_ski(ski) do
      render(conn, "show.json", ca_certificate: ca_certificate)
    end
  end
```

which then means I can add the JITP to the NervesHubAPIWeb.CACertificateView as you requested.

So I think I can make this work.

fhunleth:
@zolakeith Just wanted to give you a heads up that we have some vacations happening so there might be delays in responding. Your work is definitely appreciated!!!

zolakeith:
Thanks @fhunleth! I hope I can extract a PR in the next week or so.

jjcarstens:
Closing this in favor of #1167 which adds `jitp` to the CACertificate response (this assumes if you're able to get the SKI, then you should be able to get the serial for the lookup)

Its also worth noting that since this PR was created, the devices endpoint now returns `product_name` and so it may be easier to query `/devices/:identifier` to get the product information you need

PR #881, opened 2023-01-28
Fix typo and port configuration

francois2metz:
Without casting the string to int, the server crash on startup.

Also a small typo fixed.


fhunleth:
Thanks!

PR #890, opened 2023-03-24
Remove OrgLimits

oestrich:
Removes org limits since we couldn't edit them outside of the console and they were bumped to the point of being ignored a few years ago. So let's just ignore them completely!


jjcarstens (commented):
This removes OrgLimits and firmware TTL. Was that intentional to combine into one?

oestrich:
Intentional, firmware TTL pulls straight from org_limits if you don't provide a value. Which this case statement made it seem like it was required and could go together https://github.com/nerves-hub/nerves_hub_web/pull/890/files#diff-612c8e1fb5b3a2b86b0a3796338c4efbd7de590b7b9ab38593eacdf069595b11L488

jjcarstens (approved):

PR #891, opened 2023-03-27
Log what we rescue before redirecting

oestrich:
Some of our LiveViews completely swallow errors and make it nearly impossible to debug and figure out what went wrong. This formats the exception as an error to have it in logs for future us.


jjcarstens (approved):
LGTM. We could also put this in`nerves_hub_web.ex` macro where it is defined so that the caller doesn't have to remember that they should _also_ make a logger call, but I'm good either way

jjcarstens:
Ahhhh..Yep. Didn't think about that

PR #897, opened 2023-03-28
Don't assume that the user has an organization matching their name

oestrich:
This is no longer a requirement and is just the default one that could change if the user edits it.

I don't like loading the user's read orgs but this is what happens elsewhere currently to load the dropdown. That will be another PR to make the user's active orgs loaded via a plug so we don't have to load it multiple times (or in the view)


jjcarstens (approved):

lib/nerves_hub_web/controllers/home_controller.ex
```diff
@@ -1,12 +1,14 @@
...
 
+  alias NervesHub.Accounts
+
   def index(conn, _params) do
-    case conn.assigns do
-      %{user: user} ->
-        redirect(conn, to: Routes.product_path(conn, :index, user.username))
+    case Map.has_key?(conn.assigns, :user) && !is_nil(conn.assigns.user) do
+      true ->
+        render(conn, "index.html")
```
tonnenpinguin, line 9: Wouldn't it be enough to just write something like
```elixir
case Map.get(conn.assigns, :user) do
   nil ->  redirect(conn, to: Routes.session_path(conn, :new))
   _user -> render(conn, "index.html")
end

PR #900, opened 2023-03-29
Clean out the rel scripts folder of unused files

oestrich:
The current deployment to ECS does not use this


jjcarstens (approved):
evil-laugh

PR #903, opened 2023-03-30
Remove uninvited signups

oestrich:
Moving towards single company self-hosted version of NervesHub means that uninvited signups don't make sense. They should be invite only.


fhunleth:
How does the first user get an account?

oestrich:
Seeds or the console (likely the seeds)

fhunleth (approved):

PR #904, opened 2023-03-30
Plugs and views should load orgs/products less

oestrich:
- Top level plug that fetches everything the layout needs when loading the user
- Update plugs below the `FetchUser` plug to use what was already loaded
- Update views to use the already loaded data instead of fetching multiple times and count on a new query

### DIFF
--- a/lib/nerves_hub_web/live/device_live/index.ex
+++ b/lib/nerves_hub_web/live/device_live/index.ex
@@ -44,9 +44,20 @@ defmodule NervesHubWeb.DeviceLive.Index do
       socket.endpoint.subscribe("product:#{product_id}:devices")
     end
 
+    user = Accounts.get_user!(user_id)
+
     socket =
       socket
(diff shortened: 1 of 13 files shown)


lib/nerves_hub_web/live/device_live/index.ex
```diff
@@ -44,9 +44,20 @@ defmodule NervesHubWeb.DeviceLive.Index do
...
     end
 
+    user = Accounts.get_user!(user_id)
+
     socket =
       socket
-      |> assign_new(:user, fn -> Accounts.get_user!(user_id) end)
+      |> assign(:user, user)
+      |> assign_new(:orgs, fn ->
+        import Ecto.Query
```
jjcarstens, line 53: What is the load on potentially importing Ecto.Query on several mounts? Should we just compile this into the module rather than during requests?

oestrich: Oh I dunno, moving to the module is fine with me.

jjcarstens (approved):

PR #906, opened 2023-03-31
Select only firmware version numbers for the device filters

oestrich:
If you have multiple version numbers for differing platforms (or just different UUIDs) they show up multiple times in this list.
Screenshot 2023-03-31 at 9 36 11 AM

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -21,11 +21,9 @@ defmodule NervesHub.Firmwares do
 
   @spec get_firmwares_by_product(integer()) :: [Firmware.t()]
   def get_firmwares_by_product(product_id) do
-    from(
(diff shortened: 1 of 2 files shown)


jjcarstens (commented):
LGTM. Though we could still have duplicates if the product has multi x.x.x firmwares but for different platforms. Maybe we add the platform or architecture next to the version name in the dropdown?

oestrich:
This won't show multiple versions for different platforms. They get distinct'd away. 

This is what it looks like without these changes. 

Screenshot 2023-03-31 at 11 33 35 AM

Given the field and platform separate I think this is what we want. Or we turn it to filtering for firmware by UUID and do as you suggest

jjcarstens (approved):
O, right. Separate by platform and then see versions. 🤦

PR #907, opened 2023-04-03
Validate the version condition on a deployment

oestrich:
If we don't have a space between the version condition then the deployment finding in the database breaks and render a 500 on the device page. This validates we have a proper string before you save it.

### DIFF
--- a/lib/nerves_hub/deployments/deployment.ex
+++ b/lib/nerves_hub/deployments/deployment.ex
@@ -11,6 +11,9 @@ defmodule NervesHub.Deployments.Deployment do
 
   alias __MODULE__
 
+  # Similar to Elixir's `Version` but simplified for the match in postgres
+  @version_regex ~r/(=|<|<=|>|>=|~>) \d+\.\d+(\.\d+)?/
+
   @type t :: %__MODULE__{}
   @required_fields [:org_id, :firmware_id, :name, :conditions, :is_active, :product_id]
   @optional_fields [
@@ -52,6 +55,7 @@ defmodule NervesHub.Deployments.Deployment do
     |> cast(with_product_id, @required_fields ++ @optional_fields)
     |> validate_required(@required_fields)
     |> unique_constraint(:name, name: :deployments_product_id_name_index)
+    |> validate_conditions()
     |> validate_change(:is_active, fn :is_active, is_active ->
       creation_errors(:is_active, is_active)
     end)
@@ -138,8 +142,8 @@ defmodule NervesHub.Deployments.Deployment do
         |> validate_required([:tags])
         |> validate_length(:tags, min: 1)
         |> validate_change(:version, fn :version, version ->
-          if not is_nil(version) and Version.parse_requirement(version) == :error do
-            [version: "Must be valid Elixir version requirement string"]
+          if not is_nil(version) and !String.match?(version, @version_regex) do
+            [version: "Must match the regex #{Regex.source(@version_regex)}"]
           else
             []
           end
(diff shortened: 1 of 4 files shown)


lib/nerves_hub/deployments/deployment.ex
```diff
@@ -11,6 +11,9 @@ defmodule NervesHub.Deployments.Deployment do
 
   alias __MODULE__
 
+  # Similar to Elixir's `Version` but simplified for the match in postgres
+  @version_regex ~r/(=|<|<=|>|>=|~>) \d+\.\d+(\.\d+)?/
```
jjcarstens, line 15: So this is now requiring there to always be a conditional? (I.e. I can't just put `1.0.0`, I have to use `= 1.0.0`)

Also, this doesn't support the `and/or` clauses (`< 1.0.0 and > 2.0.0`, etc), but I'm guessing that's okay since `~>` is in there? Maybe the postgres way doesn't even support it as is?

test/nerves_hub/deployments/deployments_test.exs
```diff
@@ -90,6 +90,27 @@ defmodule NervesHub.DeploymentsTest do
...
+        firmware_id: firmware.id,
+        name: "a different name",
+        conditions: %{
+          "version" => "<1.0.0",
+          "tags" => ["beta", "beta-edge"]
+        },
+        is_active: false
+      }
+
+      assert {:error, %Changeset{}} = Deployments.create_deployment(params)
```
jjcarstens, line 112:
```suggestion
      assert {:error, %Changeset{errors: errs}} = Deployments.create_deployment(params)

Assert Keyword.has_key?(:version)
```
Should we assert the version error is indeed in there and we didn't get lucky with some other dumb error?

tonnenpinguin:
This would be a breaking change for existing installations, right? Should we add a migration script that automatically updates strings like `1.0.0` to `= 1.0.0`?

Or maybe a more general question: Can't we just add another case to the `semver_match` function?
```sql
    CREATE OR REPLACE FUNCTION semver_match(version text, req text) RETURNS boolean
    LANGUAGE SQL
    IMMUTABLE
    RETURNS NULL ON NULL INPUT
    AS $$
    SELECT CASE
    WHEN version LIKE '%-%' THEN 'f'
    WHEN req LIKE '~>%' THEN
        string_to_array(version, '.')::int[] >= string_to_array(substring(req from 4), '.')::int[]
        AND
        string_to_array(version, '.')::int[] <
        -- increment last item by one. (X.Y.Z => X.Y.(Z+1))
        array_append(
            (string_to_array(substring(req from 4), '.')::int[])[1:(array_length(string_to_array(req, '.'), 1) - 1)], -- X.Y
            (string_to_array(substring(req from 4), '.')::int[])[array_length(string_to_array(req, '.'), 1)] + 1 -- Z + 1
        )
    WHEN req LIKE '>=%' THEN string_to_array(version, '.')::int[] >= string_to_array(substring(req from 4), '.')::int[]
    WHEN req LIKE '<=%' THEN string_to_array(version, '.')::int[] <= string_to_array(substring(req from 4), '.')::int[]
    WHEN req LIKE '>%' THEN string_to_array(version, '.')::int[] > string_to_array(substring(req from 3), '.')::int[]
    WHEN req LIKE '<%' THEN string_to_array(version, '.')::int[] < string_to_array(substring(req from 3), '.')::int[]
    WHEN req LIKE '=%' THEN
        (string_to_array(version, '.')::int[])[1:array_length(string_to_array(substring(req from 3), '.'), 1)] =
        string_to_array(substring(req from 3), '.')::int[]
-- add something like this
    WHEN req LIKE '%.%.%' THEN
        (string_to_array(version, '.')::int[])[1:array_length(string_to_array(substring(req from 3), '.'), 1)] =
        string_to_array(substring(req from 3), '.')::int[]
--
    ELSE NULL
    END $$;
```

PR #909, opened 2023-04-08
Allow web registration and various fixes

axelson:
Before this only registration via mix task or an invitation was allowed.

Adds a very basic registration route/page:
image

Also adds a registration link to the login page:
image

Also fixes some other issues
- `make reset-db` was broken because of a typo
- Adds getting started commands that don't rely on make because many elixir developers may not have make installed
  - Would it make sense to remove the reliance of `make` entirely? I don't think I've seen any other Phoenix projects that require `make` for common tasks
- Points the user to `/register` after they initial get the app running

Adds the `namespace: NervesHub` to the configuration, without this `mix phx.routes` doesn't work by default

TODO:
- [ ] Add tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ iex-server:
 	iex -S mix phx.server
 
 reset-db:
-	mix exto.reset
+	mix ecto.reset
 
 reset-test-db:
 	MIX_ENV=test \
--- a/README.md
+++ b/README.md
@@ -79,17 +79,18 @@ is needed in order to run both applications. When you create certificates from `
      ```
 
 2. Fetch dependencies: `mix do deps.get, compile`
-3. Initialize the database: `make reset-db`
+3. Initialize the database: `make reset-db` (or mix `ecto.reset`)
 4. Compile web assets (this only needs to be done once and requires python2):
    `mix assets.install`
 
 ### Starting the application
 
-* `make server` - start the server process
-* `make iex-server` - start the server with the
-   interactive shell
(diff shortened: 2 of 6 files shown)


oestrich:
Hey, thanks for the pr, but I just removed registration because of where nerveshub is heading. It’s going to be aimed at self hosted by a company instead of a single hosted instance. Given this, you probably don’t want anyone to register on your company’s instance of nerves hub, haha.

If you can split out the readme and make file changes into a new pr then we can get those merged. Thanks again!

axelson:
Yeah I can split out the other fixes from registration. Also, what do you think about adding a setting to control whether registration is allowed? That would allow using NervesHub in other use-cases.

oestrich:
It used to be hidden behind an environment variable haha. I removed it as part of stripping nerves hub down to its minimal amount. Orgs are likely to go as well, which makes registration weird.

A lot of the future of nerves hub is just in conversations between @fhunleth, @jjcarstens, and me. The main driving force for the 2.0 branch is making it better for a single company (or individual/team) to run nerves hub and have it work for a large fleet of devices. Stripping out unnecessary parts makes it easier to get to this, so out they go! 😄

axelson:
Yeah that's fair and it's good to have a specific target in mind. My main goal here is to make it more seamless for a hobby dev to get NervesHub up and running. To that end, how about adding some instructions for the initial login: https://github.com/nerves-hub/nerves_hub_web/pull/911

PR #910, opened 2023-04-09
Readme improvements and various fixes

axelson:
Fixes:
- `make reset-db` was broken because of a typo
- Adds getting started commands that don't rely on make because many elixir developers may not have make installed
- Adds the `namespace: NervesHub` to the configuration
  - Without this `mix phx.routes` doesn't work by default

Split out from https://github.com/nerves-hub/nerves_hub_web/pull/909


oestrich:
Docker is being weird for other reasons, so I'm going to merge this without it passing.

PR #911, opened 2023-04-09
Add initial login instructions

axelson:
Recommend logging in (while in development mode) via the `nerveshub` root user created in the seed script.


oestrich:
Docker is being weird for external reasons, so I'm going to merge this with that failing. Thanks for the updates!

PR #912, opened 2023-04-10
Devices have a deployment

oestrich:
When a device connects, find and attach a deployment to the device. From then on, the device will only accept updates from that deployment. Tag changes on either end will force a recalculation of matches between the two.

I think this needs some more descriptions on the deployment edit page to let you know how big of a thing it will be to edit the deployment's conditions. There might also be a few cases I'm missing on when a device should move itself out of a deployment. ~I'm not sure I have version checking in place for instance.~ Done now

This uses a good amount of Phoenix channels to bounce around and have devices update themselves when a deployment changes.


lib/nerves_hub/deployments.ex
```diff
@@ -82,19 +83,64 @@ defmodule NervesHub.Deployments do
...
+            |> Repo.update_all(set: [deployment_id: nil])
+
+            broadcast(deployment, "deployments/is-active-changed")
+
+            description = "deployment #{deployment.name} is inactive and removed all devices"
+            AuditLogs.audit!(deployment, deployment, :update, description)
+          end
+        end
+
         fetch_and_update_relevant_devices(deployment)
```
jjcarstens, line 138:
```suggestion
        broadcast(deployment, "deployments/update")
```

Seems like we could just replace these function calls with the broadcast since that is all that is happening now (no fetching and things)

lib/nerves_hub/deployments.ex
```diff
@@ -82,19 +83,64 @@ defmodule NervesHub.Deployments do
...
+          else
+            Device
+            |> where([d], d.deployment_id == ^deployment.id)
+            |> Repo.update_all(set: [deployment_id: nil])
+
+            broadcast(deployment, "deployments/is-active-changed")
+
+            description = "deployment #{deployment.name} is inactive and removed all devices"
+            AuditLogs.audit!(deployment, deployment, :update, description)
+          end
```
jjcarstens, line 135: My only beef with all this is if someone updates the `:conditions` and `:is_active` in the same payload, it will trigger multiple reloads of each device in the channel.

You can't technically do that with the web right now, but it could potentially in the future. You can, however, [update both of these fields with the API](https://github.com/nerves-hub/nerves_hub_web/blob/2.0/lib/nerves_hub_web/controllers/api/deployment_controller.ex#L15)

It seems like it's less crucial for the device to know _what_ changed, but rather that _something_ changed. So maybe it would be better to use a generic topic with `deployments/changed` instead? (maybe even include the changes in the payload if the device wants to do something with it)

lib/nerves_hub_web/live/device_live/show.ex
```diff
@@ -45,9 +45,12 @@ defmodule NervesHubWeb.DeviceLive.Show do
       socket.endpoint.subscribe("device:#{socket.assigns.device.id}:internal")
     end
 
+    device = Repo.preload(socket.assigns.device, [:deployment])
```
jjcarstens, line 48: Maybe we should add this to the `Devices.get_device_by_product/3` instead of another lookup?

lib/nerves_hub_web/templates/device/show.html.heex
```diff
@@ -71,7 +71,16 @@
...
+        <%= if @deployment do %>
+          <%= link(to: Routes.deployment_path(@socket, :show, @org.name, @product.name, @deployment.name), class: "badge") do %>
+            <div class={"deployment-state state-#{if @deployment.is_active, do: "on", else: "off"}"}><%= @deployment.name %></div>
+          <% end %>
+        <% end %>
+      </div>
+      <div class="mt-4">
+        <div class="help-text mb-1">Potential Deployments</div>
         <%= if !Enum.empty?(@deployments) do %>
           <%= for deployment <- @deployments do %>
```
jjcarstens, line 85:
```suggestion
          <%= for deployment <- @deployments, deployment.id != get_in(@deployment, [Access.key(:id)]) do %>
```
We should probably filter out the assigned deployment from this list of potentials to be more pleasing on the 👀

lib/nerves_hub/deployments.ex
jjcarstens: One extra thought - Are we ever going to have a race here between publishing the messages above which alter the deployment linked in the device socket and this broadcast telling the socket to prep and send an update message?

test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -94,11 +100,10 @@ defmodule NervesHubWeb.DeviceChannelTest do
 
     {:ok, _deployment} =
       NervesHub.Deployments.update_deployment(deployment, %{
-        firmware_id: new_firmware.id,
-        is_active: true
+        firmware_id: new_firmware.id
       })
 
-    assert_push("update", %{firmware_meta: %{version: "0.0.2"}})
+    assert_push("update", %{})
```
jjcarstens, line 106: Isn't this bad? Shouldn't there be some update information in this payload that is being sent to the device? I feel like we should still be checking that part of the payload is what we expect since this is crucial for getting updates to devices (or maybe the payload is checked in another test I haven't seen yet?)

test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -427,6 +417,40 @@ defmodule NervesHubWeb.WebsocketTest do
...
+        "nerves_fw_uuid" => Ecto.UUID.generate(),
+        "nerves_fw_product" => "test",
+        "nerves_fw_architecture" => "arm",
+        "nerves_fw_platform" => "tester",
+        "nerves_fw_version" => "0.1.0"
+      })
+      SocketClient.wait_join(socket)
+
+      device = Repo.reload(device)
+      refute device.deployment_id
```
jjcarstens, line 452: I think I need some notes to understand what's going on here. The title says it matches deployment on connect, but this is refuting the same deployment_id that was asserted on previously. Is there lots of context in the fixtures I'm missing that explain what is happening here?

test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -137,6 +142,63 @@ defmodule NervesHubWeb.DeviceChannelTest do
...
+    {:ok, _deployment} =
+      NervesHub.Deployments.update_deployment(deployment, %{
+        conditions: %{"tags" => ["beta"]}
+      })
+
+    socket_alpha = :sys.get_state(socket_alpha.channel_pid)
+    assert is_nil(socket_alpha.assigns.device.deployment_id)
+
+    socket_beta = :sys.get_state(socket_beta.channel_pid)
+    refute is_nil(socket_beta.assigns.device.deployment_id)
```
jjcarstens, line 199: I missing where this is checking that a relookup happened. Should we first assert the beta device deployment_id is nil, update deployment, and refute it nil again?

jjcarstens (commented):
K, I think I got through it all but this is big and might have missed something. Lets start with all this and see where it goes

lib/nerves_hub/deployments.ex
```diff
@@ -251,14 +180,61 @@ defmodule NervesHub.Deployments do
...
     |> where([d], fragment("coalesce(semver_match(?::text, ?->>'version'), 't')", ^device.firmware_metadata.version, d.conditions))
     |> Repo.all()
   end
+
+  @doc """
+  Sync devices that are missing a deployment
+
+  Checks potential deployments for a device and sets to the first or only deployment found.
+  """
+  def sync_missing_devices() do
```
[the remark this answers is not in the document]
oestrich: I used this in the beginning to attach deployments to devices, I am fine getting rid of it now since it should be automatic when the device comes online.

lib/nerves_hub/deployments.ex
oestrich: `fetch_and_update_relevant_devices/1` is used somewhere, I don't remember exactly but getting rid of it caused a compilation warning so I left it. I can update things though.

lib/nerves_hub/deployments.ex
oestrich: I had original thoughts to check the payload before updating but it seemed like it would be trickier than needed for this first go around. Now that the first pass is in I am good to do that in this PR.

I'm not entirely sure what we'd use in the payload, which might be why I didn't? If we're sending enough stuff in the message, spreading it out to 100k+ messages is pretty hefty. 🤷

lib/nerves_hub_web/templates/device/show.html.heex
oestrich: I was split on this. I think I did remove it at one point but then it looks kinda weird so I left it all. I am fine either way. I think because it's called "potential deployments" it seemed weird to not include it again. Maybe rewording to "Other Potential Deployments" will fix that

test/nerves_hub_web/channels/device_channel_test.exs
oestrich: You are correct, this was caught in a rename of the deployments `update` to `deployments/update`. Will put back 👍

test/nerves_hub_web/channels/websocket_test.exs
oestrich: Ah, this is a really terrible title. It's showing that when the device comes back with a new version the deployment is cleared out because it no longer matches.

lib/nerves_hub/deployments.ex
jjcarstens: Ya, we probably don't need the payload. Just that it changed as a single broadcast message since the effect on the device is the same. The distinction between `conditions` vs `is_active` seems to only matter here in the context of the deployment

lib/nerves_hub_web/templates/device/show.html.heex
jjcarstens: I was on the inverse. In my mind, `assigned deployment` is no longer a "potential" since it is assigned. Maybe `Alternative deployments`? I can honestly live with whatever, but just thought I'd point out

test/nerves_hub_web/channels/device_channel_test.exs
oestrich: This test is showing a device losing the deployment and a device gaining the deployment. i can add more checks to make it clearer (and probably update the title haha)

jjcarstens (approved):
:shipit:

PR #927, opened 2023-04-21
Set code that is sent when the device connects to the console

oestrich:
For devices that need something set when they come online, add a block of code that gets sent grapheme by grapheme over the console topic.

Also removes the LiveView console because it was confusing me and shouldn't be needed anymore.


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -213,9 +213,23 @@ defmodule NervesHubWeb.DeviceChannel do
...
+    metadata = %{console_available: true, console_version: version}
     Presence.update(socket.assigns.device, metadata)
+
+    # now that the console is connected, push down the device's elixir, line by line
+    device = socket.assigns.device
+
+    if device.connecting_code do
+      device.connecting_code
+      |> String.graphemes()
+      |> Enum.map(fn character ->
```
tonnenpinguin, line 226:
```suggestion
      |> Enum.each(fn character ->
```

PR #929, opened 2023-04-26
Push deployment updates through an orchestrator process

oestrich:
Don't push deployment updates directly to the device channel, introduce a process between the two in order to have a hook for rolling updates going forward.

There's no specific tests for this because I'm not entirely sure how to test it. But I've done manual testing with my ~200 devices and they upgraded without issue. This _shouldn't_ be any different than how things currently are.

### DIFF
--- a/lib/nerves_hub/application.ex
+++ b/lib/nerves_hub/application.ex
@@ -3,6 +3,8 @@ defmodule NervesHub.Application do
 
   require Logger
 
+  @env Mix.env()
+
   def start(_type, _args) do
     case System.cmd("fwup", ["--version"]) do
       {_, 0} ->
@@ -15,7 +17,7 @@ defmodule NervesHub.Application do
     children = [
       NervesHub.Metrics,
       NervesHub.Supervisor
-    ] ++ endpoints()
+    ] ++ endpoints(@env)
 
     opts = [strategy: :one_for_one, name: NervesHub.Supervisor]
     Supervisor.start_link(children, opts)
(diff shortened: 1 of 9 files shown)


lib/nerves_hub/application.ex
```diff
@@ -3,6 +3,8 @@ defmodule NervesHub.Application do
 
   require Logger
 
+  @env Mix.env()
```
tonnenpinguin, line 6: Is this available in release mode? I remember having to add some lines to the config in order to read it from the application env

```
config :nerves_hub, env: Mix.env()
```

and then something like
```suggestion
  @env Application.compile_env(:nerves_hub, :env)
```

oestrich: Mix is available at compile time during the release so this works. We also build a release in CI so this would have failed if it was not OK to use.

tonnenpinguin: Ah yeah you're right!
I think we were using it in the function body which ended up failing

lib/nerves_hub/deployments.ex
```diff
@@ -10,6 +10,10 @@ defmodule NervesHub.Deployments do
   alias NervesHub.Repo
   alias Ecto.Changeset
 
+  def all() do
```
tonnenpinguin, line 13: I'm not sure what the consensus is but most projects I've worked on seem to adhere to something like credos [ParenthesesOnZeroArityDefs](https://hexdocs.pm/credo/Credo.Check.Readability.ParenthesesOnZeroArityDefs.html) rule. What do you think?

jjcarstens (approved):

PR #932, opened 2023-05-01
Self hosting setup

oestrich:
This PR is a bit of a work in progress, but I'm able to deploy to a local server and have devices connect. I would like to get local file system working for firmware storage before merging so we don't need s3 for self hosting.


fhunleth (commented):
This is looking great!

lib/nerves_hub_web/endpoint.ex
```diff
@@ -29,7 +29,7 @@ defmodule NervesHubWeb.Endpoint do
   )
 
   file_upload_config =
-    Application.compile_env(:nerves_hub_www, NervesHub.Firmwares.Upload.File, [])
+    Application.get_env(:nerves_hub_www, NervesHub.Firmwares.Upload.File, [])
```
oestrich, line 32: This probably will raise a warning when we upgrade elixir, but that's for later

ivanhercaz:
Hi @oestrich! I am following the development of the version 2.0 very closely, because I am really interested. Now I am deploying an instance to test it and try to link one of my Nerves devices to my NervesHub, so if I have some doubt I will ask in Elixir Forum.

But, in relation with this PR, I want to know what's the main purpose of this PR. I mean, is the purpose of the PR to ease the setup of NervesHub, leave AWS infrastructure as optional and document it? I am really interested in that because I am deploying NervesHub on a Hetzner Cloud server and, for example, with the Dockerfile was a bit hard, so I decide to deploy it with the instructions in the "First time application setup" section.

If this PR is not the correct place to ask this, just let me know and I move this comment to wherever you recommend (e.g., Elixir Forum) and, if it is necessary, I remove the comment.

Either way, I will follow (all) this PR very closely ❤️ . I don't think I can help with development tasks, but if I can help out with the documentation, I'll be delighted.

oestrich:
Hey @ivanhercaz, the main goal of this PR is to have a way to self host for individuals or smaller companies who have <100 devices or so (it'll work for more but you probably have a devops team at that point haha.) I also selfishly want to host this internally for testing a production-like server so this is how I'm getting there.

All of the AWS public setup has been deprecated because it was very complex and a daunting task in order to get NervesHub hosted. Given there's no publicly hosted version of NervesHub we can leave hosting more up to the company that wants to run it or offer this simplified version to get something small going. 

So hopefully this PR simplifies getting NervesHub hosted on a single server/VPS and it's entirely self contained. You won't get niceties like rolling deployments for the server or anything, but I think that's fine given the more hobby scale of a single VPS.

oestrich:
I think this is as close to merging on a first pass as I can get it :smile:

PR #933, opened 2023-05-01
Add an endpoint to try reconnecting to expected nodes

oestrich:
Manual way to force reconnecting nodes as a test.

image

### DIFF
--- a/lib/nerves_hub_web/api_router.ex
+++ b/lib/nerves_hub_web/api_router.ex
@@ -25,6 +25,7 @@ defmodule NervesHubWeb.APIRouter do
     pipe_through(:api)
 
     get("/", HealthCheckController, :health_check)
+    get("/nodes", HealthCheckController, :node_check)
   end
 
   scope "/users", NervesHubWeb.API do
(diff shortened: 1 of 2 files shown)


oestrich:
Good question, I think it's expected you'll be VPN'd off or not have a cluster. So should be fine?

jjcarstens (approved):

lib/nerves_hub_web/api_router.ex
```diff
@@ -25,6 +25,7 @@ defmodule NervesHubWeb.APIRouter do
     pipe_through(:api)
 
     get("/", HealthCheckController, :health_check)
+    get("/nodes", HealthCheckController, :node_check)
```
tonnenpinguin, line 28: Given that this has side effects maybe a `post` endpoint would be more appropriate?

oestrich: Yeah, it should be. This was partially thrown together to figure out an issue in our staging environment and we got around it with #934 so I'm going to close this.

PR #934, opened 2023-05-01
Update ecs-cluster script for new release name

oestrich:
We don't have names per node type now, everything is `nerves_hub`


esvinson (approved):
Yay!

PR #942, opened 2023-05-04
Truncate audit logs hourly

oestrich:
Prune beyond a date and limit the amount of data deleted each run. Run hourly to prune a little bit every hour instead of a huge amount once a day to be night to the vacuum.

### DIFF
--- a/lib/nerves_hub/config.ex
+++ b/lib/nerves_hub/config.ex
@@ -25,4 +25,17 @@ defmodule NervesHub.Config do
            {:url_port, "URL_PORT", default: "443", map: &String.to_integer/1},
            {:url_scheme, "URL_SCHEME", default: "https"}
          ])
+
+  config :audit_logs,
+         env([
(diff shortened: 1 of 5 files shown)


lib/nerves_hub/config.ex
```diff
@@ -25,4 +25,17 @@ defmodule NervesHub.Config do
            {:url_port, "URL_PORT", default: "443", map: &String.to_integer/1},
            {:url_scheme, "URL_SCHEME", default: "https"}
          ])
+
+  config :audit_logs,
+         env([
+           {:enabled, "TRUNATE_AUDIT_LOGS_ENABLED", default: "true", map: &to_boolean/1},
```
jjcarstens, line 31:
```suggestion
           {:enabled, "TRUNATE_AUDIT_LOGS_ENABLED", default: "false", map: &to_boolean/1},
```
I think this was previously defaulted to `false` to allow users to decide if when to remove data instead

jjcarstens (approved):

PR #944, opened 2023-05-05
Optimize device deployment reassignment when previously assigned

oestrich:
Check the incoming payload to see if it still matches the old deployment and set back to that deployment. If not, delay looking up for a spaced out amount of time to not cause a thundering herd of devices checking the database (they still will but it will be a little better)

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -144,46 +144,29 @@ defmodule NervesHubWeb.DeviceChannel do
   # If it matches, we can set the deployment directly and only do 3 queries (update, two preloads)
   def handle_info(%Broadcast{event: "deployments/changed", topic: "deployment:none", payload: payload}, socket) do
     device = socket.assigns.device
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -345,4 +328,32 @@ defmodule NervesHubWeb.DeviceChannel do
...
+      device.firmware_metadata.architecture == payload.architecture &&
+      Enum.all?(payload.conditions["tags"], &Enum.member?(device.tags, &1)) &&
+      Deployments.version_match?(device, payload)
+  end
+
+  defp assign_deployment(socket, device, payload) do
+    device =
+      device
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.put_change(:deployment_id, payload.id)
```
jjcarstens, line 345:
```suggestion
      |> Ecto.Changeset.put_change(:deployment_id, payload.id)
```
Pretty sure you can skip this and go straight to adding a change to the struct?

jjcarstens (approved):

PR #948, opened 2023-05-09
Add a reconnect button to device page

oestrich:
Force a NervesHub reconnect by using a button on the device page. This won't reboot the device, just force a `NervesHubLink.reconnect/0`


jjcarstens (commented):
This will require a change to `NervesHubLink` to support. Couldn't we just close the channel/socket in NH and the devices reconnect code would be forced to kick in without any NHL/firmware changes?

oestrich:
Hmm, that's also a good option. Honestly this was abusing the fact that the socket will crash on the device and force a reconnect that way 😂 So that's probably a better way

jjcarstens (approved):

PR #950, opened 2023-05-09
Add an API for forcing reconnect and sending code to a device

oestrich:
The UI has this capability and for easier scripting purposes, let's expose something similar via the API.

### DIFF
--- a/lib/nerves_hub_web/controllers/api/device_controller.ex
+++ b/lib/nerves_hub_web/controllers/api/device_controller.ex
@@ -1,12 +1,16 @@
 defmodule NervesHubWeb.API.DeviceController do
   use NervesHubWeb, :api_controller
 
-  alias NervesHub.{Devices, Devices.DeviceCertificate}
+  alias NervesHub.AuditLogs
+  alias NervesHub.Devices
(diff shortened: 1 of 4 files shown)


jjcarstens (approved):

lib/nerves_hub_web/controllers/api/device_controller.ex
```diff
@@ -1,12 +1,14 @@
...
-  alias NervesHub.{Devices, Devices.DeviceCertificate}
+  alias NervesHub.Devices
+  alias NervesHub.Devices.DeviceCertificate
+  alias NervesHubWeb.Endpoint
 
   action_fallback(NervesHubWeb.API.FallbackController)
 
   plug(:validate_role, [product: :delete] when action in [:delete])
-  plug(:validate_role, [product: :write] when action in [:create, :update])
+  plug(:validate_role, [product: :write] when action in [:create, :update, :reconnect])
```
rraub, line 11: Shouldn't we include `code` in this validation?

oestrich: Yup, missed it. Added it with the new reboot endpoint as well

PR #954, opened 2023-05-11
When moving devices between products, clear the assigned deployment

oestrich:
And check for a new matching deployment in the product

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -189,7 +189,7 @@ defmodule NervesHubWeb.DeviceChannel do
       socket.assigns.device
       |> Repo.reload()
       |> Deployments.set_deployment()
-      |> Repo.preload(deployment: [:firmware])
+      |> Repo.preload([deployment: [:firmware]], force: true)
 
     socket =
       if device.deployment_id do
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -261,10 +261,20 @@ defmodule NervesHubWeb.DeviceChannel do
...
+      %{deployment_id: device.deployment_id}
+    end)
+
+    Presence.update(device, %{
+      product_id: device.product_id,
+      deployment_id: device.deployment_id
+    })
+
+    # The old deployment is no longer valid, so let's look one up again
+    send(self(), :resolve_changed_deployment)
```
jjcarstens, line 277:
```suggestion
    handle_info(:resolve_changed_deployment, assign(socket, device: device))
```

We could save a message round trip here by just handling it directly. Not crucial though, just a thought

jjcarstens (approved):

lib/nerves_hub_web/channels/device_channel.ex
oestrich: I merged it since I tested what I had. I don't think this will be a performance problem and it feels better to me to keep it separate, but that's purely a feeling lol

PR #956, opened 2023-05-23
Rolling Device Updates

oestrich:
Adds a new field `concurrent_updates` that sets the maximum number of updating devices that NH will tell to update at any given time. There is still a spike for reconnecting devices, this is expected for now. We're going to see how it works with that and come back later.

Devices will be told one at a time on deployment update for each node. They will grab a slot and tell the device to update. When the device comes back online, it will tell its new orchestrator that it updated and the orchestrator will tell another device to update if there's a slot.

There's a 5 minute tick to trigger deployment again as a catch all if devices don't evenly connect to nodes.

### DIFF
--- a/lib/nerves_hub/deployments/orchestrator.ex
+++ b/lib/nerves_hub/deployments/orchestrator.ex
@@ -75,13 +75,15 @@ defmodule NervesHub.Deployments.Orchestrator do
           {:ok, inflight_update} ->
             send(pid, {"deployments/update", inflight_update})
 
+            # Only continue to loop if we were able to update otherwise
+            # this will continue to trigger the same device over and over
+            send(self(), :trigger)
+
           :error ->
             Logger.error(
               "An inflight update could not be created or found for the device #{device.identifier} (#{device.id})"
             )
         end
-
-        send(self(), :trigger)
       end
     end
   end
--- a/lib/nerves_hub/deployments/orchestrator.ex
+++ b/lib/nerves_hub/deployments/orchestrator.ex
@@ -67,6 +67,8 @@ defmodule NervesHub.Deployments.Orchestrator do
 
     # Get a rough count of devices to update
     count = deployment.concurrent_updates - Devices.count_inflight_updates_for(deployment)
+    # Just in case inflight goes higher than concurrent, limit it to 0
+    count = Enum.max([count, 0])
 
     devices
     |> Enum.take(count)
--- a/lib/nerves_hub/deployments/orchestrator.ex
+++ b/lib/nerves_hub/deployments/orchestrator.ex
@@ -28,18 +28,46 @@ defmodule NervesHub.Deployments.Orchestrator do
 
   def name(deployment), do: name(deployment.id)
 
+  def device_updated(deployment_id) do
+    GenServer.cast(name(deployment_id), :trigger)
+  end
+
   @doc """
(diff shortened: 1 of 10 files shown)


lib/nerves_hub/deployments/orchestrator.ex
```diff
@@ -28,27 +28,66 @@ defmodule NervesHub.Deployments.Orchestrator do
...
+    devices =
+      Enum.filter(devices, fn device ->
+        device.firmware_uuid != deployment.firmware.uuid
+      end)
 
-      case Devices.told_to_update(device, deployment) do
-        {:ok, inflight_update} ->
-          send(pid, {"deployments/update", inflight_update})
+    # Get a rough count of devices to update
+    count = deployment.concurrent_updates - Devices.count_inflight_updates_for(deployment)
```
jjcarstens, line 69: We prob should be a little more assertive here since if this were a negative number, it would still cause the `Enum.take` below to return devices and potentially go over the count (assuming there is a slight possibility for the inflight count to be ± a few)

```elixir
iex(condor@10.1.1.34)14> Enum.take [1, 2, 3, 4, 5], -1
[5]
iex(condor@10.1.1.34)15> Enum.take [1, 2, 3, 4, 5], -10
[1, 2, 3, 4, 5]
iex(condor@10.1.1.34)16> Enum.take [1, 2, 3, 4, 5], -3 
[3, 4, 5]
```

jjcarstens (approved):
LGTM - I have one comment that might be worth considering

lib/nerves_hub/deployments/orchestrator.ex
oestrich: It might go over but since it still checks inside the loop it won't trigger an update. So extra work but there's likely extra work anyways due to multiple nodes. I'll add a `Enum.max(count, 0)` anyways since it's easy

fhunleth: `max(count, 0)`?

lib/nerves_hub/deployments/orchestrator.ex
```diff
@@ -60,32 +60,35 @@ defmodule NervesHub.Deployments.Orchestrator do
...
 
-    if Devices.count_inflight_updates_for(deployment) < deployment.concurrent_updates do
-      device =
-        Enum.find(devices, fn device ->
-          device.firmware_uuid != deployment.firmware.uuid
-        end)
+    devices =
+      Enum.filter(devices, fn device ->
+        device.firmware_uuid != deployment.firmware.uuid
+      end)
```
fhunleth, line 66: Not blocking, but is it possible to include this check in the match spec for `Registry.select`. I feel like it is, but my match spec brain is rusty.

lib/nerves_hub/deployments/orchestrator.ex
oestrich: Yup, pushed that up

lib/nerves_hub/deployments/orchestrator.ex
oestrich: Probably but my brain was broken enough from this lol

oestrich: I figured it out!

fhunleth (approved):

PR #957, opened 2023-05-24
Display the current process of a deployment

oestrich:
Progress bar on the deployment page that shows a rough percentage of how far the deployment is.

image

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -716,6 +716,11 @@ defmodule NervesHub.Devices do
 
     if inflight_update != nil do
       Orchestrator.device_updated(inflight_update.deployment_id)
+
+      Deployment
+      |> where([d], d.id == ^inflight_update.deployment_id)
+      |> Repo.update_all(inc: [current_updated_devices: 1])
+
       Repo.delete(inflight_update)
     end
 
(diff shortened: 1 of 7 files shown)


lib/nerves_hub/devices.ex
```diff
@@ -716,7 +716,15 @@ defmodule NervesHub.Devices do
...
     if inflight_update != nil do
       Orchestrator.device_updated(inflight_update.deployment_id)
-      Repo.delete(inflight_update)
+
+      inflight_update
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.prepare_changes(fn changeset ->
+        query = from(Deployment, where: [id: ^inflight_update.deployment_id])
+        changeset.repo.update_all(query, inc: [current_updated_devices: 1])
+        changeset
```
jjcarstens, line 725: Is it necessary to build the whole changeset? Or could you just do the incrementing update then delete the inflight record? It seems like they are not coupled so much that they needed to be piped together?

oestrich: It's not, when I looked up prepare changes for the other part it included an `inc` in the example and I went and used it lol

jjcarstens (approved):

PR #961, opened 2023-05-30
Wrap :gproc calls in try/catch and raise a custom exception

oestrich:
:crossed_fingers:  gracefully shutdown device sockets if there's an issue during a deployment and gproc times out.

### DIFF
--- a/lib/nerves_hub_device/presence.ex
+++ b/lib/nerves_hub_device/presence.ex
@@ -1,3 +1,14 @@
+defmodule NervesHubDevice.PresenceException do
+  defexception [:message]
+
+  @impl true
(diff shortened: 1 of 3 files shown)


oestrich:
The idea is to at least make the failure that's going to happen anyways be a bit more silent. Moving to a process later approach seems like a huge pain with how this channel stands now. I've been thinking on this a bit and I think we can get to the process later as part of the move to split transport layer from NH device internals (re: MQTT.) So I dunno if we just want to sit on this since it only happens during a deploy (and wait for other larger changes) or we merge and deploy it out :shrug: I'm kinda fine with either.

PR #973, opened 2023-06-16
Migrate presence tracking to the DeviceLink process

oestrich:
Have the DeviceLink handle presence tracking as part of the migration of logic from the DeviceChannel


oestrich:
Closing because I think we need to switch up how presence is handled in general.

PR #977, opened 2023-06-20
Reconnect button disconnects the socket

oestrich:
Disconnect the socket so the device channel isn't the only thing terminated. Leaving the console channel open and having the device not try to reconnect the main channel.


jjcarstens (approved):
For posterity .... this is hidden in Phoenix
https://github.com/phoenixframework/phoenix/blob/v1.7.6/lib/phoenix/socket.ex#L518-L520

PR #978, opened 2023-06-20
Track devices with sharded ets tables

oestrich:
Instead of using gproc to have a main node that handles writes, which can be very noisy and troublesome when rolling in and out servers during a deploy, use local ets tables that stay up to date with abcast messages.

This does remove actual status tracking. So if the device publishes `status_update` messages we won't do anything with them. I _think_ this is only used for fwup failures?


lib/nerves_hub/tracker.ex
```diff
@@ -0,0 +1,349 @@
...
+    {:reply, :ok, state, {:continue, {:sync, pid}}}
+  end
+
+  # NOTE this might become an issue as we start having very large amounts of devices online
+  def handle_continue({:sync, pid}, state) do
+    # select everything out at once, because otherwise we might miss something
+    records = :ets.select(state.ets_table, [{:_, [], [{:element, 2, :"$_"}]}])
+
+    Enum.each(records, fn record ->
+      GenServer.cast(pid, {:online, record})
```
keathley, line 208: You may want to have a specific clause that handles a list of records. This would allow you to send batches of records instead of individual messages. The batches will be more efficient since these are going over the network. You could limit the batch size to 100s of records per batch or something. Here after you do the select. Its still a potential problem if you have hundreds of thousands of online devices that you need to cast to another node. But its better than sending them as individual casts.

keathley:
I believe that you have a subtle bug in your tracker. I've only read the code and didn't run it or run tests. Its something you could confirm with https://github.com/elixir-toniq/schism and local cluster. I'm also not sure exactly how likely this scenario is since it may depend on your deployment. But its maybe worth considering.

Based on my reading of the code here's what can happen:

- Given Nodes {A, B, C}
- Node B disconnects from Nodes {A,C}.
- Device connects to Node B.
- Node B adcasts to all device shard process which, in this case, is only the shard on Node B.
- Node B reconnects to Nodes {A,C}.
- User connects to Node A to examine device status.
- A believes that device is offline or does not exist since it has no record of this device.
- The state is now out of sync and as far as my reading of this code goes, it won't converge to the correct state. The status will continue to be out of sync until a node changes or the device status changes and is broadcast to the other nodes.

As a note, I suspect there are other ways to induce this failure. This is what I came up with off the top of my head and should be reproducible.

There are a few solutions to this problem. 

The easiest (and the one that groot uses) is to have each device shard process periodically sync its state to the other nodes. Its easiest to abcast to all available nodes but if the network usage is prohibitive you may want to pick subsets etc. It depends on how much data you have.

Another solution would be to use the `:nodeup` events to trigger a sync to the device that just came online. Because there is no coordination, each node that sees the nodeup event will send records to the node that has just come online, but this also ensures that you eventually converge to the most up to date state.

There are more intelligent (meaning complicated and hard to implement) solutions for this problem but these are the two I would consider as a starting point. FWIW Groot uses both: https://github.com/elixir-toniq/groot/blob/master/lib/groot/storage.ex#L82-L114.

jjcarstens:
Is it typical to have dynamic instances? I think for our case, we have a predefined list of nodes in the cluster. So using `:nodeup` would prob be worth while since we're not adding or removing nodes in the cluster dynamically. I could see that potentially not being as useful for those who are more dynamic (unless that is not a thing?)

oestrich:
I think I've fixed the two issues you pointed out @keathley. There is a full sync on nodeups just as a way to make the problem go away for now and I cast over 100 records at a time for a sync.

Once this is merged I want to beef up our telemetry game because there's not enough insight into most of the system to know if 100 is enough or we should chunk further or something else. But that's a next PR problem!

jjcarstens (approved):

PR #979, opened 2023-06-21
Always allow the device header buttons to be clicked

oestrich:
Broadcast the messages regardless of if the liveview node thinks the device is online. It might be and the tracker is out of sync. Something might be listening so let it try to happen.

### DIFF
--- a/lib/nerves_hub_web/live/device_live/show.ex
+++ b/lib/nerves_hub_web/live/device_live/show.ex
@@ -298,12 +298,7 @@ defmodule NervesHubWeb.DeviceLive.Show do
 
     socket.endpoint.broadcast_from(self(), "device:#{socket.assigns.device.id}", "reboot", %{})
 
-    socket =
-      socket
-      |> put_flash(:info, "Device Reboot Requested")
-      |> assign(:status, "reboot-requested")
-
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/live/device_live/show.ex
```diff
@@ -297,12 +297,7 @@ defmodule NervesHubWeb.DeviceLive.Show do
 
     socket.endpoint.broadcast_from(self(), "device:#{socket.assigns.device.id}", "reboot", %{})
 
-    socket =
-      socket
-      |> put_flash(:info, "Device Reboot Requested")
-      |> assign(:status, "reboot-requested")
```
jjcarstens, line 303: Is this status now assigned somewhere else?

oestrich: It's not stored anywhere. It was just weird looking if you're on a device that's offline and click the reboot button. Suddenly looks online and requested a reboot. I've been trying to peel back what the device `status` is anyways. We lost the ability to keep it up to date in the PR this is stacked on. I can add it back though, it just will get lost when you reload the page.

jjcarstens (approved):

PR #981, opened 2023-06-22
Remove gproc from dependencies

oestrich:
Remove gproc and gen_leader since they are replaced with a tracker in repo.


jjcarstens (approved):

fhunleth (approved):
🚀

PR #983, opened 2023-06-22
Remove the EctoEnum library for Ecto.Enum

oestrich:
It's built in now, no need for a library


lib/nerves_hub_web/views/org_user_view.ex
```diff
@@ -1,10 +1,10 @@
 defmodule NervesHubWeb.OrgUserView do
   use NervesHubWeb, :view
 
-  alias NervesHub.Accounts.User
-
   def role_options() do
-    User.Role.__enum_map__() |> Enum.map(fn opt -> {format_option(opt), opt} end)
+    Enum.map([:admin, :delete, :write, :read], fn opt ->
```
jjcarstens, line 5:
```suggestion
  
Ecto.Enum.values(OrgUser, :role)
|> Enum.map(fn opt ->
```
I'm pretty sure you can use the defined enum values like this (or close to this) rather than hardcoding

oestrich: You can! This is much better

jjcarstens (approved):

PR #985, opened 2023-06-27
Start otel and set up first traces on the device channel

oestrich:
Some traces:

image
image


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -110,39 +118,47 @@ defmodule NervesHubWeb.DeviceChannel do
...
-    {:noreply, socket}
+    trace("DeviceChannel.status_update", socket.device.assigns.device) do
+      # TODO store in tracker or the database?
+      {:noreply, socket}
+    end
   end
 
   def handle_in("rebooting", _payload, socket) do
-    {:noreply, socket}
+    trace("DeviceChannel.rebootign", socket.assigns.device) do
```
jjcarstens, line 128:
```suggestion
    trace("DeviceChannel.rebooting", socket.assigns.device) do
```

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -18,6 +18,9 @@ defmodule NervesHubWeb.DeviceChannel do
   alias Phoenix.Socket.Broadcast
 
   require Logger
+  require OpenTelemetry.Tracer, as: Tracer
+
+  import NervesHub.Tracer
```
jjcarstens, line 23: I don't _love_ this - We're overloading `Tracer` while also importing making the function source slightly more hidden when down in the code

config/dev.exs
```diff
@@ -86,3 +86,8 @@ config :nerves_hub, NervesHubWeb.Endpoint,
       ~r{lib/nerves_hube_www_web/live/.*(ex)$}
     ]
   ]
+
+# Export to a local collector
+config :opentelemetry_exporter,
+  otlp_protocol: :http_protobuf,
+  otlp_endpoint: "http://localhost:4318"
```
jjcarstens, line 93: If someone doesn't run this locally, will things still be okay here?

lib/nerves_hub_web/channels/device_channel.ex
oestrich: I think I did this to let the trace function be a macro do have the do block work correctly. Switched to an anonymous function instead and that module can go away for now.

config/dev.exs
oestrich: I think the answer is yes, but I'll have it be an ENV to start it otherwise be a noop

jjcarstens (approved):

PR #987, opened 2023-07-13
Add an API endpoint to upgrade a devices firmware manually

oestrich:
Similar to the UI for a device

### DIFF
--- a/lib/nerves_hub_web/controllers/api/device_controller.ex
+++ b/lib/nerves_hub_web/controllers/api/device_controller.ex
@@ -4,13 +4,16 @@ defmodule NervesHubWeb.API.DeviceController do
   alias NervesHub.AuditLogs
   alias NervesHub.Devices
   alias NervesHub.Devices.DeviceCertificate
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/controllers/api/device_controller.ex
```diff
@@ -119,4 +122,36 @@ defmodule NervesHubWeb.API.DeviceController do
 
     send_resp(conn, 200, "Success")
   end
+
+  def upgrade(conn, %{"uuid" => uuid}) do
+    product = conn.assigns.product
+    {:ok, firmware} = Firmwares.get_firmware_by_product_and_uuid(product, uuid)
```
jjcarstens, line 128: Do we need to handle if a bad/unknown UUID is sent?

oestrich: I think the 500 is fine right? If you're sending in bad data then it should be fine to fail. Assuming we don't expect this to happen often

PR #992, opened 2023-07-18
Remove tracing around oban

oestrich:
- Remove opentelemetry_oban
- Create a repo that is identical to the main repo but only for oban, so it's not traced


lib/nerves_hub/repo.ex
```diff
@@ -64,3 +64,28 @@ defmodule NervesHub.Repo do
...
+    database_config = vapor_config.database
+
+    config =
+      Keyword.merge(config,
+        pool_size: database_config.pool_size,
+        url: database_config.url
+      )
+
+    {:ok, config}
+  end
```
jjcarstens, line 90:
```suggestion
  defdelegate init(context, config), to: NervesHub.Repo
```
If the goal is to keep it the same, would you want to just delegate here so it uses the same init codepath?

lib/nerves_hub/repo.ex
```diff
@@ -64,3 +64,17 @@ defmodule NervesHub.Repo do
 
   def destroy(struct_or_changeset), do: delete(struct_or_changeset)
 end
+
+defmodule NervesHub.ObanRepo do
+  use Ecto.Repo,
+    otp_app: :nerves_hub,
+    adapter: Ecto.Adapters.Postgres
+
+  alias NervesHub.Config
```
jjcarstens, line 73:
```suggestion
```
Shouldn't need this anymore

jjcarstens (approved):

jjcarstens (approved):

PR #995, opened 2023-07-18
Start rate limiting device connections

oestrich:
Configured by ENV, but defaults to 100 / s in production. A websocket connection error is returned if the rate limit is dinged. NervesHubLink will retry the connection according to its own retry limits.

### DIFF
--- a/lib/nerves_hub/rate_limit.ex
+++ b/lib/nerves_hub/rate_limit.ex
@@ -1,0 +1,47 @@
+defmodule NervesHub.RateLimit do
+  @moduledoc """
+  Rate Limit devices connecting to the server
+  """
+
+  use GenServer
+
+  @doc """
+  Increment and check the rate limit
+
+  Return `true` if the request was under the rate limit
+  """
+  @spec increment() :: boolean()
+  def increment() do
+    bucket_key =
+      DateTime.utc_now()
+      |> DateTime.to_unix()
+
+    result = :ets.update_counter(:nerves_hub_rate_limit, bucket_key, 1, {bucket_key, 0})
+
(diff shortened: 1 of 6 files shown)


lib/nerves_hub/rate_limit.ex
```diff
@@ -0,0 +1,66 @@
...
+  def increment() do
+    bucket_key =
+      DateTime.utc_now()
+      |> DateTime.to_unix()
+
+    result = :ets.update_counter(:nerves_hub_rate_limit, bucket_key, 1, {bucket_key, 0})
+
+    config = Application.get_env(:nerves_hub, __MODULE__)
+
+    result < config[:limit]
```
jjcarstens, line 23: Should we warn somewhere that if the config value is bad, then rate limiting is basically skipped?

```elixir
iex(10)> 1 <  nil
true
iex(11)> 1 < "10"
true
iex(12)> 10 < "10"
true
iex(13)> 1000 < nil
true
```

jjcarstens (approved):

lib/nerves_hub/rate_limit.ex
oestrich: I think this will blow up and handle that for us? https://github.com/nerves-hub/nerves_hub_web/pull/995/files#diff-fc214fffb54babf6a126b25f7a4d6a0b97774cc603300c310db29de02d25e46bR12

jjcarstens: I kind of view the `release.exs` as potentially replaceable in a custom deployment. That is the point of customization for anyone hosting NervesHub. But it also seems practical that others would use the environment variable instead so this should be good for now

PR #998, opened 2023-07-21
Add device firmware update status to the API

oestrich:
- latest: updates disabled, no deployment, matching the deployment
- updating: update attempts are present and it doesn't match the above
- pending: anything else, which should be only devices that haven't started updating and don't match the current deployment firmare

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -1058,4 +1058,27 @@ defmodule NervesHub.Devices do
     |> where([iu], iu.deployment_id == ^deployment.id)
     |> Repo.one()
   end
+
+  @doc """
+  Get the firmware status for a device
+
+  "latest", "pending", or "updating"
+  """
+  def firmware_status(device) do
+    device = Repo.preload(device, deployment: [:firmware])
+
+    cond do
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/devices.ex
```diff
@@ -1058,4 +1058,30 @@ defmodule NervesHub.Devices do
...
+  """
+  def firmware_status(device) do
+    device = Repo.preload(device, deployment: [:firmware])
+
+    cond do
+      is_nil(device.deployment_id) ->
+        "latest"
+
+      device.updates_enabled == false ->
+        "latest"
```
jjcarstens, line 1075: I don't know about this one. It's not quite true. It could match a deployment and be a version behind, but the updates are just disabled. This could be misleading?

oestrich: Should we just ignore this and not have anything special for updates enabled false? That way it can be marked pending but with `updates_enabled: false` you'll know it won't happen?

jjcarstens: That might be the better route to go rather than conflate `firmware_status`

jjcarstens (approved):

PR #999, opened 2023-07-24
Update oban and related packages

oestrich:


lib/nerves_hub/devices.ex
```diff
@@ -241,7 +242,6 @@ defmodule NervesHub.Devices do
 
   def get_eligible_deployments(_), do: []
 
-  @spec create_device(map) :: {:ok, Device.t()} | {:error, Changeset.t()}
```
jjcarstens, line 244: Why remove the spec?

oestrich: This fixed my compilation error, which I don't know that it was necessary. just one of the lines elixir pointed at with incompatible types :shrug:

jjcarstens (approved):

PR #1005, opened 2023-07-27
Migrate certain device routes to be higher up in paths

oestrich:
Since we have unique identifiers across the entire NH, we can skip having the org and the product in the route. Role validation will happen in the route actions for now, and as the rest of the controller is filled out we can figure out a better way to load these and validate.

### DIFF
--- a/lib/nerves_hub_web/api_router.ex
+++ b/lib/nerves_hub_web/api_router.ex
@@ -35,6 +35,17 @@ defmodule NervesHubWeb.APIRouter do
     post("/login", UserController, :login)
   end
 
+  scope "/devices", NervesHubWeb.API do
+    pipe_through([:api, :user])
+
+    get("/:identifier", DeviceController, :show)
+    post("/:identifier/reboot", DeviceController, :reboot)
+    post("/:identifier/reconnect", DeviceController, :reconnect)
+    post("/:identifier/code", DeviceController, :code)
+    post("/:identifier/upgrade", DeviceController, :upgrade)
+    delete("/:identifier/penalty", DeviceController, :penalty)
+  end
+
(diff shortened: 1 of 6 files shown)


lib/nerves_hub_web/controllers/api/device_controller.ex
```diff
@@ -13,8 +14,7 @@ defmodule NervesHubWeb.API.DeviceController do
 
   plug(:validate_role, [org: :delete] when action in [:delete])
   plug(:validate_role, [org: :write] when action in [:create, :update])
-  plug(:validate_role, [org: :write] when action in [:reboot, :reconnect, :code, :upgrade])
```
jjcarstens, line 16: Did you mean to remove these actions from being validated?

oestrich: Yeah, they’re validated in the action now, after device lookup

lib/nerves_hub_web/api_router.ex
```diff
@@ -35,6 +35,17 @@ defmodule NervesHubWeb.APIRouter do
     post("/login", UserController, :login)
   end
 
+  scope "/devices", NervesHubWeb.API do
+    pipe_through([:api, :user])
```
jjcarstens, line 39: Do we need the `:device` pipe still used here?

oestrich: No, that loaded the device from the org in another plug. This undoes the plugs we have in place to be in the controller. I think once we’re further along in the migration I want to do we can replace the plugs with what all is going on here (maybe :shrug:)

jjcarstens (approved):

oestrich:
Merged locally

PR #1009, opened 2023-08-01
Add a reference id to audit logs

oestrich:
Mostly around device connections. So we can tie device actions to a specific connection, like a connect and disconnect message.

### DIFF
--- a/lib/nerves_hub/audit_logs/audit_log.ex
+++ b/lib/nerves_hub/audit_logs/audit_log.ex
@@ -18,7 +18,7 @@ defmodule NervesHub.AuditLogs.AuditLog do
     :resource_type,
     :org_id
   ]
-  @optional_params [:changes]
+  @optional_params [:reference_id]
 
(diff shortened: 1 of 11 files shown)


lib/nerves_hub/audit_logs/audit_log.ex
```diff
@@ -18,7 +18,7 @@ defmodule NervesHub.AuditLogs.AuditLog do
     :resource_type,
     :org_id
   ]
-  @optional_params [:changes]
```
jjcarstens, line 21: Why are we also removing the changes field? If we are, shouldn't this also be dropped from the DB?

oestrich: We can’t drop fields from the database until it’s all deployed, I have a list going that we can clean out soon. This could probably move to another pr but I noticed nothing used it and deleted it when I saw it. Action and params were used enough I did a new pr

jjcarstens (approved):

PR #1017, opened 2023-08-08
Switch to debian in the Dockerfile

oestrich:
Alpine uses musl which can cause an issue to overrun the stack and heap in OTP 26.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,10 @@
 ARG ELIXIR_VERSION=1.15.4
 ARG ERLANG_VERSION=26.0.2
-ARG ALPINE_VERSION=3.18.2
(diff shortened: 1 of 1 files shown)


jjcarstens (approved):

Dockerfile
```diff
@@ -37,23 +36,18 @@ COPY --from=assets /build/priv/static priv/static
 RUN mix do phx.digest, release nerves_hub --overwrite
 
 # Release Container
-FROM alpine:${ALPINE_VERSION} as release
+FROM debian:buster as release
```
esvinson, line 39: @oestrich This should probably match the tag above?  Using immutable tags can sometimes result in weird cache issues.
`debian:buster-20230612`

PR #1025, opened 2023-09-05
Allow setting role when inviting new user

jjcarstens:
Adds `role` to the invite stored for new users so they are created with the expected role.

image

### DIFF
--- a/lib/nerves_hub/accounts.ex
+++ b/lib/nerves_hub/accounts.ex
@@ -440,7 +440,7 @@ defmodule NervesHub.Accounts do
   def add_or_invite_to_org(%{"email" => email} = params, org) do
     case get_user_by_email(email) do
       {:error, :not_found} -> invite(params, org)
-      {:ok, user} -> add_org_user(org, user, %{role: :admin})
+      {:ok, user} -> add_org_user(org, user, %{role: params["role"]})
     end
   end
 
@@ -507,7 +507,7 @@ defmodule NervesHub.Accounts do
 
     Repo.transaction(fn ->
       with {:ok, user} <- create_user(user_params),
-           {:ok, user} <- add_org_user(org, user, %{role: :admin}),
+           {:ok, user} <- add_org_user(org, user, %{role: invite.role}),
            {:ok, _invite} <- set_invite_accepted(invite) do
         # Repo.transaction will wrap this in an {:ok, user}
         user
--- a/lib/nerves_hub/accounts/invite.ex
+++ b/lib/nerves_hub/accounts/invite.ex
@@ -4,6 +4,7 @@ defmodule NervesHub.Accounts.Invite do
   import Ecto.Changeset
 
   alias NervesHub.Accounts.Org
+  alias NervesHub.Accounts.OrgUser
   alias __MODULE__
 
   @type t :: %__MODULE__{}
@@ -14,7 +15,7 @@ defmodule NervesHub.Accounts.Invite do
     field(:email, :string)
     field(:token, Ecto.UUID)
     field(:accepted, :boolean)
-    field(:role, Ecto.Enum, values: Ecto.Enum.values(NervesHub.Accounts.OrgUser, :role))
+    field(:role, Ecto.Enum, values: Ecto.Enum.values(OrgUser, :role))
 
     timestamps()
   end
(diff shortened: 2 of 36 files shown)


fhunleth:
What's the difference between the write, delete and admin roles?

jjcarstens:
It just depends. They are defined in each controller. I think the next bigger task is to really define that and have it more central so it's easy to overview what each can/cannot do.

In the meantime, the typical means are:

* `read` - mostly `show` and `index` pages. Allows creation of users items
* `delete` - Like `read`, but allows deleting user items
* `write` - CRUD actions for more shared items (device, CA certificate, etc)
* `admin` - All the things, especially altering org details (invite users, create new products, etc)

fhunleth:
Since these are "roles" and we have a "role" called "admin", would different words make sense? I find read, write, and delete a little confusing - especially since "read" allows creation? 

Since our immediate requirement is to support a user that has access to device status and nothing else, can the roles be "admin" and "status"? Or another word besides "status"? And then delete the "read", "delete", and "write" roles?

jjcarstens:
I misspoke. I went through some controllers and `read` does not seem able to create.

Yes, I do think that makes sense, but I also think it's a much bigger change than this PR.`role` in the DB is an Enum type and that is just difficult enough to change that it requires thought and I'd rather it be separate.

jjcarstens:
Thought right after submitting - the quick hack could be displaying the role as `status`, but not changing the Enum value in the DB. I don't love how that disconnects the view and code handling so much, but I could be convinced

fhunleth:
Wouldn't it be better if there was a layer of indirection here? I.e., the role would point to a set of fine grained permissions that would determine what could be done. Perhaps the set is trivially small now and having "if" statements all over won't be terrible. 

I'm ok with whatever you all think best on this. My desire would be to expose fewer roles while we still can so that we don't accidentally lock in confusing names as people start using them.

jjcarstens:
Chatted offline. Going to only display the `Admin` and `Read` roles for now until all the permissions are adjusted (see https://github.com/nerves-hub/nerves_hub_web/pull/1025/commits/e0594f48c9a8301593ae0a2936aa1bb764c70984)

joshk (approved):

joshk:
Upon reflection, I think we should use `admin`, `manage`, and `view` as the roles.

`manage` is a good middle ground for cases where you want to restrict actions like 'delete a product'

I also think these names are more descriptive.

joshk:
@fhunleth @jjcarstens 

What are your thoughts on three user roles:

`admin` : can create and delete products, and can invite users and set roles
`manage` : do everything except the above
`view` : can view all information

jjcarstens:
I like those role names and these basic starting restrictions

fhunleth:
Same. Those roles sound good to me.

joshk:
@fhunleth @jjcarstens this is ready for another review.

priv/repo/migrations/20231221105046_add_new_role_values.exs
jjcarstens, line 1: My only concern with all this is that Postgres Enums are hairy, especially when trying to change. Do we have major benefit by using a database enum? Maybe we should give up the `:role` key for something different, make it a simple `string` field with `Ecto.Enum` in the application to manage the type and simplify the db setup a bit?

joshk: That’s fair. Just so you know, I tested these migrations locally and they work perfectly. I had to split them into three as the enum changes need to be committed for them to be usable.

jjcarstens: Josh and I brainstormed this a bit ultimately deciding that our future selves and those migrating would prob prefer a much simpler setup changing the roles to simple string field in the DB.

The main benefits we get from Enum types in the DB is database validation and maybe more efficient lookups. However, we don't look up the role field individually as it comes with the whole user field so it seems worth the trade off to handle validation in the application instead to make it much simpler to maintain going forward

joshk:
from a PR review by @jjcarstens 

```
Josh and I brainstormed this a bit ultimately deciding that our future selves and those migrating would prob prefer a much simpler setup changing the roles to simple string field in the DB.

The main benefits we get from Enum types in the DB is database validation and maybe more efficient lookups. However, we don't look up the role field individually as it comes with the whole user field so it seems worth the trade off to handle validation in the application instead to make it much simpler to maintain going forward
```

PR #1027, opened 2023-09-07
Don't allow adjusting the iex console window

oestrich:
Everyone will have the same terminal size and not be allowed to adjust it. This will prevent oddities when multiple people are connected.

Screenshot 2023-09-07 at 5 01 48 PM

### DIFF
--- a/assets/css/_console.scss
+++ b/assets/css/_console.scss
@@ -1,108 +1,82 @@
-// Console V1
-
-.console-wrapper {
-  background: black;
-  padding: 1rem;
-  border-radius: 4px;
-}
-
-.console {
-  font-family: monospace;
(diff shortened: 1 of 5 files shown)


oestrich:
The theme for the console has been updated, along with adding in a chat to show who all loaded the console to someone active in the console

Screenshot 2023-09-08 at 10 07 03 AM

fhunleth:
This looks great! 

I mean, now that you've started this, I can come up with a ton of things that would be super helpful to have handy when I'm at the console window. I'm going to hold off since I don't want to mess up the key part of this: fixing the screen blanking all the time when others connect.

jjcarstens (approved):

PR #1029, opened 2023-09-08
Use unique channels to limit the chat to that device

oestrich:
It was a global chat previously, so any device could chat with another device. This limits to specific devices.

### DIFF
--- a/lib/nerves_hub_web/channels/user_console_channel.ex
+++ b/lib/nerves_hub_web/channels/user_console_channel.ex
@@ -3,9 +3,8 @@ defmodule NervesHubWeb.UserConsoleChannel do
 
   alias Phoenix.Socket.Broadcast
 
-  def join("user_console", %{"device_id" => device_id, "product_id" => product_id}, socket) do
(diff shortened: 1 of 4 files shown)


lib/nerves_hub_web/channels/user_console_channel.ex
```diff
@@ -3,9 +3,8 @@ defmodule NervesHubWeb.UserConsoleChannel do
 
   alias Phoenix.Socket.Broadcast
 
-  def join("user_console", %{"device_id" => device_id, "product_id" => product_id}, socket) do
+  def join("user:console:" <> device_id, _, socket) do
```
jjcarstens, line 6: Is device_id used elsewhere? I think now it would be a string compared to being an integer before?

oestrich: It's only for channel topics in this file, want me to parse it so we stay consistent? I'm fine either way

jjcarstens (approved):

PR #1040, opened 2023-09-29
Remove `firmware:<uuid>` topic from DeviceChannel

jjcarstens:
This topic was used in the earliest version of NervesHub and the device client connection. However, the lib it was used in ([`nerves_hub`](https://github.com/nerves-hub/nerves_hub)) has been deprecated since 2020-04-14. The last version of that lib to use this topic was [v0.4.0](https://github.com/nerves-hub/nerves_hub/releases/tag/v0.4.0) published 2019-02-14. The latest version available in Hex is v0.7.4 released 2019-06-21.

Since this was completely replaced by `nerves_hub_link` and it's been 4 years, I think it's safe to remove support for this topic which we don't have intentions on reusing in 2.0

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -20,14 +20,6 @@ defmodule NervesHubWeb.DeviceChannel do
   require Logger
   require OpenTelemetry.Tracer, as: Tracer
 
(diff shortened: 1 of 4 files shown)


oestrich (approved):
I’ve been too afraid to delete this 🚮

PR #1043, opened 2023-10-02
Add metrics around SSL connections

oestrich:
More statsd metrics than spans to keep overhead lower for this very critical section.

### DIFF
--- a/lib/nerves_hub/ssl.ex
+++ b/lib/nerves_hub/ssl.ex
@@ -3,6 +3,8 @@ defmodule NervesHub.SSL do
   alias NervesHub.Certificate
   alias NervesHub.RateLimit
 
+  require OpenTelemetry.Tracer, as: Tracer
+
   @type pkix_path_validation_reason ::
           :cert_expired
           | :invalid_issuer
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/ssl.ex
```diff
@@ -42,30 +44,32 @@ defmodule NervesHub.SSL do
...
-        # cert next and expiration will be checked later if registration of a new
-        # device cert needs to happen.
-        {:valid, state}
-
-      true ->
-        # The signer CA was not included in the request, so this is most
-        # likely a device cert that needs verification. If it isn't, then
-        # this is some other unknown CA that will fail
-        do_verify(otp_cert, state)
+    Tracer.with_span "SSL.bad_cert" do
```
jjcarstens, line 47: Do we want to be any more specific here with `unknown_ca` or `cert_epxired`?

oestrich: Dunno, we can set an attribute in each of the cond's so we know what branch we hit, I think without generating more spans and overloading things. I was mostly curious how often this was happening and the general length of time it took.

jjcarstens (approved):

PR #1046, opened 2023-10-03
Update :env for statsd metrics to use DEPLOY_ENV

oestrich:
Instead of using the compile time environment, load it from the
environment variables.

### DIFF
--- a/config/release.exs
+++ b/config/release.exs
@@ -9,6 +9,8 @@ nerves_hub_app =
 
 config :nerves_hub, app: nerves_hub_app
 
+config :nerves_hub, deploy_env: System.get_env("DEPLOY_ENV")
+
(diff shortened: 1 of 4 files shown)


config/release.exs
```diff
@@ -9,6 +9,8 @@ nerves_hub_app =
 
 config :nerves_hub, app: nerves_hub_app
 
+config :nerves_hub, deploy_env: System.get_env("DEPLOY_ENV")
```
jjcarstens, line 12:
```suggestion
config :nerves_hub, deploy_env: System.get_env("DEPLOY_ENV", "prod")
```
Do you think we need a default here?

oestrich: It's also later on down in the sentry config, so everyone should have it. It's in the self hosted env template as well.

jjcarstens (approved):

PR #1055, opened 2023-10-06
Include the org and product names for a device API response

oestrich:
Request from @jjcarstens

### DIFF
--- a/lib/nerves_hub_web/views/api/device_view.ex
+++ b/lib/nerves_hub_web/views/api/device_view.ex
@@ -27,7 +27,9 @@ defmodule NervesHubWeb.API.DeviceView do
       firmware_update_status: Devices.firmware_status(device),
       deployment: render_one(device.deployment, __MODULE__, "deployment.json", as: :deployment),
       updates_enabled: device.updates_enabled,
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/views/api/device_view.ex
```diff
@@ -27,7 +27,9 @@ defmodule NervesHubWeb.API.DeviceView do
       firmware_update_status: Devices.firmware_status(device),
       deployment: render_one(device.deployment, __MODULE__, "deployment.json", as: :deployment),
       updates_enabled: device.updates_enabled,
-      updates_blocked_until: device.updates_blocked_until
+      updates_blocked_until: device.updates_blocked_until,
+      org_name: device.org.name,
+      product_name: device.product.name
```
jjcarstens, line 32: Does `_name` need to be in here? It seems like we could just do `org` and `product` then be fine? I don't know how API's usually try to handle that?

oestrich: I'd rather have it be `_name` or nest it under an `org` field as `name`

jjcarstens (approved):

PR #1056, opened 2023-10-14
Start to a load balancer for devices

oestrich:
Simply logs if the current node would shutdown due to increased device counts. This will help balance devices during a deploy, refusing devices to join an older node that is saturated with devices while the new nodes have very few devices on them.

### DIFF
--- a/lib/nerves_hub/load_balancer.ex
+++ b/lib/nerves_hub/load_balancer.ex
@@ -1,0 +1,83 @@
+defmodule NervesHub.LoadBalancer do
+  @moduledoc """
+  Determines whether the node is overloaded by scanning the online devices
+  and if the node is running high, flips a switch to stop accepting new
+  devices on this node. This is useful in a rolling deploy where the older
+  a node is, the more likely it is to run significantly higher than new nodes
+  coming in.
+
+  Listens for nodes in the cluster to go offline which will result in a large
+  device state sync. Accepting devices will be temporarily disabled across
+  the cluster to help prevent state issues from popping up.
+
+  *Note*: Currently this only logs when devices would be blocked to let it run
+  in a production environment to see that devices would be blocked at
+  appropriate times.
+  """
+
+  use GenServer
+
+  require Logger
+
+  @doc """
+  Start the load balancer
+
+  There are currently no options.
+  """
+  @spec start_link(keyword()) :: GenServer.on_start()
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/load_balancer.ex
```diff
@@ -0,0 +1,71 @@
...
+  *Note*: Currently this only logs when devices would be blocked to let it run
+  in a production environment to see that devices would be blocked at
+  appropriate times.
+  """
+
+  use GenServer
+
+  require Logger
+
+  def start_link(opts \\ []) do
```
fhunleth, line 22:
```suggestion
  @doc """
  Start the load balancer
  
  There are currently no options.
  """
  @spec start_link(keyword()) :: GenServer.on_start()
  def start_link(opts) do
```

lib/nerves_hub/load_balancer.ex
```diff
@@ -0,0 +1,71 @@
...
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  def init(_) do
```
fhunleth, line 26:
```suggestion
  @impl GenServer
  def init(_opts) do
```

lib/nerves_hub/load_balancer.ex
```diff
@@ -0,0 +1,71 @@
...
+    Process.send_after(self(), :tick, 60_000)
+
+    :net_kernel.monitor_nodes(true)
+
+    {:ok, :undefined}
+  end
+
+  def handle_info(:tick, state) do
```
fhunleth, line 34:
```suggestion
  @impl GenServer
  def handle_info(:tick, state) do
```

lib/nerves_hub/load_balancer.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  def handle_info({:nodedown, _node}, state) do
+    Logger.info("[LoadBalancer] Node went offline")
+    {:noreply, state}
+  end
+
+  def handle_info({:nodeup, _node}, state) do
+    Logger.info("[LoadBalancer] Node came online")
+    {:noreply, state}
+  end
```
fhunleth, line 70:
```suggestion
  end
  
  def handle_info(_other, state) do
    {:noreply, state}
  end
```

fhunleth (approved):

PR #1058, opened 2023-10-18
Fix OTP & TLS link

jjcarstens:
And internal link was previously used which could not be shared outside my company which wasn't discovered until now.

The OTP & TLS document was moved to my private and shared to anyone with the link


oestrich (approved):
🔗

PR #1059, opened 2023-10-18
Send files to a device through the web console

oestrich:
Web side of sending files to a device through the NervesHub console. Transfers files similar to how file upload chunking happens from a device.

Related: https://github.com/nerves-hub/nerves_hub_link/pull/131


oestrich:
One thing I'm not entirely sure on is how to make this aware to the user. You just drag the file onto the window and it uploads to the device with no progress or anything else. The file is simply there in the configured folder when it's done. We could add a progress bar for that at least but how do we make it known that you can even drag files on? 🤷

jjcarstens (approved):

PR #1062, opened 2023-10-23
Add dependabot automerge

jjcarstens:
Adds an extra job that will run only on dependabot PR's, require all previous jobs to be successful, and require the PR was created by dependabot. If all conditions are met, it will automatically merge the dependabot PR


oestrich (approved):
:shipit:

PR #1090, opened 2023-11-02
Consolidate configuration into Vapor

coop:
Initially I was wanting to deploy to fly.io but I found it hard to figure out what configuration was required. The current state of affairs has configuration living in environment variables (good) but it is a mix of `System.get_env/1` in config files that could be runtime, vapor vs `System.get_env/1` and some stuff (e.g. device endpoint) which wasn't configurable.

I don't know if this is where you were headed but I've pushed as much config as possible into Vapor and validated in the application start callback that whatever is required is set.

Maybe it's an ambitious goal but I'd like to think that eventually nerves_hub_web could be distributed as a Docker container which is configured exclusively through environment variables. This PR gets really close to that goal, the only remaining change is moving `config :nerves_hub, firmware_upload: NervesHub.Firmwares.Upload.File` from compile time to runtime which I can implement if you'd like in a separate PR.

If this is a valuable change I will rebase with master and fixup any outstanding issues.

NOTE: I renamed some environment variables for consistency between endpoint and device endpoint - if you're running this in production you'll want to review those changes.


.env
```diff
@@ -1,6 +1,12 @@
...
-SECRET_KEY_BASE=3uRRXQ9FUMc3dd7NFTFWQU2xRXcVFXWpPLHZ4fdG/toP9/jJL92Mtjmq8Fd/ijZe
-
-HOST=nerves-hub.org
-URL_SCHEME=http
-URL_PORT=4000
+DEVICE_ENDPOINT_HTTPS_CACERTFILE=../../../../test/fixtures/ssl/ca.pem
+DEVICE_ENDPOINT_HTTPS_CERTFILE=../../../../test/fixtures/ssl/device.nerves-hub.org.pem
+DEVICE_ENDPOINT_HTTPS_KEYFILE=../../../../test/fixtures/ssl/device.nerves-hub.org-key.pem
+DEVICE_ENDPOINT_HTTPS_PORT=4001
+DEVICE_ENDPOINT_SECRET_KEY_BASE=3uRRXQ9FUMc3dd7NFTFWQU2xRXcVFXWpPLHZ4fdG/toP9/jJL92Mtjmq8Fd/ijZe
```
coop, line 5: It wasn't obvious to me if this secret should be shared with web endpoint or not. At the moment I've allowed configuring them separately but it's possible that isn't desired.

config/config.exs
```diff
@@ -24,8 +18,27 @@ config :phoenix,
 # NervesHub Device
 #
 
-# Configures the endpoint
+# Configures the DeviceEndpoint
 config :nerves_hub, NervesHubWeb.DeviceEndpoint,
+  url: [],
+  http: false,
+  https: [
```
coop, line 25: From what I can see this endpoint _cannot_ be run over http (https is required) so I've disabled it and required https configuration.

config/config.exs
```diff
@@ -55,23 +67,21 @@ config :nerves_hub, Oban,
...
   queue_target: 500,
   queue_interval: 5_000
 
 ##
 # NervesHubWWW
 #
-config :nerves_hub,
-  ecto_repos: [NervesHub.Repo],
-  # Options are :ssl or :header
-  websocket_auth_methods: [:ssl]
```
coop, line 68: This is not used.

config/config.exs
```diff
@@ -55,23 +67,21 @@ config :nerves_hub, Oban,
...
+
 config :nerves_hub, NervesHub.Repo,
   queue_target: 500,
   queue_interval: 5_000
 
 ##
 # NervesHubWWW
 #
-config :nerves_hub,
-  ecto_repos: [NervesHub.Repo],
```
coop, line 66: Duplicated several lines above.

config/config.exs
```diff
@@ -82,5 +92,12 @@ config :opentelemetry,
 
 config :swoosh, :api_client, Swoosh.ApiClient.Finch
 
+config :sentry,
+  dsn: System.get_env("SENTRY_DSN_URL"),
```
coop, line 96: TODO: this can be moved to runtime config.

config/config.exs
```diff
@@ -82,5 +92,12 @@ config :opentelemetry,
...
+  environment_name: System.get_env("DEPLOY_ENV"),
```
coop, line 97: This value is validated by the sentry lib as part of startup BUT can be changed when nerves hub boots. Which means the config can be moved to runtime.

config/dev.exs
```diff
@@ -1,78 +1,41 @@
...
+#     mix phx.gen.cert
+#
+# Note that this task requires Erlang/OTP 20 or later.
+# Run `mix help phx.gen.cert` for more information.
 config :nerves_hub, NervesHubWeb.DeviceEndpoint,
   debug_errors: true,
   code_reloader: false,
   check_origin: false,
-  watchers: [],
-  https: [
```
coop, line 23: Moved to `config/config.exs` - no need to duplicate in each config.

config/prod.exs
```diff
@@ -1,14 +1,15 @@
 import Config
 
 # Do not print debug messages in production
-config :logger, level: :info, backends: [:console, Sentry.LoggerBackend]
```
coop, line 4: Sentry says to configure this in the application start callback.

config/release.exs
```diff
@@ -1,55 +1,9 @@
...
-    {:ok, sync_nodes_optional} ->
-      sync_nodes_optional
-      |> String.split(" ", trim: true)
-      |> Enum.map(&String.to_atom/1)
-
-    :error ->
-      []
-  end
-
-config :kernel,
```
coop, line 33: This setting seems to be deployment specific. When deploying to fly we need something different so I've removed for now - if you want this back lmk.

lib/nerves_hub/application.ex
```diff
@@ -20,19 +22,25 @@ defmodule NervesHub.Application do
       :opentelemetry_cowboy.setup()
     end
 
+    Application.put_env(:nerves_hub, :host, vapor_config.web_endpoint.url_host)
+    Application.put_env(:nerves_hub, :port, vapor_config.web_endpoint.url_port)
+    Application.put_env(:nerves_hub, :from_email, vapor_config.nerves_hub.from_email)
+    Application.put_env(:nerves_hub, :app, vapor_config.nerves_hub.app)
+    Application.put_env(:nerves_hub, :deploy_env, vapor_config.nerves_hub.deploy_env)
```
coop, line 29: Ideally this isn't required. In the code both `Vapor.load!/1` and `Application.get_env/2` are used so I "just made it work".

lib/nerves_hub/load_balancer.ex
```diff
@@ -56,10 +56,12 @@ defmodule NervesHub.LoadBalancer do
 
     average_count = Enum.sum(Map.values(device_counts)) / Enum.count(Map.values(device_counts))
 
-    current_node_ratio = device_counts[node()] / average_count
```
coop, line 59: With no devices connected I was getting an error here.

lib/nerves_hub/release/tasks.ex
```diff
@@ -1,57 +1,42 @@
 defmodule NervesHub.Release.Tasks do
```
coop, line 1: I will revert these changes - please ignore.

lib/nerves_hub/load_balancer.ex
jjcarstens: I'm pretty sure this was addressed in main now. Would you mind rebasing?

lib/nerves_hub/config.ex
```diff
@@ -3,36 +3,62 @@ defmodule NervesHub.Config do
...
+           {:url_host, "WEB_ENDPOINT_URL_HOST"},
+           {:url_port, "WEB_ENDPOINT_URL_PORT", default: 443, map: &String.to_integer/1},
+           {:url_scheme, "WEB_ENDPOINT_URL_SCHEME", default: "https"},
+           {:server, "WEB_ENDPOINT_SERVER", default: true, map: &to_boolean/1}
+         ])
+
+  config :device_endpoint,
+         env([
+           {:live_view_signing_salt, "DEVICE_ENDPOINT_LIVE_VIEW_SIGNING_SALT", required: false},
+           {:secret_key_base, "DEVICE_ENDPOINT_SECRET_KEY_BASE"},
```
jjcarstens, line 45: These should not be needed for device endpoint

jjcarstens (commented):
Thanks @coop !

I think this is a great start. I've breezed through a bit, but need to break for now and come back later with a comb

config/release.exs
oestrich: This is required for existing deployments, it needs to be left

coop: From my understanding these values can be configured as environment variables. Given they're specific to one kind of deployment target maybe we can figure out a way to set them at runtime rather than forcing everyone to have these settings.

I am working through the changes to support deploying to fly and I've found that `rel/env.sh.eex` may be a good place to hardcode if you don't want to specify them at runtime.

jjcarstens:
Can we break this up into smaller chunks? I like the idea, but it's currently a massive change with lots of risk towards breaking current deployments. I think taking a step back and first clearly defining deployment requirements is step one. Then we can start dealing with generic vs specific configuration based on deployment type (aws, fly, etc) and once we have a handle on that we can implement pieces at a time to shift how this is configured.

If we have that, I think we'd be able to better document using your own `runtime.exs` or `env.sh.ex` just custom env vars to control configuration at deployment more

coop:
Yep no problems.

I am just about finished my work to get fly.io working - I'll push that up and change this PR to draft so you can see where I think this should roughly end up. From there I can summarise the changes, you tell me what is worth doing / what you'd do differently and then I can submit individual PRs based on that feedback. Does that sound ok?

coop:
Sorry, my first attempt at responding was poorly worded. Take 2.

>Can we break this up into smaller chunks?

Yep no problems.

I am just about finished my work to get fly.io working - I'll push that up and change this PR to draft so you can see where I think this should roughly end up. From there I can summarise the changes / challenges I made.

joshk:
I'm really sorry, but I think we might want to close this PR as #1111 standardized configuration around `runtime.exs`

I'm also working on some docs updates on how to run Nerves Hub on Fly.io using the new `runtime.exs` setup (and updated `Dockerfile`)

I'd be happy to work on more of this with you, @coop 

Feel free to ping me in the Nerves Hub Discord.

PR #1091, opened 2023-11-03
Assign an archive to a deployment

oestrich:
And send a socket message on device connect to tell about the archive.

Related: https://github.com/nerves-hub/nerves_hub_link/pull/135

### DIFF
--- a/lib/nerves_hub/devices/device_link.ex
+++ b/lib/nerves_hub/devices/device_link.ex
@@ -8,6 +8,7 @@ defmodule NervesHub.Devices.DeviceLink do
 
   use GenServer
 
+  alias NervesHub.Archives
   alias NervesHub.AuditLogs
   alias NervesHub.Deployments
   alias NervesHub.Devices
@@ -30,6 +31,7 @@ defmodule NervesHub.Devices.DeviceLink do
       :reference_id,
       :transport_pid,
       :transport_ref,
+      :device_api_version,
       :update_started?
     ]
   end
@@ -184,6 +186,8 @@ defmodule NervesHub.Devices.DeviceLink do
         Tracer.set_attribute("nerves_hub.device.id", device.id)
         Tracer.set_attribute("nerves_hub.device.identifier", device.identifier)
 
+        state = %{state | device_api_version: Map.get(params, "device_api_version", "1.0.0")}
+
         description = "device #{device.identifier} connected to the server"
 
         AuditLogs.audit_with_ref!(
@@ -197,7 +201,7 @@ defmodule NervesHub.Devices.DeviceLink do
           device
           |> Devices.verify_deployment()
           |> Deployments.set_deployment()
-          |> Repo.preload(deployment: [:firmware])
+          |> Repo.preload(deployment: [:archive, :firmware])
 
         # clear out any inflight updates, there shouldn't be one at this point
         # we might make a new one right below it, so clear it beforehand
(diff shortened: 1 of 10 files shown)


lib/nerves_hub_web/views/deployment_view.ex
```diff
@@ -77,6 +77,27 @@ defmodule NervesHubWeb.DeploymentView do
...
+            %Version{major: 0, minor: 0, patch: 0}
+        end
+      end,
+      {:desc, Version}
+    )
+    |> Enum.map(&[value: &1.id, key: archive_display_name(&1)])
+  end
+
+  def archive_display_name(%{} = a) do
+    "#{a.version} #{a.platform} #{a.architecture} #{a.uuid}"
```
jjcarstens, line 98:
```suggestion
    "#{a.version} #{a.uuid} #{a.platform} #{a.architecture}"
```

I feel like version and UUID are probably the most important here. Not sure if we even need the rest, but at least have those 2 together and leftmost

lib/nerves_hub_web/templates/deployment/edit.html.heex
```diff
@@ -50,6 +50,15 @@
     <div class="has-error"><%= error_tag f, :firmware_id %></div>
   </div>
 
+  <div class="form-group">
+    <label for="firmware_id" class="tooltip-label">
```
jjcarstens, line 54:
```suggestion
    <label for="archive_id" class="tooltip-label">
```

jjcarstens (changes requested):

lib/nerves_hub_web/views/deployment_view.ex
oestrich: This was a copy paste from firmware, so we might want to update that one in another PR

lib/nerves_hub/devices/device_link.ex
```diff
@@ -30,6 +31,7 @@ defmodule NervesHub.Devices.DeviceLink do
       :reference_id,
       :transport_pid,
       :transport_ref,
+      :transport_version,
```
jjcarstens, line 34: Why `transport_version` vs `device_api_version`? Seems like it might be better to match so it's less confusing?

test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -93,7 +93,7 @@ defmodule NervesHubWeb.WebsocketTest do
 
       {:ok, socket} = SocketClient.start_link(@socket_config)
       SocketClient.wait_connect(socket)
-      SocketClient.join(socket, "device")
+      SocketClient.join(socket, "device", %{"device_api_version" => "2.0.0"})
```
jjcarstens, line 96: Should we revert all these `device_api_version` checks for now? Its probably better that is the default, but I think we'll want to also have tests for <= 1.0.0 for backwards compatibility

oestrich: Yeah, that works for me

lib/nerves_hub/devices/device_link.ex
oestrich: I think I just picked it since there were other transport keys, so I'm not sold on it at all. I'll swap them to match

jjcarstens (commented):
Sorry, I forgot one more thing. Can we also get a test in device link for the archive push?

jjcarstens (approved):

PR #1093, opened 2023-11-09
Remove API from ecs-service, and allow overriding service names via ENV

esvinson:
This removes the references to API tasks which are no longer used.

This also adds the option to override the service name while maintaining backwards compatibility with the previous values.


DevBaca (approved):
🚀

PR #1094, opened 2023-11-10
Add minor deployment update optimization

jjcarstens:
The version change is the limiting factor of when a device can be matched to a deployment fully in a db transaction. This adds a minor optimization to do the deployment calculation for devices in the DB when the deployment does not have a version condition.

Doing this also helps devices that might be offline when the deployment is updated by reducing the load of DB lookups when a device connects. It is especially useful for cases where many devices are coming online at once


oestrich (approved):
I think this is good, we might do double updates for online device but I think that's a fair trade off.

PR #1096, opened 2023-11-13
Start on an OpenAPI specification for the API

oestrich:
Starting simple with just fetching information about a device.

### DIFF
--- a/api-spec.yml
+++ b/api-spec.yml
@@ -1,0 +1,165 @@
+openapi: '3.0.2'
+info:
+  title: NervesHub Web
+  version: '2.0'
+servers:
+  - url: http://localhost:4000/api
+security:
+  - bearerAuth: []
+components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
+  schemas:
+    deployment:
+      type: object
+      nullable: true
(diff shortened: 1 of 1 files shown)


api-spec.yml
```diff
@@ -0,0 +1,165 @@
...
+  version: '2.0'
+servers:
+  - url: http://localhost:4000/api
+security:
+  - bearerAuth: []
+components:
+  securitySchemes:
+    bearerAuth:
+      type: http
+      scheme: bearer
```
jjcarstens, line 13: This won't work. The header will end up being `Authorization: Bearer 1235` and it will need to be `Authorization: token 1234`

oestrich: It does work, I've been using this for a while. Looks like we accept both: https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/plugs/api/user.ex#L40

jjcarstens: lol. Go me 🤦 https://github.com/nerves-hub/nerves_hub_web/blame/main/lib/nerves_hub_web/plugs/api/user.ex#L40

jjcarstens (approved):

PR #1106, opened 2023-11-22
Ensure eligbile deployments returns stable results

### LINKED ISSUE #1105: Indeterminate deployment when device connects to socket and multiple deployments available
When a device connects to the socket a [list of possible deployments](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/deployments.ex#L251-L272) are found and the [first one is assigned to the device](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/devices/device_link.ex#L199) - the problem is the deployments are not ordered so "the first" becomes indeterminate. I believe deployments can be sorted using `Version.compare/2` e.g.:

```elixir
iex(1)> versions = ["1.0.0-rc.2", "1.0.0-rc.1", "1.0.2", "1.0.0", "1.0.1"]
["1.0.0-rc.2", "1.0.0-rc-1", "1.0.2", "1.0.0", "1.0.1"]
iex(2)> Enum.sort(versions, {:desc, Version})
["1.0.2", "1.0.1", "1.0.0", "1.0.0-rc.2", "1.0.0-rc.1"]
```

If this is done after filtering for `version_match?/2` the device _should_ get the most recent deployment available.

I saw this because I had multiple eligible deployments available and each device ended up with a different version. I haven't been able to reproduce the issue (naturally) but when reading through the code this codepath looks very suspicious.

coop:
This change ensures that `Deployments.alternate_deployments/2` returns deployments sorted by the most recent firmware version.

Fixes #1105

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -259,6 +259,7 @@ defmodule NervesHub.Deployments do
   def alternate_deployments(device, active) do
     Deployment
     |> join(:inner, [d], assoc(d, :firmware), as: :firmware)
+    |> preload([_, firmware: f], firmware: f)
     |> where([d], d.product_id == ^device.product_id)
     |> where([d], d.is_active in ^active)
     |> ignore_same_deployment(device)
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/deployments.ex
```diff
@@ -259,16 +259,25 @@ defmodule NervesHub.Deployments do
...
-    |> Enum.filter(fn deployment ->
-      version_match?(device, deployment)
-    end)
+    |> Enum.filter(&version_match?(device, &1))
+    |> Enum.sort_by(
+      &{&1.firmware.version, &1.id},
+      fn {a_vsn, a_id}, {b_vsn, b_id} ->
+        case Version.compare(a_vsn, b_vsn) do
+          :lt -> false
+          :eq -> a_id <= b_id
```
coop, line 276: It is possible that two deployments contain the same firmware, in that case prefer the one with the lower ID. At the end of the day it doesn't matter because the firmware will be the same but predictable sorting is always nice.

coop:
FYI this does preload the firmware for each deployment which it wasn't doing previously. If this is undesirable I can map the results and remove the preloaded association.

joshk:
This is very cool and a great find!

jjcarstens (approved):
🎉 Thanks @coop !

PR #1107, opened 2023-11-22
Allow configuring device endpoint with ENV VARS

coop:
This PR allows configuring the device endpoint through environment variables. Previously all values were hardcoded _except_ in `config/release.exs` were host and port were configurable.

The new env vars are:

`DEVICE_ENDPOINT_HTTPS_CACERTFILE`
`DEVICE_ENDPOINT_HTTPS_CERTFILE`
`DEVICE_ENDPOINT_HTTPS_KEYFILE`
`DEVICE_ENDPOINT_HTTPS_PORT` (previously `DEVICE_PORT` in production only)
`DEVICE_ENDPOINT_URL_HOST` (previously `DEVICE_HOST` in production only)
`DEVICE_ENDPOINT_HTTPS_VERSIONS`
`DEVICE_ENDPOINT_HTTPS_CERTIFICATE_AUTHORITIES`

From what I can see you were previously setting `DEVICE_HOST` and `DEVICE_PORT`. These two values can be changed to one of the env vars mentioned above.

With these changes anyone can setup certificates that match their system without requiring changes to the codebase.


test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -9,48 +9,33 @@ defmodule NervesHubWeb.WebsocketTest do
...
   alias NervesHub.Repo
   alias NervesHub.Tracker
-  alias NervesHubWeb.DeviceEndpoint
 
   @valid_serial "device-1234"
   @valid_product "test-product"
 
-  @device_port Application.compile_env(:nerves_hub, DeviceEndpoint) |> get_in([:https, :port])
-
-  @bad_socket_config [
```
coop, line 19: These attributes were used in a single test so I moved them to that test. This seems inline with other tests that have custom options passed to the socket.

I did create `socket_config/0` which could optionally take an overrides argument that deep merges allowing us to remove duplicating config - I will leave that for another day though.

lib/nerves_hub/application.ex
```diff
@@ -50,9 +50,11 @@ defmodule NervesHub.Application do
   end
 
   defp endpoints(:test) do
+    vapor = NervesHub.Config.DeviceEndpoint.load!()
+
     [
       NervesHub.Devices.Supervisor,
-      NervesHubWeb.DeviceEndpoint,
+      {NervesHubWeb.DeviceEndpoint, vapor},
```
coop, line 57: Phoneix allows you to pass configuration in via the suprtvisor and it will keep merge with `Application.get_env(:app_name, Endpoint)` - this means you no longer need the `init/2` callback in Endpoint either.

jjcarstens (commented):
This is a good start. I want to think on the env var names a bit. I'd also like to make sure the old ones are still supported and not just replaced just to reduce risk for current deploys

coop:
@jjcarstens over in https://github.com/nerves-hub/nerves_hub_web/pull/1090 you mentioned:

>I think taking a step back and first clearly defining deployment requirements is step one.

This was my attempt to start the conversation around real code. So with that in mind here are my thoughts and some followup regarding your comments.

>I'm not a huge fan of having another config module. I'd rather it all be in `NervesHub.Config` or stay in their current `*.exs` files. Part of the issue is settings spread between multiple files and this just shifts that spread from one place to another

This is the crux of the problem IMO - once we agree on this implementation should be easy(er).

I would like to make an argument for Vapor by showing where I find mix config falls down.

I selfishly and without asking I was working towards the goal of distributing nerves hub as a docker container that you configure with environment variables. Nothing about using mix config preclude this outcome BUT they come at the cost of duplication and in non-release environments requires configuration to be set at compile time.

>they come at the cost of duplication

This one may not be entirely fair and probably says more about my skills as a developer but as an example, `DeviceEndpoint` requires key, cert and ca files - the location of these files will be different for dev, test and production (and likely they will be different for each developer). This means in each of the `config/*.exs` files you will end up with something similar to:

```elixir
keyfile = System.get_env("DEVICE_ENDPOINT_KEYFILE") || "/etc/ssl/blah-key.pem"
certfile = System.get_env("DEVICE_ENDPOINT_CERTFILE") || "/etc/ssl/blah.pem"
cacertfile = System.get_env("DEVICE_ENDPOINT_CACERTFILE") || "/etc/ssl/ca.pem"

config :nerves_hub, DeviceEndpoint, https: [
  keyfile: keyfile,
  certfile: certfile,
  cacertfile: cacertfile
]
```

With mix config there is no other way (please correct me if I'm wrong) to set per env defaults (assuming different values per env) without duplicating this configuration and redefining the default value (`"/etc/ssl/"` portion) with different values. This also inevitably leads to environment specific configuration (e.g. hardcoding `verify: :verify_peer`) because you've got free reign to override any configuration options you want.

The copy pasta that needs to happen to keep these files in sync is a complete nightmare most of the time as it is easily forgotten.

I believe vapor solves these problems with a single definition of what is accepted in `config/config.exs`, per env defaults in `.env.<env_name>` and developer specific overrides in `.env.<env_name>.local`. All the configuration is passed to the supervisor(s) at runtime which ultimately defers "what do I want the configuration to look like" to the operator when they boot the app (and I don't want to get into the confusing mess that is `config/runtime.exs` that only gets loaded when using releases only adding to the confusion).

>and in non-release environments requires configuration to be set at compile time

Assuming that `System.get_env/1` calls end up in `config/*.exs` any time you change the value (admittedly not often) parts of your application need to be recompiled which is probably "fast" but unnecessary.

>I'm not a huge fan of having another config module.

This is the downfall of my approach that someone else may have a better solution to. Vapor does not allow conditionally requiring environment variables based on a value. It only has a `required` option which takes a boolean. NervesHub supports booting up in two modes: "all"; both endpoint and device endpoint and "web|device"; only load one of the endpoints. Because of this we need to conditionally require the `DEVICE_ENDPOINT_*` environment variables and right now the only way I know how to do that is in a separate module.

>I want to think on the env var names a bit.

I don't mind what the variables names are. I tried to replicate the naming structure from the existing vapor config files. Whatever it ends up being as long as it is consistent and predictable. 

>I'd also like to make sure the old ones are still supported and not just replaced just to reduce risk for current deploys

Makes sense.

jjcarstens:
Ah, I see. I should clarify my comment - I like the vapor env approach, but was missing the link of how it was proving to be different that configs after seeing the shift to `NervesHub.Config.*`. I was not considering the conditional need based on vapor restrictions.

More thoughts to come, but I need to break for now. I do like the start of ideas though 🎉

joshk:
With our move to using `runtime.exs` for runtime configuration, lots has changed and needs to be documented.

I think we should close this for the time being and open a new PR which adds in the configurability of some of the Device endpoint bits, like the TLS versions supported.

PR #1111, opened 2023-11-24
`runtime.exs` replaces `release.exs`

joshk:
**Disclaimer**: _This PR wasn't meant to be this big, and some extra bits got included as I got this up and running on Fly.io_

This change has Nerves Hub uses the recommended config approach of `runtime.exs` for anything that might need customization for different deployments.

Bits included in this PR:
- `release.exs` removed in favor of `runtime.exs`
- change all compile time configs to runtime configs
- remove Vapor and focus on keeping everything in `runtime.exs` and environment variables
- only log expired inflight updates if any were expired
- allow for the NodeReporter to be enabled/disabled (disabled by default)
- allow for the LoadBalancer to be ignored, it isn't needed for small setups (skipped by default)
- added `dns_cluster` for easy clustering support (using a fork for now, needed for multi app deployments, eg. Fly.io)
- added support for auto db migrations on app startup (enabled by default)
- renamed `Dockerfile` to `Dockerfile.aws`
- and added a new `Dockerfile` which should work for everyone out of the box, and not bound to AWS

I have tested this all on Fly.io with two apps (web and device), each with two instances, and all clustering as expected. I'll write this up in some docs instead of adding anything cloud specific to the repo.

This new Dockerfile should also allow us to publish a public image for people to use. We can then focus on some docs on what environment variables are required to get a full setup deployed on a handful of clouds.

I'm sorry for how big this is, and I'm very sure things need further cleanup and improvements.

This PR does have some overlap with #1107 and #1090


joshk:
the error in the circle run build is a little confusing :/

joshk:
Ok, this failure is correct, because previously `release.exs` was only being used when the app was being run after a distribution had been built.

I think there is a broader question here, is everything in `release.exs`/`runtime.exs` a sane default for when starting up the app, even in dev mode?

oestrich (approved):

oestrich:
Thanks! 🎉

PR #1113, opened 2023-11-24
update footer links

joshk:
- remove the status page link, which goes nowhere
- remove the support email link, GitHub issues are the preferred means of help
- link to this repos for the issues link, and not the smartrent repo, which 404s

and these last two might be a bit presumptuous 
- remove the `nerves_hub` hexdocs link
- and remove the nerves key link

these both are very specific and are better served by using the docs.


fhunleth:
Thanks for the rebase! Trying to merge the easy ones on holiday from my phone and made an oops.

Issue #815, opened 2022-04-08
Ability to set a default org


dnrce:
After login, users are always taken to their personal org. In many (most? nearly all?) cases, particularly within companies, a user will interact more with other orgs than with their personal org. It would be helpful if a user could designate an org other than their personal org as the default, and be automatically taken to this other org after login.

joshk:
After thinking about this issue more, I think this needs to go into a bigger unit of work related to auth and orgs.

I'm going to keep this issue to track some of the wider puzzle I'm still piecing together.

PR #1115, opened 2023-11-24
Don't show the orgs and projects drop down on the home page

joshk:
This might be a bit nit picky of me, but it bugged me that on the home page, where you see your profile drop down at the top right, and your orgs in the main section, you are shown your username again in the nav bar, where the org drop down goes.

This PR hides the org drop down if an org hasn't been selected yet.

**What the nav bar looks like when you have selected an org to navigate to:**
Screenshot 2023-11-25 at 9 05 57 AM

**What the nav looks like when you are on the home page (your username is in the nav bar twice)**
Screenshot 2023-11-25 at 9 06 44 AM

**After this PR**
Screenshot 2023-11-25 at 9 05 47 AM

### DIFF
--- a/lib/nerves_hub_web/templates/layout/_navigation.html.heex
+++ b/lib/nerves_hub_web/templates/layout/_navigation.html.heex
@@ -8,7 +8,7 @@
     <%= if logged_in?(@conn) do %>
       <ul class="navbar-nav mr-auto flex-grow">
         <li class="nav-item dropdown switcher">
-          <a class="nav-link dropdown-toggle org-select arrow-primary" href="#" id="navbarDropdownMenuLink" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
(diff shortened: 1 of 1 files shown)


jjcarstens:
Not sure on this one - I like having the dropdown there so I can quickly navigate to the product I want to save a few clicks. In the end, both are incredibly minor so it'd really be fine either way

The reason why you see your username twice is because every user automatically has an org created with their username. So what you're really seeing is your org on the left and user on the right.

What might be a better fix here is to allow defaulting to a different org (see https://github.com/nerves-hub/nerves_hub_web/issues/815) rather than the user default?

joshk:
The code shows something different. If you have selected an org then it displays the org, otherwise it displays `@conn.assigns.user.username`. 

If you find this useful, then I would suggest changing it to say 'Select an org' or similar.

joshk:
Screenshot 2023-12-01 at 10 48 32 PM

Updated the PR to show "Select Org" if an Org isn't selected.

PR #1117, opened 2023-11-24
Remove the sponsors page

joshk:
It doesn't make much sense to have this page when the app is self hosted.


fhunleth:
Completely agree. Thanks for the PR.

PR #1133, opened 2023-12-07
Build and publish Docker images to GitHub Container Registry

joshk:
This PR uses GitHub Actions and Container Registry for testing and publishing Docker images.

By publishing Docker images we can reduce the complexity for setting up your own Nerves Hub.

This opens the door for people to use PaaS's to run Nerves Hub without having to clone the repo.

In fact, people can run Nerves Hub locally with a simple `docker pull ghcr.io/nerves-hub/nerves-hub` 

There is a little duplication with Circle CI as we are also running the tests here, but on the plus side, this whole test and build workflow takes half the time than Circle.

Docker images are only pushed to the Container Registry for the main branch (latest) and tags.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,138 @@
+name: Test + Docker
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  compile-and-test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 14
+
+    permissions:
+      contents: read
+      packages: read
+
+    env:
+      OTP_VERSION: "26.1.2"
+      ELIXIR_VERSION: "1.15.7"
+      FWUP_VERSION: "1.10.1"
+      MIX_ENV: "test"
+      DATABASE_URL: postgres://postgres:postgres@localhost:5432/nerves_hub_test
+
+    services:
+      db:
+        image: postgres:15
+        ports: ['5432:5432']
+        env:
+          POSTGRES_PASSWORD: postgres
+        options: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Set up Elixir
+        uses: erlef/setup-beam@v1
+        with:
+          otp-version: ${{env.OTP_VERSION}}
+          elixir-version: ${{env.ELIXIR_VERSION}}
+
+      - name: Install system deps
+        run: |
+          wget https://github.com/fwup-home/fwup/releases/download/v1.10.1/fwup_1.10.1_amd64.deb
+          sudo dpkg -i fwup_1.10.1_amd64.deb && rm fwup_1.10.1_amd64.deb
```
jjcarstens, line 45: Is it worth specifying the fwup version as a var as well?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,135 @@
+name: Test + Docker
+
+on:
+  push:
```
jjcarstens, line 4:
```suggestion
  push:
    branches:
      - main
```

Should we restrict on push a bit so it doesn't double trigger jobs on pull requests (which may have another push)

.github/workflows/ci.yml
joshk: good idea!

joshk: done!

.github/workflows/ci.yml
joshk: done!

jjcarstens (approved):

PR #1137, opened 2023-12-08
Runtime config improvements + add support for a `.env`

joshk:
Adding the [Dotenvy](https://hexdocs.pm/dotenvy/readme.html) improves the development and deployment experience.

You can specify any environment variable used in `runtime.exs` in a local (and `.gitignore`d) `.env` file, allowing for easy dev environment customizations. Or you can use this in production for configuring the production runtime.

One notable change is that I have removed https://github.com/nerves-hub/nerves_hub_web/blob/74e410e756ab9b700e51426f46e359552f6b7303/config/runtime.exs#L37-L54 as using `sys.config` should be the preference. See:
- https://github.com/slashdotdash/til/blob/master/elixir/distributed-erlang-sync-nodes.md
- https://www.erlang.org/doc/man/config.html#sys.config

This can be added to the release before starting the application.


joshk:
Closing this for the time being

Issue #1136, opened 2023-12-08
Device Token Auth


joshk:
**A little background first...**

Devices talk to Hub using SSL client certs. This requires Hub to have a Devices Endpoint which does its own TLS termination so it has access to the client cert. The by product of this is that running Hub on a PaaS is not possible for 90% of the PaaS's, and of the one PaaS that I have personally got it running on (Fly.io) I needed to run two apps (one for web, one for devices), and I had to purchase an SSL cert for the endpoint to use. 

**Ok, it's a little annoying, so what?**

I believe we can make the first time experience simpler, easier, and more approachable for the community. And much of this comes down to simplifying how devices connect to Hub. If we allow for devices to use tokens via HTTP headers then we can add device connectivity support to the web endpoint. This would allow people to deploy to a PaaS, use an SSL cert generated by the PaaS, and only need to run one app for everything.

**Ahhh, ok, tell me more!**

The outcome of this change would be a Nerves Hub that anyone can run from behind a load balancer using 90% of PaaS's, from Fly.io, to Heroku, to Render, to `[insert name of your favourite PaaS]`.

**What about security?**

Using client certificates is the industry standard, and this would be the recommendation when running Nerves Hub for larger deployments. For home use, R&D, and prototyping, a simpler token auth is a great place to start.

**What apps are affected by this change?**

NervesHub, NervesHub Link, and NervesHub CLI

**What are the next steps?**

- Decide on the taste and shape of the PSK/PKIs we use
- Decide on how this is passed to NervesHub (headers? and if so, what do we call the headers?)
- And decide on how this is configured in NervesHub.

jjcarstens:
Documenting some offline discussion:

# Idea 1 - Access ID + Secret

Sometimes referred to as Pre-shared Identity (PSI) & Pre-shared Key (PSK)

Core idea is that a PSI and PSK are generated for each device which is supposed to be saved to device "securely" (This is up to the user, but typically just manually including in the firmware).

* PSK is not to be transmitted
* PSI is included with the request. NervesHub will use the PSI to lookup the corresponding PSK
* NervesHub calculates the expected salt based on specific headers and then verifies it with the found PSK
* The verified signature will result in the device identifier being decoded

<details><summary>Example signature calculation</summary>
<p>

```elixir
##
# Generated from NervesHub
access_id = "nhd_123456789abcdefghijklmnop"
secret = "I-am-super-secret!-don't-share!!"

key_digest = :sha256
[...]

PR #1139, opened 2023-12-10
Device Shared Secret Authentication

joshk:
This adds support for Shared Secret authentication, which allows Devices to connect to Nerves Hub using a shared key and secret. This is similar to the AWS HMAC signatures used for S3. 

This is based on #1136 

### Some key highlights:

- If you use the Shared Secrets strategy, you won't need to run the Device Endpoint, and thus you won't need to setup custom SSL certs.
- Which also means you can run this entirely behind a load balancer and let it take care of TLS termination.
- Which makes this a lot easier to run on a PaaS. (eg. Heroku, Fly.io, Render)
- When a Device connects for the first time the Device will be registered with the Product ("Just-in-Time registration").

This feature is opt-in, on the server level, using the environment variable `DEVICE_SHARED_SECRET_AUTH`.

### Some tech bits

I've added a series of useful mounts which are used by a `live_session`.  As we expand LiveView usage we can reuse a lot of this structure.

And I've added a centralized permissions checking module, abstracting away from checking for `:admin`, `:manage`, and `:view`.

### Not included in this PR, but to be added in a future PR...

- When setting up a Product, you have to select if the it will use Cert auth or Shared Secret auth.

### Screenshots

_Updated Product Settings page_ (now using LiveView)
Screenshot 2023-12-15 at 9 43 26 PM

_One active Shared Secret_
Screenshot 2023-12-15 at 9 44 11 PM

_Two active Shared Secrets, one deactivated_
Screenshot 2023-12-15 at 9 44 40 PM

_Feature not enabled for the server_
Screenshot 2023-12-15 at 10 12 28 PM


lib/nerves_hub_web/channels/device_socket_token_auth.ex
```diff
@@ -0,0 +1,111 @@
...
+         {:ok, device} <-
+           Devices.get_or_create_device(token_auth: token_auth, identifier: identifier) do
+      socket =
+        socket
+        |> assign(:device, device)
+        |> assign(:reference_id, generate_reference_id())
+
+      {:ok, socket}
+    else
+      _ -> {:error, :invalud_auth}
```
jjcarstens, line 41: I'd probably just leave this out and let any non-match close the socket as is rather wrap in a generic error.

Or, at the very least, change to `invalid_auth` 😉

lib/nerves_hub_web/channels/device_socket_token_auth.ex
```diff
@@ -0,0 +1,111 @@
...
+      token_auth -> {:ok, token_auth}
+    end
+  end
+
+  defp generate_reference_id() do
+    Base.encode32(:crypto.strong_rand_bytes(2), padding: false)
+  end
+
+  def handle_error(conn, :invalud_auth),
+    do: Plug.Conn.send_resp(conn, 403, "Invalid authorization")
```
jjcarstens, line 110: Do we need a custom error handler? I believe it will 403 by default.

If we keep, do we need to handle the other case of when a non-matching `connect/3` returns only `:error`?

lib/nerves_hub_web/channels/device_socket_token_auth.ex
```diff
@@ -0,0 +1,111 @@
...
+          {:signature, v}
+
+        _ ->
+          # TODO: What should we do with this unknown?
+          nil
+      end
+    end
+  end
+
+  defp verify_options(parsed_data) do
```
jjcarstens, line 79:
```suggestion
  defp verification_opts(parsed_data) do
```
Just a nit, but `verify_opts` makes me think we're verifying the options are okay before we use them

lib/nerves_hub_web/channels/device_socket_token_auth.ex
```diff
@@ -0,0 +1,111 @@
...
+  @default_max_age 900
+
+  def connect(_params, socket, %{x_headers: headers}) do
+    parsed_data = parse_headers(headers)
+    verify_opts = verify_options(parsed_data)
+    salt = expected_salt(headers)
+
+    with {:ok, true} <- Application.fetch_env(:nerves_hub, __MODULE__)[:enabled],
+         {:ok, access_id} <- Keyword.fetch(parsed_data, :access_id),
+         {:ok, token_auth} <- get_product_token_auth(access_id),
```
jjcarstens, line 29: Just a note - We'll need to split this up based on the incoming ID to decide if we're looking up a product or device token

lib/nerves_hub/devices.ex
```diff
@@ -27,8 +28,24 @@ defmodule NervesHub.Devices do
...
-  def get_device!(device_id), do: Repo.get!(Device, device_id)
+  def get_device!(device_id) do
+    Repo.get!(Device, device_id)
+  end
+
+  def get_device(device_id) when is_integer(device_id) do
+    Repo.get(Device, device_id)
+  end
+
+  def get_device(product_id: product_id, identifier: identifier) do
```
jjcarstens, line 39: If we want to use keyword lists, we could extend `Repo.get_by` a bit here instead:

```elixir
def get_by(clauses, opts \\ []) do
  case Repo.get_by(Device, clauses, opts) do
    nil -> {:error, :not_found}
    device -> {:ok, device}
  end
end
```

lib/nerves_hub/devices.ex
```diff
@@ -223,6 +240,22 @@ defmodule NervesHub.Devices do
     end
   end
 
+  def get_or_create_device(token_auth: token_auth, identifier: identifier) do
```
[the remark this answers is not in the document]
joshk: Urg, I didn't know that, that sucks. 

Should I change it to `get_or_create_device_by_token_auth_and_identifier(token_auth, identifier)`? (which is ugly and brittle, imo)

lib/nerves_hub_web/channels/device_socket_token_auth.ex
joshk: I thought a `with` non match would cause an exception to be raised, and thus would likely cause this to be become a 500 error? (I haven't dived into it to check)

(just dived in, it isn't clear from the docs)

joshk: I'll at least fix the spelling:D

lib/nerves_hub_web/channels/device_socket_token_auth.ex
joshk: maybe we can do some quick testing and confirm what error code and message is returned by default, then decide what we need to do?

lib/nerves_hub_web/channels/device_socket_token_auth.ex
joshk: thats what I wanted to talk over with you today

lib/nerves_hub_web/channels/device_socket_token_auth.ex
joshk: I can confirm that the default error code is 403, with an empty body.

https://github.com/phoenixframework/phoenix/blob/a66e7c33019a3d564c4f7f2e48b77efb63c54ad7/lib/phoenix/transports/websocket.ex#L26

https://github.com/phoenixframework/phoenix/blob/a66e7c33019a3d564c4f7f2e48b77efb63c54ad7/lib/phoenix/transports/websocket.ex#L80

joshk: I could ditch this error handler and we just stick with the default (403 error code but no description)

lib/nerves_hub/devices.ex
joshk: thanks for the feedback on this, fix incoming

config/runtime.exs
```diff
@@ -150,6 +150,9 @@ if config_env() == :prod do
         signing_salt: System.fetch_env!("LIVE_VIEW_SIGNING_SALT")
       ],
       server: true
+
+    config :nerves_hub, NervesHubWeb.DeviceSocketTokenAuth,
```
jjcarstens, line 154:
```suggestion
    config :nerves_hub, NervesHubWeb.DeviceSocketSharedSecretAuth,
```

fhunleth:
All least from my level of review, this looks great, and I think it's really going to help people get started. Thank you so much for all of the effort to push this feature through.

joshk:
> All least from my level of review, this looks great, and I think it's really going to help people get started. Thank you so much for all of the effort to push this feature through.

Thanks @fhunleth, its been my pleasure to get involved. More to come soon!

config/runtime.exs
```diff
@@ -256,3 +259,12 @@ config :nerves_hub, :audit_logs,
   max_records_per_run:
     String.to_integer(System.get_env("TRUNCATE_AUDIT_LOGS_MAX_RECORDS_PER_RUN", "10000")),
   days_kept: String.to_integer(System.get_env("TRUNCATE_AUDIT_LOGS_MAX_DAYS_KEPT", "30"))
+
+if System.get_env("OTEL_ENABLED", "false") == "true" do
```
oestrich, line 263: I think this is wrong for being in runtime, but I also merged removing otel entirely so we can get rid of these, haha.

lib/nerves_hub_web/helpers/authorization.ex
```diff
@@ -0,0 +1,17 @@
+defmodule NervesHub.Helpers.Authorization do
+  alias NervesHub.Accounts.OrgUser
+  alias NervesHub.Accounts.User
+
+  def authorized!(org_user, permission) do
+    true = authorized?(org_user, permission)
```
jjcarstens, line 6: Should we raise a more helpful error here? This will raise `MatchError`

jjcarstens (approved):

PR #1144, opened 2023-12-15
Update earmark to fix a compile issue

joshk:
A dependabot merge broke the build with an earmark compile issue, this should fix it.


joshk:
This isn't directly needed, it seems the issue with the broken build was related to a cache that needed to refreshed.

jjcarstens (approved):

PR #1146, opened 2023-12-18
Load the device key/cert via ENV or existing file path

oestrich:
Check to make sure the files exist before allowing to run.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -156,11 +156,35 @@ if config_env() == :prod do
     host = System.get_env("DEVICE_HOST") || System.fetch_env!("WEB_HOST")
     https_port = String.to_integer(System.get_env("DEVICE_PORT", "443"))
 
-    ssl_key = System.fetch_env!("DEVICE_SSL_KEY") |> Base.decode64!()
-    :ok = File.write("/app/tmp/ssl_key.crt", ssl_key)
+    keyfile =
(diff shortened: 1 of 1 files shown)


config/runtime.exs
```diff
@@ -156,11 +156,35 @@ if config_env() == :prod do
...
 
-    ssl_key = System.fetch_env!("DEVICE_SSL_KEY") |> Base.decode64!()
-    :ok = File.write("/app/tmp/ssl_key.crt", ssl_key)
+    keyfile =
+      if System.get_env("DEVICE_SSL_KEY") do
+        ssl_key = System.fetch_env!("DEVICE_SSL_KEY") |> Base.decode64!()
+        :ok = File.write("/app/tmp/ssl_key.crt", ssl_key)
+        "/app/tmp/ssl_key.crt"
+      else
+        ssl_keyfile = System.fetch_env!("DEVICE_SSL_KEYFILE")
```
jjcarstens, line 165:
```suggestion
        ssl_keyfile = System.fetch_env!("DEVICE_SSL_KEYFILE", "/etc/ssl/#{host}-key.pem")
```
Do we want to keep the old default here for the most backwards compatibility?

oestrich: I think that makes sense. I'll update both and push up

jjcarstens (approved):

PR #1149, opened 2023-12-19
Don't require S3 access key id or secret access key

oestrich:
Only load if provided, ECS uses role based permissions and this won't be set in that deployment environment.


joshk:
@oestrich is there a way to detect if this has been configured by ECS? I think it would be good to fail fast if it hasn't been setup correctly.

oestrich:
At the moment I'm trying to get this booted on our environment, so I guess we could set something in the env to skip this once it's booting again.

PR #1151, opened 2023-12-19
A `libcluster_postgres` discussion

joshk:
In a previous PR I added `dns_cluster` to help set up clustering, but this discovery mechanism won't work for everyone as it relies on dns node discoverability.

This PR swaps in `libcluster_postgres`, using the postgres db for node discoverability, which encapsulates the node discover and connect logic into something that should work for 99% of people.

This is for discussion.


joshk:
@jjcarstens @oestrich @fhunleth i think this is a better set up when compared to dns cluster. This is a lot more flexible.

jjcarstens (commented):
I played with this locally and I think it's a solid option. Looks like there are no extra DB table needed and just relies on built-ins which have been around Postgres for quite a few years. If a notification is missed, there is a heartbeat that would catch the node later on.

Are we required to parse the DB URL or can we simply pass that to the strategy?

joshk:
If we want to use a URL then we have to parse it, which is a little annoying.

joshk:
Also, I’m using my fork of the strategy as I needed to add support for ssl configs

oestrich (approved):
I have another PR to merge to remove the old clustering once this is merged

PR #1154, opened 2023-12-20
Wrap the SocketDrano child in a list

oestrich:
Needs to be a list for appending to the list of application children


jjcarstens (approved):

joshk:
ah, great catch, thank you

PR #1156, opened 2023-12-20
Remove inet6 option from the DeviceEndpoint

oestrich:
This doesn't appear to be an option in ThousandIsland and it's messing up starting the DeviceEndpoint because this needs to be a keyword list.

Update: it is an option deeper down in gen_tcp and ssl but something in Bandit is treating it as a keyword list still


jjcarstens (approved):

jjcarstens:
For posterity, fixed in bandit 1.1.2 - https://github.com/mtrudel/bandit/pull/277

joshk:
now that this is fixed in bandit, did we want to add it back?

PR #1158, opened 2023-12-21
Prune Oban jobs older than a week 

### LINKED ISSUE #776: Suggestion to use Oban.Plugins.Pruner
Currently rows of oban_jobs table increases monotonically, so I suggest to use [Oban.Plugins.Pruner](https://hexdocs.pm/oban/Oban.Plugins.Pruner.html).

This plugin does

> Periodically delete completed, cancelled and discarded jobs based on age.

I can make a simple PR for this, but before that I would like to ask the development team what the duration of the max_age option should be.

joshk:
resolves #776

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -36,6 +36,7 @@ config :nerves_hub, Oban,
   log: false,
   queues: [delete_firmware: 1, firmware_delta_builder: 2, truncate: 1],
   plugins: [
+    {Oban.Plugins.Pruner, max_age: 3_600},
     {Oban.Plugins.Cron,
      crontab: [
        {"0 * * * *", NervesHub.Workers.TruncateAuditLogs, max_attempts: 1},
--- a/config/config.exs
+++ b/config/config.exs
@@ -36,7 +36,8 @@ config :nerves_hub, Oban,
   log: false,
   queues: [delete_firmware: 1, firmware_delta_builder: 2, truncate: 1],
   plugins: [
-    {Oban.Plugins.Pruner, max_age: 604_800}, # 1 week 
+    # 1 week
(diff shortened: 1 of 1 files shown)


config/config.exs
```diff
@@ -36,6 +36,7 @@ config :nerves_hub, Oban,
   log: false,
   queues: [delete_firmware: 1, firmware_delta_builder: 2, truncate: 1],
   plugins: [
+    {Oban.Plugins.Pruner, max_age: 3_600},
```
jjcarstens, line 39: And hour seems aggressive? I can see a case where I want to see some of the jobs history for longer than an hour. Thoughts?

joshk:
The default is 1 min.

What about 24 hours? 3 days?

You pick.

jjcarstens:
O geez. lol. How about 7 days? We don't use oban a whole lot and that gives introspection time to start with

config/config.exs
joshk, line 39:
```suggestion
    {Oban.Plugins.Pruner, max_age: 604_800}, # 1 week 
```

joshk:
@jjcarstens  review/approval please

jjcarstens (approved):

config/config.exs
```diff
@@ -36,6 +36,7 @@ config :nerves_hub, Oban,
   log: false,
   queues: [delete_firmware: 1, firmware_delta_builder: 2, truncate: 1],
   plugins: [
+    {Oban.Plugins.Pruner, max_age: 604_800}, # 1 week
```
joshk, line 39:
```suggestion
    # 1 week
    {Oban.Plugins.Pruner, max_age: 604_800},
```

PR #1160, opened 2023-12-21
Device link fixes

jjcarstens:
* Remove `DeviceLink.disconnect/1` from channel terminate
  * The device channel is monitored by DeviceLink. The extra call to `DeviceLink.disconnect/1` would sometimes timeout causing bigger issues with restarts elsewhere or race getting to the link process potentially producing DOWN messages after the monitor was cleared from link causing unmatched `handle_info` callbacks to be evaluated 
* Ignore messages from unknown socket in DeviceLink
  * TCP sockets have longer timeouts. There is a chance the old socket was still around when the new one started which could result in getting this message later than expected. For cases like that and when we no longer know the ref, simply ignore

### DIFF
--- a/lib/nerves_hub/devices/device_link.ex
+++ b/lib/nerves_hub/devices/device_link.ex
@@ -358,11 +358,17 @@ defmodule NervesHub.Devices.DeviceLink do
   end
 
   @impl GenServer
-  def handle_info(
-        {:DOWN, transport_ref, :process, _pid, _reason},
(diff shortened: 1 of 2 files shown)


fhunleth:
It looks like the change to ignore unknown messages didn't make it.

fhunleth:
Ok, I understand. Nevermind.

fhunleth (approved):

PR #1162, opened 2023-12-22
Ignore unhandled messages in DeviceLink

jjcarstens:
Crashing DeviceLink could cause cascading problems. To fix that, this change ignores unhandled messages in `handle_info` and logs it


fhunleth (approved):

oestrich (approved):

joshk (approved):
It might be worth also sending the message to Sentry, but that can always come later.

jjcarstens:
Tested with and without Sentry configured. As long as everyone is happy with the message, then we're 👍🏼

joshk (approved):
Awesome!

PR #1165, opened 2023-12-24
Config cleanup

joshk:
This supersedes #1137 

- Removes unused env files
- Adds basic docs to config files
- Removes unused config items
- Reorder some config items for clarity
- allow for `inet_dist_listen_min` and `inet_dist_listen_max` config to be opt out


config/prod.exs
```diff
@@ -6,20 +6,24 @@ config :logger, level: :info, backends: [:console, Sentry.LoggerBackend]
...
 
 ##
-# NervesHubWWW
+# Database and Oban
 #
-config :nerves_hub, NervesHubWeb.Endpoint,
-  server: true,
-  force_ssl: [rewrite_on: [:x_forwarded_proto]]
+config :nerves_hub, NervesHub.Repo,
+  ssl: true,
```
jjcarstens, line 24: This was not previously defaulted in `prod.exs`. We should probably leave it be and handle in runtime.exs

config/runtime.exs
```diff
@@ -246,35 +181,125 @@ if config_env() == :prod do
   databse_socket_options = if System.get_env("DATABASE_INET6") == "true", do: [:inet6], else: []
 
   config :nerves_hub, NervesHub.Repo,
-    ssl: System.get_env("DATABASE_SSL", "true") == "true",
-    ssl_opts: database_ssl_opts,
     url: System.fetch_env!("DATABASE_URL"),
-    pool_size: String.to_integer(System.get_env("DATABASE_POOL_SIZE", "20")),
```
jjcarstens, line 252: Looks like `DATABASE_POOL_SIZE` and `DATABASE_SSL` are now missing which we may still want to be configurable

joshk: I found out that these can be set via the database URL, I can add them back if you prefer, I thought it was good to remove unneeded configs

config/prod.exs
joshk: I'm torn, its a sensible default, but i'm fine with changing this back

jjcarstens (approved):
After chatting a bit we decided to move the `inet_dist_listen_*` changes to their own PR as it was the only behavior being altered. The rest is reorganization and cleanup

PR #1171, opened 2024-01-04
Migration fixes

jjcarstens:
### Adjust auto migrations to Ecto recommendations
`Ecto.Migrator` is setup to be able to run as part of the application start up, so switch to that for cases when auto migration is preferred

This removes the need for `NervesHub.Release.DatabaseAutoMigrator`

See https://hexdocs.pm/ecto_sql/3.11.1/Ecto.Migrator.html#module-example-running-migrations-on-application-startup

### Fix release migrations
Previously, release migrations were attempting to be handled in `Enum.each` without any return checks. This means failures could be returned but were unhandled and false positive of success could be relayed.

Instead, this updates to match the Ecto and Phoenix recommendations for dealing with migrations in a release. See also:

* https://hexdocs.pm/ecto_sql/3.11.1/Ecto.Migrator.html#module-example-running-migrations-in-a-release
* https://hexdocs.pm/phoenix/releases.html#ecto-migrations-and-custom-commands


oestrich (approved):

joshk:
Oh, this is cool, thanks for this change.

PR #1172, opened 2024-01-05
Remove SYNC_NODES_OPTIONAL from runtime config

oestrich:
And remove the reboot_system_after_config option since we aren't adjusting the kernel config.

To be merged along with https://github.com/nerves-hub/nerves_hub_web/pull/1151


jjcarstens (approved):

joshk:
You also removed the ports, is that intended?

jjcarstens:
Yep. Going off some of our other discussions of making the distribution specific configuration up to the user with their own env.sh/vm.args

PR #1173, opened 2024-01-07
Add `Devices.SharedSecretAuth`

jjcarstens:
Next part of #1136

Adds support for individual device secret auth. These can also be associated with a production shared secret auth to track when a device auth is created from the product auth (coming next). This will be helpful in the future to easily deactive all device tokens associated with a deactivated production token


joshk (approved):
This.
Is.
Awesome.
!.

PR #1176, opened 2024-01-09
Shared Secret Auth

joshk:
This PR is based upon the ideas in https://github.com/nerves-hub/nerves_hub_web/issues/1136

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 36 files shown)


.github/workflows/ci.yml
jjcarstens, line 1: I think we won't need this anymore, right?

joshk: fixed!

jjcarstens (approved):

PR #1187, opened 2024-01-16
Changes to how DeviceLinks are supervised

oestrich:
- Partition supervisor for the DynamicSupervisor, with thousands of devices connected the single DynamicSupervisor could get overloaded, partition it to help reduce processing changing state quickly
- Start a DeviceLink with the ID only, to help reduce state inside the supervisor, only store the device's ID in the supervisor state (since it's part of the child_spec that's stored in the supervisor state)
- Remove the supervisor module to simplify, put the one function that matters in the DeviceLink module directly
- Set to transient for restart since we do want the DeviceLink to terminate and get cleared out of the child_spec state


test/nerves_hub/devices/device_link_test.exs
```diff
@@ -10,18 +10,6 @@ defmodule NervesHub.DeviceLinkTest do
...
-    assert {:noreply, updated} = DeviceLink.handle_continue(:boot, state)
-    assert updated.deployment_channel == "deployment:none"
-  end
-
-  test "device with deployment subscribes deployment:\#{id}" do
-    state = %DeviceLink.State{device: %Device{id: 1, deployment_id: 1}}
-    assert {:noreply, updated} = DeviceLink.handle_continue(:boot, state)
-    assert updated.deployment_channel == "deployment:1"
-  end
-
```
jjcarstens, line 24: Why did these get removed?

oestrich: It was going to require enough fiddling with test startup and the tests didn't seem worthwhile enough to make work with the new boot function. I can add them back if we think it's worth it

lib/nerves_hub/devices/device_link.ex
```diff
@@ -6,7 +6,7 @@ defmodule NervesHub.Devices.DeviceLink do
   e.g. websockets, MQTT, etc
   """
 
-  use GenServer
+  use GenServer, restart: :transient
```
jjcarstens, line 9: I don't know about this? It kind of makes sense from the context of sockets, but future backends (like mQTT) would suffer from this.

oestrich: Even for mqtt we want to be able to stop these device links permanently right? Transient will restart everything except for when we gracefully shutdown in the force disconnect or from the monitor

test/nerves_hub/devices/device_link_test.exs
oestrich: Yeah, the issue was create device makes a deployment so you'd have to undo that before calling handle_continue. Just enough friction to make me consider if we needed the tests. I'll add them back

lib/nerves_hub/devices/device_link.ex
jjcarstens: 🤔 For MQTT, an event has to start/stop this link. If the process crashes and goes away even though the device still has that MQTT connection, this link will just be missing and there is nothing that would bring it back up

With this new setup, the child_spec data is just the id of the device which is way less than before, so maybe not as much of a problem.

On the inverse, there is logic that will attempt to create this link from the socket and might be racing the supervisor recreating it 🤔 

Probably worth trying, but maybe we leave a note/TODO for ourselves here that we'll need to consider this for other transports besides web sockets?

oestrich: I can leave a note for sure. Also not sure what DynamicSupervisor does for `{:stop, :normal, state}` with `:permanent` restart anyways. But maybe that was part of the overloaded supervisor :shrug:

jjcarstens (approved):

PR #1191, opened 2024-01-19
Add a custom device transport for ThousandIsland

oestrich:
Delegates to the SSL transport from ThousandIsland but adds rate limiting before the handshake. To forcefully close before SSL is allowed to start doing CPU intensive tasks.

### DIFF
--- a/config/dev.exs
+++ b/config/dev.exs
@@ -44,6 +44,7 @@ config :nerves_hub, NervesHubWeb.DeviceEndpoint,
     port: 4001,
     otp_app: :nerves_hub,
     thousand_island_options: [
+      transport_module: NervesHub.DeviceSSLTransport,
       transport_options: [
         # Enable client SSL
         # Older versions of OTP 25 may break using using devices
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -106,6 +106,7 @@ if config_env() == :prod do
         port: https_port,
         otp_app: :nerves_hub,
         thousand_island_options: [
+          transport_module: NervesHub.DeviceSSLTransport,
           transport_options: [
             # Enable client SSL
             # Older versions of OTP 25 may break using using devices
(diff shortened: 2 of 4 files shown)


lib/nerves_hub/device_transport.ex
```diff
@@ -0,0 +1,69 @@
+defmodule NervesHub.DeviceTransport do
+  @moduledoc false
```
jjcarstens, line 2:
```suggestion
  @moduledoc """
  SSL transport for device certificate authentication
  
  The main purpose for this transport is to allow us to rate limit incoming socket connections
  early on to prevent memory issues when sockets come thundering in. When an SSL
  handshake is started with a device, that process (`ssl_gen_statem`) starts accumulating
  state including certificates and other data. Previously, we were rate limiting at that layer
  (in `NervesHub.SSL`) which was too late. Even though we were preventing the connect,
  this SSL socket would have built up state that was not garbage collected. Replicated by
  N sockets attempting to connect and this can eventually cause memory issues.
  
  See `handshake/1` for the main change. All other function are delegated back to
  `ThousandIsland.Transports.SSL`
```

lib/nerves_hub/device_transport.ex
```diff
@@ -0,0 +1,69 @@
+defmodule NervesHub.DeviceTransport do
```
jjcarstens, line 1:
```suggestion
defmodule NervesHub.DeviceSSLTransport do
```
Maybe we want SSL in there somewhere? Im good either way

joshk: +1 on adding `SSL`

joshk:
Awesome work and find

lib/nerves_hub/device_transport.ex
oestrich: the main reason i wanted this here was to limit wasted cpu on SSL, i'll add docs to reflect that

jjcarstens (approved):

PR #1195, opened 2024-01-23
Configure Phoenix socket drainer for devices

oestrich:
https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-common-configuration

This should shed devices in a more controlled manner than dropping all N sockets at the same time when a SIGTERM is received.


jjcarstens (approved):
LGTM - If I'm reading right, this says we have 30sec to drain 15,000 sockets at 500/sec

oestrich:
Yeah, which is a lot per node currently

PR #1204, opened 2024-02-01
Remove DeviceLink

oestrich:
We're going to go back to the drawing board for this process, so move everything back into the DeviceChannel. Removes the supervisor for DeviceLinks as well since it's not needed.


jjcarstens (approved):
😢 Fret not friends - This is all being reworked.

PR #1212, opened 2024-02-10
Revert to `:cowboy` as the webserver

jjcarstens:
While using Bandit, it was discovered that a herd of _disconnects_ could cause catastrophic failures leading to node crashes (in our testing, these were decently sized 4-core, 8GB instances with 5-6k devices connected each). When testing again with Cowboy, the problem was resolved.

This change reverts to use cowboy to allow progression of other fixes and features while Bandit/Thousand Island is investigated to find the root cause

<details>
<summary>Long Version</summary>
<p>
Based on errors we'd see and various server graphs, the assumption was that NervesHub was struggling to handle "the thundering herd" of sockets as the node would go down dropping thousands of connections and those quickly turning around trying to reconnect.

We created test scenarios based on this assumption. Fixes were along the lines of:

* More variable reconnect logic on devices via `:nerves_hub_link`
* Rate limiting incoming socket connections
* Reducing memory footprint of device socket state
* Increasing the number of devices nodes to handle connections
  * increasing memory and vCPU allocated as well
*  Manually managing a deploy to more closely match a blue/green deployment
* Creating a fully matching test environment with a large number of fake "device" sockets to mimic connections
* Stepping through small number of commits at a time from our last known good deployment until seeing the behavior

We've found our ability to handle the herd in the test environment was now highly optimized based on these adjustments. While stepping through commits, we got all the way to the latest on main with success in the test environment.

Then deploying to production would completely crash and burn 🔥 🚒 Specifically, all systems would look good (CPU, memory, device connection count, etc) _**except**_ the Erlang `run queue`. This would sky rocket and eventually become the sign of death for a node once we saw lines sustain over time.

In the latest round of effort, we isolated commits even further focusing the smallest pieces to code we deemed potentially problematic. One recent change was to use Bandit/Thousand Island. We isolated, deployed, and the test environment handled the herd very well.

Somewhat dejected, we decided to break for bit and tear down our test environment while we evaluate. That's when we noticed "the run queues of death" ¬

---
**Bandit with abrupt disconnects**

![bandit_disconnects](https://github.com/nerves-hub/nerves_hub_web/assets/11321326/4be21e59-8819-4a2d-a162-80d4d9872738)

---

The main thing to note in the image above is that all of the test sockets were completely off for the entirety of this time frame. That raised some questions:

* Why does `Successful Connections` still show anything?
* What is actually being restricted as indicated by `Device Connection Rate Limiting`??
* How does `NervesHub Online Devices` keep fluctuating up and down???
  * In fact, why isn't that just 0????
* What is blowing up the run queues????

We got onto a couple nodes and started inspecting the system. All the elixir processes were "frozen" - So much so, that the processes were unable to even increment their uptime as displayed by `htop`. But not frozen enough to fail our health checks or prevent other work from coming in. Eventually, the node would crash and the new node could get into the same problem. The only noticeable problem was the run queue. But why?

The next step was to revert the addition of Bandit and use the usual Cowboy+Ranch webserver combo. The results looked as expected and servers handled connects/disconnects at high rates like a champ

---
**Cowboy with abrupt disconnects**

![cowboy_disconnects](https://github.com/nerves-hub/nerves_hub_web/assets/11321326/cbc3c430-81d9-47ea-b5bf-644e06125a57)

---

In these graphs, we started all sockets at once and there is a nice gradual climb to connectivity. Then the same test killing them all at once with typical cliff of disconnects, ecto queries (for some audit events being logged), and balanced run queues that recover very quickly.

The cause is not completely known, but we have some ideas now about why we saw such drastically different behavior with our production devices vs our fake herd. The main thing we noticed was there is a continual flow of 60-80 connects and disconnects every 10 seconds in production. This could be because many devices are on LTE which can be inconsistent, or we need to optimize sockets more. The assumption is that if disconnects with Bandit are indeed the problem, this constant rate of disconnects coupled with mass disconnects as nodes go down seem to push things right over the edge.

The PR to revert to Cowboy is an unfortunate one as we need to take a break from this problem area to make advancements that have been delayed because of it. We hope to help contribute in finding the root cause and a fix in the future
</p>
</details>


oestrich (approved):
😭 :shipit:

joshk (approved):
a smaller set of changes than i was expecting

nice work!

mtrudel:
So this is the first of what I suspect will be a long run of PRs, about how the unspecified parts of bandit compare to the unspecified parts of cowboy. I strongly suspect (though please give me the day) that this is some particular bandit pathology that's just different from a cowboy one. The good part is that bandit is really simple, so whatever's wrong here can likely get fixed forthwith. 

Lemme dig into this and report back.

mtrudel:
A couple of questions to help diagnose this:

* These are websocket connections, correct? I assume that the majority of the clients are hitting the same socket? Which one is it?
* Do the two screenshots above represent the same set of actions? I'm a bit confused as to how the Bandit screenshot starts with a bunch of connections already established and then slowly decreases, whereas the shape of the cowboy graphs clearly show the various stages at play. I'm not certain based on the info above whether this is an issue with conneciton startup, connection maintenance, or connection shutdown
* How are you inducing the mass disconnection?

jjcarstens:
* Yes, web socket connections are going through the [`NervesHubWeb.DeviceSocketCertAuth` via the `device` channel](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/channels/device_socket_cert_auth.ex#L10)
* Sorry about the screenshot. Yes, both represent the same set of actions. And the first image had almost identical climb to connectivity. I'll see if I can get another image of that graph with more time to show that climb. For our observation it seems to be connection shutdown as startup and maintenance look really similar
* We have a little app that launches an Elixir runtime and makes all the connections under a dynamic supervisor. To mass disconnect, we just kill that runtime entirely (or totally stop the node it is running on) - @oestrich can work on making that public incase you'd like to use for testing

mtrudel:
> We have a little app that launches an Elixir runtime and makes all the connections under a dynamic supervisor. To mass disconnect, we just kill that runtime entirely (or totally stop the node it is running on) - @oestrich can work on making that public incase you'd like to use for testing

The dynamic supervisor supervises the clients? Like, the shutdowns are all initiated from the client-side?

jjcarstens:
Yes, the shutdowns are initiated by the client-side to reproduce the behavior. Usually ungracefully (either `Supervisor.terminate_child` or just killing the whole runtime).

The clients are [`NervesHubLink.Socket`s](https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/socket.ex) which are using [`:slipstream`](https://hex.pm/packages/slipstream) as the websocket client and handles the underlying reconnect logic when connection is lost.

In practice, these device clients are usually shutdown due to internet connectivity being lost, or manually calling `NervesHubLink.reconnect()` because the device changed from one interface to a better one (i.e. switched `LTE` -> `Ethernet`), so we bump the connection to ensure the routing is going through the better interface

mtrudel:
It's starting to seem to me as if Bandit's websocket connections aren't being shut down when you think they are. Depending on how the shutdown is initiated from the client, I'm wondering if what's actually happening is the server's connection process is sticking around, and only shutting down later via timeout.

mtrudel:
> Sorry about the screenshot. Yes, both represent the same set of actions. And the first image had almost identical climb to connectivity. I'll see if I can get another image of that graph with more time to show that climb. For our observation it seems to be connection shutdown as startup and maintenance look really similar

That would actually be great to see. I'm narrowing in on the massive size of the run queue on Bandit & I'd love to see how it tracks through all the lifecycle of your tests

jjcarstens:
<details><summary>Updated image of connect, maintain, and disconnect using Bandit adapter</summary>
<p>

image


</p>
</details>

mtrudel:
Hopefully solved by https://github.com/mtrudel/thousand_island/pull/110

PR #1213, opened 2024-02-10
Socket rate limiting with cowboy/ranch

jjcarstens:
With Bandit, we were able to rate limit at the socket handshake which is much more efficient. This is an attempt to do so with `ranch` and `cowboy`. However, there are currently limitations such as how [cowboy has a hard match on {:ok, socket} = ranch:handshake(Ref)](https://github.com/ninenines/cowboy/blob/2.10.0/src/cowboy_tls.erl#L43)

Opening this as draft for now in case we want to continue investigating and finding a different spot we can potentially limit at


jjcarstens:
No longer needed

PR #1214, opened 2024-02-12
Allow overriding the cacertfile

oestrich:
CAStore is good for most things but existing production environments rely on a custom ca.pem for now.


joshk (approved):
nice!

PR #1215, opened 2024-02-12
Add config to rate limit any more devices from connecting

oestrich:
For use during deployments as a stop gap for preventing devices from reconnecting to nodes that are going to be torn down still.

### DIFF
--- a/lib/nerves_hub/rate_limit.ex
+++ b/lib/nerves_hub/rate_limit.ex
@@ -12,15 +12,19 @@ defmodule NervesHub.RateLimit do
   """
   @spec increment() :: boolean()
   def increment() do
-    bucket_key =
-      DateTime.utc_now()
-      |> DateTime.to_unix()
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/rate_limit.ex
```diff
@@ -12,15 +12,19 @@ defmodule NervesHub.RateLimit do
   """
   @spec increment() :: boolean()
   def increment() do
-    bucket_key =
-      DateTime.utc_now()
-      |> DateTime.to_unix()
+    if Application.get_env(:nerves_hub, :allow_devices, true) do
```
jjcarstens, line 15: In looking this over, it looks like this might actually be overkill. Since this is a manual action for a specific situation, we could simple update to `config :nerves_hub, NervesHub.RateLimit, limit: -1` and that would effectively stop all traffic

oestrich: Hmm excellent point, if we’re forcing app config anyways then that’s just as good

PR #1218, opened 2024-02-14
Remove connect/disconnect messages from the audit logs

oestrich:
It's a very large portion of what's in the audit logs table. We should probably be storing these in another way because devices tend to flap online/offline enough with LTE connections.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -76,7 +76,9 @@ defmodule NervesHubWeb.DeviceChannel do
     end
 
     ## After join
-    :telemetry.execute([:nerves_hub, :devices, :connect], %{count: 1})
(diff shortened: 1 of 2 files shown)


jjcarstens (approved):
For those following - When scaling to lots of devices, this audit event bloats the table and is problematic when trying to load other pages that include audit logs.

Adding an identifier to the telemetry event at least allows some metric and introspection until we figure out how to best handle connect/disconnect

fhunleth:
I'm debugging an issue right now where having a record of when the hub was online is really nice. I hope we can get this back soon. I can make the "Last Handshake" field to work in the interim.

fhunleth:
If anyone is following this, I've learned that there are just fundamental issues with how the audit log is stored and index. Jon alluded to it in his comment. I did not realize how unscaleable the audit log was right now. Manual intervention is necessary to prune logs aggressively to prevent timeouts. This PR provides breathing room to make a proper fix.

PR #1229, opened 2024-02-26
More logging around device events

oestrich:
Add logging to devices connecting, disconnecting, receiving an update, and successfully updating. Outputs metadata as key/value pairs for parsing with something like datadog.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,9 +1,9 @@
 import Config
 
 # Configures Elixir's Logger
-config :logger, :console,
-  format: "$time $metadata level=$level $message\n",
-  metadata: [:user_id, :request_id, :trace_id, :span_id]
+config :logger, :default_formatter,
+  format: {NervesHub.LoggerFormatter, :format},
+  metadata: :all
 
 config :phoenix,
   json_library: Jason,
--- a/lib/nerves_hub/logger_formatter.ex
+++ b/lib/nerves_hub/logger_formatter.ex
@@ -1,0 +1,7 @@
+defmodule NervesHub.LoggerFormatter do
+  @pattern Logger.Formatter.compile("$time [$level] $metadata$message\n")
+  def format(level, message, timestamp, metadata) do
+    metadata = Keyword.drop(metadata, [:line, :file, :domain, :application])
+    Logger.Formatter.format(@pattern, level, message, timestamp, metadata)
+  end
+end
--- a/lib/nerves_hub/metrics.ex
+++ b/lib/nerves_hub/metrics.ex
@@ -119,6 +119,7 @@ defmodule NervesHub.Metrics.Reporters do
   def handle_continue(:initialize, state) do
     reporters = [
       NervesHub.EctoReporter,
+      NervesHub.DeviceReporter,
       NervesHub.NodeReporter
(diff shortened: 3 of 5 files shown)


lib/nerves_hub/logger_formatter.ex
```diff
@@ -0,0 +1,7 @@
+defmodule NervesHub.LoggerFormatter do
+  def format(level, message, timestamp, metadata) do
+    pattern = Logger.Formatter.compile("$time [$level] $metadata$message\n")
+    metadata = Keyword.drop(metadata, [:line, :file, :domain, :application])
+    Logger.Formatter.format(pattern, level, message, timestamp, metadata)
+  end
```
jjcarstens, line 6:
```suggestion
  @pattern Logger.Formatter.compile("$time [$level] $metadata$message\n")
  def format(level, message, timestamp, metadata) do
    metadata = Keyword.drop(metadata, [:line, :file, :domain, :application])
    Logger.Formatter.format(@pattern, level, message, timestamp, metadata)
  end
```

Just a small optimization to precompile the format rather than compile it on each log/formatting event

config/config.exs
```diff
@@ -2,8 +2,8 @@ import Config
 
 # Configures Elixir's Logger
 config :logger, :console,
```
jjcarstens, line 4:
```suggestion
config :logger, :default_formatter,
```
We should prob change this to match the newer Elixir versions we're using since `:console` is soft-deprecated

lib/nerves_hub/metrics.ex
```diff
@@ -153,6 +154,49 @@ defmodule NervesHub.EctoReporter do
...
+      [:nerves_hub, :devices, :connect],
+      [:nerves_hub, :devices, :disconnect],
+      [:nerves_hub, :devices, :update, :automatic]
+    ]
+  end
+
+  def handle_event([:nerves_hub, :devices, :connect], _, metadata, _) do
+    Logger.info("Device connected",
+      event: "nerves_hub.devices.connect",
+      device_identifier: metadata[:identifier],
```
jjcarstens, line 171:
```suggestion
      identifier: metadata[:identifier],
```

I think since it's scoped to `devices` in the event key, we can match the metadata key here and save a few bytes. What do you think?

oestrich: Works for me!

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -228,7 +229,10 @@ defmodule NervesHubWeb.DeviceChannel do
   end
 
   def handle_info({"deployments/update", inflight_update}, %{assigns: %{device: device}} = socket) do
-    :telemetry.execute([:nerves_hub, :devices, :update, :automatic], %{count: 1})
+    :telemetry.execute([:nerves_hub, :devices, :update, :automatic], %{count: 1}, %{
```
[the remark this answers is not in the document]
oestrich: Yeah, but we should only ever get an inflight update if there's a payload to send. So this might be a future PR to trust the inflight update and build from that

jjcarstens (approved):

PR #1254, opened 2024-03-25
Configure tls_certificate_check for tls config in swoosh

oestrich:
OTP 26 changed the error level of what we're seeing when trying to send email to SES via TLS. This additional configuration should let us communicate via TLS without issues.

See https://github.com/gen-smtp/gen_smtp/issues/328


config/runtime.exs
```diff
@@ -293,6 +293,7 @@ if config_env() == :prod do
       password: System.fetch_env!("SMTP_PASSWORD"),
       ssl: System.get_env("SMTP_SSL", "false") == "true",
       tls: :always,
+      tls_options: :tls_certificate_check.options(System.fetch_env!("SMTP_SERVER")),
```
jjcarstens, line 296: Will this affect if `System.get_env("SMTP_SSL", "false") == "false"`? As in if its disabled and we still include SSL options, will things blow up?

oestrich: We do have `tls: :always` right below `ssl: false` which I'm not sure what the difference is. 

It seems like the difference is `ssl: false` won't immediately try to start ssl, vs using starttls since we have `tls: :always` That's my read of the docs though https://github.com/gen-smtp/gen_smtp?tab=readme-ov-file#options

jjcarstens (approved):

PR #1271, opened 2024-05-01
Prewritten commands for a device

oestrich:
Pre-written commands help enable self-service for support level users. Admins can create new commands that are sent via the console channel and the output is returned in an API response. Users can get access to the console in a very limited ability with this.

### DIFF
--- a/lib/nerves_hub_web/templates/command/index.html.heex
+++ b/lib/nerves_hub_web/templates/command/index.html.heex
@@ -1,0 +1,51 @@
+<%= if @commands == [] do %>
+  <div class="no-results-blowup-wrapper">
+    <img src="/images/command.svg" alt="No commands" />
+    <h3 style="margin-top: 2.75rem"><%= @product.name %> doesn’t have any commands yet</h3>
+    <div class="flex-row align-items-center mt-3">
+      <a class="btn btn-outline-light" aria-label="Create command" href={Routes.command_path(@conn, :new, @org.name, @product.name)}>
+        <span class="action-text">Create Command</span>
+        <span class="button-icon add"></span>
+      </a>
+    </div>
+  </div>
+<% else %>
+  <div class="action-row">
+    <h1>Commands</h1>
+    <a class="btn btn-outline-light btn-action" aria-label="Create command" href={Routes.command_path(@conn, :new, @org.name, @product.name)}>
+      <span class="button-icon add"></span>
+      <span class="action-text">Create Command</span>
+    </a>
+  </div>
+
+  <table class="table">
+    <tr>
+      <th>Name</th>
+      <th>Text</th>
+      <th>Actions</th>
+    </tr>
+    <%= for command <- @commands do %>
+      <tr class="item">
+        <td>
+          <%= command.name %>
+        </td>
+        <td>
+          <pre><code><%= command.text %></code></pre>
+        </td>
+        <td class="actions">
+          <div class="mobile-label help-text">Actions</div>
+          <div class="dropdown options">
+            <a class="dropdown-toggle options" data-target="#" id={"command-#{command.id}"} data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+              <div class="mobile-label pr-2">Open</div>
+              <img src="/images/icons/more.svg" alt="options" />
+            </a>
+            <div class="dropdown-menu dropdown-menu-right">
+              <%= link("edit", class: "dropdown-item", to: Routes.command_path(@conn, :edit, @org.name, @product.name, command.id)) %>
+              <%= link("delete", class: "dropdown-item", method: :delete, to: Routes.command_path(@conn, :delete, @org.name, @product.name, command.id)) %>
+            </div>
+          </div>
+        </td>
+      </tr>
+    <% end %>
+  </table>
+<% end %>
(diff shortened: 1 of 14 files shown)


fhunleth:
Can these be called something else? Maybe "automations" or "shortcuts" or maybe there's another word?

oestrich:
I’m up for any name, once we decide I’ll rename everything. I think shortcuts is better of the two but still feels a bit off

joshk:
I like “scripts” as they are device scripts.

test/nerves_hub/commands_test.exs
```diff
@@ -0,0 +1,38 @@
...
+      assert command.product_id == product.id
+    end
+
+    test "no newline allowed", %{product: product} do
+      {:error, changeset} =
+        Commands.create(product, %{
+          name: "MOTD",
+          text: """
+          Hello there
+          """
```
fhunleth, line 32: I think newlines have to be allowed here since it's not possible to make some things in Elixir one liners. Restricting the name to a one line makes sense.

oestrich: It's a lot uglier in the output but I made it work. Pushing up shortly

lib/nerves_hub_web/templates/command/index.html.heex
```diff
@@ -0,0 +1,51 @@
...
+    <h3 style="margin-top: 2.75rem"><%= @product.name %> doesn’t have any commands yet</h3>
+    <div class="flex-row align-items-center mt-3">
+      <a class="btn btn-outline-light" aria-label="Create command" href={Routes.command_path(@conn, :new, @org.name, @product.name)}>
+        <span class="action-text">Create Command</span>
+        <span class="button-icon add"></span>
+      </a>
+    </div>
+  </div>
+<% else %>
+  <div class="action-row">
```
joshk, line 13: it might be useful to include a description section somewhere here which explains what this feature is and does.

joshk (changes requested):

lib/nerves_hub/scripts/script.ex
```diff
@@ -0,0 +1,27 @@
...
+  def create_changeset(struct, params) do
+    struct
+    |> cast(params, [:name, :text])
+    |> validate_required([:name, :text])
+    |> validate_length(:name, lte: 255)
+  end
+
+  def update_changeset(struct, params) do
+    create_changeset(struct, params)
+  end
```
jjcarstens, line 26: If we're just replicating the create change set, could we just consolidate to `changeset`?

lib/nerves_hub_web/controllers/script_controller.ex
```diff
@@ -0,0 +1,105 @@
+defmodule NervesHubWeb.ScriptController do
+  use NervesHubWeb, :controller
+
+  alias NervesHub.Scripts
+  alias NervesHub.Repo
+
+  plug(:validate_role, [org: :manage] when action in [:new, :create, :delete])
```
jjcarstens, line 7:
```suggestion
  plug(:validate_role, [org: :manage] when action in [:new, :create, :edit, :delete])
```
Do we need the edit role in here to make changes?

lib/nerves_hub_web/controllers/api/script_controller.ex
```diff
@@ -0,0 +1,60 @@
...
+
+  def send(conn, %{"identifier" => identifier, "id" => id}) do
+    %{user: user} = conn.assigns
+
+    case Devices.get_by_identifier(identifier) do
+      {:ok, device} ->
+        if Accounts.has_org_role?(device.org, user, :view) do
+          case Scripts.get(device.product, id) do
+            {:ok, command} ->
+              text(conn, Scripts.Runner.send(device, command))
```
jjcarstens, line 39: I think we might need more structure to this return. The error view returns JSON right? But here we just send a blob of text?

oestrich: That's what the `Content-Type` header is for, this sends `Content-Type: text/plain` to tell clients what's in there. If we did add structure, what else would we want to include?

lib/nerves_hub_web/controllers/script_controller.ex
oestrich: Good catch, also `:update`!

lib/nerves_hub/scripts/runner.ex
```diff
@@ -0,0 +1,65 @@
...
+
+  alias NervesHubWeb.Endpoint
+
+  defmodule State do
+    defstruct [:buffer, :from, :receive_channel, :send_channel]
+  end
+
+  def send(device, command) do
+    {:ok, pid} = start_link(device)
+    GenServer.call(pid, {:send, command.text}, 10_000)
```
jjcarstens, line 12:
```suggestion
  def send(device, command, timeout \\ 10_000) do
    {:ok, pid} = start_link(device)
    GenServer.call(pid, {:send, command.text}, timeout)
```

Should we allow this to be configurable since scripts will be quite variable?

joshk (approved):
All in all I love where this is going. 
I would suggest that either for this PR, or as a followup, that each 'run' is saved and creates an audit trail of who ran what script and what the output was.
This would allow the API to return straight away and return a 'script run' id which can be polled for completion and output.

PR #1284, opened 2024-05-10
the start of a handy health check route for both endpoints

joshk:
I use these for my deployments to Fly.io, which require a health check for blue green deployments

### DIFF
--- a/lib/nerves_hub_web/device_endpoint.ex
+++ b/lib/nerves_hub_web/device_endpoint.ex
@@ -15,6 +15,8 @@ defmodule NervesHubWeb.DeviceEndpoint do
     ]
   )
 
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -0,0 +1,129 @@
+defmodule NervesHubWeb.DeviceSocketUnified do
```
oestrich, line 1: Should this be included in this PR?

joshk: oops, no, that snuck its way in

joshk: fixed!

oestrich (approved):

PR #1286, opened 2024-05-11
support shared auth on the devices endpoint

joshk:
With Nerves Cloud I want to support a Shared Secrets socket on the devices endpoint (devices.nervescloud.com)

This PR unifies the cert auth and shared secret auth sockets, and tweaks the runtime settings so that `peer_cert` isn't required to get to the endpoint.


joshk:
Because Slipstream doesn't pass the body of the connection error to the socket, I duplicated the error message in the body to the response headers to make it easier for Link to use it.

I don't love it, but I've made my peace with it.

jjcarstens (approved):
🎉 🎉

PR #1292, opened 2024-05-14
Remove phoenix_markdown from project along with policy pages

oestrich:
Locally and in CI compiling breaks due to this library. It seems like any time any deps change something with earmark breaks and you can't compile. Since these are policy pages I'm not sure we need them. Alternatively I can convert them to heex if we think they're still required.


joshk (approved):
I agree with these being removed, they are more of a concern for anyone running a public NervesHub service.

PR #1294, opened 2024-05-15
correct the unhandled message function

joshk:
`handle_in` handles incoming channel messages, not `handle_info`, my bad


jjcarstens (commented):
I actually think it would behoove us to keep both and this is just another great catch of where we could crash were we don't want to.

joshk:
hahaha, ok, I'll add the other one back and look into a test for it

joshk:
@jjcarstens feedback addressed

jjcarstens (approved):

PR #1295, opened 2024-05-15
add support for devices checking if an update is available

joshk:
this is related to https://github.com/nerves-hub/nerves_hub_link/issues/47

**open questions:**

do we want to add an audit log for this?

should the event name be something more like `update:check` or `update:check_available`?

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -400,6 +400,20 @@ defmodule NervesHubWeb.DeviceChannel do
     {:noreply, socket}
   end
 
+  def handle_in("check_update_available", _params, socket) do
+    device =
+      socket.assigns.device
+      |> Devices.verify_deployment()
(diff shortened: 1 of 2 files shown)


oestrich:
My only comment is we should remember to add this to the orchestrator sometime, maybe respond with "update pending" if the device will eventually receive an update? I'd like to keep the only loophole to the orchestrator being on websocket connect if we can. I'm totally fine with another PR for that because we have the note.

joshk:
I didn’t know about the orchestrator, including what it is or does.

If you can give a me summary I can look into the changes you are
suggesting.

jjcarstens (approved):
The orchestrator might not be the right thing to do _yet_ - Also, there are improvements we can make here.

I have experiments based on the discussion with HCA about waterpark that might help us get to a GenServer for every device and make these checkins less aggressive. I'd rather work towards that than spend a lot of time worrying about the coordination for a new feature that won't be on a lot of devices for a while

PR #1299, opened 2024-05-16
API to move devices between products and organizations

oestrich:


lib/nerves_hub_web/controllers/api/device_controller.ex
```diff
@@ -260,4 +261,36 @@ defmodule NervesHubWeb.API.DeviceController do
...
+
+          {:error, changeset} ->
+            # fallback controller will render this
+            {:error, changeset}
+        end
+      else
+        conn
+        |> put_resp_header("content-type", "application/json")
+        |> send_resp(403, Jason.encode!(%{status: "missing required role: write"}))
+      end
```
jjcarstens, line 293: This is a bit gnarly for my taste. I could be convinced, but with all the nesting it takes a bit of effort to get to the `else` clause and know at what level that is (`with` vs `if`). How would you feel about a limited `else` in the `with` clause?

oestrich: The problem with adding anything to the with as an else is this is relying on the fallback controller for rendering a 404. I also don’t like that so I’m fine making everything in one place. The downside for the with to have this as an else is the match is true/false 🫤

lib/nerves_hub_web/controllers/api/device_controller.ex
```diff
@@ -260,4 +261,36 @@ defmodule NervesHubWeb.API.DeviceController do
...
+        "org_name" => org_name,
+        "product_name" => product_name
+      }) do
+    %{user: user} = conn.assigns
+
+    with {:ok, device} <- Devices.get_by_identifier(identifier),
+         {:ok, org} <- Accounts.get_org_by_name(org_name),
+         {:ok, product} <- Products.get_product_by_org_id_and_name(org.id, product_name) do
+      if Accounts.has_org_role?(device.org, user, :manage) &&
+           Accounts.has_org_role?(org, user, :manage) do
```
joshk, line 276: I'd suggest moving this into a helper function, which would allow it to cleanly go into the `with` if desired, or it can stay where it is but be more concise.

lib/nerves_hub_web/controllers/api/device_controller.ex
jjcarstens: Ah, I see. The lookup results need to be `404` and the inability to move would be `403` (and the fallback for the changes error is ?)

I think with that we should also check the role _before_ we do DB lookup work. So maybe its break that into a bigger `if` block all the work is actually within the `with`?

oestrich: I'm not sure we can move the lookups to be higher. We don't know the device's org until we look it up, and the new org we also need to lookup

jjcarstens: Ah, I see. I understand now 🤦🏼

jjcarstens (approved):

PR #1306, opened 2024-05-26
Device Socket: turn on compression, and increase the timeout

joshk:
compression has been enabled to save some data, and the device socket timeout has been increased to a max of 3 minutes, which allows devices to choose a heartbeat which best suits their connection and data usage preferences.


fhunleth:
Aside from the conflicts, this all seems harmless since the device needs to advertise that it supports compression and change its keepalive period to support.

fhunleth (approved):

joshk:
Thanks for the review @fhunleth 
@jjcarstens, any comments or thoughts?

jjcarstens (approved):

PR #1308, opened 2024-05-30
Redirect past orgs page more conveniently

lawik:
Added two cases where we can skip the org page for the user:

1. The user only has one org.
2. The user has previously selected an org and we have it in session.

### DIFF
--- a/lib/nerves_hub_web/controllers/home_controller.ex
+++ b/lib/nerves_hub_web/controllers/home_controller.ex
@@ -4,7 +4,24 @@ defmodule NervesHubWeb.HomeController do
   def index(conn, _params) do
     case Map.has_key?(conn.assigns, :user) && !is_nil(conn.assigns.user) do
       true ->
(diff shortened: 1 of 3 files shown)


lawik:
This should help this case: https://github.com/nerves-hub/nerves_hub_web/issues/815

jjcarstens (commented):
I click the NervesHub icon frequently to take me back to an orgs list and pick the one I want. Either this change, I would only be able to use the drop-down org/product switcher, correct?

lawik:
Yes. That would be the debatable part. But the switcher is there for you.

The skip for users with a single org thing seems like it would be straight-forward at least. The drop-down has the Create Org button for that need.

I understand if the session one is less certain.

joshk:
my preference is for the top left icon to ...

1. if you have only one org, take you to that org
2. if you have more than one org, then take you to the org list page

PR #1309, opened 2024-05-30
Set fullsweep_after on device channel

oestrich:
Remove drainer since it seems to be in the wrong spot and there's a default for it anyways

### DIFF
--- a/lib/nerves_hub_web/device_endpoint.ex
+++ b/lib/nerves_hub_web/device_endpoint.ex
@@ -12,7 +12,8 @@ defmodule NervesHubWeb.DeviceEndpoint do
     NervesHubWeb.DeviceSocket,
     websocket: [
       connect_info: [:peer_data, :x_headers],
-      error_handler: {WebsocketConnectionError, :handle_error, []}
+      error_handler: {WebsocketConnectionError, :handle_error, []},
+      fullsweep_after: 0
     ]
   )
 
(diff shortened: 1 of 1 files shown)


joshk:
has this change been tested on your setup? were their any noticeable improvements?

also, why do you think the drainer is in the wrong spot?

oestrich:
This has been tested with 1 device thus far. I suspect the drainer is in the wrong spot because of the docs, they say it should be at the same level as `websocket` and not nested below. 

In production like load tests we hang on to about 200KB per device, which is mostly garbage waiting to be collected. In local profiling with the one device we go from 40KB-60KB to ~5KB in the socket process. This is following the guide from Phoenix here https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-websocket-configuration

fhunleth:
Could the PR be split? The drainer one seems like it could be verified and merged independent of the GC change.

joshk:
Interesting on both accounts.

I'm just triple checking with Mr Bandit if `drainer` is supported by Bandit, as the docs only mention Cowboy support.

Could you also remove the `drainer` config from the `endpoint.ex` file (its also defined there)

oestrich:
PR is split and this is pointed at #1311 as a base

joshk:
should we also set it here? https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/endpoint.ex#L19-L31

joshk (approved):
seems good to try out

PR #1311, opened 2024-05-31
Remove drainer config from device endpoint

oestrich:
It seems to be in the wrong spot per the docs for `socket/3`

### DIFF
--- a/lib/nerves_hub_web/device_endpoint.ex
+++ b/lib/nerves_hub_web/device_endpoint.ex
@@ -12,11 +12,6 @@ defmodule NervesHubWeb.DeviceEndpoint do
     NervesHubWeb.DeviceSocket,
     websocket: [
       connect_info: [:peer_data, :x_headers],
-      drainer: [
-        batch_size: 500,
-        batch_interval: 1_000,
-        shutdown: 30_000
-      ],
       error_handler: {WebsocketConnectionError, :handle_error, []}
     ]
   )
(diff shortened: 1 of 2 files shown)


jjcarstens (commented):
Where is it supposed to be?

oestrich:
Up a level higher than it is, I'm pretty sure this has never worked and we just rely on whatever the default is. Also Josh pointed out in the other thread that this might not even work with bandit 😅

jjcarstens:
Nvm. I see the other PR. Though I'm confused why we're completely removing vs moving to the correct spot?

oestrich:
Since it's not working now, it seems better to remove than move because deploys are going well at the moment

joshk:
From the marvelous Mr Bandit

```
Draining 'just works' in Bandit
no config needed
https://hexdocs.pm/thousand_island/ThousandIsland.html#module-connection-draining-shutdown
https://hexdocs.pm/thousand_island/ThousandIsland.html#t:options/0 shutdown_timeout is the only configuration  option pertaining to draining
```

so its safe to remove

joshk:
regarding the `shutdown_timeout`, from the docs:

```
shutdown_timeout: How long to wait for existing client connections to complete before forcibly shutting those connections down at server shutdown time, in milliseconds. Defaults to 15_000. May also be :infinity or :brutal_kill as described in the [Supervisor](https://hexdocs.pm/elixir/Supervisor.html) documentation
```

the default is half of what we were using before the switch to Bandit.

joshk (approved):

PR #1314, opened 2024-06-04
Upgrade to Elixir 17.0 and Erlang 27.0

joshk:


oestrich:
I’m all for this but we gotta add to the spreadsheet @jjcarstens made to make sure older devices will connect

PR #1320, opened 2024-06-13
Change pattern matching on headers for Device connection

lawik:
The current way of matching was dependent on the order of headers in the list of tuples and I hit problems when testing other clients against it.

### DIFF
--- a/lib/nerves_hub_web/channels/device_socket.ex
+++ b/lib/nerves_hub_web/channels/device_socket.ex
@@ -36,30 +36,31 @@ defmodule NervesHubWeb.DeviceSocket do
   end
 
   # Used by Devices connecting with HMAC Shared Secrets
-  def connect(_params, socket, %{x_headers: [{"x-nh-signature", _} | _] = headers}) do
-    headers = Map.new(headers)
+  def connect(_params, socket, headers) do
+    case Enum.find(headers.x_headers, fn {key, _} ->
+      key == "x-nh-signature"
+    end) do
+      nil -> {:error, :no_auth}
+      {_key, signature} ->
+        headers = Map.new(headers.x_headers)
 
-    with true <- shared_secrets_enabled?(),
-         {:ok, key, salt, verification_opts} <- decode_from_headers(headers),
-         {:ok, auth} <- get_shared_secret_auth(key),
-         {:ok, signature} <- Map.fetch(headers, "x-nh-signature"),
-         {:ok, identifier} <- Crypto.verify(auth.secret, salt, signature, verification_opts),
-         {:ok, device} <- get_or_maybe_create_device(auth, identifier) do
-      socket =
-        socket
-        |> assign(:device, device)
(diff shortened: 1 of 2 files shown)


lawik:
@joshk you might enjoy this?

joshk:
was `def connect(_params, socket, %{x_headers: [{"x-nh-signature", _} | _] = headers}) do` expecting `x-nh-signature` to be first in the `x_headers` list?

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -36,32 +36,33 @@ defmodule NervesHubWeb.DeviceSocket do
...
-      {:ok, socket}
-    else
-      error ->
-        Logger.info("device authentication failed : #{inspect(error)}")
-        {:error, :invalid_auth}
+  def connect(_params, socket, headers) do
+    case Enum.find(headers.x_headers, fn {key, _} ->
+      key == "x-nh-signature"
+    end) do
+      nil -> {:error, :no_auth}
```
jjcarstens, line 43: I don't think this check and `:no_auth` return is totally necessary. I think we could add a `when is_list(headers) and length(headers) > 0` guard to the original function and call it good. Let it go through the check that shared secrets are enabled and error if it is but headers are missing

lawik:
> was `def connect(_params, socket, %{x_headers: [{"x-nh-signature", _} | _] = headers}) do` expecting `x-nh-signature` to be first in the `x_headers` list?

Yup:)

lib/nerves_hub_web/channels/device_socket.ex
lawik: Fair

joshk: can `x_headers` be used for other information that isn't related to what we are using it for?

lawik: The HTTP client can send a bit of whatever and the order should probably be considered undefined.

lawik: @jjcarstens I've taken your notes on the match, much nicer :)

joshk: I'm just wondering if the guard Jon proposed is the best path forward if `x_headers` could exist but not contain any auth bits.

lawik: If x_headers is empty it cannot contain the necessary auth bits and we can't really do anything useful with it. And it can fall through to the no_auth case from before. That's what I did for now.

If there are `x_headers` and no shared secret auth headers the with statement will fail and give a better error.

lawik:
Added a regression test. Tried one when I just Enum.shuffle'd the headers but that would be flaky. This is consistent. I don't know if we need it, but it proves out the problem.

joshk:
@lawik you have a format error

PR #1331, opened 2024-06-15
Org level audit log data retention policy 

joshk:
This addresses https://github.com/nerves-hub/nerves_hub_web/issues/795

A few highlights:
- The query used to delete audit logs has been simplified, and optimized with the added index
- An app-wide default for the number of days to keep is set in the `runtime.exs` config
- But the `max_records_per_run` to delete has been removed
- Each Org can be set a custom number of days of audit logs to keep by setting `audit_log_days_to_keep` on the Org

I decided not to add a way for the Org admin to change the number of days of audit logs to keep because an Org admin will always want to keep as many audit logs as possible. If the platform admin wants to set this to less or more days then they will do it application wide, or they can change this on an Org level via the DB if they want.


fhunleth (approved):
Seems good to my level of knowledge on this subject.

priv/repo/migrations/20240615075328_add_org_id_and_inserted_at_index_to_audit_logs.exs
```diff
@@ -0,0 +1,8 @@
+defmodule NervesHub.Repo.Migrations.AddOrgIdAndInsertedAtIndexToAuditLogs do
+  use Ecto.Migration
+  @disable_ddl_transaction true
```
jjcarstens, line 3:
```suggestion
  @disable_ddl_transaction true
  @disable_migration_lock true
```
According to the [Safe Ecto Migrations: Adding an index](https://fly.io/phoenix-files/migration-recipes/#adding-an-index), we may also want to disable the migration lock here as well

lib/nerves_hub/workers/schedule_org_audit_log_truncation.ex
```diff
@@ -0,0 +1,42 @@
...
+          org
+          |> truncation_args()
+          |> OrgAuditLogTruncation.new()
+          |> Oban.insert()
+      end)
+
+      {:ok, Enum.count(orgs)}
+    else
+      {:ok, :disabled}
+    end
```
jjcarstens, line 25: These return values are ignored. Maybe we simplify and always return `:ok` and only run the iteration when it is enabled?

lib/nerves_hub/workers/schedule_org_audit_log_truncation.ex
```diff
@@ -0,0 +1,42 @@
...
+      org_id: org.id,
+      days_to_keep: days_to_keep(org)
+    }
+  end
+
+  def days_to_keep(org) do
+    org.audit_log_days_to_keep || config()[:default_days_kept]
+  end
+
+  defp enabled?(), do: config()[:enabled]
```
jjcarstens, line 39:
```suggestion
  defp enabled?(), do: config()[:enabled] == "true"
```

If the config value is `"false"` then this would still evaluate to run

lib/nerves_hub/workers/schedule_org_audit_log_truncation.ex
```diff
@@ -0,0 +1,42 @@
+defmodule NervesHub.Workers.ScheduleOrgAuditLogTruncation do
+  use Oban.Worker,
+    max_attempts: 5,
```
jjcarstens, line 3: The [config value is setting `max_attempts: 1`](https://github.com/nerves-hub/nerves_hub_web/pull/1331/files#diff-fcb6e9bf2546587075d6850070c1b4d26f061f0c199e6ea3ac4d8fd76d9fa8f6R59) which I believe will take precedence here?

lib/nerves_hub/audit_logs.ex
```diff
@@ -78,19 +78,16 @@ defmodule NervesHub.AuditLogs do
...
   end
 
-  def truncate(opts) do
-    oldest = DateTime.add(DateTime.utc_now(), -24 * 60 * 60 * opts[:days_kept], :second)
-
-    to_delete =
-      from(a in AuditLog,
-        where: a.inserted_at < ^oldest,
-        order_by: [asc: :inserted_at],
-        limit: ^opts[:max_records_per_run],
```
jjcarstens, line 88: The `max_records_per_run` is a bit of a loss. We can try this new index, but we may want this back since it was an optimization to handle large datasets

test/nerves_hub/firmwares/firmwares_test.exs
```diff
@@ -1,6 +1,5 @@
 defmodule NervesHub.FirmwaresTest do
   use NervesHub.DataCase, async: true
-  use Oban.Testing, repo: NervesHub.Repo
```
jjcarstens, line 3: Unrelated and just noticed while scrubbing through?

test/support/fixtures.ex
```diff
@@ -390,50 +407,6 @@ defmodule NervesHub.Fixtures do
...
-      deployment: deployment,
-      device: device,
-      device_certificate: device_certificate,
-      firmware: firmware,
-      org: org,
-      org_key: org_key,
-      product: product
-    }
-  end
-
```
jjcarstens, line 436: Same here - Just noticed these weren't being used?

test/nerves_hub/firmwares/firmwares_test.exs
joshk: I removed that because I added https://github.com/nerves-hub/nerves_hub_web/pull/1331/files#diff-50c4752c16c7d9d604c325ddd6c36216abaedad717befdfcffd8b50d48ead444R27

test/support/fixtures.ex
joshk: these are very old and aren't being used

lib/nerves_hub/workers/schedule_org_audit_log_truncation.ex
joshk: in `runtime.exs`
```
config :nerves_hub, :audit_logs,
  enabled: System.get_env("TRUNATE_AUDIT_LOGS_ENABLED", "false") == "true",
```
so we should always have a boolean

lib/nerves_hub/audit_logs.ex
joshk: i'm happy to add it back if its needed, although I don't think it will be needed. 

I don't think this was the right approach, and instead the query can be optimized

joshk:
@jjcarstens I've updated everything based on your feedback

lib/nerves_hub/audit_logs.ex
oestrich: Yeah we likely want the `max_records_per_run` because there's a good chance we delete so much every single time that a vacuum would hamper database performance. The goal with max records was to slowly chip away at it and let postgres handle the deleted data spread out over a day

oestrich:
I think we do want to keep max records, it's there as a way to attempt to prevent PG vacuum problems (which if there's a better way I'm up for that)

joshk:
Can you explain a bit more about the problems you were seeing? 

PG vacuum happens later, and happens in stages later, so a big delete shouldn't do anything to perf. 

Deletes should be quick, as its just telling the DB to clean it up later.

oestrich:
The big delete issue was letting postgres have to deal with a large amount of data instead of vacuuming more often with smaller amounts. I can't remember the full context of why I added this but it seemed useful at the time and was helpful

joshk:
Vacuum is something that doesn't happen straight away, but when Postgres believes its time for a clean up.

The other trouble is that the cron was only running once a day, so if you are adding 20,000 audit logs a day, only 10,000 will deleted a day.

Also, the previous query was inefficient as it was one query nested in another. (https://github.com/nerves-hub/nerves_hub_web/pull/1331/files#diff-2c7b03c045b3220c3089907edf0dacaf557c7a71e1c5cb0f07862a3258ba7cb4L84-L94)

I do believe this will perform better, and I'm happy to spend time on it to make it perform or implement mitigations if performance issues are discovered.

I would highly recommend hooking up https://pganalyze.com/ so we can better understand the db issues and address them as we find them.

PR #1335, opened 2024-06-17
Start moving the UI/UX to LiveView

joshk:
The core of this PR is the transition from the older style Phoenix controllers to LiveViews for web interactions.

I've started with all Org focused pieces first. This includes the following URLs:

```
/orgs
/orgs/new
/org/:org_name
/org/:org_name/new
/org/:org_name/settings/keys
/org/:org_name/settings/keys/new
/org/:org_name/settings/users
/org/:org_name/settings/users/invite
/org/:org_name/settings/users/:user_id/edit
/org/:org_name/settings/certificates
/org/:org_name/settings/certificates/new
/org/:org_name/settings/certificates/:serial/edit
/org/:org_name/settings
```

I've also added tests for all these pages and interactions. These tests are more complete than the previous ones as forms are being filled in and submitted.

I've also included a few features along the way:
- storing who a Signing Key belongs to
- storing who invited a team member

And fixed some bugs which exist in the current UI.


lib/nerves_hub/accounts.ex
```diff
@@ -224,15 +222,10 @@ defmodule NervesHub.Accounts do
...
-    product_query = from(p in Product, where: is_nil(p.deleted_at))
-
-    orgs_preload = {org_query, products: product_query}
-
     User
-    |> where([u], u.id == ^user_id)
+    |> where(id: ^user_id)
     |> Repo.exclude_deleted()
-    |> preload(orgs: ^orgs_preload)
+    |> preload(orgs: [:products])
```
lawik, line 228: This seems to drop the `deleted_at` soft delete check. Is this fine? Or taken care of elsewhere?

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -117,16 +117,17 @@ defmodule NervesHubWeb.DeviceSocket do
...
   end
 
-  def shared_secrets_enabled?() do
-    enabled =
-      Application.get_env(:nerves_hub, __MODULE__, [])
-      |> Keyword.get(:shared_secrets, [])
-      |> Keyword.get(:enabled, false)
-
-    if enabled do
+  def shared_secrets_enabled_with_error?() do
```
lawik, line 120: Yo dawg, we can't do `?` at the end if it doesn't strictly return a bool :)

I suggest `check_shared_secret_enabled` and return `:ok` on success.

lib/nerves_hub_web/components/form_helpers.ex
```diff
@@ -1,5 +1,5 @@
-defmodule NervesHubWeb.OrgKeyView do
-  use NervesHubWeb, :view
+defmodule NervesHubWeb.Components.FormHelpers do
```
lawik, line 1: This module took a big trip into a different name and role. But seems legit :)

lawik (commented):
Have looked at the stuff that sticks out. Have read and skimmed a lot of it. Good job on doing all this.

lib/nerves_hub_web/channels/device_socket.ex
joshk: goooooooood point

lib/nerves_hub/accounts.ex
joshk: Good spotting. That should be taken case by User and Org. 

User has:
```
has_many(:orgs, through: [:org_users, :org], where: [deleted_at: nil])
```
But Org is missing a similar `where`. I'll get that fixed.

joshk: oh, no, I'm wrong, I already fixed that.

lib/nerves_hub_web/components/form_helpers.ex
joshk: yeah, its kinda funny how git did this.

jjcarstens (approved):
Looks good enough to try. Left one comment about separating out the Elixir 1.17.1 change

.tool-versions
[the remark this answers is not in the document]
joshk: done!

Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG ELIXIR_VERSION=1.17.0
 ARG OTP_VERSION=27.0
-ARG DISTRO=noble-20240530
+ARG DISTRO=noble-20240605
```
fhunleth, line 3: Not sure if this was intended. Elixir 1.17.1 is out now if you're going to do a Dockerfile update.

fhunleth (approved):

Dockerfile
joshk: https://github.com/nerves-hub/nerves_hub_web/pull/1335#pullrequestreview-2143086188

jjcarstens: Sorry, I prob should've specified that we could probably include this docker image change wherever we're doing the Elixir 1.17 change

Issue #1337, opened 2024-06-18
Invites to join an Org require the user to accept the invite


joshk:
When a user is invited to join an Org, the user is added straight away, even if they don't want to join the Org.

It should be required for users to accept Org invites before they become Org members.

PR #1346, opened 2024-06-23
Invites to join an Org require the user to accept the invite

beltranaceves:
Addressing #1337.

Hi, first time contributing to this repo. I wanted to validate the changes and style before writing tests.

Checklist:
1. [x] When inviting existing users, they are not automatically added to the org until they accept the invite. 
2. [x] Invites for existing users cannot be accepted if the user is not logged in.
3. [x] Invites for existing users cannot be accepted if the logged in user is not the recipient.
4. [x] Invites for new users cannot be accepted if a user is logged in.
5. [x] Show pending invites on login.
6. [ ] Invites list view.
7. [ ] Cleaned up temp changes.
8. [ ] Broken test are fixed.
9. [ ] New test have been writen.


beltranaceves:
I was tempted to add a route for logged in users that lists pending received invites, accessible from the user dropdown menu. But it seems that issues and PRs are fairly atomic, so I kept it pretty barebones

lib/nerves_hub/accounts.ex
```diff
@@ -442,8 +442,12 @@ defmodule NervesHub.Accounts do
...
   def add_or_invite_to_org(%{"email" => email} = params, org) do
     case get_user_by_email(email) do
-      {:error, :not_found} -> invite(params, org)
-      {:ok, user} -> add_org_user(org, user, %{role: params["role"]})
+      {:error, :not_found} ->
+        invite(params, org)
+
+      {:ok, _user} ->
+        invite(params, org)
+        # add_org_user(org, user, %{role: params["role"]})
```
joshk, line 450: I guess this can be simplified to always invite, and not having to search for the user

lib/nerves_hub_web/router.ex
```diff
@@ -198,6 +198,7 @@ defmodule NervesHubWeb.Router do
 
     get("/invite/:token", AccountController, :invite)
     post("/invite/:token", AccountController, :accept_invite)
+    post("/invite_existing/:token", AccountController, :accept_invite_existing)
```
joshk, line 201: this should all be possible without another route.

joshk (changes requested):
Thanks for jumping into this issue! Its great to have new people onboard.

You are on the right track. I have a few suggestions which should clear up the UX of the feature:
- My suggestion would be to have a new banner in the Orgs list (`/`) with a message like "You have x pending invites to Organizations". 
- When you click the message, or link by the message ("View pending invites"), you would then be taken to a page which lists all pending invites. 
- This list would include the Organization name, who invited you (I have this in another PR which I'm just finishing), when they invited you, and two buttons, accept and decline.

Let me know if there is anything I can help with.

joshk:
Hi @beltranaceves, thanks for your initial work on this PR. I'm going to close it as it's become stale.

PR #1359, opened 2024-06-28
When updating a deployment don't clear existing devices' deployment

oestrich:
When we have very large deployments we can run into timeouts as all of the devices recalculate at the same time. Instead let the device channel recalculate and generally sit and do nothing


oestrich:
This is obsolete with the experimental changes in the calculator PR

PR #1360, opened 2024-06-28
Add jitter to devices' setting their deployment on deployment change

oestrich:
If we have a large fleet of devices, we could cause cascading ecto pool issues if thousands+ of devices go to set their deployment at the same time. Jitter over 5 minutes to help avoid slamming the pool. Also space out resolving a deployment for attached devices over the same time frame.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -172,10 +172,24 @@ defmodule NervesHubWeb.DeviceChannel do
         %{assigns: %{device: device}} = socket
       ) do
     if device_matches_deployment_payload?(device, payload) do
-      {:noreply, assign_deployment(socket, payload)}
+      if timer = Map.get(socket.assigns, :assign_deployment_timer) do
+        Process.cancel_timer(timer)
+      end
+
+      # jitter over five minutes to attempt to not slam the database when
+      # any matching devices go to set their deployment. This is for very large
+      # deployments, to prevent ecto pool contention.
+      jitter = :rand.uniform(300) * 1000
+      timer = Process.send_after(self(), {:assign_deployment, payload}, jitter)
+      {:noreply, assign(socket, :assign_deployment_timer, timer)}
     else
       {:noreply, socket}
     end
+  end
+
+  def handle_info({:assign_deployment, payload}, socket) do
+    socket = assign(socket, :assign_deployment_timer, nil)
+    {:noreply, assign_deployment(socket, payload)}
   end
 
   def handle_info(
@@ -186,9 +200,10 @@ defmodule NervesHubWeb.DeviceChannel do
       :telemetry.execute([:nerves_hub, :devices, :deployment, :changed], %{count: 1})
       {:noreply, assign_deployment(socket, payload)}
     else
-      # jitter over a minute but spaced out to attempt to not
-      # slam the database when all devices check
-      jitter = :rand.uniform(30) * 2 * 1000
+      # jitter over five minutes to attempt to not slam the database when
+      # any matching devices go to resolve their deployment. This is for very large
+      # deployments, to prevent ecto pool contention.
+      jitter = :rand.uniform(300) * 1000
       Process.send_after(self(), :resolve_changed_deployment, jitter)
       {:noreply, socket}
     end
--- a/test/nerves_hub_web/channels/device_channel_test.exs
+++ b/test/nerves_hub_web/channels/device_channel_test.exs
@@ -239,6 +239,8 @@ defmodule NervesHubWeb.DeviceChannelTest do
     socket_alpha = :sys.get_state(socket_alpha.channel_pid)
     assert is_nil(socket_alpha.assigns.device.deployment_id)
 
+    # skip the jitter
+    send(socket_beta.channel_pid, :resolve_changed_deployment)
     socket_beta = :sys.get_state(socket_beta.channel_pid)
     refute is_nil(socket_beta.assigns.device.deployment_id)
   end


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -172,12 +172,26 @@ defmodule NervesHubWeb.DeviceChannel do
...
     if device_matches_deployment_payload?(device, payload) do
-      {:noreply, assign_deployment(socket, payload)}
+      if timer = Map.get(socket.assigns, :assign_deployment_timer) do
+        Process.cancel_timer(timer)
+      end
+
+      # jitter over five minutes to attempt to not slam the database when
+      # any matching devices go to set their deployment. This is for very large
+      # deployments, to prevent ecto pool contention.
+      jitter = :rand.uniform(300) * 1000
```
joshk, line 182: minor suggestion: maybe move the jitter calculation to its own function to isolate the calculation logic for future changes

joshk:
Question: would this potentially mean that Deployments with just 1 Device in them might have a 5-minute wait until the Device updates (max jitter time for that device)?

If so, I think this jitter should not apply when the device count for a deployment is under a certain threshold, and maybe it should be smaller than 5 minutes for cases with different device counts.

oestrich:
It could take 5 minutes for 1 device if that’s all you have. There’s no way of knowing how many devices the deployment is targeting until you let something calculate attachment and that’s a very heavy thing to do. I also don’t know that it’s worth having branches for something like this. There was already a minute jitter for devices that were attached so this makes things more even

joshk:
In this case, I would prefer this to be a runtime config.

When a deployment has A LOT of devices in it, well, this makes sense. For the 99% of NervesHub installs, this jitter could feel like an unexpected delay. And with Lars and I doing more demos and conf talks, including showing off how deployments work, having features feel snappy can lead to people getting to an 'aha' moment.

fhunleth:
I don't have strong opinions here, but I really like the idea of a runtime config to allow the large scale deployment needs to not degrade small and demo deployments.

config/runtime.exs
```diff
@@ -19,7 +19,9 @@ config :nerves_hub,
   admin_auth: [
     username: System.get_env("ADMIN_AUTH_USERNAME"),
     password: System.get_env("ADMIN_AUTH_PASSWORD")
-  ]
+  ],
+  device_deployment_change_jitter_seconds:
+    String.to_integer(System.get_env("DEVICE_DEPLOYMENT_CHANGE_JITTER_SECONDS", "10"))
```
fhunleth, line 24: Are environment variables documented somewhere? It seems like there should be a corresponding description to what this does and why it's important somewhere, but I don't know where.

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -545,4 +560,9 @@ defmodule NervesHubWeb.DeviceChannel do
     |> assign(:device, device)
     |> assign(:deployment_channel, deployment_channel)
   end
+
+  defp device_deployment_change_jitter_seconds() do
```
fhunleth, line 564:
```suggestion
  defp device_deployment_change_jitter_ms() do
```
Changed to `_ms` since this returns milliseconds rather than seconds.

config/runtime.exs
oestrich: I don't believe any of these are documented, and I'm not sure where we'd put them at the moment. README I guess?

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -545,4 +560,9 @@ defmodule NervesHubWeb.DeviceChannel do
...
+    jitter = Application.get_env(:nerves_hub, :device_deployment_change_jitter_seconds)
+    :rand.uniform(jitter) * 1000
```
fhunleth, line 566:
```suggestion
    :rand.uniform(jitter * 1000)
```
Adding even more jitter since we can. No idea if this helps...

config/runtime.exs
fhunleth: @joshk @jjcarstens Any preference where environment variables are documented?

lib/nerves_hub_web/channels/device_channel.ex
oestrich: Yeah, I'm not sure either. I think we want second separation to let queries complete so we don't overload?

fhunleth: I thought the first way selected 1 out of 10 slots to put queries by default. Won't that result in a pattern where queries get issued on 1 second boundaries rather than all over the place? My expectation in other context is that smearing things out is better. What I didn't know is if the database did some smart batching and databases just liked it when you thinned things out by 1/10, for example.

A meta-comment here is that this all seems hard to tune. Wouldn't it be better to queue these up to something that can feed the database at whatever speed it can take?

oestrich: Probably but it would not be as easy to switch to as jittering is, so that's how I started here. We can't change certain deployments at the moment so I was going for something that could work fast, but we're already a week out from that so :shrug: I can figure out how to migrate to a queue that we zip through probably per node since I think this thing will need to communicate with connected device channels

fhunleth: Cool. I'm done with my comments. Thanks for the explanations.

config/runtime.exs
joshk: sooooo, I have had this on my backlog for a while. 

Leave this with me. This falls within a docs update Lars and I will attack soon.

joshk (approved):

PR #1365, opened 2024-07-01
User: switch from username to name

joshk:
- Username is removed
- And Name is added
- For the Users table, this is done as a column rename
- Validations have been modified, and templates that used `username` have been updated

### DIFF
--- a/lib/nerves_hub/accounts/user.ex
+++ b/lib/nerves_hub/accounts/user.ex
@@ -24,7 +24,8 @@ defmodule NervesHub.Accounts.User do
     has_many(:org_users, OrgUser, where: [deleted_at: nil])
     has_many(:orgs, through: [:org_users, :org], where: [deleted_at: nil])
 
-    field(:name, :string)
+    # The username column has been repurposed as a name field
+    field(:name, :string, source: :username)
     field(:email, :string)
     field(:password, :string, virtual: true)
     field(:password_confirmation, :string, virtual: true)
--- a/lib/nerves_hub/accounts/user.ex
+++ b/lib/nerves_hub/accounts/user.ex
@@ -43,7 +43,7 @@ defmodule NervesHub.Accounts.User do
     |> hash_password()
     |> password_validations()
     |> update_change(:name, &trim/1)
-    |> validate_name()
+    |> validate_format(:name, ~r/^[a-zA-Z\'\- ]+$/, message: "has invalid character(s)")
     |> validate_required(@required_params)
     |> unique_constraint(:email)
   end
@@ -68,15 +68,6 @@ defmodule NervesHub.Accounts.User do
     |> generate_password_reset_token_expires()
     |> email_password_update_valid?(user, params)
   end
-
-  defp validate_name(%Changeset{changes: %{name: name}} = changeset) do
-    case Regex.match?(~r/^[a-zA-Z\'\- ]+$/, name) do
-      true -> changeset
-      false -> add_error(changeset, :name, "invalid character(s) in name")
-    end
-  end
-
-  defp validate_name(changeset), do: changeset
 
   def with_all_orgs(%User{} = u) do
     u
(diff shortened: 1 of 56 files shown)


priv/repo/migrations/20240701004550_change_username_to_name.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHub.Repo.Migrations.ChangeUsernameToName do
+  use Ecto.Migration
+
+  def up do
+    drop index("users", [:username], name: "users_username_index")
+
+    rename table("users"), :username, to: :name
```
esvinson, line 7: This is not a safe operation for a production system.
https://fly.io/phoenix-files/migration-recipes/#renaming-a-column

joshk: Thanks for the review @esvinson 

Although I don't love Strategy 1, its the easiest option for now, so I will implement those changes.

Thank you!

lib/nerves_hub/accounts/user.ex
```diff
@@ -68,33 +69,14 @@ defmodule NervesHub.Accounts.User do
     |> email_password_update_valid?(user, params)
   end
 
-  defp validate_username(%Changeset{changes: %{username: username}} = changeset) do
-    case Regex.match?(~r/^[A-Za-z0-9-_]+$/, username) do
+  defp validate_name(%Changeset{changes: %{name: name}} = changeset) do
+    case Regex.match?(~r/^[a-zA-Z\'\- ]+$/, name) do
```
jjcarstens, line 73: Ecto has come a long way since this was put in. We could prob just do `Ecto.Changeset.validate_format/3` with this regex now

lib/nerves_hub/accounts/user_token.ex
```diff
@@ -29,7 +29,7 @@ defmodule NervesHub.Accounts.UserToken do
 
   defp generate(user) do
     secret =
-      <<user.username::binary, user.email::binary, DateTime.to_unix(DateTime.utc_now())::32>>
+      <<user.email::binary, DateTime.to_unix(DateTime.utc_now())::32>>
```
jjcarstens, line 32: Leaving the name in here adds more entropy. Prob fine without, but maybe worth keeping since it is just for generation

lib/nerves_hub/application.ex
```diff
@@ -12,6 +12,10 @@ defmodule NervesHub.Application do
         raise "fwup could not be found in the $PATH. This is a requirement of NervesHubWeb and cannot start otherwise"
     end
 
+    :logger.add_handler(:my_sentry_handler, Sentry.LoggerHandler, %{
+      config: %{metadata: [:file, :line]}
+    })
```
jjcarstens, line 17: Leftover experimentation?

lib/nerves_hub_web/controllers/account_controller.ex
```diff
@@ -25,11 +25,9 @@ defmodule NervesHubWeb.AccountController do
   end
 
   def accept_invite(conn, %{"user" => user_params, "token" => token} = _) do
-    clean_params = whitelist(user_params, [:password, :username])
-
     with {:ok, invite} <- Accounts.get_valid_invite(token),
          {:ok, org} <- Accounts.get_org(invite.org_id) do
-      _accept_invite(conn, token, clean_params, invite, org)
+      _accept_invite(conn, token, user_params, invite, org)
```
jjcarstens, line 30: Is this atomize needed? It seems like we should just adjust that create function as Ecto is going to immediately convert to string when casting. I'm usually all for atoms, but were sort of blindly allowing anything to be `String.to_atom` via the implicit using macro which seems not great

lib/nerves_hub/application.ex
joshk: oops, yes, this was a Sentry improvement, but it wasn't supposed to be added here.

lib/nerves_hub_web/controllers/account_controller.ex
joshk: agree, also, converting to atoms is also a memory issue. I will clean this up.

jjcarstens (approved):

PR #1369, opened 2024-07-02
Allow for new user registrations

joshk:
This is built upon #1365 

Once we have this merged in we can start allowing people to sign up for NervesCloud.


lib/nerves_hub/application.ex
```diff
@@ -12,6 +12,10 @@ defmodule NervesHub.Application do
         raise "fwup could not be found in the $PATH. This is a requirement of NervesHubWeb and cannot start otherwise"
     end
 
+    :logger.add_handler(:my_sentry_handler, Sentry.LoggerHandler, %{
+      config: %{metadata: [:file, :line]}
+    })
```
jjcarstens: Another accidental?

lib/nerves_hub_web.ex
```diff
@@ -47,6 +47,10 @@ defmodule NervesHubWeb do
         |> Enum.filter(fn x -> !is_nil(params[to_string(x)]) end)
         |> Enum.into(%{}, fn x -> {x, params[to_string(x)]} end)
       end
+
+      def atomize_params(params) do
+        for {key, val} <- params, into: %{}, do: {String.to_atom(key), val}
+      end
```
jjcarstens, line 53: Same here?

priv/repo/migrations/20240701004550_change_username_to_name.exs
jjcarstens, line 1: I think this is left over from the previous pr

jjcarstens (commented):
I don't think the register route is in these changes?

joshk:
> I don't think the register route is in these changes?

they were never removed: https://github.com/nerves-hub/nerves_hub_web/blob/allow-for-new-user-registrations/lib/nerves_hub_web/router.ex#L193-L194

lib/nerves_hub_web/controllers/account_controller.ex
```diff
@@ -3,9 +3,31 @@ defmodule NervesHubWeb.AccountController do
 
   alias Ecto.Changeset
   alias NervesHub.Accounts
-  alias NervesHub.Accounts.SwooshEmail
+  alias NervesHub.Accounts.{User, SwooshEmail}
```
jjcarstens, line 6:
```suggestion
  alias NervesHub.Accounts.SwooshEmail
  alias NervesHub.Accounts.User
```
One thing we've been doing recently is actually removing alias expansions when we run into them. Its not a huge deal, but makes grepping/finding a little bit easier when future us has to poke around and change, so we opt for a few more lines as the tradeoff.

Not a huge deal, but food-for-thought

jjcarstens (approved):

lib/nerves_hub_web/controllers/account_controller.ex
joshk: cool, I'll keep this in mind and switch them over as I come across them

PR #1376, opened 2024-07-05
Move Archives to LiveView

joshk:
(based on top of #1375, so that should be merged in first)

A 1-for-1 LiveView implementation, with two added bits:

1. awesome progress reporting
2. you can actually delete archives now

BUT, I need some of your magic @jjcarstens, the failing test has me stumped


joshk:
Weird, it fails locally but passes on CI.

This is the error I see:

```
  1) test upload archive error if corrupt firmware uploaded (NervesHubWeb.Live.ArchivesTest)
     test/nerves_hub_web/live/archives_test.exs:105
     Expected path to be "/org/Jeff/CoolProduct/archives/upload" but got "/org/Jeff/CoolProduct/archives"

     code: |> assert_path("/org/#{org.name}/#{product.name}/archives/upload")
     stacktrace:
       (phoenix_test 0.3.1) lib/phoenix_test/assertions.ex:201: PhoenixTest.Assertions.assert_path/2
       test/nerves_hub_web/live/archives_test.exs:137: (test)
```

joshk:
my only idea is that the code to corrupt a firmware file doesn't behave the same on my mac (intel) (13.3)

joshk:
And fixed, after adjusting the block size of the `dd` command (used for corrupting `fw` files)

jjcarstens (approved):

PR #1382, opened 2024-07-08
Device location goodness

joshk:
Note: This PR is built upon #1380 and should be merged first

- save the request IP of the device to a new field, `connection_metadata`
- geolocate the device using the ip address
- display a map on the device page based on the geolocation

I also included:

- remove the 'alternative deployments' area in the Deployments section
- show a message if there is no assigned deployment
- don't show 'send update' if there are no firmware to send
- move tags into the Information section
- refresh the device list page (only if there is 1 page total)
- and move the sections about

image


fhunleth:
I'm very excited about this!

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -130,4 +119,38 @@ defmodule NervesHubWeb.DeviceSocket do
...
+      true ->
+        nil
+    end
+  end
+
+  defp x_forwarded_for(connect_info) do
+    (connect_info[:x_headers] || [])
+    |> Enum.find(fn header ->
+      elem(header, 0) == "x-forwarded-for"
+    end)
```
jjcarstens, line 144:
```suggestion
    (connect_info[:x_headers] || [])
    |> Enum.find(
      fn {"x-forwarded-for", val} ->
        hd(String.split(val, ",")
        _ -> nil
    end)
```

jjcarstens (approved):

PR #1383, opened 2024-07-10
Dialyzer fixes, Ecto queries, and connection status broadcasts

joshk:
- **Dialyzer fixes**

- **Separate connection change and connection check**

this makes it clearing if the device is coming online, or is already online but we just needed to check.

this also allows us to decide when to reload the device, as a `connection:status: is just a status confirmation, but a `connection:change` could include more information, like a firmware change, so we need to reload the device.

- **prefer join + preload vs just preload**

with join + preload you can use the data from the join, but using just preload will cause an extra query


jjcarstens (approved):
Wow. Dialyzer has been a long time coming. Thanks!

joshk:
I haven't added it to the CI pipeline yet, but its a start

PR #1384, opened 2024-07-10
Console layout bug

joshk:
This bug was introduced when fixing dialyzer errors.

### DIFF
--- a/lib/nerves_hub_web/templates/layout/console.html.heex
+++ b/lib/nerves_hub_web/templates/layout/console.html.heex
@@ -7,7 +7,7 @@
     <meta name="description" content="" />
     <meta name="author" content="" />
 
-    <title>Device Console : <%= @device.identifier %></title>
+    <title><%= String.slice(@device.identifier, -6..-1) %> - Console Session</title>
     <link rel="stylesheet" href={Routes.static_path(@conn, "/css/app.css")} />
   </head>
 
--- a/lib/nerves_hub_web/templates/layout/console.html.heex
+++ b/lib/nerves_hub_web/templates/layout/console.html.heex
@@ -7,7 +7,7 @@
     <meta name="description" content="" />
     <meta name="author" content="" />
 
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/templates/layout/console.html.heex
```diff
@@ -7,7 +7,7 @@
     <meta name="description" content="" />
     <meta name="author" content="" />
 
-    <title><%= @device.identifier %></title>
+    <title>Device Console : <%= @device.identifier %></title>
```
fhunleth, line 10: When you have a lot of tabs, does this cause all of your console tabs to show "Device C..." rather than an abbreviated serial number?

joshk: I would tend to argue that maybe you have a few too many consoles open, but I also see how that can happen.

I'm happy to hear suggestions.

jjcarstens: I actually find the device identifier valuable there. When I'm in a lot of consoles, I use the browser sidebar to keep track of them and switch around.

image

joshk: Would something like `[last 6 characters] - Console` or `[last 6 characters] - Device Console` work?

fhunleth: Seems good to me. If I hate it, I'll complain. ;)

fhunleth: Scratch that. I know the line of code, so I'll just send a PR to make it the way I like.

joshk: hahahahaha

I'll update the PR today.

joshk:
@fhunleth @jjcarstens updated, please review

fhunleth (approved):

PR #1386, opened 2024-07-11
sync a devices connection status

joshk:
- sync when a device establishes a connection, when it disconnects, and its 'heartbeat' (powered by the device channel)
- sync 'connected', 'disconnected', and 'not_seen' based on the device connecting, disconnecting, and the heartbeat
- add a connection status filter to the device list
- an oban worker that runs each minute to clean up device connection statuses if they haven't been seen in 6 minutes (configurable)
- updated the device show UI to include the last connected at and last seen at

This change will mean that each connected device will write to the devices table once each 5 minutes (configurable) based on a heartbeat timer in `DeviceChannel`. 

While this change isn't expected to be a performance issue, the impact is a little unknown. For example, 100,000 devices updating every 5 minutes would mean 333 DB updates every second. It's very likely the `DEVICE_LAST_SEEN_UPDATE_INTERVAL_MINUTES` environment variable will need to be tweaked to find the right setting for a platform's installation.

Screenshot 2024-07-11 at 4 23 35 PM

### DIFF
--- a/api-spec.yml
+++ b/api-spec.yml
@@ -1,7 +1,7 @@
-openapi: '3.0.2'
+openapi: "3.0.2"
 info:
   title: NervesHub Web
-  version: '2.0'
+  version: "2.0"
 servers:
   - url: http://localhost:4000/api
 security:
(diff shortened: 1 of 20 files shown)


api-spec.yml
```diff
@@ -1,7 +1,7 @@
-openapi: '3.0.2'
+openapi: "3.0.2"
```
fhunleth, line 1: I don't know why this should be single quotes or double quotes, but I feel capable of reviewing/merging the change in a separate PR.

joshk: ah, sorry, this was my code editor, Zed, formatting on save.

joshk:
@fhunleth FWIW, I'll be having a bigger discussion with Jon about this PR as there are legitimate concerns that the updating could be a bit much.

PR #1392, opened 2024-07-11
Rely on the power of `Ecto.NoResultsError` for 404s

joshk:
by using Ectos `!` functions, we can use Phoenix's error handling to show 404 pages.

This is cleaner, and follow convention.

This PR also includes some Sentry tidy ups, including the ignoring of all errors with a code below 500.

### DIFF
--- a/lib/nerves_hub_web/errors.ex
+++ b/lib/nerves_hub_web/errors.ex
@@ -1,19 +1,3 @@
 defmodule NervesHubWeb.NotFoundError do
   defexception message: "not found", plug_status: 404
 end
-
-defmodule NervesHubWeb.SentryEventFilter do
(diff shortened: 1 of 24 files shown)


lib/nerves_hub_web/errors.ex
```diff
@@ -1,3 +1,19 @@
 defmodule NervesHubWeb.NotFoundError do
   defexception message: "not found", plug_status: 404
 end
+
+defmodule NervesHubWeb.SentryEventFilter do
```
jjcarstens, line 5: If I'm being totally honest, I don't love jamming multiple modules of the same level into files. Especially in this case where I had to really work to find this module in the `errors.ex` which I really didn't even think to look in cause I assumed this file was only for exceptions/errors

jjcarstens (approved):

PR #1404, opened 2024-07-15
add a test for 'set tags' on the Devices page

joshk:
I thought it wasn't working for me, so I added a test, but it seems to be working.

None the less, tests are good

### DIFF
--- a/lib/nerves_hub_web/live/devices/index.html.heex
+++ b/lib/nerves_hub_web/live/devices/index.html.heex
@@ -156,7 +156,7 @@
         </form>
 
         <form id="bulk-tag-input" class="col-lg-6" phx-submit="tag-devices" phx-change="validate-tags">
-          <label for="move_to">Set tag(s) to:</label>
+          <label for="input_tags">Set tag(s) to:</label>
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/live/devices/index.html.heex
```diff
@@ -156,7 +156,7 @@
         </form>
 
         <form id="bulk-tag-input" class="col-lg-6" phx-submit="tag-devices" phx-change="validate-tags">
-          <label for="move_to">Set tag(s) to:</label>
+          <label for="input_tags">Set tag(s) to:</label>
```
fhunleth, line 159: This looks like it might be a fix or it might be an accidental commit.

joshk: nope, that is correct, the `for` needs to match the id of the input it is linked to. It was wrong, now its right.

fhunleth: This change is unrelated to the test, right?

fhunleth (approved):

PR #1408, opened 2024-07-16
Address dialyzer warnings

joshk:
Now we have no dialyzer warnings!

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -41,7 +41,9 @@ defmodule NervesHub.MixProject do
         :logger,
         :os_mon,
         :runtime_tools,
-        :timex
+        :timex,
+        :crypto,
+        :public_key
       ]
     ]
   end
--- a/mix.exs
+++ b/mix.exs
@@ -23,7 +23,7 @@ defmodule NervesHub.MixProject do
         ]
       ],
       dialyzer: [
-        plt_add_apps: [:ex_unit]
+        plt_add_apps: [:ex_unit, :mix]
       ]
     ]
   end
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -1,0 +1,4 @@
+[
+  {"lib/x509/certificate.ex", "Unknown type: X509.ASN1.record/1."},
+  {"lib/x509/certificate/extension.ex", "Unknown type: X509.ASN1.record/1."}
+]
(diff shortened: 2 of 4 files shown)


mix.exs
```diff
@@ -23,7 +23,7 @@ defmodule NervesHub.MixProject do
         ]
       ],
       dialyzer: [
-        plt_add_apps: [:ex_unit]
+        plt_add_apps: [:ex_unit, :mix]
```
fhunleth, line 26: This is weird. In fact, I'd remove `:ex_unit`, too. The `:mix` addition here makes it sound like `Mix.env()` is being called in a place where `Application.compile_env/3` should be called instead.

mix.exs
```diff
@@ -41,7 +41,9 @@ defmodule NervesHub.MixProject do
         :logger,
         :os_mon,
         :runtime_tools,
-        :timex
+        :timex,
+        :crypto,
+        :public_key
```
fhunleth, line 46: Almost all of these `:extra_applications` should be automatically included because they're Elixir libraries or transitively included by dependencies.

Here's a list that works for me:

```
      extra_applications: [
        :logger,
        :os_mon,
        :runtime_tools
      ]
```

mix.exs
fhunleth: I see why `:ex_unit` is in there. It's because of the test infrastructure which looks like it's useful for development too.

fhunleth:
I really like that you're cleaning up Dialyzer warnings!!!

At the risk of sounding ungrateful, the default Dialyzer checks are pretty weak. Here are the ones I normally enable:

```
      dialyzer: [
        flags: [:missing_return, :extra_return, :unmatched_returns, :error_handling, :underspecs],
        plt_add_apps: [:ex_unit, :mix]
       ]
```

Just looking at the results, there are a lot of unmatched returns. Perhaps it's a second pass to enable that flag.

mix.exs
joshk: `:mix` was also required due to some Mix tasks.

mix.exs
joshk: I can remove the extra ones. `:public_key` was added due to a `next-ls` set of warnings, but that wasn't specifically dialyzer.

joshk:
I'm happy to do a second pass with your updates.

fhunleth:
Spoke offline. Going to merge this set since they're going in the right direction and keep iterating in future PRs.

PR #1420, opened 2024-07-18
`FWUP` metadata spec and docs

joshk:


lib/nerves_hub/fwup.ex
```diff
@@ -23,10 +32,12 @@ defmodule NervesHub.Fwup do
       }
 
       {:ok, metadata}
+    else
+      {:error, {_, :not_found}} -> {:error, :invalid_metadata}
```
fhunleth, line 36: Since `metadata_value/2` is private, how about changing it to return `{:error, :invalid_metadata}` so the `else` clause can be deleted? This would also fix a crash if `get_metadata/1` returns an error.

joshk: i can do that, but then i should also apply some similar logic to `get_metadata` as that returns all possible errors from `System.cmd`, thoughts?

joshk: Done. Ready for re-review.

lib/nerves_hub/fwup.ex
```diff
@@ -26,34 +37,34 @@ defmodule NervesHub.Fwup do
...
-      {error, _} ->
-        {:error, error}
+      {_error, _} ->
+        {:error, :invalid_fwup_file}
     end
   end
 
-  def metadata_value(metadata, key) when is_binary(key) do
+  defp metadata_value(metadata, key) when is_binary(key) do
     {:ok, regex} = "#{key}=\"(?<value>[^\n]+)\"" |> Regex.compile()
```
fhunleth, line 51: I suppose whoever wrote this owned up to explicitly compiling the regex every call. I know this isn't the intention of this particular PR, but it seems like moving this to a regex sigil would not only reduce the amount of code, but save a few nanoseconds in not doing all the recompiles.

fhunleth: Oh, I see why they did it. Nevermind. I would have redone the code so that the fwup metadata is returned in a map, but I guess this works.

fhunleth (approved):
The code here hurts me, but your PR is an improvement so I'm approving.

joshk:
> The code here hurts me, but your PR is an improvement so I'm approving.

I was going to look into improving the code, but then I found that this code is duplicated in `firmwares.ex`, and I gave up for now. I'll do this in another PR.

PR #1423, opened 2024-07-21
Fwup metadata deduplication + tests

joshk:
This is for @fhunleth 

I removed the multiple Regex compiles and named captures and switched to a single compile and single run.

I also removed duplicated code in `Firmwares` which was doing the exact same thing, line for line.

And added tests.

### DIFF
--- a/lib/nerves_hub/fwup.ex
+++ b/lib/nerves_hub/fwup.ex
@@ -3,6 +3,11 @@ defmodule NervesHub.Fwup do
   Helpers for dealing with files created by FWUP.
   """
 
+  @required_metadata_keys [:architecture, :platform, :product, :uuid, :version]
+  @optional_metadata_keys [:author, :description, :misc, :vcs_identifier]
+
+  @metadata_regex Regex.compile!("meta-(?<key>[^\n]+)=\"(?<value>[^\n]+)\"")
+
   @doc """
   Decode and parse metadata from a FWUP file.
   """
@@ -12,28 +17,11 @@ defmodule NervesHub.Fwup do
           | {:error, :invalid_metadata}
   def metadata(file_path) do
     with {:ok, metadata} <- get_metadata(file_path),
-         {:ok, uuid} <- metadata_value(metadata, "meta-uuid"),
-         {:ok, architecture} <- metadata_value(metadata, "meta-architecture"),
-         {:ok, platform} <- metadata_value(metadata, "meta-platform"),
-         {:ok, product} <- metadata_value(metadata, "meta-product"),
-         {:ok, version} <- metadata_value(metadata, "meta-version"),
-         {:ok, author} <- metadata_value(metadata, "meta-author", nil),
-         {:ok, description} <- metadata_value(metadata, "meta-description", nil),
-         {:ok, misc} <- metadata_value(metadata, "meta-misc", nil),
-         {:ok, vcs_identifier} <- metadata_value(metadata, "meta-vcs-identifier", nil) do
-      metadata = %{
-        architecture: architecture,
-        author: author,
-        description: description,
-        misc: misc,
-        platform: platform,
-        product: product,
-        uuid: uuid,
-        vcs_identifier: vcs_identifier,
-        version: version
-      }
-
-      {:ok, metadata}
+         parsed_metadata <- parse_metadata(metadata),
+         {:ok, required_metadata} <- required_values(parsed_metadata),
+         optional_metadata <- optional_values(parsed_metadata),
+         complete <- Map.merge(required_metadata, optional_metadata) do
+      {:ok, complete}
     end
   end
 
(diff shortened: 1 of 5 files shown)


fhunleth (approved):
So much nicer!

joshk:
I'm working on an improvement that uses structs for the metadata. I'll ping you for a review.

joshk:
@fhunleth done, re-review please. 

I think this is nicer, plus it gives a defined structure to the Fwup metadata.

lib/nerves_hub/fwup.ex
```diff
@@ -47,25 +45,25 @@ defmodule NervesHub.Fwup do
     end
   end
 
-  defp metadata_value(metadata, key) when is_binary(key) do
-    {:ok, regex} = "#{key}=\"(?<value>[^\n]+)\"" |> Regex.compile()
+  defp parse_metadata(metadata) do
+    Regex.scan(@metadata_regex, metadata)
```
fhunleth, line 49:
```suggestion
    Regex.scan(~r/meta-(?<key>[^\n]+)=\"(?<value>[^\n]+)\"/, metadata)
```
`~r` is precompiled so the regex can be here if you like that better.

lib/nerves_hub/fwup.ex
```diff
@@ -47,25 +45,25 @@ defmodule NervesHub.Fwup do
...
+    |> Enum.reduce(%{}, fn line, acc ->
+      [_, key, value] = line
 
-    case Regex.named_captures(regex, metadata) do
-      %{"value" => value} ->
-        {:ok, value}
+      key =
+        key
+        |> String.replace("-", "_")
+        |> String.to_atom()
```
fhunleth, line 56: `String.to_atom/1` always makes me think really hard when seeing it. I feel like this has to be safe, but I'd probably have a string->atom map to explicitly skip over unknown keys. This seems useful if new metadata keys ever get added to `fwup`.

test/support/fwup.ex
```diff
@@ -126,4 +152,18 @@ defmodule NervesHub.Support.Fwup do
...
     """
   end
+
+  defp build_conf_contents(%InvalidMetaParams{} = meta_params) do
+    """
+    meta-description = "#{meta_params.description} "
+    meta-version = "#{meta_params.version}"
+    meta-platform = "#{meta_params.platform}"
+    meta-architecture = "#{meta_params.architecture}"
+    meta-author = "#{meta_params.author}"
```
fhunleth, line 162: Is this file invalid since it doesn't have a product? `fwup` doesn't enforce any metadata except for UUID which is automatically generated, but it does seem that requiring at least product, platform, architecture, and version are reasonable for NH to do.

lib/nerves_hub/fwup.ex
```diff
@@ -47,25 +45,25 @@ defmodule NervesHub.Fwup do
...
-      {:ok, metadata_item} ->
-        {:ok, metadata_item}
-
-      {:error, :invalid_metadata} ->
-        {:ok, default}
-    end
+  defp transform_to_struct(metadata) do
+    keys = Map.keys(Map.from_struct(Metadata))
+    filtered = Map.take(metadata, keys)
+    {:ok, struct!(Metadata, filtered)}
```
jjcarstens, line 65: I thought `struct/2` supported passsing string keys and also ignored any unknown keys so there shouldn't be a need to filter

Also, what is the goal here? To completely blow up when the required keys aren't there or to return an error that the metadata is invalid?

If it's the latter, then we could simply use `embedded_schema` and run all of this through the Ecto tooling to cast and enforce

lib/nerves_hub/fwup.ex
joshk: Keys need to be atoms, so this is needed

lib/nerves_hub/fwup.ex
joshk: https://hexdocs.pm/elixir/Kernel.html#struct/2

as shown in the example:

```
# String keys are ignored
struct(User, %{"name" => "meg"})
#=> %User{name: "john"}
```

So, we need to use `struct!` so that the struct enforces keys. But, the one downside of `struct!` is that any unrecognized keys will also raise an error, so we first need to filter out keys we dont recognize, and then use `struct!` so that `@enforce_keys` is, well, enforced.

test/support/fwup.ex
joshk: if you look at the current version of this code, https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/fwup.ex#L18 and https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/fwup.ex#L50-L60, you will see that product is required and if it isn't found in the metadata we return `{:error, :invalid_metadata}`

joshk: In the future, I would like to relax the requirement for product metadata to exist. I think that should be a Product setting, eg. `Strict firmware upload checking`.

lib/nerves_hub/fwup.ex
jjcarstens: I mistakenly thought this was a mixed behavior or exceptions and error tuple, but there is a rescue clause to handle it so all good 👍

PR #1424, opened 2024-07-23
Fixes to Dialyzer warnings from recent PRs

joshk:


lib/nerves_hub/fwup.ex
```diff
@@ -22,9 +22,8 @@ defmodule NervesHub.Fwup do
   Decode and parse metadata from a FWUP file.
   """
   @spec metadata(String.t()) ::
-          {:ok, map()}
-          | {:error, :invalid_fwup_file}
-          | {:error, :invalid_metadata}
+          {:ok, struct()}
```
fhunleth, line 25:
```suggestion
          {:ok, Metadata.t()}
```
Can you make the `t()` work here so Dialyzer can know about the fields?

lib/nerves_hub/products.ex
```diff
@@ -182,7 +182,7 @@ defmodule NervesHub.Products do
     end
   end
 
-  @spec load_shared_secret_auth(Product.t()) :: Product.t()
+  @spec load_shared_secret_auth(Product.t()) :: Product.t() | [Product.t()] | nil
```
fhunleth, line 185: This change probably has ramifications, but when I hit this situation, I like calling `List.wrap/1` at the end. That way the return would only be `[Product.t()]` and you'd have an empty list if nothing was found rather than `nil`.

fhunleth (approved):

lib/nerves_hub/products.ex
jjcarstens: I don't even understand how it can return nil or a list of product? Shouldn't it be product arg and product return? If you're getting a dialyzer error here, maybe it is being called incorrectly somewhere else?

joshk:
It’s because of Repo.preload. Those are possible return types for that
function

lib/nerves_hub/products.ex
joshk: It’s because of `Repo.preload` 

Those are the possible return types for that function.

joshk:
@jjcarstens I've merged in this PR, along with https://github.com/nerves-hub/nerves_hub_web/pull/1424/commits/4890374310e952aa6cb92f793e841b05c1e73b26. 

Although I'm not super in love with it, it does get around the issue with dialyzer complaining about `nil` or `[%Product{}]` being return types.

jjcarstens:
Thanks! I started looking into it after merge and then never came back with a 👍. All is well for the time being

PR #1429, opened 2024-07-24
Use hooks to update the Device connection times

joshk:
This is partially cherry picked from the Health PR

Screenshot 2024-07-25 at 11 16 07 AM

The last connected at and last seen at now update using JS, eg. "1 minute ago" -> "2 minutes ago"


joshk:
Confirmed this was working locally and on NervesCloud

PR #1439, opened 2024-08-05
Remove `Application.load/1` in migrations

oestrich:
We're running in a release and the match added for `Application.load`
will fail with `{:error, {:already_loaded, :nerves_hub}}` causing
migrations to fail.


oestrich:
Yeah, we can't migrate with `nerves_hub eval "NervesHub.Release.Tasks.migrate()"` at the moment from the latest commit on `main`

fhunleth:
Could you update the comment and PR, so that it's more obvious that this is a fix rather than code cleanup?

PR #1440, opened 2024-08-05
Don't send `:health_check` on device connect

oestrich:
Devices with older NervesHubLink versions will crash because it doesn't know what a `check_health` message is and doesn't have the catch all. If the NervesHub instance has the feature disabled it will still send a message because this call was not checked for application config. If the instance does have it enabled then it will be scheduled and can start collecting health metrics as expected.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -99,8 +99,10 @@ defmodule NervesHubWeb.DeviceChannel do
 
     send(self(), :boot)
 
-    send(self(), :health_check)
-    schedule_health_check()
+    if device_health_check_enabled?() do
+      send(self(), :health_check)
+      schedule_health_check()
+    end
 
     {:noreply, socket}
   end


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -99,7 +99,6 @@ defmodule NervesHubWeb.DeviceChannel do
 
     send(self(), :boot)
 
-    send(self(), :health_check)
     schedule_health_check()
```
jjcarstens, line 102: Can we add an argument to `schedule_health_check` to allow immediately scheduling the health check to keep this same behavior if it is enabled? Otherwise we're potentially minutes out to getting a check for those that have it enabled

joshk: +1 to what Jon is suggesting.

oestrich: I really didn't like how that looked so I added a check around the block that was there

oestrich:
This feature also needs to be device api versioned like the other features in this channel once we merge the PR in link

jjcarstens (approved):
Agreed. I think we'll either get a version check in or go with the idea of devices reporting what extensions they support and NervesHub can hold that in context to know what to send to devices

PR #1441, opened 2024-08-05
Remove migration that removes `last_communication` field

oestrich:
This causes genserver crashes when deploying to a large production environment because the field is still being used on running servers while the rollout is occurring.


joshk:
@oestrich thank you for catching this.

I'm good with this change. Can you please create a PR with this migration added back which we can delay merging in till a later date?

joshk:
(we have already run this migration on NervesCloud, so keeping the same migration file will be important so that Ecto is smart enough to skip it as it knows it has already run it)

oestrich:
I don’t think we should redo this migration because it will just strike the next person. There’s a good number of fields that are hanging around unused because of stuff like this. If we bring it back we need something documenting what stages to deploy to. Otherwise something like this will happen again

jjcarstens:
I think we should be fine having the removal of a column. The column is unneeded and the problem was removing it while processes of old nodes might still be using (even then, it's a very temporary flood of errors and not the end of the world)

Applying the removal in a follow-up deploy separated by a commit should be no problem

fhunleth:
I'm not the expert on migrations, but isn't what Eric saying is that anyone else running NervesHub will run into this same issue. NervesCloud and SmartRent will be past it, but company X might not be. If we merge the column deletion, then that will mess them up when they upgrade. Or do we know that there are no other multi-node NervesHub users out there that we need to care about?

It seems like there should be a way to clean up unused columns. I hate the idea of leaving cruft like this around forever when it's such a temporary need.

oestrich:
Whenever we get a changelog we should include migrations like this and what docker images should be deployed in what steps. That way we're not breaking (even temporarily) everyone else but the two instances being mentioned here

PR #1442, opened 2024-08-06
Don't broadcast device connection changes

oestrich:
When performing a heartbeat we were noticing a matching volume of select statements for a device by id. This is likely the heartbeat broadcasting out a device change and the device channel reloading based on its own message.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -576,28 +576,40 @@ defmodule NervesHub.Devices do
   end
 
   def device_connected(device) do
-    update_device(device, %{
-      connection_status: :connected,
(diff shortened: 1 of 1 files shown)


jjcarstens (commented):
Do we need these broadcasts for the UI to be updated without refresh?

Do we know where the extra queries are coming from specifically? If it is the device channel, should we just adjust to be `broadcast_from!` so the message is excluded?

oestrich:
In talking with Josh last night he thought the broadcasts weren’t needed for the UI. I’m also fine updating to broadcast from to avoid getting any other surprise messages

jjcarstens (approved):

PR #1443, opened 2024-08-06
Remove `last_communication` and allow more migration control

jjcarstens:
Adding the migration for removing `last_communication` which was included in https://github.com/nerves-hub/nerves_hub_web/pull/1386 and then removed by https://github.com/nerves-hub/nerves_hub_web/pull/1441 in order to support deploying the removal in a separate step following the guide in https://fly.io/phoenix-files/migration-recipes/#removing-a-column

This also adds more migration control for others using the mix release so that they can step through whatever migration versions needed, whether via commit isolation or manually utilizing the migration task.

This can be separated into 2 PRs, but they seemed related enough to review together

### DIFF
--- a/lib/nerves_hub/release/tasks.ex
+++ b/lib/nerves_hub/release/tasks.ex
@@ -7,14 +7,25 @@ defmodule NervesHub.Release.Tasks do
     start_apps_before_migration: [:logger, :ssl, :postgrex, :ecto_sql]
   ]
 
-  def migrate() do
+  @doc """
+  Run the Ecto.Migrator for each defined repo
+
+  You can optionally pass options supported Ecto.Migrator.run/3 for each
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/release/tasks.ex
```diff
@@ -7,14 +7,25 @@ defmodule NervesHub.Release.Tasks do
...
 
-  def migrate() do
+  @doc """
+  Run the Ecto.Migrator for each defined repo
+
+  You can optionally pass options supported Ecto.Migrator.run/3 for each
+  repo in order to control the migration a bit more, like for specifying
+  a migration to stop at:
+
+    migrate([{NervesHub.Repo, [to: 20240806112233]}])
```
oestrich, line 17: 🎉

oestrich (approved):

PR #1446, opened 2024-08-09
Use `ssl: options` for postgrex config to fix deprecation

### LINKED ISSUE #1445: `ssl_opts` deprecated and db server certificate not being verified
**Describe the bug**
After `postgrex` was [upgraded to 0.18.0](https://github.com/nerves-hub/nerves_hub_web/pull/1305), `ssl_opts` was deprecated in favor of `ssl` and `ssl` set to true does not verify server certificate. If previously `verify_peer` was configured in `ssl_opts`, we're now getting warnings:
```
setting ssl: true on your database connection offers only limited protection, as the server's certificate is not verified. Set "ssl: [cacertfile: path/to/file]" instead
```
Refs:
https://hexdocs.pm/postgrex/changelog.html#v0-18-0-2024-05-18

**To Reproduce**
1. Pass in `DATABASE_PEM` to verify db server cert
2. See warnings in logs on startup 

**Expected behavior**
Verify database server certificate. No warnings.  

**Additional context**
https://github.com/nerves-hub/nerves_hub_web/blob/5e6b3370e158e83a418e2530115e137971a23542/config/runtime.exs#L205-L222

oestrich:
Per postgrex changelog from 0.18.0
https://hexdocs.pm/postgrex/changelog.html#v0-18-0-2024-05-18

Fixes #1445

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -178,34 +178,37 @@ end
 #
 if config_env() == :prod do
   database_ssl_opts =
-    if System.get_env("DATABASE_PEM") do
-      db_hostname_charlist =
-        ~r/.*@(?<hostname>[^:\/]+)(?::\d+)?\/.*/
-        |> Regex.named_captures(System.fetch_env!("DATABASE_URL"))
-        |> Map.get("hostname")
-        |> to_charlist()
+    if System.get_env("DATABASE_SSL", "true") == "true" do
+      if System.get_env("DATABASE_PEM") do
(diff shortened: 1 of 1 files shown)


joshk (changes requested):
I believe there needs to be a way to disable ssl even for production setups. This requires `ssl: false`, which is also the default.

joshk (approved):
Awesome! Thank you @oestrich !

PR #1448, opened 2024-08-10
update readme to use email address for default dev user

thejaywhy:
username was changed to email in PR #1365


jjcarstens (approved):
Thanks! 🙏

PR #1457, opened 2024-08-16
Add dashboard with devices marked on world map

elinol:
Adds a dashboard to project site, with device locations marked on world map.

The dashboard is disabled from menu by default and is enabled with application variable `:enable_dashboard?`


lawik:
We should add a way to set the setting in runtime.exs via en environment variable as well. Forgot about that.

elinol:
image

lib/nerves_hub_web/live/dashboard/index.html.heex
```diff
@@ -0,0 +1,9 @@
+<link rel="stylesheet" href="https://unpkg.com/leaflet@1.7.1/dist/leaflet.css"
+    integrity="sha512-xodZBNTC5n17Xt2atTPuE1HxjVMSvLVW9ocqUKLsCC5CXdbqCmblAshOMAS6/keqq/sMZMZ19scR4PsZChSR7A=="
+    crossorigin="" />
+<script src="https://unpkg.com/leaflet@1.7.1/dist/leaflet.js"
```
lawik, line 4: We should probably add this to package.json via npm install instead :/ 

This was my quick way of getting started. It should be able to provide both the leaflet.js and the CSS.

We may want to put it outside of app.js and make it like a map.js that we only load on that view?

lawik: Not sure it's necessary at this stage, but later?

mix.exs
```diff
@@ -79,6 +79,7 @@ defmodule NervesHub.MixProject do
       {:gettext, "~> 0.24.0"},
       {:hackney, "~> 1.16"},
       {:hlclock, "~> 1.0"},
+      {:hpax, "~> 1.0", override: true},
```
lawik, line 82: This was some dep acting up, not sure why.

lawik (commented):
Will let Josh and Jon check in. I just need the runtime.exs so we can deploy :D

lawik:
If we want to hold off on merging this while we tweak it, Josh and I could always deploy it from branch to get the demo done. Do need the runtime.exs for that.

lib/nerves_hub_web/live/dashboard/index.html.heex
elinol: let's do it now!

mix.exs
elinol: Removed in latest commit

elinol:
Runtime settings and leaflet npm package was added with latest commit

joshk (changes requested):
I love where this is going, and think this is a great foundation.

I think for a future iteration we could look into using the Mapbox JS lib (https://docs.mapbox.com/mapbox-gl-js/example/cluster/) as we can do useful things like cluster counts when at different zoom levels, as well as offering a satellite view.

We could also pre zoom in based on the bounding box of the map markers, so if you only have a handful of points in an isolated area, we can zoom in. 

One thing before we merge and ship, what does the 'empty state' look like? Can we show a message over the map to guide users on adding their first device. And we might want to show a legend of what red and green dots means.

lawik:
Leaflet has plugins that do clustering and pretty much everything else. It should be able to do the fit-zoom around the marker layer though I haven't checked. Mapbox seems to require API access to even use their library and I think that's mildly detrimental to using it in an open source project.

There seems to be an open fork: https://github.com/maplibre/maplibre-gl-js

Mapbox us fine for the detail view since we need tile data.

What makes you prefer them? I am open to the idea but not clear on why they'd be generally preferable :)

lawik:
Good call on legend and empty state of course.

elinol:
image
image

joshk (approved):
Looks good to me!

It would be good to add some form of liveview test for this page, even just to check it shows the empty state if there are no devices with location data, and a map with a dot if there are devices with location data.

Also, as a follow up PR, it would be good to add a summary of how many devices are shown on the map, and how many don't have location data.

eg. 50 devices aren't shown on the map as they don't have any location data.

PR #1458, opened 2024-08-16
Change devices list to use query params

lawik:
Filtering and pagination will now be persisted to the query params which also means we get history and linkable filtering.

### DIFF
--- a/lib/nerves_hub_web/live/devices/index.ex
+++ b/lib/nerves_hub_web/live/devices/index.ex
@@ -74,6 +74,8 @@ defmodule NervesHubWeb.Live.Devices.Index do
     }
 
     socket
+    |> assign(:current_sort, Map.get(unsigned_params, "sort", "identifier"))
+    |> assign(:sort_direction, Map.get(unsigned_params, "sort_direction", :asc) |> sort_dir())
     |> assign(:current_filters, filters)
     |> assign(:paginate_opts, pagination_opts)
     |> assign(:currently_filtering, filters != @default_filters)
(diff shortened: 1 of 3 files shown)


lawik:
I was terribly annoyed by how this view worked. So I fixed it :)

There may be some corners I've missed but I think this is the path forward. I didn't spend any effort on generalizing the pattern which we might want to later.

lawik:
Ach, think I missed `sort`

lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -51,9 +54,90 @@ defmodule NervesHubWeb.Live.Devices.Index do
...
+      end)
+
+    pagination_opts = %{
+      page_number:
+        Map.get(unsigned_params, "page_number", socket.assigns.paginate_opts.page_number) |> num(),
+      page_size:
+        Map.get(unsigned_params, "page_size", socket.assigns.paginate_opts.page_size) |> num(),
+      page_sizes: socket.assigns.paginate_opts.page_sizes,
+      total_pages: socket.assigns.paginate_opts.total_pages
+    }
```
jjcarstens, line 74: We do a lot of map rebuilding and custom filtering here and around this file now. We could potentially leverage Ecto to aid in this casting and safe params:

```elixir
defaults = %{
    page_number: 1,
    page_size: 25,
    page_sizes: [25, 50, 100],
    total_pages: 0
  }

types = %{
    page_number: :integer,
    page_size: :integer,
    page_sizes: {:array, :integer},
    total_pages: :integer
  }

# Safely cast, filter non-allowed params, and set defaults
Ecto.Changeset.cast({defaults, types}, params, Map.keys(defaults))
|> Ecto.Changeset.apply_changes()

# Can also be used to only get the changes from the defaults
Ecto.Changeset.cast({defaults, types}, params, Map.keys(defaults)).changes
```

lawik: Cool, will look at that because I'm not happy about the map wrangling. But the functionality is sweet :)

jjcarstens: O yes, you are going to make a lot of people cry the tears of joy with this 😂

lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -51,9 +54,90 @@ defmodule NervesHubWeb.Live.Devices.Index do
...
+                # Remove all default pagination options
+                value == val or value == to_string(val)
+              else
+                false
+              end
+            end
+        end
+      end)
+
+    NervesHubWeb.Router.Helpers.live_path(
```
joshk, line 134: can't we use a verified route string for this?

joshk (changes requested):
This is awesome

If you want to take it one step further you can change from using lots of `handle_event`s for all the filtering and sorting to instead using patch links in the html.

lib/nerves_hub_web/live/devices/index.ex
lawik: I can try. The most important bit here is that we clean out all default values so we don't get a lot of empty query params.

lawik:
Good suggestions that just slot in and make the darned thing better. Fancy that.

lawik:
Not tackling patch links right now, though I imagine that'd be rad. Just hit the sort that I had broken again. Think this is good to go.

joshk (approved):
aaaaaaaamazing!

PR #1461, opened 2024-08-19
Safe access when checking uuids in `Devices.firmware_status/1`

jjcarstens:
This is mostly for race conditions between finishing the DB write of related firmware metadata or deployment update and reading the record back in with a firmware_status (i.e via API request) In those rare cases we might have missing data and need to be able to handle it

Discovered when hitting production with a series of API updates followed by reads


lawik (approved):
Looks very reasonable. I still haven't gotten use to the macro versions of `get_in`.

joshk (approved):

PR #1467, opened 2024-08-25
Prevent warnings and sentry errors in unhandled message

lawik:


jjcarstens (approved):

joshk:
This is a great fix.

Since this message is being sent by the `DeviceChannel`, is there a way to ignore this message coming from itself?

Would `broadcast_from` help at all?

lawik:
Good thinking. We should look into that.

PR #1470, opened 2024-08-27
Handle unknown messages in dashboard liveview

elinol:


lib/nerves_hub_web/live/dashboard/index.ex
```diff
@@ -46,6 +46,9 @@ defmodule NervesHubWeb.Live.Dashboard.Index do
     |> noreply()
   end
 
+  # Ignore unknown messages
```
joshk, line 49:
```suggestion
  # Ignore unused messages from the subscription
```

joshk (approved):
Great find!

I have one suggestion, then happy to merge in.

PR #1476, opened 2024-08-28
Queue deployment recalculation

oestrich:
Create inflight deployment checks to verify a device is still valid or
is now valid for a deployment. These are processed in a single genserver
per deployment per node. This will help prevent database pool contention
since there should only be a single query per device check per node.

Devices are still checked as fast as possible and will continue to look
for more inflight checks without being told of new updates.

Because deployment updates are processed inside a transaction, any
pubsub messages must be delayed until after the transaction is
successful.

Finding the next device to check is performed with a postgres lock
inside of a transaction. This way we only process a device once.


jjcarstens (commented):
Thanks! I need a bit to review this. It seems there are 2 big changes in here and I'm still catching up how they correlate together. It also is very aggressively locking the rate device updates can be checked. Is this something we could have a knob to tweak at runtime in order to increase the flow if needed?

Also, there is a lot of nesting conditions and blocking. It takes a bit to grok through it to understand what exactly is happening

oestrich:
We purposefully don't want to increase the rate that we check devices to match a deployment though, since that's what was causing issues. There's no reason we couldn't have multiple calculator processes going at the same time on each node other than running into ecto pooling issues so I've left it at a single per node

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -150,95 +150,12 @@ defmodule NervesHubWeb.DeviceChannel do
...
-  end
-
-  # manually pushed
-  def handle_info(
-        %Broadcast{event: "deployments/update", payload: %{deployment_id: nil} = payload},
-        socket
-      ) do
-    :telemetry.execute([:nerves_hub, :devices, :update, :manual], %{count: 1})
-    push(socket, "update", payload)
+  # TODO bring back manually pushed
```
fhunleth, line 153: Just catching up here. I assume the plan is to add manually pushed updates back in this PR, right?

fhunleth (changes requested):
Marking as "request changes" until the question about manual firmware updates gets resolved.

jjcarstens (commented):
One quick piece to consider. 

There have been a few side discussions around this that I've failed to keep up with and communicate back and forth. There are some concerns and unknowns I keep circle around with this and the complexity it brings, but I also am not being useful in identifying ways around it.

What I've suggested is to allow this to progress as a new calculation path without affecting the old one to allow us to experiment with the SmartRent scale, but have an escape hatch should the unforeseen be worse than the old way this is replacing. That should also give us more path for experimentation and adaptation.

lib/nerves_hub/deployments.ex
```diff
@@ -108,118 +109,105 @@ defmodule NervesHub.Deployments do
...
+          |> Deployment.changeset(params)
+          |> Ecto.Changeset.put_change(:total_updating_devices, device_count)
+
+        case Repo.update(changeset) do
+          {:ok, deployment} ->
+            messages = []
+
+            deployment = Repo.preload(deployment, [:firmware], force: true)
+
+            if Enum.any?([:conditions, :is_active], &Map.has_key?(changeset.changes, &1)) do
```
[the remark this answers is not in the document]
oestrich: No? Changing the firmware shouldn't cause a recalculation of the deployment for a device. It will just get the new firmware whenever the orchestrator decides to push it out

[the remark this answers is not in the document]
jjcarstens: Ah yes, I get it now. This is just for recalculation checks, not what firmware to actually get 👍🏼

priv/repo/migrations/20240910194615_add_recalculation_type_to_deployments.exs
```diff
@@ -0,0 +1,11 @@
+defmodule NervesHub.Repo.Migrations.AddRecalculationTypeToDeployments do
+  use Ecto.Migration
+
+  def change do
+    execute "create type recalculation_type as enum ('broadcast', 'calculator');", "delete type recalculation_type;"
+
+    alter table(:deployments) do
+      add(:recalculation_type, :recalculation_type, default: "broadcast", null: false)
```
jjcarstens, line 8:
```suggestion
      add(:recalculation_type, :text, default: "broadcast", null: false)
```
I think we can keep this simple as a text field so we don't have to deal with the woes of enum columns in postgres. We're not really going to need the benefits of enum column since it's a few records and we're not search for the field. We can still use the `Ecto.Enum` type in the schema to get nice atoms and controlling a bit of what gets saved in there.

What do you think?

test/nerves_hub/deployments_test.exs
```diff
@@ -124,107 +124,6 @@ defmodule NervesHub.DeploymentsTest do
 
       assert_broadcast("deployments/update", %{}, 500)
     end
-
-    test "changing tags resets device's deployments and causes a recalculation", state do
```
jjcarstens, line 128: Should we leave this test in place for the `broadcast` recalculation?

test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -965,60 +965,6 @@ defmodule NervesHubWeb.WebsocketTest do
       SocketClient.close(socket)
     end
 
-    @tag :tmp_dir
-    test "assigned a deployment", %{user: user, tmp_dir: tmp_dir} do
```
jjcarstens, line 969: Same here?

jjcarstens (commented):
Thanks! A couple comments about testing.

This will work if we really want to move it forward. However, it's still very intermixed and a little hard to follow in the `update_deployment` function. I suggest clearly separating the code paths into functions and not mixing the two, including the topic and event names. Even if there is some code duplication, I think it will make it easier to adjust and review in the future going forward without having to keep so much context in mind and potentially affecting both routes.

What do you think?

priv/repo/migrations/20240910194615_add_recalculation_type_to_deployments.exs
oestrich: Works for me, you mentioned enum yesterday so I went that path haha.

jjcarstens: Ya I knew exactly what I had done when seeing this vs what was in my head with "Enum" at the time 🤦🏼

test/nerves_hub_web/channels/websocket_test.exs
oestrich: Brought it back

oestrich:
Ok, there's a branch higher up at the update_deployment function. Ends up leaving most everything in place as it was, which seems like what we want? I don't think we need to change any of the device channel things because the calculator won't broadcast `deployment/changed` which triggers recalculation in the device channel itself

lib/nerves_hub/deployments.ex
```diff
@@ -222,6 +230,104 @@ defmodule NervesHub.Deployments do
...
+          Device
+          |> select([d], count(d))
+          |> where([d], d.deployment_id == ^deployment.id)
+          |> Repo.one()
+
+        changeset =
+          deployment
+          |> Deployment.with_firmware()
+          |> Deployment.changeset(params)
+          |> Ecto.Changeset.put_change(:total_updating_devices, device_count)
```
joshk, line 247: it looks like these lines are the same for `update_deployment_calculator` and `update_deployment_broadcast`, maybe its better to move this logic to `update_deployment`

jjcarstens: This is my fault. I initially stated I'd rather have duplicated code than complicated code. I think we could improve reusing this in both cases, but also fine with the simple diff I'm presented with

lib/nerves_hub/deployments.ex
```diff
@@ -108,16 +109,23 @@ defmodule NervesHub.Deployments do
...
-    change, any devices that don't have a deployment and are online will be told about
-    the conditions changing to check for a deployment again
-  - If now active, any devices without a deployment will be told to reevaluate
-  - If now inactive, devices will have the deployment removed and told about the change
+  - Records audit logs depending on changes
+  - Will force a recalculation
   """
   @spec update_deployment(Deployment.t(), map) :: {:ok, Deployment.t()} | {:error, Changeset.t()}
   def update_deployment(deployment, params) do
+    case deployment.recalculation_type do
```
jjcarstens, line 117: I feel totally bipolar, but one thing I'm realizing now that I'm staring at this clean function is that we may not need to store this in the database at all. This could be part of the params or a function to the argument to specify which calculator route to take maybe?

On that note, we probably want a way to do this via the UI. I'm happy to add that change if you would like?

oestrich: I'm good with whatever in regards to the database vs param, and please do! Would make things easier haha

jjcarstens:
@oestrich @joshk Look at the changes of 21675de. I initially started by just adding the UI to set, but then found some ways to make things a little more readable and still get the `Repo.transaction` optimization.

I also kept it in the DB because when digging deeper I saw it would still be needed for the `Turn On/Off` button, so seemed better to leave in there for now

image

joshk (approved):

PR #1479, opened 2024-09-01
Add device extensions support

jjcarstens:
Start of device extensions which allow for specialized functionality on device to report data and interactions safely outside the update mechanism.

Requires https://github.com/nerves-hub/nerves_hub_link/pull/228

Based on conversations the last while, these are some of the core design decisions so far:
* NervesHub selectively tells a device what extensions to enable
* Extensions are off by default on device
* Extensions should have no affect on socket connectivity or ability to update a device

With that in mind, this is the basic communication cycle for supporting extensions on device.

```mermaid
sequenceDiagram
    participant Device as Device (NHL)
    participant NH as NervesHub

    Device ->> NH: Connect
    NH ->> NH: Check `extensions_allowed?`
    NH ->> Device: Send "extensions:get"
    Device ->> NH: Join "extensions" channel with [%{<extension> => <version>}]
    NH ->> NH: Determine extensions to enable based on report
    NH ->> Device: Send "extensions:enable" to with list of extensions to enable

    loop For each extension to enable
        Device ->> NH: Report "<extension_name>:<event>" where event is enabled or error
    end
```

TODO: All the bits to selectively decide what extensions can be enabled for a device

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -87,6 +87,8 @@ config :nerves_hub, NervesHub.SwooshMailer, adapter: Swoosh.Adapters.Test
 
 config :nerves_hub, NervesHub.RateLimit, limit: 100
 
+config :sentry, environment_name: :test
+
 config :phoenix_test, :endpoint, NervesHubWeb.Endpoint
 
 # Initialize plugs at runtime for faster test compilation
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -1394,4 +1394,45 @@ defmodule NervesHub.Devices do
         "pending"
     end
   end
+
+  def enable_extension_setting(%Device{} = device, extension_string) do
+    device = get_device(device.id)
+
+    Device.changeset(device, %{"extensions" => %{extension_string => true}})
+    |> Repo.update()
+    |> tap(fn
+      {:ok, _} ->
+        topic = "device:#{device.id}:extensions"
+
+        NervesHubWeb.DeviceEndpoint.broadcast(topic, "attach", %{
+          "extensions" => [extension_string]
+        })
+
+      _ ->
+        :nope
+    end)
+  end
+
+  def disable_extension_setting(%Device{} = device, extension_string) do
+    device = get_device(device.id)
+
+    Device.changeset(device, %{"extensions" => %{extension_string => false}})
+    |> Repo.update()
+    |> tap(fn
+      {:ok, _} ->
+        topic = "device:#{device.id}:extensions"
+
+        NervesHubWeb.DeviceEndpoint.broadcast(topic, "detach", %{
+          "extensions" => [extension_string]
+        })
+
+      _ ->
+        :nope
+    end)
+  end
+
+  def preload_product(%Device{} = device) do
+    device
+    |> Repo.preload(:product)
+  end
 end
--- a/lib/nerves_hub/extensions.ex
+++ b/lib/nerves_hub/extensions.ex
@@ -1,0 +1,68 @@
+defmodule NervesHub.Extensions do
+  @moduledoc """
+  An "extension" is an additional piece of functionality that we add onto the
+  existing connection between the device and the NervesHub service. They are
+  designed to be less important than firmware updates and requires both client
+  to report support and the server to enable support.
+
+  This is intended to ensure that:
+
+  - The service decides when activity should be taken by the device meaning
+    the fleet of devices will not inadvertently swarm the service with data.
+  - The service can turn off extensions in various ways to ensure that disruptive
+    extensions stop being enabled on subsequent connections.
+  - Use of extensions should have very little chance to disrupt the flow of a
+    critical firmware update.
+  """
+
+  @callback handle_in(event :: String.t(), Phoenix.Channel.payload(), Phoenix.Socket.t()) ::
+              {:noreply, Phoenix.Socket.t()}
+              | {:noreply, Phoenix.Socket.t(), timeout() | :hibernate}
+              | {:reply, Phoenix.Channel.reply(), Phoenix.Socket.t()}
+              | {:stop, reason :: term(), Phoenix.Socket.t()}
+              | {:stop, reason :: term(), Phoenix.Channel.reply(), Phoenix.Socket.t()}
+
+  @callback handle_info(msg :: term(), Phoenix.Socket.t()) ::
+              {:noreply, Phoenix.Socket.t()} | {:stop, reason :: term(), Phoenix.Socket.t()}
+
+  @callback attach(Phoenix.Socket.t()) :: {:noreply, Phoenix.Socket.t()}
+  @callback detach(Phoenix.Socket.t()) :: {:noreply, Phoenix.Socket.t()}
+  @callback description() :: String.t()
+
+  require Logger
+
+  @supported_extensions [:health, :geo]
+  @type extension() :: :health | :geo
+
+  @doc """
+  Get list of supported extensions as atoms with descriptive text.
+  """
+  @spec list() :: list(extension())
+  def list do
+    @supported_extensions
+  end
+
+  @spec module(extension()) :: module() | :unsupported
+  def module(:health), do: NervesHub.Extensions.Health
+  def module(:geo), do: NervesHub.Extensions.Geo
+  def module(_key), do: :unsupported
+
+  @spec module(extension(), String.t()) :: module() | :unsupported
+  def module(:health, ver) do
+    cond do
+      Version.match?(ver, "~> 0.0.1") -> NervesHub.Extensions.Health
+      true -> :unsupported
+    end
+  end
+
+  def module(:geo, ver) do
+    cond do
+      Version.match?(ver, "~> 0.0.1") -> NervesHub.Extensions.Geo
+      true -> :unsupported
+    end
+  end
+
+  def module(_key, _ver) do
+    :unsupported
+  end
+end
--- a/lib/nerves_hub/extensions/health.ex
+++ b/lib/nerves_hub/extensions/health.ex
@@ -1,0 +1,115 @@
+defmodule NervesHub.Extensions.Health do
+  @behaviour NervesHub.Extensions
+
+  alias NervesHub.Devices
+  alias NervesHub.Devices.Metrics
+
+  require Logger
+
+  @impl NervesHub.Extensions
+  def description do
+    """
+    Reporting of fundamental device metrics, metadata, alarms and more.
+    Also supports custom metrics. Alarms require an alarm handler to be set.
+    """
+  end
+
+  @impl NervesHub.Extensions
+  def attach(socket) do
+    extension_config = Application.get_env(:nerves_hub, :extension_config, [])
+
+    health_interval =
+      case get_in(extension_config, [:health, :interval_minutes]) do
+        i when is_integer(i) and i > 0 -> i
+        _ -> 60
+      end
+
+    send(self(), {__MODULE__, :check})
+
+    socket =
+      if health_interval > 0 do
+        timer =
+          health_interval
+          |> :timer.minutes()
+          |> :timer.send_interval({__MODULE__, :check})
+
+        socket
+        |> Phoenix.Socket.assign(:health_interval, health_interval)
+        |> Phoenix.Socket.assign(:health_timer, timer)
+      else
+        socket
+      end
+
+    {:noreply, socket}
+  end
+
+  @impl NervesHub.Extensions
+  def detach(socket) do
+    _ = if socket.assigns[:health_timer], do: :timer.cancel(socket.assigns.health_timer)
+    {:noreply, Phoenix.Socket.assign(socket, :health_timer, nil)}
+  end
+
+  @impl NervesHub.Extensions
+  def handle_in("report", %{"value" => device_status}, socket) do
+    device_meta =
+      for {key, val} <- Map.from_struct(socket.assigns.device.firmware_metadata),
+          into: %{},
+          do: {to_string(key), to_string(val)}
+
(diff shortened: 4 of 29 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -360,14 +365,9 @@ defmodule NervesHubWeb.DeviceChannel do
...
-  def handle_in("location:update", location, %{assigns: %{device: device}} = socket) do
-    metadata = Map.put(device.connection_metadata, "location", location)
-
-    {:ok, device} = Devices.update_device(device, %{connection_metadata: metadata})
-
-    device_internal_broadcast!(device, "location:updated", location)
-
-    {:reply, :ok, assign(socket, :device, device)}
+  def handle_in("location:update", location, socket) do
+    # Backwards compatibility for Geo feature
```
lawik, line 369: I think the only devices that use Geo are ours from testing and we could absolutely break that contract and fix those devices.

The functionality has been added in `nerves_hub_web` but the other end only exists in branches of `nerves_hub_link` extension stuff.

lawik (commented):
Looks like a good approach to me. Want me to flesh out the allow mechanisms as a branch off of this branch or something?

Or we merge this once we tidy up the dbg and TODO notes?

lib/nerves_hub_web/channels/device_channel.ex
lawik: Dropping this to not create debt before we ship the link version.

lawik:
Fixing the rebase. But ready for review @joshk and I've cut anything that would cause legacy for the yet-unreleased health and geo features. So two env vars have changed to match. Messaging internally has changed. Message to the link-side features have been aligned to all do similar stuff.

lawik:
I need to manually test this locally to make sure everything works correctly. But would love review @joshk so when I've tried it properly we can  move this and the matching nerves_hub_link along.

lawik:
@jjcarstens was this just a rebase?

lawik:
Running this and nerves_hub_link locally, catching a bunch of trickiness :)

lawik:
Marking as draft while I churn through renaming and so on.

lawik:
Re-wrote Jon's words above to be all about EXTENSIONS! 🤘🏻 👀 🤘🏻 

I've renamed all the things. The sibling nerves_hub_link PR seems healthy and solid. I put one test on pending because it is just messy to update for now.

I think everything is good to go. Or at least ready to be picked over :D

lawik:
So paging @joshk and @jjcarstens for reviewing :)

lib/nerves_hub/devices.ex
```diff
@@ -1383,4 +1383,33 @@ defmodule NervesHub.Devices do
         "pending"
     end
   end
+
+  def enable_extension_setting(%Device{} = device, extension_string) do
+    device = Repo.get(Device, device.id)
```
jjcarstens, line 1388: Why are we fetching the device if we are requiring it as an arg?

lib/nerves_hub/devices.ex
```diff
@@ -1383,4 +1383,33 @@ defmodule NervesHub.Devices do
...
+
+    Device.changeset(device, %{"extensions" => %{extension_string => true}})
+    |> Repo.update()
+    |> tap(fn _ ->
```
jjcarstens, line 1392: I believe we will need to check that the update was successful before broadcasting, otherwise we will be out of sync with the setting

lib/nerves_hub/devices.ex
```diff
@@ -1383,4 +1383,33 @@ defmodule NervesHub.Devices do
         "pending"
     end
   end
+
+  def enable_extension_setting(%Device{} = device, extension_string) do
+    device = Repo.get(Device, device.id)
+
+    Device.changeset(device, %{"extensions" => %{extension_string => true}})
```
jjcarstens, line 1390: Does this actually work if I have multiple extensions? To me it looks like it will completely overwrite all extension settings with just this one?

lib/nerves_hub/extensions.ex
```diff
@@ -0,0 +1,52 @@
...
+  @supported_extensions %{
+    health: """
+    Reporting of fundamental device metrics, metadata, alarms and more.
+    Also supports custom metrics. Alarms require an alarm handler to be set.
+    """,
+    geo: """
+    Reporting of GeoIP information or custom geo-location information sources
+    you've set up for your device.
+    """
+  }
```
jjcarstens, line 42: Since this is a behavior module, I feel like we should keep as much of this in the extension as possible. Maybe we need a `description` callback for the module to define it?

lib/nerves_hub_web/live/product/settings.ex
```diff
@@ -71,4 +73,30 @@ defmodule NervesHubWeb.Live.Product.Settings do
...
+
+        {true, _} ->
+          Products.disable_extension_setting(socket.assigns.product, extension)
+      end
+
+    socket =
+      case result do
+        {:ok, _pf} ->
+          # reload extensions
+          assign(socket, :extensions, Extensions.list())
```
jjcarstens, line 94: `Extensions.list()` is static. Do we need to reload it? In fact, we probably don't even need to hold it in the assigns and just render with the view. Save some memory

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -94,7 +95,7 @@ defmodule NervesHubWeb.DeviceSocket do
     |> Devices.get_device_certificate_by_x509()
     |> case do
       {:ok, %{device: %Device{} = device}} ->
-        socket_and_assigns(socket, device)
+        socket_and_assigns(socket, Devices.preload_product(device))
```
jjcarstens, line 98: Do we need to include this preload in a more effect way? Like with getting the device certificate? I know Josh did a lot of work optimizing preloads so I'm hesitant to put such a small function in without thinking through

lib/nerves_hub_web/live/devices/device_health.ex
```diff
@@ -34,6 +34,7 @@ defmodule NervesHubWeb.Live.Devices.DeviceHealth do
 
     if connected?(socket) do
       socket.endpoint.subscribe("device:#{device.identifier}:internal")
+      socket.endpoint.subscribe("device:#{device.identifier}:extensions")
```
jjcarstens, line 37: What are we doing with this subscription here?

lib/nerves_hub_web/channels/extensions_channel.ex
```diff
@@ -0,0 +1,159 @@
...
+         %{extensions: device_extensions, product: %{extensions: product_extensions}},
+         extension_versions
+       ) do
+    allowed_extensions =
+      product_extensions
+      |> Map.from_struct()
+      |> Enum.filter(fn {extension, enabled?} ->
+        enabled? == true and Map.get(device_extensions, extension) != false
+      end)
+      |> Enum.map(&elem(&1, 0))
```
jjcarstens, line 35:
```suggestion
      for {extension, true} <- Map.from_struct(product_extensions),
          {^extension, device_enabled?} <- Map.from_struct(device_extensions,
          device_enabled? != false,
          do: extension
```
Slight optimization. Less looping, more direct filtering. Gets you list of extensions only allowed at product _and_ device

test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -18,11 +19,18 @@ defmodule NervesHubWeb.DeviceChannelTest do
...
+    {:ok, _, _socket} =
+      subscribe_and_join(socket, ExtensionsChannel, "extensions", %{
+        "geo" => "1.0.0",
+        "health" => "1.0.0"
+      })
   end
 
   describe "device location" do
+    # This test needs rework due to the extensions mechanism
+    @tag :pending
```
nshoes, line 33: I see you did some work on this test, is anything more needed?

test/support/socket_client.ex
```diff
@@ -132,16 +137,29 @@ defmodule SocketClient do
...
+      socket
+      |> assign(:joined_extensions?, true)
       |> assign(:reply, reply)
 
     {:ok, socket}
   end
 
   @impl true
+  def handle_message("device", "extensions:get", _message, socket) do
+    {:ok, socket}
```
nshoes, line 160: Why do we do nothing here?

nshoes (commented):
Looking good so far; I left some comments, questions, and concerns.

lib/nerves_hub/devices.ex
lawik: Previous way I modeled the extensions it mattered a lot to have the updated state. This would be resistant to that so yeah, unnecessary now.

lib/nerves_hub/devices.ex
lawik: Clever :D

lib/nerves_hub/devices.ex
lawik: I could have sworn I added it...

lawik: Ah, that was the file that had a super weird rebase conflict. I am no longer surprised that the change disappeared. Will fix of course.

lib/nerves_hub_web/live/product/settings.ex
lawik: Well, this was true before we moved the descriptions. Now it makes a little sense to hold it. But yeah, refreshing it is unnecessary. This was just there and I kept it.

lib/nerves_hub_web/channels/device_socket.ex
lawik: To check feature-support we need the product information. We either load it here or in ExtensionsChannel. And this means all the channels can rely on it. We could load and throw it out as well I suppose but it wouldn't get rid of the query. We could include the preload in the device preload for `Devices.get_device_certificate_by_x509` which I guess would save a DB roundtrip?

lib/nerves_hub_web/live/devices/device_health.ex
lawik: Not sure. I don't know when that hopped in there. Can remove for now. I think the point is to track health updates. Can fix it later if it makes the graphs stop updating. Doesn't seem quite right regardless.

lib/nerves_hub_web/channels/extensions_channel.ex
lawik: I find for comprehensions very messy to parse visually. I typically prefer a pipeline even if it is less efficient. I'll take it though :)

config/test.exs
```diff
@@ -87,6 +87,8 @@ config :nerves_hub, NervesHub.SwooshMailer, adapter: Swoosh.Adapters.Test
 
 config :nerves_hub, NervesHub.RateLimit, limit: 100
 
+config :sentry, environment_name: :test
```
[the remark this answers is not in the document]
lawik: This showed up in the branch @jjcarstens did. I would hazard a guess that something was spitting warnings in test runs but I don't know.

lib/nerves_hub/extensions/health.ex
```diff
@@ -0,0 +1,113 @@
...
+    socket
+  end
+
+  @impl NervesHub.Extensions
+  def attach(socket) do
+    extension_config = Application.get_env(:nerves_hub, :extension_config, [])
+
+    health_interval =
+      case get_in(extension_config, [:health, :interval_minutes]) do
+        i when is_integer(i) -> i
```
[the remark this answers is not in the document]
lawik: It covers nil and ensures the right type of value I suppose?
Actually I'd add another guard to ensure it is `> 0`.

Prevent random misconfigs.

Other suggestions welcome :)

lib/nerves_hub/products.ex
```diff
@@ -269,4 +269,28 @@ defmodule NervesHub.Products do
...
+  def disable_extension_setting(%Product{} = product, extension_string) do
+    product = Repo.get(Product, product.id)
+
+    Product.changeset(product, %{"extensions" => %{extension_string => false}})
+    |> Repo.update()
+    |> tap(fn _ ->
+      topic = "product:#{product.id}:extensions"
+
+      NervesHubWeb.DeviceEndpoint.broadcast(topic, "detach", %{"extensions" => [extension_string]})
+    end)
```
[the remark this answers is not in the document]
lawik: I think a toggle introduces a small risk that you end up doing the opposite of what you did in the UI due to UI and data gliding out of phase. It would certainly be doable with less code that way but I like the clarity of enable/disable.

Not married to it :)

test/nerves_hub_web/channels/device_channel_test.exs
lawik: Yes and there has been a bit of churn in what we expect a channel back-and-forth to be. This test tooling could be tooled up to support extensions and the fixtures could be set up to handle it. But the test was testing something fairly brittle to begin with so I'd rather address that in a later PR as this has dragged for a bit.

test/support/socket_client.ex
lawik: Because the socket client test tooling is currently not built out to do the extensions song and dance. It should be and we can make some nice additional tests then.

lawik:
I think I've hit all of them. I need to do a manual test or two to ensure things still work right. Then squash etc. Feel free to resolve anything if you look through it, otherwise I will later :)

lawik:
While this gets any additional feedback I will take a look at that test I marked `pending`.

test/nerves_hub_web/channels/device_channel_test.exs
lawik: Fixed so this can run. It doesn't feel quite complete. But it works.

nshoes (approved):
:shipit: :shipit: :shipit:

PR #1481, opened 2024-09-03
Allow wildcard prefix and suffix searching for device id

lawik:


lawik:
As someone who gets partial serials (last 4 of an RPi serial) and often wants to look it up this thing never worked for me. Any problems with this on y'all's ends?

lawik:
Added tests :)

joshk:
soooooo, I'm wondering if we remove `%` entirely and let this be something the user has to define?

I also wonder if we use `SIMILAR TO` for a bit more power https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP

oooooor `~` for the full power of regular expressions https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-SIMILARTO-REGEXP

I'd be +1 on moving to `SIMILAR TO` and removing the current `%` in the query, leaving it up to the user.

joshk (approved):
I've added some comments above, but I'm also fine with this change.

lawik:
I think removing the % is an inconvenience that only benefits power users. Happy to accept a SIMILAR TO but not feeling the urge to change it with the current use :)

lawik:
Merging, there was discussion on Slack as well that reassured me that there is no historical context retained that prefix-only was important.

PR #1482, opened 2024-09-03
Add devices filter to list untagged devices

lawik:
If all deployments are tagged it becomes relevant to add tags to new devices. Finding devices which are untagged has been impossible using the current set of filters.

This adds a filter for Untagged devices.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -170,6 +170,13 @@ defmodule NervesHub.Devices do
             {:error, _} ->
               query
           end
+
+        {:has_no_tags, value} ->
+          if value do
+            where(query, [d], fragment("array_length(?, 1) = 0 or ? IS NULL", d.tags, d.tags))
+          else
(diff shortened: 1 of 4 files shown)


joshk (approved):
In general, love this. 

One option would be to rename `only_untagged` to `has_tags`

lawik:
Both flip it and reverse it? Absolute madman :)

Not sure the inverse is a large improvement.

joshk:
for me `only_untagged` doesn't read well. `has_tags` feels nicer.

lawik:
I prefer how that reads but not what it implies for defaults. What's the default?
has_tags: true, implies showing only things with tags
has_tags: false, implies showing only things without tags
has_tags: ??, implies left blank

I guess we could do empty string to blank it as with other defaults but I think I prefer a boolean. How about `has_no_tags`?

joshk:
+1 on `has_no_tags` or `empty_tags`

PR #1483, opened 2024-09-03
Improve devices list style and function

lawik:


lawik:
Screenshot_20240903_120428

Screenshot of the new style and fields.

lawik:
I think this will do a lot to make it easier to find devices (the sorting), see activity at a glance (the fields) and actually have some clue what you have selected (the changed density and the selected row highlight.

jjcarstens:
Is the added at time useful in this list? We might be the exception, but we bulk import thousands at a time. Most views will have basically the same value and not as useful to know since it's expected a device could be added months before it ever connects since it is sitting on a warehouse shelf.

If it's useful to others I could concede.

lawik:
At the stage where you add tens or a few hundred I would expect it is still useful. Definitely for singles which is what new users will experience.
Open to trimming back if it gets disruptive.

joshk:
I love the 'selected' highlighting, and the addition of 'last seen'

I'm not sure about 'added'. I wonder if we wait until the new UI, possibly thinking about a way to customize the columns?

joshk (approved):
as mentioned in my comment, love row selection highlighting and the addition of the 'last seen' column. I think we should drop 'added' for now, but I'm fine either way.

lawik:
If it is not a strong preference I would keep added for now. Easy change to remove if it gets annoying.

joshk:
I think my main concern is how the list will look when a device has more tags, since these two new columns eat into our available space. 

We can merge and test it out.

PR #1487, opened 2024-09-05
Fix the oban config for the audit log truncation queue

joshk:
This is fixing a mistake I introduced into the code base.

The Audit Log truncation worker is using a different name, which means we aren't actually processing this queue, and jobs are backing up.

This fixes it.

Sorry.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -64,7 +64,9 @@ config :nerves_hub, Oban,
     delete_firmware: 1,
     device: 1,
     firmware_delta_builder: 2,
-    truncate: 1
+    truncate: 1,
+    # temporary, will remove in November
+    truncation: 1
   ],
   plugins: [
     # 1 week


lawik (approved):

lawik:
@joshk I had a test fail. Related to this you think?

```
  1) test audit log truncation enabled scheduling audit log truncation for an org (NervesHub.Workers.ScheduleOrgAuditLogTruncationTest)
     test/nerves_hub/workers/schedule_org_audit_log_truncation_test.exs:25
     Assertion with == failed
     code:  assert Enum.count(all_truncation_jobs) == 2
     left:  3
     right: 2
     stacktrace:
       test/nerves_hub/workers/schedule_org_audit_log_truncation_test.exs:32: (test)
```

lawik:
Okay, ecto.reset for MIX_ENV=test seems to have fixed it.

PR #1488, opened 2024-09-05
Fix broken directory creation for firmware deltas

lawik:
This also improves the error information and logging around deltas.

It does not hook them up to be enabled. This code is currently dead.

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -293,6 +293,10 @@ defmodule NervesHub.Firmwares do
           | {:error, Changeset.t()}
 
   def create_firmware_delta(source_firmware, target_firmware) do
+    Logger.info(
+      "Creating firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}."
+    )
+
     %Firmware{org: org} = source_firmware |> Repo.preload(:org)
     {:ok, source_url} = firmware_upload_config().download_file(source_firmware)
     {:ok, target_url} = firmware_upload_config().download_file(target_firmware)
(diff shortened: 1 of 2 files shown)


joshk:
Lars and I went on a deep dive into Delta land and can confirm that no code path generates or uses deltas.

So right now we have a tick box to enable delta firmware updates, but it does nothing.

As a note for later, we should look into changing all the `{_, 0}` matches as these raise match errors, which makes logging difficult and flow control a harder.

oestrich:
Yeah I deleted the part that sends delta updates to devices a long time ago, smartrent needs some kind of work on firmwares to enable them (something with encryption? i dunno) so I haven't been pressed to add it back. I thought they were still generated, but I might have deleted that as well haha

fhunleth:
It looks like there was a big misunderstanding. We'll need to get this feature back.

lib/nerves_hub/firmwares.ex
```diff
@@ -313,10 +317,19 @@ defmodule NervesHub.Firmwares do
...
+            "Created firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}, successfully."
+          )
+
           firmware_delta
         else
           {:error, error} ->
             delta_updater().cleanup_firmware_delta_files(firmware_delta_path)
+
+            Logger.error(
+              "Failed to create firmwa delta between #{source_firmware.uuid} and #{target_firmware.uuid}: #{inspect(error)}"
```
fhunleth, line 330:
```suggestion
              "Failed to create firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}: #{inspect(error)}"
```

fhunleth (approved):

oestrich:
For reference, this is the commit and code block that didn't survive a rewrite of the `resolve_update` function as I was figuring out what was going on in NervesHub at the time and trying to make deployments more coherent. 

https://github.com/nerves-hub/nerves_hub_web/commit/1c39a4f5ed154cf81dd22f6ffcf1cda4281b12c0#diff-03454c312d5c891b1ae7d916a1eb171cdf023ca00f9aa5155ea5633f7c5cbe81L525-L591

PR #1489, opened 2024-09-05
Add delta support to orchestrator

lawik:


lawik:
This can be the start of discussion. I think this approach would work. My biggest concern is really that I didn't find any Orchestrator tests?

And that makes groking how it operates difficult.

oestrich:
Yeah the orchestrator was created with a lot of nerves vms going and tested manually, along with the integration level of testing in `websockets_test.exs`.

This is all going to be rewritten very shortly anyways, it's my next step once we get #1476 in. There's a lot of people in smartrent that would like all of this queue shaped so that's what I was going to do next. I'm not sure if it makes more sense to get this in now or wait until after this has been rewritten :shrug:

lawik:
Here is where I think we need to coordinate. We have a lot of plans for deployments with NervesCloud and the health information from devices.

If we actually synced up about what you are trying to do for SmartRent right now we could take a load off you and you from us potentially. Rather than what I think will happen now which is arguably working redundantly on the same stuff.

lawik:
@jjcarstens this is delta support for the current orchestrator, is this worth updating and getting done or should we be focusing on the new variant of firmware orchestration you think?

PR #1496, opened 2024-09-07
Let the orchestrator manage device updates

joshk:


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -57,15 +57,6 @@ defmodule NervesHubWeb.DeviceChannel do
     # we might make a new one right below it, so clear it beforehand
     Devices.clear_inflight_update(device)
```
jjcarstens, line 58: Do we also need to remove this since it should be handled by the orchestrator?

jjcarstens (commented):
Thanks! This will help a lot.

One comment for a piece I'm unsure we need or not. Also, do we know if the orchestrator handles devices currently downloading a firmware as to not mess things up? (I believe this was fixed on device quite a while ago and should be good, but didn't dig deep)

Also for others, we chatted a bit offline about having some way to jump devices to the head of the line since there might be situations we want some devices to update quicker. Not 100% sure what that would look like or if it belongs here, but would also need to consider devices in an online or offline state

lib/nerves_hub_web/channels/device_channel.ex
joshk: Hmmmm, I'm unsure. It seems correct to keep this code as a restarted device is no longer updating.

In fact, searching the code base, this is the only place that the `clear_inflight_update` function is used, so I think we have to keep this here, at least for now.

joshk:
> Also, do we know if the orchestrator handles devices currently downloading a firmware as to not mess things up?

Before a Device is told to update we create an InflightUpdate DB row. This can be found on line 262 in `device_channel.ex`.

lib/nerves_hub_web/channels/device_channel.ex
oestrich: Yeah Josh is right, this is still needed since it should happen after reboot and should be gone allowing others to start updating. Previously this would send another update if it failed so it might be worth rethinking if it failed and should be tried again while it's the devices turn

joshk:
@jjcarstens This is ready for re-review

I've added a UI for pushing an update if its available.

There is one failure, but that is because of the Registry test, which I'm going to fix next (this helps simplify things)

jjcarstens (approved):

PR #1497, opened 2024-09-09
Split metrics into separate context

elinol:
Adds context for metrics, and support to liveviews to use metrics instead of health report (for metrics purposes).

No visual changes are made, the Contex graphs are still used but with a new data fetching approach which will make it easier to create more varied graphs from here, even for product metrics.


joshk:
I love where this is going, and honestly, I don't think this needs to be a 'draft'

Awesome work @elinol !

joshk (approved):
I think this is a great step forward!

lawik:
Very nice!

PR #1499, opened 2024-09-09
Add firmware delta generation

lawik:


lawik:
@joshk think we could test this out by pushing it and seeing how it does. Could make tests for it but not sure it's that worthwhile.

lib/nerves_hub/devices.ex
```diff
@@ -638,6 +638,48 @@ defmodule NervesHub.Devices do
...
+    from deployments as dep
+    inner join devices as dev
+        on dev.deployment_id = dep.id
+    inner join firmwares as f
+        on f.uuid = dev.firmware_metadata->>'uuid'
+    where
+        dep.product_id = $1
+        and
+        f.id != dep.firmware_id
+    """
```
joshk, line 658: Is there a reason you where for raw sql for this?

lib/nerves_hub/products.ex
```diff
@@ -116,9 +116,38 @@ defmodule NervesHub.Products do
...
+        |> Enum.uniq()
+        |> Enum.each(fn {source_id, target_id} ->
+          NervesHub.Workers.FirmwareDeltaBuilder.start(source_id, target_id)
+        end)
+
+        :ok
+
+      error ->
+        error
+    end
```
joshk, line 150: I'm not sure we need to care about generating deltas for previous deployments. 
Eg. It's highly likely deployments will be completed, so this just generates deltas for deployments but which are not likely to be used.

joshk (commented):
I'm wondering if we should add a 'state' or 'status' field to Deployment which would allow us to say 'generating deltas' aka 'preparing deployment' ('preparing'). Once the Deltas are generated then the Deployment can be set to active and the Orchestrator can take over.

lib/nerves_hub/devices.ex
lawik: I worked it out in raw SQL and the initial SQL was a bit more complex. At this point this could probably be done fine in Ecto though it would be a bit clumsy with fragments for the `uuid` fields.

lib/nerves_hub/products.ex
lawik: Oh, I think you are thinking of Deployments the way you want them to work rather than the way they do maybe?

A deployment is a firmware, a platform + some tags and indicates which firmware should be blasted out continuously as needed.

My Berlin product has many deployments, all active:

- RPi4 RC
- RPi4 Stable
- RPi3 RC
- RPi3 Stable

And so on. All of these are active and should get deltas. They all point to a current firmware.

lawik:
> I'm wondering if we should add a 'state' or 'status' field to Deployment which would allow us to say 'generating deltas' aka 'preparing deployment' ('preparing'). Once the Deltas are generated then the Deployment can be set to active and the Orchestrator can take over.

I would hold off on this or separate that work as the deployment is not a one-time thing. And this PR tries to be as dumb and simple as possible to enable delta firmwares to work without changing anything about how we orchestrate things.

lib/nerves_hub/devices.ex
lawik: Fixing that to be Ecto

lib/nerves_hub/products.ex
joshk: I follow that. I just mean that if you are enabling delta updates for paste firmware deployments, its unlikely they are needed as devices are likely already up to date.

lawik: This grabs all firmware versions that were last reported on devices in deployments for this product. They should all be reasonably relevant. It will generate a few that might never show up. But I _think_ this is a good way to do it.

It excludes any that are on the latest firmware for their deployment.

joshk:
For MVP testing purposes, I'd probably prefer to not generate deltas for deployments when the delta setting is updated on the Product. I feel its both simplier and a more isolated change. But not a hill I'm going to die on either.

I do think we need a 'deployment state' of 'preparing', so we don't send full firmware images when deltas will be ready in a couple of minutes, buuuuuut also not something I'm going to push.

lawik:
Too eager there. Fixed it :)

lawik:
@joshk so should try this with a spicy non-main deploy on NervesCloud and see how it does?

I want it to generate the backlog of firmware deltas partially to see how well it does.
But then I know that matters to at least 3 companies with connectivity challenges and concerns about transfer cost.

lawik:
We could separate out the step of generating a firmware delta backlog to be a button on the deployment if you are really concerned. I think it should be fine :)

joshk:
if you want to deploy this to NC Prod, nows the right time to do it

lawik:
Will give it a whirl :)

lawik:
Replaced by #1582

Issue #1501, opened 2024-09-09
`/devices` page with filters cause DB load with thousands of devices


jjcarstens:
**Describe the bug**
First let me say the feature to support saved and shareable search params in #1458 is awesome! 🎉 

But I do think it is causing us issues 😱 

We've recently had a situation where our DB load was very high causing lots of alerts and these 2 queries were the big offenders:

<details><summary>SQL Things</summary>
<p>

**Filtering `/devices`**
```SQL
SELECT 
d0."id", d0."org_id", d0."product_id", d0."deployment_id", d0."firmware_metadata", d0."identifier", d0."description", d0."updates_enabled", d0."tags", d0."deleted_at", d0."update_attempts", d0."updates_blocked_until", d0."connection_status", d0."connection_established_at", d0."connection_disconnected_at", d0."connection_last_seen_at", d0."connection_types", d0."connecting_code", d0."connection_metadata", d0."inserted_at", d0."updated_at", p2."id", p2."org_id", p2."name", p2."deleted_at", p2."delta_updatable", p2."inserted_at", p2."updated_at", d3."id", d3."firmware_id", d3."product_id", d3."org_id", d3."archive_id", d3."conditions", d3."device_failure_threshold", d3."device_failure_rate_seconds", d3."device_failure_rate_amount", d3."failure_threshold", d3."failure_rate_seconds", d3."failure_rate_amount", d3."is_active", d3."name", d3."healthy", d3."penalty_timeout_minutes", d3."connecting_code", d3."concurrent_updates", d3."total_updating_devices", d3."current_updated_devices", d3."inflight_update_expiration_minutes", d3."inserted_at", d3."updated_at", f4."id", f4."org_id", f4."product_id", f4."org_key_id", f4."architecture", f4."author", f4."delta_updatable", f4."description", f4."misc", f4."platform", f4."size", f4."upload_metadata", f4."uuid", f4."vcs_identifier", f4."version", f4."inserted_at", f4."updated_at", o1."id", o1."name", o1."deleted_at", o1."audit_log_days_to_keep", o1."inserted_at", o1."updated_at" 
FROM "devices" AS d0 
LEFT OUTER JOIN "orgs" AS o1 ON (o1."id" = d0."org_id") AND (o1."deleted_at" IS NULL) 
LEFT OUTER JOIN "products" AS p2 ON (p2."id" = d0."product_id") AND (p2."deleted_at" IS NULL) 
[...]

PR #1504, opened 2024-09-10
Improve the filtering query, and indexes needed

joshk:
A few things came up while looking into #1501 
- The devices query included lots of nested joins, and none of this data was being used in the device list or filters
- Most of the filters didn't have corresponding indexes to go with them, including tags
- The new double wildcard identifier change means the current index can't be used

This PR simplifies the filter query, adds indexes, including some `pg_trgm` goodness.

The `pg_trgm` bits have been inspired by:
- https://stackoverflow.com/questions/32951793/how-to-index-a-varchar-array-in-postgresql-using-array-to-string/33016333#33016333
- https://stackoverflow.com/questions/39480580/how-to-index-a-string-array-column-for-pg-trgm-term-any-array-column-que


lawik:
Looks very nice!
My only question is whether we have enough test coverage on devices LiveView to know if we break any of the filters :)

joshk:
Thats a fair question. I've tested it locally, and the filters work as expected, but the test suite should be expanded.

lawik (approved):

oestrich:
Leaving a note post close because I'm a little concerned with adding in 6 indexes on a table that gets hammered with heartbeat updates. Something to look out for once this gets deployed at smartrent @jjcarstens

lawik:
Yeah, that's a valid concern. I would imagine it won't have to touch indexes unless the indexed value actually changes?

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
```diff
@@ -0,0 +1,13 @@
+defmodule NervesHub.Repo.Migrations.AddIndexesToDevices do
+  use Ecto.Migration
+  @disable_ddl_transaction true
```
jjcarstens, line 3: Is there a specific reason we did not want to also `@disable_migration_lock true` according to https://fly.io/phoenix-files/migration-recipes/#adding-an-index?

priv/repo/migrations/20240910012345_enable_pg_trgm_extension.exs
```diff
@@ -0,0 +1,20 @@
+defmodule NervesHub.Repo.Migrations.EnablePgTrgmExtension do
+  use Ecto.Migration
+
+  def change do
+    execute "CREATE EXTENSION pg_trgm", "DROP EXTENSION pg_trgm"
```
rraub, line 5: Verified that pg_trgm is a supported AWS RDS Postgres Extension ✅ 
https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-extensions.html
and GCP ✅  https://cloud.google.com/sql/docs/postgres/extensions

If it wasn't then this would limit what database providers could be used.

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
esvinson: Depending on the version of Ecto, more modern ones only require the one tag.

jjcarstens: Ah, I see. Just looked through the `ecto_sql` docs and we should be good. We also [have `migration_lock: :pg_advisory_lock`](https://github.com/nerves-hub/nerves_hub_web/blob/main/config/config.exs#L57) so should be good

lib/nerves_hub/devices.ex
```diff
@@ -163,7 +176,7 @@ defmodule NervesHub.Devices do
                 where(
                   query,
                   [d],
-                  fragment("array_to_string(?, ',') ILIKE ?", d.tags, ^"%#{tag}%")
+                  fragment("array_to_string(?, ' ', ' ') ILIKE ?", d.tags, ^"%#{tag}%")
```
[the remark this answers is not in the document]
esvinson: If this isn't using the `string_array_to_string` function that we're creating in the migration it's probably not going to use that index.

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
```diff
@@ -0,0 +1,13 @@
...
+
+  def change do
+    create index("devices", [:product_id], concurrently: true)
+    create index("devices", [:updates_enabled], concurrently: true)
```
esvinson, line 7: Indexes on boolean fields probably won't be used.

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
```diff
@@ -0,0 +1,13 @@
+defmodule NervesHub.Repo.Migrations.AddIndexesToDevices do
+  use Ecto.Migration
+  @disable_ddl_transaction true
+
+  def change do
+    create index("devices", [:product_id], concurrently: true)
```
esvinson, line 6: Additional note, if we create the indexes using [create_if_not_exists](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#create_if_not_exists/1) we can create some of the indexes in advance to help with performance.

lib/nerves_hub/devices.ex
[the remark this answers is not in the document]
jjcarstens: Ah, I get it. A third value added 🤦🏼

joshk:
> Leaving a note post close because I'm a little concerned with adding in 6 indexes on a table that gets hammered with heartbeat updates. Something to look out for once this gets deployed at smartrent @jjcarstens

As indexes are separate, this should be fine. It would only be a concern if the indexed data is updated often.

lib/nerves_hub/devices.ex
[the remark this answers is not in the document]
joshk: Thanks for catching this @esvinson

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
joshk: My general thought was "index more stuff, see what is used, remove indexes as we see fit"

This was some general advice from some Postgres contributors at Heroku that I was friendly with.

priv/repo/migrations/20240910054628_add_indexes_to_devices.exs
joshk: Great idea!

PR #1507, opened 2024-09-10
Rename the manual push message for devices

oestrich:
Split out of the same message that normal inflight deployments will send.

Breaking some things out of #1476


jjcarstens (approved):
🎉

PR #1510, opened 2024-09-10
Indexes followup

jjcarstens:
Based on the previous PR review. Actually use the new tag function and do some safe creates to allow us (or anyone else) to manually create indexes as needed

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -176,7 +176,7 @@ defmodule NervesHub.Devices do
                 where(
                   query,
                   [d],
-                  fragment("array_to_string(?, ' ', ' ') ILIKE ?", d.tags, ^"%#{tag}%")
+                  fragment("string_array_to_string(?, ' ', ' ') ILIKE ?", d.tags, ^"%#{tag}%")
                 )
               end)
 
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/devices.ex
```diff
@@ -176,7 +176,7 @@ defmodule NervesHub.Devices do
                 where(
                   query,
                   [d],
-                  fragment("array_to_string(?, ' ', ' ') ILIKE ?", d.tags, ^"%#{tag}%")
+                  fragment("string_array_to_string(?, ' ', ' ') ILIKE ?", d.tags, ^"%#{tag}%")
```
fhunleth, line 179: What issue does this fix?

oestrich: This is related to @esvinson's comment in the previous indices PR that you should use the same function as the index to trigger using the index

jjcarstens: Re: [this comment](https://github.com/nerves-hub/nerves_hub_web/pull/1504#discussion_r1752294710) - We created the index, but didn't end up using it. I think it was just missed in the original PR

fhunleth: Thanks

joshk: I can confirm that this change is needed, in the sense that the index isn't being used unless we use the new function.

PR #1513, opened 2024-09-11
Add support for custom metrics

elinol:
Allows displaying other metrics than default in health liveview.


lawik (approved):

lawik:
@joshk maybe you take a look as well. This will let custom metrics be graphed as well with the existing graph. Should help demos and so on.

joshk (approved):
I see a few things we can tweak once we move to chart.js, but all in all this looks good to me!

PR #1517, opened 2024-09-12
Add device count and estimated device count for Deployment views

lawik:
This allows seeing how big a deployment is in device count in listings
and detail view.

Beyond that the Edit form for Deployments will estimate the matching
devices based on the conditions set.

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -26,6 +26,24 @@ defmodule NervesHub.Deployments do
     |> Repo.all()
   end
 
+  @spec get_deployment_device_counts_by_product(integer()) :: %{integer() => integer()}
+  def get_deployment_device_counts_by_product(product_id) do
+    Device
+    |> select([d], {d.deployment_id, count(d.id)})
+    |> where([d], d.product_id == ^product_id)
+    |> group_by([d], d.id)
+    |> Repo.all()
+    |> Map.new()
+  end
+
+  @spec get_deployment_device_count(integer()) :: %{integer() => integer()}
+  def get_deployment_device_count(deployment_id) do
+    Device
+    |> select([d], count(d.id))
+    |> where([d], d.deployment_id == ^deployment_id)
+    |> Repo.one()
+  end
+
   @spec get_deployments_by_firmware(integer()) :: [Deployment.t()]
   def get_deployments_by_firmware(firmware_id) do
     from(d in Deployment, where: d.firmware_id == ^firmware_id)
@@ -328,6 +346,11 @@ defmodule NervesHub.Deployments do
     :ok
   end
 
+  @spec change_deployment(Deployment.t(), map()) :: Changeset.t()
+  def change_deployment(deployment, params) do
+    Deployment.changeset(deployment, params)
+  end
+
   @spec create_deployment(map) :: {:ok, Deployment.t()} | {:error, Changeset.t()}
   def create_deployment(params) do
     changeset = Deployment.creation_changeset(%Deployment{}, params)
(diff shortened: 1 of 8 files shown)


lib/nerves_hub/deployments.ex
```diff
@@ -25,6 +25,33 @@ defmodule NervesHub.Deployments do
...
+  @spec get_deployment_device_counts_by_product(integer()) :: %{integer() => integer()}
+  def get_deployment_device_counts_by_product(product_id) do
+    from(
+      d in Deployment,
+      join: dev in Device,
+      on: d.id == dev.deployment_id,
+      where: d.product_id == ^product_id,
+      select: {d.id, count(dev.id)},
+      group_by: d.id
+    )
```
joshk, line 37: I think this query can be simplified to only use the `devices` table since that table has both `product_id` and `deployment_id`. Also, it would be good to have this in the pipe-able format for consistency.

lib/nerves_hub/deployments.ex
```diff
@@ -25,6 +25,33 @@ defmodule NervesHub.Deployments do
...
+  @spec get_deployment_device_count(integer()) :: %{integer() => integer()}
+  def get_deployment_device_count(deployment_id) do
+    from(
+      d in Deployment,
+      join: dev in Device,
+      on: d.id == dev.deployment_id,
+      where: d.id == ^deployment_id,
+      select: count(dev.id),
+      group_by: d.id
+    )
```
joshk, line 51: Simple to my comment above.

lib/nerves_hub/deployments.ex
```diff
@@ -299,6 +331,21 @@ defmodule NervesHub.Deployments do
     )
   end
 
+  @doc """
+  Find all potential devices for a deployment
+
+  Based on the product, firmware platform, firmware architecture, and device tags
+  """
+  def estimate_devices_hit_by_conditions(product_id, platform, conditions) do
```
joshk, line 339: maybe rename this to `estimate_devices_matched_by_conditions`

joshk (changes requested):
All in all this looks great. Just some small change suggestions.

lib/nerves_hub_web/live/deployments/edit.ex
```diff
@@ -13,23 +13,57 @@ defmodule NervesHubWeb.Live.Deployments.Edit do
...
     |> assign(:firmware, deployment.firmware)
     |> assign(:firmwares, firmwares)
     |> assign(:form, to_form(changeset))
     |> ok()
   end
 
   @impl Phoenix.LiveView
+  def handle_event("recalculate", %{"deployment" => params}, socket) do
+    params = inject_conditions_map(params)
+    dbg(params["conditions"])
```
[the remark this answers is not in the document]
lawik: That's a load-bearing dbg!

lib/nerves_hub/deployments.ex
lawik: Huh! Did not notice that. Good catch :)

And hadn't realized there was a preferred way to Ecto. Can fix :)

lib/nerves_hub/deployments.ex
lawik: Sure thing :)

lawik:
Gotta make it not break also...

lawik:
And fixed :D

lawik:
Test hit a deadlock in an unrelated test so I gave it a re-run. Worth figuring out but very unlikely to be this PR?

lib/nerves_hub/deployments.ex
```diff
@@ -405,6 +428,21 @@ defmodule NervesHub.Deployments do
     )
   end
 
+  @doc """
+  Find all potential devices for a deployment
+
+  Based on the product, firmware platform, firmware architecture, and device tags
+  """
+  def estimate_devices_matched_by_conditions(product_id, platform, conditions) do
```
joshk, line 436: @lawik it just occurred to me to ask, do you think this will perform well with result sets in the 200k+ range?

joshk (commented):
Is it worth us adding tests for this?

lib/nerves_hub/deployments.ex
lawik: I think it will do well enough. It will only happen as you change a deployment. So not very frequent. If it turns out to be heavy we might want to tune the debounce.

lawik:
> Is it worth us adding tests for this?

Possibly but not enough that I think anyone would care if it broke and it would take kind of weird stuff to break it.

You can make a call on this. I'll write the tests if you prefer them. For me they are smack in the do/don't split.

lib/nerves_hub/deployments.ex
joshk: Sweet

joshk:
Same, this is smack bang on that line. 

I'm going to skip the need for a test as we are looking at making changes to Deployments in the long run.

PR #1518, opened 2024-09-13
Support Grafana via `prom_ex`

joshk:
Using Grafana has become more of the norm, and this is the start of supporting all the Grafanas

Both the Web and Devices Endpoints are monitored, which gives us some initial insight into whats going on within the app.

I've implemented the `/metrics` route via a new `MetricsEndpoint`, as this allows us to expose it on its own port which isn't exposed, without SSL, which Fly can monitor via its Prometheus service.

I think these dashboards require a little more work until I'm happy with the data, and insight, they provide.


oestrich:
I like the idea of this a lot, do we know how to have prometheus scrape ECS tasks directly? Last time I had a chance to work with prometheus it didn't handle dynamic pods/tasks/containers well (at least with whatever set up the company had haha)

joshk:
Sorry, I'm not sure on the best way to solve the prometheus scape part of the equation. 

Fly.io makes it super super easy by providing you a prometheus out of the box. We have this up and running with NervesCloud, including the correct taging of apps, instances, and hosts.

pxp9:
> I like the idea of this a lot, do we know how to have prometheus scrape ECS tasks directly? Last time I had a chance to work with prometheus it didn't handle dynamic pods/tasks/containers well (at least with whatever set up the company had haha)

i would say , to create a DNS record by using Route53 for your app and make prometheus spawn by an ECS to point this guy.

this is not application responsability @oestrich

joshk:
Closing as the PR has become stale.

PR #1522, opened 2024-09-16
Deployment calculator fixes around which devices are looked at

oestrich:
- Skip any device that hasn't connected previously, which also means they don't have firmware metadata to look at
- If a device has a deployment that doesn't match the inflight check's deployment, return a different value than `:skipped` to avoid waiting 15s before processing the next device

### DIFF
--- a/lib/nerves_hub/deployments/calculator.ex
+++ b/lib/nerves_hub/deployments/calculator.ex
@@ -29,9 +29,10 @@ defmodule NervesHub.Deployments.Calculator do
           if !is_nil(device.deployment_id) && device.deployment_id != deployment.id do
             Repo.delete!(inflight_check)
 
-            :skipped
+            :ignored
           else
             if deployment.is_active &&
+                 !is_nil(device.connection_last_seen_at) &&
                  device.product_id == deployment.product_id &&
                  device.firmware_metadata.platform == deployment.firmware.platform &&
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/deployments/calculator.ex
```diff
@@ -102,6 +106,11 @@ defmodule NervesHub.Deployments.Calculator do
 
         {:noreply, %{state | process_timer_ref: nil}}
 
+      :ignored ->
+        send(self(), :process_next_device)
+
+        {:noreply, %{state | process_timer_ref: nil}}
+
       :skipped ->
         timer_ref = Process.send_after(self(), :process_next_device, 15_000)
```
fhunleth, line 115: What's the reason to wait 15 seconds between processing devices?

Also, it seems like the wait time is decided when the device is skipped/ignored. How about returning a tuple `{:skipped, timeout}` to make the handling more generic between the two cases?

oestrich: It's 15s if there's no device found that needs to be looked at. Just to continually be scanning in case there's a missed phoenix channel broadcast. The skipped timeout seems good 👍

oestrich: I did something a little different, but I think it's good. Renamed the not found so it makes more sense and `:ignored` sends back `:ok` since it's still fine

fhunleth (approved):

PR #1523, opened 2024-09-17
Use Chart.js for metrics

elinol:
Replaces Contex with Chart.js for better charts.


lib/nerves_hub_web/live/devices/device_health.ex
```diff
@@ -90,177 +81,145 @@ defmodule NervesHubWeb.Live.Devices.DeviceHealth do
...
-    case type do
-      :load_1min -> get_cpu_load_max_value(data)
-      :load_5min -> get_cpu_load_max_value(data)
-      :load_15min -> get_cpu_load_max_value(data)
-      :used_mb -> memory_size
-      _ -> 100
-    end
+  defp get_max_per_hour(metrics, _unit) do
+    metrics
+    |> Enum.group_by(& &1.inserted_at.day)
```
elinol, line 188: should be `inserted_at.hour`

PR #1528, opened 2024-09-19
Tell duplicate device connections to shutdown

joshk:
Its possible for websocket connections to be established to multiple devices using the same identifier. This either happens when a websocket connection dies (eg. network error) but the server takes a while to know about it and shut down the connection, while in the meantime the device has already reconnected. Its also possible to mess up your device settings and set the same identifier for multiple physical devices.

This PR send a PubSub event that a device has connected, and responds to the event by shutting down. I've used `broadcast_from` so `self()` doesn't get the `connected` message.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -91,7 +91,7 @@ defmodule NervesHubWeb.DeviceChannel do
     _ =
       NervesHubWeb.DeviceEndpoint.broadcast_from(
         self(),
-        "device:#{device.identifier}:internal",
+        "device:#{device.id}",
         "connected",
         %{}
       )
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -87,6 +87,15 @@ defmodule NervesHubWeb.DeviceChannel do
         Map.merge(value, update)
       end)
 
+    # disconnect devices using the same identifier
+    _ =
+      NervesHubWeb.DeviceEndpoint.broadcast_from(
+        self(),
+        "device:#{device.identifier}:internal",
+        "connected",
+        %{}
+      )
+
     # Cluster tracking
     Tracker.online(device)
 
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -87,6 +87,15 @@ defmodule NervesHubWeb.DeviceChannel do
         Map.merge(value, update)
       end)
 
+    # disconnect devices using the same identifier
+    _ =
+      NervesHubWeb.DeviceEndpoint.broadcast_from(
+        self(),
+        "device:#{device.identifier}:internal",
```
jjcarstens, line 94: I don't DeviceChannel ever subscribes to this topic? All I can find is `device:#{device.id}` subscription and the rest are deployments

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -141,6 +150,12 @@ defmodule NervesHubWeb.DeviceChannel do
     {:noreply, socket}
   end
 
+  # Listen for devices which have connected using the same device identifier
+  # and trigger a clean shutdown
+  def handle_info(%Broadcast{event: "connected"}, socket) do
+    {:stop, :shutdown, socket}
+  end
```
jjcarstens, line 157: Since metrics are so hot right now....Is there anything we'd want to mark and glean from this or not care if behavior is working?

joshk: Great point! I'll add a metric for a forced disconnection

lib/nerves_hub_web/channels/device_channel.ex
joshk: Thanks for spotting this, I'll fix this now.

Small question, when should the `"device:#{device.identifier}:internal"` vs `"device:#{device.identifier}"` topics be used? What is the core difference?

jjcarstens: I actually don't know the answer here without going digging. Maybe @oestrich can answer better?

joshk: Looking through our code, only `DeviceChannel` subscribes to the `device:#{device.id}` channel, so that should be safe for me to use.

jjcarstens (approved):

PR #1529, opened 2024-09-19
Rework duplicated connection handling

joshk:
Because items (in this case, devices) can only be removed from a Registry by the process that added them, the handling of duplicated connections and registry items has to change.

This PR adds a function before `:after_join` that checks for the device's existence in the registry. If the same device is found in the registry, the check is rescheduled to happen again in five seconds.

If the registry check fails three times, the connection is closed, forcing the device to reconnect and retrigger the broadcast which closes duplicate device connections again.

I've also added some metrics for the failed retries case.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -25,7 +25,17 @@ defmodule NervesHubWeb.DeviceChannel do
          {:ok, device} <- Devices.device_connected(device) do
       socket = assign(socket, :device, device)
 
-      send(self(), {:after_join, params})
+      # broadcast that the device has connected so we can disconnect
+      # devices using the same identifier
+      _ =
+        NervesHubWeb.DeviceEndpoint.broadcast_from(
+          self(),
+          "device:#{device.id}",
+          "connected",
+          %{}
+        )
+
+      send(self(), {:check_for_duplicate_connections, params, 0})
 
       {:ok, socket}
     else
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -25,7 +25,17 @@ defmodule NervesHubWeb.DeviceChannel do
...
-      send(self(), {:after_join, params})
+      # broadcast that the device has connected so we can disconnect
+      # devices using the same identifier
+      _ =
+        NervesHubWeb.DeviceEndpoint.broadcast_from(
+          self(),
+          "device:#{device.id}",
+          "connected",
+          %{}
+        )
```
jjcarstens, line 36: I would suggest flipping the situation a bit. Rather than pubsub and hope the process gets it to shut down, tell the other process to shutdown when attempting to register

```elixir
def handle_info({:register, attempt}, socket) do
  socket =
  case Registry.register(NervesHubDevices, device.id, %{}) do
    {:error, {:already_registered, pid}} ->
      send pid, :close
      if timer = socket.assigns[:registration_timer], do: Process.cancel_timer(timer)
      timer = Process.send_after(self(), {:register, attempt + 1}, 500)
      assign(socket, registration_timer: timer)
    {:ok, _} -> socket
  end

  {:noreply, socket}
end
```

oestrich: You can't use the registry to check for duplicated devices, it's for the local node only and I'm going to guess any production system will have at least two nodes that allow devices to connect. This is a large part why duplicate device checks were removed a while ago while the tracker was in flux

jjcarstens: O I see. So we still want the broadcast for other processes across nodes to potentially stop themselves. We also probably still want this registry code to retry since if device happens to be reconnecting to the same node, this issue still exists and we need the registry to be accurate

joshk:
Closing in favor of #1533

PR #1530, opened 2024-09-19
Always save/update the Devices connection status fields

joshk:
This short circuits Ecto change tracking by assigning dummy values before updating the data.

This is one part of the full fix related to a race condition in which two device connections could exist and both update the connection's status.

It could be argued that this isn't needed now that we are enforcing single connections per device, but I still think this is a good thing to have.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -602,13 +602,6 @@ defmodule NervesHub.Devices do
       connection_last_seen_at: DateTime.utc_now()
     })
     |> Repo.update()
-    |> case do
-      {:ok, device} ->
-        device
-
-      {:error, changeset} ->
-        {:error, changeset}
-    end
   end
 
   def device_heartbeat(device) do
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/devices.ex
```diff
@@ -593,40 +593,46 @@ defmodule NervesHub.Devices do
...
-        connection_status: :connected,
-        connection_established_at: DateTime.utc_now(),
-        connection_disconnected_at: nil,
-        connection_last_seen_at: DateTime.utc_now()
-      },
-      broadcast: false
-    )
+    device
+    |> clear_connection_information()
+    |> Device.changeset(%{
```
jjcarstens, line 598: This works. Though I do think you can get rid of the clearing step and skip the changeset to instead use `Ecto.Changeset.change/2` directly since you control the data here and it's a simple update.

jjcarstens: Nvm. That still ignores existing values.

jjcarstens (approved):

joshk:
The test failure is being addressed in another PR

PR #1533, opened 2024-09-21
Move connection tracking to `socket.ex`

joshk:
This PR fixes a fairly big problem, and that is a device being able to have two connections open at the same time, which happens fairly often when a device is on an unstable network, which causes it to disconnect and reconnect, but the server hasn't noticed the dead connection yet.

The big issue is that the Device Registry requires uniqueness, and we weren't enforcing the return values of registration, which meant the device would reconnect, not register, and not get updates from its deployment assigned to it (the Device Registry is used by the Orchestrator).

This PR makes sure devices can only have one connection, that Registry registration is retried, and that connection heartbeats (which helped find this issue) are correct and consistent (it's better for them to be attached to the socket as it is the core of the connection, and it reduces the code in channel to be more focused on device updates.)

As for the heartbeats, Phoenix heartbeats are every 30 seconds, so I added an interval checker/restriction
https://github.com/nerves-hub/nerves_hub_web/pull/1533/files#diff-a69fa65fc9d9ecb22dc9a2da65d0ca86dc22a2cd290d584ac5a178c2ae34d7a7R69-R73

This PR is live on NervesCloud and working a treat. 

Oh, another issue this PR fixes, the terminate call on DeviceChannel was not being called in normal websocket usage, as documented by Phoenix. What I've added is more reliable and consistent way to do some clean up.


lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -15,7 +16,44 @@ defmodule NervesHubWeb.DeviceSocket do
   # Default 90 seconds max age for the signature
   @default_max_hmac_age 90
 
+  defoverridable init: 1, handle_info: 2, terminate: 2
+
+  @impl true
```
jjcarstens, line 21:
```suggestion
  @impl Phoenix.Socket.Transport
```
Love me some explicit over implicit definitions 😉 https://hexdocs.pm/credo/Credo.Check.Readability.ImplTrue.html

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -52,6 +90,7 @@ defmodule NervesHubWeb.DeviceSocket do
     {:error, :no_auth}
   end
 
+  @impl true
```
jjcarstens, line 93:
```suggestion
  @impl Phoenix.Socket
```

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -15,7 +16,44 @@ defmodule NervesHubWeb.DeviceSocket do
...
   @default_max_hmac_age 90
 
+  defoverridable init: 1, handle_info: 2, terminate: 2
+
+  @impl true
+  def init(state) do
+    {:ok, {other, socket}} = super(state)
+    socket = on_connect(socket)
+    {:ok, {other, socket}}
+  end
```
jjcarstens, line 26:
```suggestion
  def init(state_tuple) do
    {:ok, {state, socket}} = super(state_tuple)
    socket = on_connect(socket)
    {:ok, {state, socket}}
  end
```
Naming suggestion to help our future selves

jjcarstens: Also, am I understanding right that `init` callback is coming _after_ `connect`? If that's the case, maybe we want to order our callback functions here in the perceived order?

jjcarstens (commented):
Got through some of it but timed out. Super exciting so far! 🎉

lib/nerves_hub_web/channels/device_socket.ex
joshk: You are right, and hmmmm. I put `on_connect` after `init` to make sure the socket was initialized correctly first,. and so we can pass the socket (with its assigns) to `on_connect`

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -119,36 +61,59 @@ defmodule NervesHubWeb.DeviceChannel do
...
+    send(self(), {:device_registation, 0})
+    {:noreply, socket}
+  end
+
+  def handle_info({:device_registation, 3}, socket) do
+    # lets make sure we deregister any other connected devices using the same device id
+    [:nerves_hub, :devices, :registry, :retries_exceeded]
+    |> :telemetry.execute(%{}, %{device: socket.assigns.device})
 
     {:stop, :shutdown, socket}
```
joshk, line 93: @jjcarstens I'm not sure if this function return will close the socket, or just the channel.
Maybe we should send a socket disconnect broadcast? eg.
`NervesHubWeb.DeviceEndpoint.broadcast("device_socket:#{device.id}", "disconnect", %{})`

joshk:
This does move the 'Device connected' and 'Device disconnected' updates to `DeviceSocket`. 

I've been able to avoid using a timer for the heartbeats by hooking into the Phoenix heartbeats. These are sent by Slipstream, and required by Phoenix as it maintains its own heartbeats mechanism. 

Channels is a Phoenix concept, and there is nothing (from what I've come across) that requires a socket connection to join a channel. This so something for us to consider handling, since this is an open source project and anyone could build their own NervesHubLink like SDK.

lawik:
Thought that might be what we got with the heartbeat changes and that's great :)

Yeah, it gets interesting. Is a device actually "connected" in NervesHub's meaning if it can't be sent firmware?
We could skip requiring that the DeviceChannel be joined for firmware updates specifically? But that'll probably set us into friction with the norm of how people using Phoenix Channels. So not eager to do that.

joshk:
Phoenix forces us to use channels, so we can't just us a socket.

What I would like to do, at the very least, is disconnect the client if they don't connect to the `device` channel within x period. I also think it is fair to disconnect the client if they close the `device` channel, which makes sure they don't stay connected when they shouldn't be. But thats for a different PR.

oestrich:
I like the idea of moving heartbeats to the phoenix heartbeat that is required anyways, but I'm concerned that this is device controlled. Currently we tuned the "device is connected" heartbeat to be 30 minutes given the size of our fleet. If we make this change then that is no longer configurable and will be the default 30 seconds that phoenix/slipstream uses. This is something we could change but it will take a very long time to filter out to every device through firmware, which means we'll have a potentially extremely high database load until then.

joshk:
```
Enum.count(Process.list(), fn p -> Process.info(p)[:dictionary][:"$initial_call"] == {NervesHubWeb.DeviceChannel, :join, 3} end)
38
Registry.count(NervesHub.Devices.Registry)
38
```
Things look to be working well, the number of devices in the Registry match the number of DeviceChannel connections

And when I look at our logs I see:
```
01:31:14.999 [info] pid=<0.5702.0> identifier=00000000f0d7e5f9 event=nerves_hub.devices.disconnect ref_id=CRPQ Device disconnected
01:31:15.000 [info] pid=<0.5702.0> identifier=00000000f0d7e5f9 event=nerves_hub.devices.duplicate_connection ref_id=CRPQ Device duplicate connection detected
01:31:15.000 [info] pid=<0.5728.0> identifier=00000000f0d7e5f9 event=nerves_hub.devices.connect firmware_uuid=a2e91b70-9b71-58ed-c95c-36de490d9fe7 Device connected
```
Which shows duplicate connections being killed correctly.

lawik (approved):
It makes sense to me :)

oestrich:
My only extra comment is that we're aware that is a best effort at making sure a device has a single connection. It definitely guarantees on a single node there will only be one connection, but we're broadcasting out to close any other open connections in the cluster, which will be eventually closed.

PR #1534, opened 2024-09-21
Remove `cachex` as it is no longer used

joshk:


joshk:
The test failure is being addressed in a different PR

PR #1535, opened 2024-09-21
Improve the Device Registry name

joshk:
There is naming conflict with the Devices context, which is confusing at times.


joshk:
The failure is being addressed in another PR

PR #1537, opened 2024-09-23
Allow the deployment calculator interval to be changed

joshk:
I noticed that the deployment calculator changes were spamming the NervesCloud DB due to the 'not_found` timer.

The number of queries hitting our DB was 450 a minute, and while the queries are less than 1ms, it seems a bit excessive.

We have 5 nodes running, and around 22 deployments active. There is a calculator running for each deployment, and for each node, and the timer runs 4 times a mins, thus 450 queries.

This will only grow as more customers and deployments are created.

This PR at least reduces the issue, but the real fix will be to not running a timer per deployment per node.


joshk:
FYI: The test failure is being addressed in #1533

config/runtime.exs
```diff
@@ -33,6 +33,8 @@ config :nerves_hub,
     String.to_integer(System.get_env("DEVICE_DEPLOYMENT_CHANGE_JITTER_SECONDS", "10")),
   device_last_seen_update_interval_minutes:
     String.to_integer(System.get_env("DEVICE_LAST_SEEN_UPDATE_INTERVAL_MINUTES", "5")),
+  deployment_calculator_interval_seconds:
+    String.to_integer(System.get_env("DEPLOYMENT_CALCULATOR_INTERVAL_SECONDS", "15")),
```
oestrich, line 37: this could likely be in the hours, this is here as a backup in case the process missed a broadcast for the deployment updating

oestrich (approved):

config/runtime.exs
joshk: If its in hours I'd question if we need it to be configurable. Is SmartRent happy with this being an hourly check?

I think this check only benefits single node deployments as its unlikely all nodes in a multi node setup would not receive the broadcast.

oestrich: I'm fine with it being an hourly check

fhunleth, line 37:
```suggestion
    String.to_integer(System.get_env("DEPLOYMENT_CALCULATOR_INTERVAL_SECONDS", "3600")),
```
For ease of updating. :)

fhunleth (approved):

PR #1538, opened 2024-09-23
Use Oban for calculating the deployment for a device

joshk:
This is an experiment and would still need some bits investigated for improvement, eg. removing duplicate jobs

Some advantages of this approach:
- We aren't implementing a custom job queue just for device/deployment checks
- It allows for better scaling options using Oban workers
- We don't need more supervisors and genservers. Right now there is a calculator genserver per deployment per node, and each calculator process is polling the DB every x seconds. So the more deployments you have, or the more nodes you have, or worse, the more of both you have, then the more you are hitting the db. 

This is a screenshot of the number of db calls for 5 nodes and 22 deployments. Initially we had 450 DB calls per minute, and 99.999% of the time these deployments aren't doing anything. I changed the polling to once per 60 seconds, which reduced the number of calls by 75%, which is an improvement, but it still feels like we are duplicating Oban instead of using it to its advantage.

Screenshot 2024-09-23 at 5 37 48 PM

It is fair to note that each DB call is taking next to no time at all, less than 0.1ms. My concern is that this won't scale as we grow our nodes and customers (deployments). Also, this was consuming around 1.5% of DB CPU time on a DB that had no load (only 88 devices connected).

It would be interesting to test this against a large data set. If it is performant, I think we should consider making it the default "recalculation type."

(This PR doesn't clean up some of the DB bits, I can do that once this is tested)


oestrich:
fwiw I did a calculator per deployment to "single thread" each of the deployments per node and still allow other deployments to be adjusted. given each calculator is doing very small amounts of work they shouldn't block the ecto queue even if we have a ton of them. I personally don't like sticking things in oban that can be done in other ways, so that's why I wrote what I did. I wouldn't expect the amount of polling for the calculator to be anywhere near as "bad" for the database as saving device heartbeats, even for nervescloud which will probably have a higher deployment to device ratio than smartrent.

joshk:
The calculator approach doesn't scale for NervesCloud. As customers grow and deployments grow, and nodes grow, the number of per-deployment calculators will grow, and with each polling is the db for work, then 99.99999% of the time there is no work, it will add load (even though it is insignificant) for no reason.

The calculator approach is essentially recreating Oban but just for this one task.

In regards to the topic of the load device heartbeats put on the system, if we had 10,000 devices connected, and the heartbeat interval set to 5 mins, we can expect 2,000 updates a min.

If we have 7 nodes supporting those 10,000 devices, and 100 deployments, with the standard 15 second interval, we can expect 2,800 select queries per minute.

Additionally, this PR removes the need to use locking and a transaction, which simplifies some of the work it is doing.

lib/nerves_hub/deployments.ex
```diff
@@ -293,15 +293,18 @@ defmodule NervesHub.Deployments do
...
-    delete_inflight_checks(deployment)
-
+  def schedule_deployment_calculations(deployment) do
     query =
       Device
       |> select([d], %{
-        deployment_id: ^deployment.id,
-        device_id: d.id,
-        inserted_at: ^DateTime.utc_now()
+        worker: "NervesHub.Workers.DeviceCalculateDeployment",
```
oestrich, line 300: while we're here, is there a better way to mass import jobs in oban? i'm guessing probably not because it's inserting from another table?

oestrich (approved):

lib/nerves_hub/deployments.ex
joshk: I think your way of doing it was inspired. 
Leaving the DB to do the work of adding jobs is incredibly performant.

joshk:
Thank you for reviewing the PR @oestrich .

Maybe we can schedule some time to test this PR on your QA system before we consider merging?

oestrich:
Yeah, we deploy late at night on Thursdays, which I think will give us the best crossing of timezones. Are you good tomorrow sometime after 7pm Eastern? (which I think is 11am your time maybe? lol)

joshk:
@oestrich that works perfectly for me. Sorry that its so late for you.

oestrich:
I'm around anyways, so no worries. We have to deploy to production off hours to avoid people in the field using NervesHub to update devices as they're installed

joshk:
Is that because of devices updating when they boot?

nshoes:
@joshk 👋 getting up to speed with this body of work, a few questions:

- Did you ever get connected with Eric to test this?
- What else needs to be done here besides removing `InflightDeploymentCheck` related things? I can handle getting this over the finish line.

joshk:
hey @nshoes , great to have you on the NervesHub team!

Sadly we didn't get to test this together. Getting this deployed to your QA environment would be great, and I'd be happy to join a zoom and learn what this takes.

Sorry, what do you mean by `removing InflightDeploymentCheck related things`?

nshoes:
@joshk thank you! Stoked to start learning and working on this project.

> Sadly we didn't get to test this together

Dang! I'll have to connect with Jon to understand how to load-test things in QA; I know it exists, not sure how to operate it :)

> Sorry, what do you mean by removing InflightDeploymentCheck related things?

This PR removes the [only insertion of `InflightDeploymentCheck`s](https://github.com/nerves-hub/nerves_hub_web/pull/1538/files#diff-48cec1d45c7f0dac953ea8a89cecbf1c568e6a9b61b02a863a3ac4c23fb68920L335). Alongside your PR description with "(This PR doesn't clean up some of the DB bits, I can do that once this is tested)", I assumed we would want to remove that table and any more references to the schema. Is that incorrect?

Other than that, I noticed this call to [remove any `InflightDeploymentCheck`s here](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/deployments.ex#L217) when `deployment.recalculating_type == :device`. I'm not sure if this is even close to right, but should we remove Oban rows here?

I have lots of questions and uncertainties because I'm new to the project, so I'm happy to hop on a Zoom call as well.

joshk:
Ahhhhhh, true, silly me for not clicking on what you were talking about!

Once we test this and confirm it works on QA, and works well, any help with cleaning up the final bits would be greatly appreciated.

nshoes:
We're running into an issue with how `actions/checkout` behaves. We use the sha it checks out to pull the image to be able to deploy PRs to our QA env. The commit it's checking out isn't matching our expectations so I'll be doing some experimentation on this branch.

joshk:
@nshoes hey Nate, hows it all going? is there anything I can help with?

nshoes:
@joshk it's going! I recalculated a deploy on Friday with 800 devices and it performed well; all devices got updated as expected. There's two things I'm still working through: the query that inserts job rows is timing out when editing a large deployment, and the inflight checks aren't getting created so the inflight progress doesn't work properly (still just a correlation but I believe it's correct). I'm making progress on both so I'm not blocked.

joshk:
@nshoes if you can get the raw sql for the query that is failing due to timing out we can run a `EXPLAIN ANALYZE` to see what is slow about the query.

Hmmmm, inflight checks are getting created? that is really odd.

Anything I can help you with?

nshoes:
@joshk yeah good call, that's on the end of my debugging list 😆 I'll report back once I find something.

> Hmmmm, inflight checks are getting created? that is really odd.

No, they _aren't_ getting created: [they used to be](https://github.com/nerves-hub/nerves_hub_web/pull/1538/files#diff-48cec1d45c7f0dac953ea8a89cecbf1c568e6a9b61b02a863a3ac4c23fb68920L356).

> Anything I can help you with?

Nothing specific atm, I'll ping you when I do!

joshk:
Ohhhhh yes yes, they aren't needed anymore! Inflight checks were a schema which was added just for this new way of calculating deployments. Now we are using Oban it isn't needed. 
Happy to jump on a call and go over it with you.

nshoes:
> Inflight checks were a schema which was added just for this new way of calculating deployments

Oh, okay! I thought inflight checks were important for the `Orchestrator` when calculating what batch of devices to update concurrently, but I guess I need to go back and read more carefully :) 

The query may still be an issue, but it's happening on deployments nearly 8x larger than the sets we have in production (800 vs 62k) and you can't even edit a deployment that large currently (it's on my todo list to figure out why). In QA I could successfully activate, deactivate, change tags, and change recalculate type with no issues. All the expected devices would associate or de-associate with the deployment. I saw expected ticks in CPU usage and ecto query activity:

CleanShot 2024-11-13 at 08 12 39@2x
CleanShot 2024-11-13 at 08 14 17@2x

nshoes (approved):

nshoes:
Note that the above tests were done with 20k devices connected and deployments that had 400 (first spike) and 800 (second spike) devices.

joshk:
@nshoes Thanks for testing this!

In production, how many devices are in a typical deployment?

lib/nerves_hub/deployments.ex
```diff
@@ -352,8 +354,9 @@ defmodule NervesHub.Deployments do
       )
       |> where([d], d.firmware_metadata["platform"] == ^deployment.firmware.platform)
       |> where([d], d.firmware_metadata["architecture"] == ^deployment.firmware.architecture)
+      |> where([d], fragment("? <@ ?", ^deployment.conditions["tags"], d.tags))
 
-    Repo.insert_all(InflightDeploymentCheck, query)
+    Repo.insert_all(Oban.Job, query)
```
joshk, line 359: @nshoes regarding the timeouts you were seeing, we could change this to 
```suggestion
    Repo.insert_all(Oban.Job, query, timeout: 30_000)
```
and see if that helps?

joshk:
@jjcarstens any concerns with us merging this in?

jjcarstens:
Nope. Not if everyone else is happy with it

joshk:
> Nope. Not if everyone else is happy with it

Cool, I'll wait to hear from @nshoes regarding one of my comments, then I can look into merging this in.

jjcarstens:
100-600 is a typical deployment size. It's not uncommon to have mega deployments with 60k+

joshk:
> 100-600 is a typical deployment size. It's not uncommon to have mega deployments with 60k+

Ahhhh, ok, so we should test this with a deployment that has 80k devices associated with it and make sure it doesn't time out. I'd love to know the `EXPLAIN ANALYZE` for this query too so we can make sure we have the right indexes for it.

nshoes:
@joshk originally the query timeout was an issue, but filtering on tags fixed it. We then ran into a separate issue that looked similar which was our QA setup running out of memory when editing and saving a deployment with 60k devices. This is happening on `main` so it has nothing to do with this PR.

joshk:
@nshoes oh, thanks for the information, that is cool to hear that this PR isn't having the issues being seen on main. 

I'm going to merge this in!

jjcarstens:
Did you mean to merge instead of close?

Issue #1532, opened 2024-09-21
Downloading the certificate for a device goes to an error screen


fhunleth:
1. Go to a device's settings
2. Click on '...'
3. Click download

<img width="1042" alt="image" src="https://github.com/user-attachments/assets/6ac8cbc5-b121-4b1a-82ae-662197b806b7">

If this seems to work, then it might be because this device was moved from one org to a different one.

nshoes:
This looks like it was fixed in the associated PR, closing this issue.

PR #1539, opened 2024-09-23
Device certificate fixes

elinol:
Fixes bug where certificates could not be downloaded due to certificates not preloaded on device. (https://github.com/nerves-hub/nerves_hub_web/issues/1532)

Also fixes certificate upload bug (discovered when trying to reproduce the download bug) by adding phx-change handler to live upload. This is required for uploads to work. For now it only does `{:noreply, socket}`, maybe not optimal but we can discuss proper validation later on if wanted.


joshk:
Awesome find and fix!

I'm surprised that we didn't have a test failure for this, were you able to replicate this in a test?

elinol:
There are no tests for this yet, I'll go ahead and add some. Would you prefer it in a separate PR on included here?

joshk:
Same PR, please. I really appreciate it.

elinol:
Added tests for certificate download (in controller only) and upload via device settings liveview. My ambition was to write liveview tests for download and deleting certificates, but haven't manage to toggle the options dropdown within the test flow, I think perhaps the view has to be rewritten a bit for that to work.

joshk (approved):
Awesome work!

PR #1540, opened 2024-09-23
Add Changelog and move it to v2.0.0

lawik:


oestrich (approved):
:shipit:

I roll up a weekly changelog for SmartRent when I deploy to production, I can take those and bundle them up into a changelog update afterwards

lawik:
Sweet!

PR those here then we all get an eye on what goes into a version I reckon :)

lawik:
Think we should hold this until main actually passes CI :)

joshk (approved):

joshk:
Once this is merged in, I think we should tag main and tag the docker image as well.

PR #1545, opened 2024-09-25
Increase the Repo `queue_target` in the `test` config

joshk:
This has eliminated test failures locally


lawik (approved):
Have not had the problem, will take your word for it :D

PR #1546, opened 2024-09-25
Device channel cleanups

joshk:
This is 100% a `DeviceChannel` tidy up for readability, with a couple of bits of duplicated logic encapsulated into their own functions.


lawik (approved):
Looks reasonable to me :)

PR #1549, opened 2024-09-25
Add container version tagging

lawik:
I believe this should make new docker containers get tagged with the version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - main
+    tags:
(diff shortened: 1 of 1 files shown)


oestrich (approved):
I think this will work with this down below to do what we want? Probably can only figure out by merging 😅 

https://github.com/nerves-hub/nerves_hub_web/pull/1549/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL132

lawik:
For science!

PR #1553, opened 2024-09-25
Some minor dialyzer fixes

joshk:


lib/nerves_hub/deployments/calculator.ex
```diff
@@ -64,11 +64,12 @@ defmodule NervesHub.Deployments.Calculator do
         :ok
 
       {:ok, device} ->
-        Phoenix.PubSub.broadcast(
-          NervesHub.PubSub,
-          "device:#{device.id}",
-          %Phoenix.Socket.Broadcast{event: "devices/updated"}
-        )
+        _ =
```
fhunleth, line 67:
```suggestion
        :ok =
```

Didn't we recently learn the hard way about ignoring return values? I don't know why `Phoenix.PubSub.broadcast/3` would fail, but it seems like if it does, it should at least log an error since this would be unexpected, right?

lib/nerves_hub/deployments/calculator.ex
```diff
@@ -121,9 +122,10 @@ defmodule NervesHub.Deployments.Calculator do
 
     Logger.info("[InflightDeploymentCheck] reloaded deployment")
 
-    if state.process_timer_ref do
-      Process.cancel_timer(state.process_timer_ref)
-    end
+    _ =
+      if state.process_timer_ref do
+        Process.cancel_timer(state.process_timer_ref)
+      end
```
fhunleth, line 128: This is how I've been fixing this lately. Ignoring `if`'s always looks weird to me.

```suggestion 
    if state.process_timer_ref do
      _ = Process.cancel_timer(state.process_timer_ref)
      :ok
    end
```

lib/nerves_hub/deployments/calculator.ex
joshk: Sounds good to me. I was iffy if we should ignore the return of not, only because it had me think 'if the broadcast wasn't sent, would it cause an issue for us somewhere else?' 
I think one thing to take into account is that this is happening after the device is updated and outside of the transaction, so do we gain anything from this match error?

fhunleth: We gain knowledge about how `Phoenix.PubSub.broadcast` can fail in our system. I did notice that if you check the return value in the `broadcast` helper function then you can skip all of the `_ = `'s when calling the helper. That might be worth the code aesthetics if nothing else.

I don't have enough knowledge on this to push either way. I'm kind of curious of the failure cases, but my curiousity doesn't need to be satisfied in this project.

fhunleth (approved):
Approving since I've expressed my opinions and am ok with whatever you or others decide.

PR #1554, opened 2024-09-26
Remove the EctoReporter

oestrich:
This trips monitors during deploys because we have a sudden surge of warning logs. Since we're exporting repo query time via telemetry_metrics this isn't required any more.

### DIFF
--- a/lib/nerves_hub/metrics.ex
+++ b/lib/nerves_hub/metrics.ex
@@ -119,7 +119,6 @@ defmodule NervesHub.Metrics.Reporters do
 
   def handle_continue(:initialize, state) do
     reporters = [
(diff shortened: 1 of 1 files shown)


joshk (approved):
We also see a bunch of these messages when we deploy, especially with our device node based in Spain (our DB is on the east coast)

Does SmartRent use the NodeReporter?

oestrich:
We have it enabled, and at one point it was helpful in debugging things. I haven't looked at it in a while though, so we can probably delete it if you don't use it either.

joshk:
@oestrich we don't use it, I felt it was just going to be noise from all nodes.

PR #1564, opened 2024-09-27
Add task for generating randomized device metrics

elinol:
Inserts randomized metrics for a device every 20 minutes between now and a week ago.

Usage: `mix fake.metrics DEVICE_ID`


joshk (changes requested):
This is awesome! Thank you!

There is one change I think would be useful, instead of passing in the device database id, it would be nicer/easier to pass in the device identifier, as this can be found in the UI.

joshk (approved):
Thank you very much!

PR #1565, opened 2024-09-27
Only render connection tooltip when timestamp is present

elinol:
Fixes javascript error yielded when device connection timestamp was missing.

### DIFF
--- a/lib/nerves_hub_web/components/device_header.ex
+++ b/lib/nerves_hub_web/components/device_header.ex
@@ -33,8 +33,8 @@ defmodule NervesHubWeb.Components.DeviceHeader do
       <div>
         <div class="help-text mb-1 tooltip-label help-tooltip">
           <span>Last connected</span>
-          <span class="tooltip-info"></span>
-          <span class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
+          <span :if={@device.connection_established_at} class="tooltip-info"></span>
+          <span :if={@device.connection_established_at} class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
             <%= @device.connection_established_at %>
           </span>
         </div>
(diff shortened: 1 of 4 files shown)


lib/nerves_hub_web/components/device_header.ex
```diff
@@ -33,8 +33,8 @@ defmodule NervesHubWeb.Components.DeviceHeader do
       <div>
         <div class="help-text mb-1 tooltip-label help-tooltip">
           <span>Last connected</span>
-          <span class="tooltip-info"></span>
-          <span class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
+          <span :if={@device.connection_established_at} class="tooltip-info"></span>
+          <span :if={@device.connection_established_at} class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
```
joshk, line 37: Since we are adding an `:if` here, should also take into account an empty value for line 38, and instead display a friendly message like "No connection" or maybe even just a dash?

lib/nerves_hub_web/components/device_health/health_header.ex
```diff
@@ -44,8 +44,8 @@ defmodule NervesHubWeb.Components.HealthHeader do
       <div>
         <div class="help-text mb-1 tooltip-label help-tooltip">
           <span>Last connected</span>
-          <span class="tooltip-info"></span>
-          <span class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
+          <span :if={@device.connection_established_at} class="tooltip-info"></span>
+          <span :if={@device.connection_established_at} class="tooltip-text" id="connection-establisted-at-tooltip" phx-hook="LocalTime">
```
joshk, line 48: Same as previous message

lib/nerves_hub_web/components/device_header.ex
elinol: This is just for the tooltip that is showing the timestamp (when there is one). It will still show a text saying "Never". A tooltip that doesn't add more information is unnecessary imo.

elinol: I might just `:if` the whole tooltip div. What do you think?

joshk: Ahhhhhhhh, true true

PR #1567, opened 2024-09-27
Add alternative way to get shared secret

lawik:
Exclusively a UI fix with the slightest security implication.

The Copy to clipboard way hasn't worked on Chromium for Linux in some time and this fix adds a way to show the secret as an option.

This has been reported by users running into it and wasting time using the wrong credentials. I've personally hit it about a million times.


joshk:
Thanks for the PR. I did fix the "copy to clipboard" button earlier this week in #1541 and tested it across Firefox, Safari, and Chrome.

Can you please share a screenshot what it looks like when you click "Show".

lawik:
I will test that fix on Linux. If it works. No need for this.

joshk:
@lawik were you able to test if copy and paste was fixed?

lawik:
It works! Closing this.

PR #1572, opened 2024-10-01
Device Connections

elinol:
Adds support for device connection history.


elinol:
I'm not sure if we want device location on `DeviceConnection` as `metadata`, or if it should be stored on `Device`, maybe as it's own field? A location map?

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -42,9 +44,10 @@ defmodule NervesHubWeb.DeviceSocket do
 
   defp heartbeat(
          %Phoenix.Socket.Message{topic: "phoenix", event: "heartbeat"},
-         %{assigns: %{device: device}} = socket
+         %{assigns: %{device: device, connection_established_at: established_at}} = socket
        ) do
     if heartbeat?(socket) do
+      {:ok, _device_connection} = Connections.device_heartbeat(device.id, established_at)
       {:ok, _device} = Devices.device_heartbeat(device)
```
elinol, line 51: To be removed

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -203,50 +201,66 @@ defmodule NervesHubWeb.DeviceSocket do
     socket =
       socket
       |> assign(:device, device)
-      |> assign(:reference_id, generate_reference_id())
```
joshk, line 206: Instead of removing this assigns, you can change it to `:connection_information` and store the `DeviceConnection`, which we can then use for the device heartbeat.
eg. `Connections.device_heartbeat(socket.assigns.connection_information)`

joshk: Ahhhh, I see this is being done on line 210.

The `reference_id` can be used in the `device_heartbeat` function

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -203,50 +201,66 @@ defmodule NervesHubWeb.DeviceSocket do
...
 
-    {:ok, device} = Devices.device_connected(socket.assigns.device)
-
     Tracker.online(device)
 
-    assign(socket, :device, device)
+    socket
+    |> assign(:device, device)
+    |> assign(:reference_id, connection_id)
+    |> assign(:connection_established_at, established_at)
```
joshk, line 225: I don't think we need to store this or use it, thoughts?

lib/nerves_hub/devices/device_connection.ex
```diff
@@ -0,0 +1,41 @@
+defmodule NervesHub.Devices.DeviceConnection do
+  use Ecto.Schema
+
+  import Ecto.Changeset
+
+  alias NervesHub.Devices.Device
+
+  @type t :: %__MODULE__{}
+  @primary_key {:id, :binary_id, autogenerate: true}
```
joshk, line 9: I would recommend using https://github.com/martinthenth/uuidv7 as v7 UUIDs are sortable due to how the timestamp is at the front of the generated code

joshk (changes requested):
All in all I love the direction. 
I'm more than happy to jump on a zoom and go over the feed.

lib/nerves_hub/devices.ex
```diff
@@ -44,14 +52,10 @@ defmodule NervesHub.Devices do
...
-        d in Device,
-        where: d.org_id == ^org_id,
-        where: d.product_id == ^product_id
-      )
-
-    query
+    Device
+    |> where([d], d.org_id == ^org_id)
+    |> where([d], d.product_id == ^product_id)
+    |> preload(device_connections: ^Connections.latest_connection_preload_query())
```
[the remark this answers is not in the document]
elinol: this functions has two callers only: the dashboard which is using the connection info, and the index function from device controller which I figured also should have this info included (not completely sure on this). That's why I added the preload here.

lib/nerves_hub/devices.ex
```diff
@@ -82,6 +87,7 @@ defmodule NervesHub.Devices do
 
     Device
     |> where([d], d.product_id == ^product_id)
+    |> preload(device_connections: ^Connections.latest_connection_preload_query())
```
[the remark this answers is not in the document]
elinol: As above, this is because connection info is wanted when asking for the device list.

lib/nerves_hub/deployments/calculator.ex
```diff
@@ -23,7 +23,11 @@ defmodule NervesHub.Deployments.Calculator do
           |> Repo.one()
 
         if !is_nil(inflight_check) do
-          device = Repo.get!(Device, inflight_check.device_id)
+          device =
+            Device
+            |> where(id: ^inflight_check.device_id)
+            |> preload(:device_connections)
```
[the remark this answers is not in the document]
elinol: If we'll use a status field on device for seen/not seen (just registered/connected) we would not need a preload at all.

lib/nerves_hub_web/channels/device_socket.ex
elinol: I think it's good to have the established_at information on every DeviceConnection record to not need to make separate calls to db for `last_seen_at`and `established_at`.

lib/nerves_hub/devices/device_connection.ex
elinol: oh I thought this was set as default when using binary ids, couldn't find any other schemas where it was defined

lib/nerves_hub/deployments/calculator.ex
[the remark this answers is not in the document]
joshk: Following up on a chat we had, lets add two new fields to the Device schema, `status` (`registered` and `provisioned`) and `first_seen_at` (DateTime)

lib/nerves_hub_web/live/devices/device_health.ex
```diff
@@ -40,6 +41,7 @@ defmodule NervesHubWeb.Live.Devices.DeviceHealth do
     |> page_title("Device #{device.identifier} - Health")
     |> assign(:device, device)
     |> assign(:status, Tracker.status(device))
+    |> assign(:connection_established, Connections.get_established_at(device.id))
```
joshk, line 44: This should probably be changed to something like:
```elixir
|> assign(:latest_connection, Connections.get_latest_for_device(device.id))
```

lib/nerves_hub_web/live/devices/index.html.heex
```diff
@@ -249,7 +249,7 @@
             <%= if @device_statuses[device.identifier] == "online" do %>
               <img src="/images/icons/check.svg" alt="connected" class="table-icon" />
             <% else %>
-              <span title={if device.connection_last_seen_at, do: "Last seen at #{DateTimeFormat.from_now(device.connection_last_seen_at)}", else: "Not seen yet"} }>
+              <span title={if device.device_connections != [], do: "Last seen at #{device.device_connections |> hd |> Map.get(:last_seen_at) |> DateTimeFormat.from_now()}", else: "Not seen yet"} }>
```
joshk, line 252: The multiple piping is a bit messy, can you create a helper function for this instead.

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -53,6 +55,10 @@ defmodule NervesHubWeb.Live.Devices.Show do
 
     {:ok, device} = Devices.get_device_by_identifier(org, device.identifier)
 
+    socket
+    |> assign(:device_connection, Connections.get_latest_for_device(device.id))
+    |> noreply()
+
     {:noreply, assign(socket, :device, device)}
```
joshk, line 62: I think line 62 can be merged into the piping above:
```elixir
    socket
    |> assign(:device, device)
    |> assign(:device_connection, Connections.get_latest_for_device(device.id))
    |> noreply()
```

joshk (changes requested):
I think we are nearly there!

This is great work!

lib/nerves_hub/devices/device.ex
```diff
@@ -45,6 +49,20 @@ defmodule NervesHub.Devices.Device do
     field(:update_attempts, {:array, :utc_datetime}, default: [])
     field(:updates_blocked_until, :utc_datetime)
 
+    field(:status, Ecto.Enum,
+      values: [:registered, :provisioned],
+      default: :registered
+    )
```
jjcarstens, line 55: This might be offloading `:status` a bit. With this, it will not match the [device#show `:status` assigns](https://github.com/nerves-hub/nerves_hub_web/pull/1572/files#diff-7aac782381f55c9784ae6d3201da689e8c17034ae359b1d4c8a8a58cdbc42c25R36) or the `Tracker.status/1` which is always `"online" | "offline"`

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -31,6 +32,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> page_title("Device #{device.identifier} - #{product.name}")
     |> assign(:tab_hint, :devices)
     |> assign(:device, device)
+    |> assign(:device_connection, Connections.get_latest_for_device(device.id))
     |> assign(:status, Tracker.status(device))
```
joshk, line 36: could we rename this to `connection_status`?

joshk (changes requested):

lib/nerves_hub/devices/device.ex
joshk: I've suggested we change the assigns you have linked to to instead be named `connection_status`.

priv/repo/migrations/20241104095030_add_status_and_first_seen_at_to_devices.exs
```diff
@@ -0,0 +1,22 @@
+defmodule NervesHub.Repo.Migrations.AddStatusAndFirstSeenAtToDevices do
+  use Ecto.Migration
+
+  def up do
+    alter table(:devices) do
+      add(:status, :string)
+      add(:first_seen_at, :utc_datetime)
```
joshk, line 7: We should add an `execute` to set a value for `first_seen_at`. We can use the `inserted_at` field if the devices `connection_status` value isn't `not_seen`. Thoughts?

joshk (changes requested):

lib/nerves_hub/tracker.ex
```diff
@@ -71,7 +72,7 @@ defmodule NervesHub.Tracker do
   """
   def online?(device) do
     _ = Phoenix.PubSub.broadcast(NervesHub.PubSub, "device:#{device.id}", :online?)
-    device.connection_status == :connected
+    Connections.get_current_status(device.id) == :connected
```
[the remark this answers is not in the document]
joshk: @elinol we preload the latest device connection, can we use that instead of loading from the DB?

priv/repo/migrations/20241104095030_add_status_and_first_seen_at_to_devices.exs
elinol: sounds good

joshk:
@elinol I've left a couple of comments, and I see a couple of comments not resolved. Happy to jump on a Zoom if it helps. I think we are nearly there. I need to check with Jon regarding the use of Tracker, but for now I think we should keep it.

lib/nerves_hub_web/live/devices/show.ex
elinol: I ended up with removing the status assign since the status is already included in `:device_connection`

lib/nerves_hub/tracker.ex
```diff
@@ -66,13 +66,11 @@ defmodule NervesHub.Tracker do
   @doc """
   Check if a device is currently online
 
-  Returns `false` immediately but sends a message to the device's channel asking if it's
-  online. If the device is online, it will send a connection state change of online.
+  Returns `true` if device's latest connections has a status of `:connected`,
+  otherwise `false`.
   """
-  def online?(device) do
-    _ = Phoenix.PubSub.broadcast(NervesHub.PubSub, "device:#{device.id}", :online?)
```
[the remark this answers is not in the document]
joshk: @jjcarstens can we chat about this tomorrow? (your today)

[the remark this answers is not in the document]
joshk: Spoke with Jon, we are fine with removing this and doing a bigger refactor and rethink of Tracker in the very near future.

joshk (approved):
I think this is good to go!

We can monitor it in prod and make sure we have covered all our bases.

joshk:
@elinol sorry, during my rebase I've broken the test suite, could you please help fix it? sorry.

elinol:
@joshk no problems, it's done!

joshk:
Awesome, thank you!

joshk:
🎉

PR #1574, opened 2024-10-01
Source a known env.sh if it is present

oestrich:
Since we switched to `VERSION+SHA` it is more challenging to replace the env.sh file with a custom version. Sourcing a known file if present would help make things easier to override.

### DIFF
--- a/rel/env.sh.eex
+++ b/rel/env.sh.eex
@@ -1,15 +1,12 @@
 #!/bin/sh
 
-# Sets and enables heart (recommended only in daemon mode)
-# case $RELEASE_COMMAND in
-#   daemon*)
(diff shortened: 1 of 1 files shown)


jjcarstens:
I think this is a bit of an anti pattern with releases in general which are setup to more isolated releases. I just pushed a commit to our internal builder which should handle this case of variable version+sha by using the `./bin/nerves_hub version` command to fetch it

joshk:
Soooooo this is above my pay grade as it seems a bit particular to the SmartRent setup. I'll leave this for you both to figure out.

jjcarstens:
Seems we've already merged a change on our side so shouldn't be needed anymore

PR #1576, opened 2024-10-02
More logging tweaks

joshk:
The commit messages explain the changes.
Switching the timestamp and level to key=value is important for how we are using vector.dev to ship our logs from Fly to BetterStack.
The rest are improvements I came across during my setup of BetterStack


lib/nerves_hub/logger.ex
```diff
@@ -1,18 +1,30 @@
...
+    :application,
+    :pid,
+    :mfa,
+    :time,
+    :gl,
+    :ansi_color,
+    :__sentry__
+  ]
 
   def format(level, message, timestamp, metadata) do
```
lawik, line 17: So to address @oestrich's not I think we could have two different functions for two different formats.

And switch in the runtime config based on an env var.

jjcarstens (approved):

joshk:
For transparency, @jjcarstens and I had a quick catch-up, he believes the dev ops team can filter out the timestamp and level, so he doesn't see the problem with this now being key=value (logfmt)

jjcarstens:
I checked with a few places and it seems having a key to the message and timestamp could be beneficial with index searching. I think it's worth a try and if QA finds that to not be the case, we follow up with formatting

PR #1577, opened 2024-10-02
Allow the socket drainer to be configured at runtime

joshk:
This PR is for discussion.

When NervesHub hosts are shutdown (usually for updating) the device connection telemetry and database updates may need some extra time to complete.

This previously wasn't an issue as we weren't hooking into clean shutdowns, and in the case of Fly.io, a hard kill was being used by default.

I think having these dials available is useful, and I think it's worth discussing this approach.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -39,6 +39,12 @@ config :nerves_hub,
   mapbox_access_token: System.get_env("MAPBOX_ACCESS_TOKEN"),
   dashboard_enabled: System.get_env("DASHBOARD_ENABLED", "false") == "true"
 
+config :nerves_hub, :device_socket_drainer,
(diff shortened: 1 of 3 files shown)


oestrich (approved):
As far as I can tell we've been using the default drainer config and it works but tends to spike the ecto queue as expected. So I don't think it's a bad idea to allow configuration. I think the smaller batch size is good but should we keep `DEVICE_SOCKET_DRAINER_BATCH_INTERVAL` to be `2000`? I don't have any strong feelings towards either way

joshk:
I thought 4 secs, as a max cut off (I believe it acts as a timeout) would give time for all the sockets in the group to be closed and update the db. 

If you are ok with keeping this PR as it, I'll merge it in and we can reevaluate after it has had some testing?

oestrich:
For posterity - chatted in Slack and I was :100: on merging since it's configurable. If it's a timeout then I'm totally cool with 4s

PR #1578, opened 2024-10-03
Reintroduce missing custom metrics and metadata to device page.

elinol:
Also stop saving firmware metadata on device health.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -409,18 +409,12 @@ defmodule NervesHubWeb.DeviceChannel do
   end
 
   def handle_in("health_check_report", %{"value" => device_status}, socket) do
-    device_meta =
-      for {key, val} <- Map.from_struct(socket.assigns.device.firmware_metadata),
-          into: %{},
-          do: {to_string(key), to_string(val)}
-
     # Separate metrics from health report to store in metrics table
     metrics = device_status["metrics"]
 
     health_report =
       device_status
       |> Map.delete("metrics")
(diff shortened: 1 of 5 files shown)


elinol:
image

joshk:
Thanks for including the screenshot. 

I would recommend not showing custom metrics in the health section on the Device show page as it will mess with formatting. 

Can we keep custom metrics on the Device Metrics page only?

lawik:
As clarified on Slack, this restores functionality that was broken with metrics-changes. It needs to be nicer in the next iteration of the UI. But it does matter.

joshk (approved):
I approve this under duress 😆

lawik:
I would prefer if we remove the "stop saving firmware metadata" bit from this part for now. That's mixing two different topics. That merge is intentional and at some point I can explain more of my thinking there. It may turn out to be unnecessary but I think it will be helpful to have.

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -409,18 +409,12 @@ defmodule NervesHubWeb.DeviceChannel do
...
-          into: %{},
-          do: {to_string(key), to_string(val)}
-
     # Separate metrics from health report to store in metrics table
     metrics = device_status["metrics"]
 
     health_report =
       device_status
       |> Map.delete("metrics")
-      |> Map.put("metadata", Map.merge(device_status["metadata"], device_meta))
```
lawik, line 423: I'd rather this not be in this PR :)

lawik (changes requested):
Small thing about that merge bit.

Thanks for fixing the thing :)

lawik (approved):

lawik:
Thank you very much ✨

PR #1582, opened 2024-10-04
Add firmware delta generation back

lawik:
Remade [this branch](https://github.com/nerves-hub/nerves_hub_web/pull/1499) on a fresh main.

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -162,7 +162,7 @@ defmodule NervesHub.Deployments do
 
     case result do
       {:ok, {deployment, changeset}} ->
-        maybe_trigger_delta_generation(deployment, changeset)
+        _ = maybe_trigger_delta_generation(deployment, changeset)
         broadcast_deployment_updates(deployment, changeset)
 
         {:ok, deployment}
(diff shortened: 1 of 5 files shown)


oestrich (commented):
Run the formatter and I think we're good to go :shipit:

lawik:
Good catches, addressed :)

lib/nerves_hub/deployments.ex
```diff
@@ -305,6 +306,36 @@ defmodule NervesHub.Deployments do
...
+    end
+  end
+
+  defp trigger_delta_generation_for_deployment(deployment) do
+    case NervesHub.Devices.get_device_firmware_for_delta_generation_by_deployment(deployment.id) do
+      {:ok, %{rows: rows}} ->
+        rows
+        |> Enum.map(fn [source_id, target_id] ->
+          {source_id, target_id}
+        end)
```
oestrich, line 326:
```suggestion
```

This should be removed now I think?

oestrich (approved):

joshk (commented):
I'm leaving this as a comment vs an approve or request changes as I think it can be merged in as is, or some feedback could be addressed.

https://github.com/nerves-hub/nerves_hub_web/pull/1582/files#diff-03454c312d5c891b1ae7d916a1eb171cdf023ca00f9aa5155ea5633f7c5cbe81R608-R619
I'm not sure if its the naming of the function, or the return of the function, or maybe it just needs a spec. The function doesn't return a deployment or firmware, but a tuple where one of the items is the same across all results. I think this could just return Firmware instead of a tuple, but at the very least it would be good to add a spec to document what is going on. (same for the function above as well)

I would also love to see a 'state' added to deployments which tracks if deltas have been created or not. At the very least it would be good to document this as an issue so we can track it.

lawik:
I'll add specs and some types to try and clarify the usage. The `by_deployment` one would likely end up with the same target firmware ID, yeah. The product one is likely to pull multiple deployments. These two functions are used in the same way wrt to delta generation so I think it makes sense that they work in the same way. But I could see changing the one for deployments to only return a single target firmware ID and a list of sources. Not sure that's worth the ceremony.

I made it select specific fields to keep it lean for when there are a bunch of deployments and a bunch of firmwares. I'm not heavily opposed to doing the whole struct but I don't see much advantage there.

Can make an issue about tracking whether deltas have been generated.

lawik:
The specs clarified that the return had changed from when it was a SQL query and I fixed that. Looking at the failing test.

PR #1592, opened 2024-10-09
Add option for showing updated devices on the map

lawik:
This adds a way to indicate if the firmware is fully up to date on devices for the map.

It also cleans up a few performance aspects of the map:
- Don't block initial load on fetching the devices/markers as it can be slow for many devices
- Change query to be more specific and not fetch as many unused devices
- Change data-plumbing so that markers are sent over push_event instead of via attributes with extra JSON encode/decode.

The map should load significantly faster and will indicate "Loading..." while loading if delayed.

We should further check index usage for these new queries as I think they could be made even faster if they don't have indexes today.


lawik:
This should be both useful, nice for people sending out deployments (see them take effect live) and it should make for a heck of a demo. Unfortunately there are a number of tests that are broken by the change in initial load behavior. The tests are all for the dashboard so I'm inclined to disable them for now as we know we will change the dashboard a lot.

lib/nerves_hub/devices.ex
```diff
@@ -88,6 +88,24 @@ defmodule NervesHub.Devices do
     |> Repo.paginate(pagination)
   end
 
+  def get_minimal_device_location_by_org_id_and_product_id(org_id, product_id) do
+    from(d in Device)
```
joshk, line 92: this can be just `Device`

assets/js/app.js
```diff
@@ -257,15 +259,23 @@ Hooks.WorldMap = {
 
     // initialize the map
     this.map = L.map(mapId, mapOptionsNoZoom).setView([40.5, 10], 2);
+    window.addEventListener(
```
joshk, line 262: There is a nicer way of doing this. Instead of adding an event listener on the winder, you can add d something like this:
```javascript
    this.handleEvent('update-markets', (data) => {
      this.map.updateMarkers(data);
    });
```
Then you can have an `updateMarkers` function attached to the map class.

joshk: Similar to what is done on https://github.com/nerves-hub/nerves_hub_web/pull/1592/files#diff-2126216ecd2ff9ae9b5f0c1e84d8fecb4ee4e54dc21685f69a590c2162dec186R147-R152

joshk (changes requested):
I haven't tested this locally.
Why was the test file for the dashboard removed?
I don't think this is ready to be merged in, but I would be happy to deploy this to its own app for demo use?

lawik:
> I haven't tested this locally. Why was the test file for the dashboard removed? I don't think this is ready to be merged in, but I would be happy to deploy this to its own app for demo use?

I mentioned the tests breaking in the initial comment. The loading order was changed to start blank as a large initial load could hang the LiveView for longer. The change in behavior invalidsted the tests. I am not sure how to fix those tests (maybe easy if you are used to testing LiveViews but what I tried didn't work) and I think the dashboard view will churn a lot so not sure if those tests are worth reviving. But if it is easy, we can certainly do it.

I tagged this so a container got built and have tried it on NervesCloud. Works well. Performs better than the previous version.

lib/nerves_hub/devices.ex
lawik: Right you are

assets/js/app.js
lawik: Cool. I grabbed what they suggested in LiveView docs :D but will fix that for sure

lib/nerves_hub_web/live/dashboard/index.ex
```diff
@@ -42,15 +43,23 @@ defmodule NervesHubWeb.Live.Dashboard.Index do
 
     socket
     |> assign_devices_and_markers()
-    |> subscribe_to_devices()
```
joshk, line 45: I have an idea!
don't remove this, but follow my second comment

joshk: create a new function, similar to this:
```elixir
defp subscribed?(device) do
  Registry.count_select(NervesHub.PubSub, [{{"device:#{device.identifier}:internal", self(), :_}, [], [true]}]) > 0
end
```
and then on https://github.com/nerves-hub/nerves_hub_web/pull/1592/files#diff-4e1dd445a5a9fb5ca7c2ab2056a5820eea96b0bb22995b6e423c3cf818978fd1R69, check if the device is subscribed first.

joshk:
> > I haven't tested this locally. Why was the test file for the dashboard removed? I don't think this is ready to be merged in, but I would be happy to deploy this to its own app for demo use?
> 
> I mentioned the tests breaking in the initial comment. The loading order was changed to start blank as a large initial load could hang the LiveView for longer. The change in behavior invalidsted the tests. I am not sure how to fix those tests (maybe easy if you are used to testing LiveViews but what I tried didn't work) and I think the dashboard view will churn a lot so not sure if those tests are worth reviving. But if it is easy, we can certainly do it.
> 
> I tagged this so a container got built and have tried it on NervesCloud. Works well. Performs better than the previous version.

could we undelete the test file and add a tag to ignore the tests temporarily?

lawik:
I should still undelete the test file and I need to test these changes on an environment where I have an easier time triggering changes. But this **should** do what we want and never refresh more frequently than 500ms.

joshk (approved):
Looks great.

I'd still like to bring back the test file and just skip the tests, but I'm also fine with removing it as we will need to revamp this page soon.

lawik:
We can revive it if we find it useful later :)

Issue #1596, opened 2024-10-10
Device counts on deployment list is not accurate


lawik:
Something has gone wrong there because it always lists 1 if it lists anything.

PR #1600, opened 2024-10-14
Fix broken device count for deployments list

elinol:
Fixes https://github.com/nerves-hub/nerves_hub_web/issues/1596.


elinol:
It's only the devices that has to be counted from my understanding. In this list:
image

We'd might add a `0` or `-` instead of displaying nothing when deployments aren't connected to any devices, what do you think?

jjcarstens:
O, I'm happy with whatever you choose. I frequently mix up select+group statements in how they function and this could be another one of those times.

joshk (approved):
Ha, nice find!

Issue #1519, opened 2024-09-16
Indicate when console is disabled/not available


lawik:
If a device does not have `remote_iex: true` in it's config the terminal comes up blank.

The button should be disabled and an information icon should indicate that it is not configured for remote_iex.

Does the device currently report this bit of config? It should :)

We could possibly check if it has joined the console channel.

oestrich:
Some historical context: previously we disabled the button if the device was offline (and maybe also if the device wasn't connected to the console, I can't remember this part) but throughout determining if the device was online and connected it was wrong enough that the button was always enabled.

PR #1605, opened 2024-10-18
Disable console button when device console is not available

elinol:
An attempt to solve https://github.com/nerves-hub/nerves_hub_web/issues/1519.

Explanation for disabled button is not added, we'd might wait with that until new UI is ready?


lawik (approved):
Looks good to me. Can you check what the deal is with the CI thing?

elinol:
Don't know why that test suddenly behaves differently but the fix solved it, locally at least

PR #1613, opened 2024-10-26
Batch insert device metrics

joshk:
This decreases the metrics event processing time by half and reduces the number of Ecto queries, including no longer needing a transaction.

Screenshot 2024-10-26 at 8 12 21 PM

(the 3 and 12 relate to the number of telemetry spans)

(the 400,000 and 1,000,000 are the number of milliseconds used for processing)


lawik (approved):
Love it!

PR #1614, opened 2024-10-26
Optimize the queries needed for `Accounts.list_org_keys`

joshk:
- The Org related to the Device doesn't need to be preloaded
- Don't load the user (`created_by`) when sending the keys to devices


lawik (approved):
Seems reasonable :)

PR #1615, opened 2024-10-28
Increase the Orchestrator timer, and add a jitter

joshk:
This randomly spaces out the timers, which also reduces DB spikes. (minor spikes)


joshk:
This is a temporary addition until we change over to Oban for the calculator

lawik (approved):

PR #1611, opened 2024-10-25
Add marker cluster to map

elliotjon:
https://github.com/user-attachments/assets/9d8be70a-d774-4d89-ac7e-1743e2edf226

PR #1619, opened 2024-10-29
World Map clustering

lawik:
See video here: https://github.com/nerves-hub/nerves_hub_web/pull/1611

I tested this with 180.000 active and connected devices (faked). It loaded consistently but took a decent number of seconds. It was much faster down towards 50.000 and snappy below those types of numbers.

Once loaded it becomes easy to navigate and performant. I think this works. We could look at faster options. But the clustering makes 180K possible where it isn't without it.

We could try various things to constrain how many we load. I've looked at recommendations for performance and applied what I found. chunkedLoading option keeps it responsive while loading, renderer using Canvas makes it more performant.

I cut down heavily on non-used values because the first JSON transfer was 18 Mb and I think I got it down to about 1Mb for 180K. Maybe we could win more there with compression. That wasn't the slow part for me.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -7,6 +7,7 @@ import { LiveSocket } from 'phoenix_live_view'
 import L from 'leaflet/dist/leaflet.js'
 import Chart from 'chart.js/auto'
 import 'chartjs-adapter-date-fns';
+import 'leaflet.markercluster/dist/leaflet.markercluster.js'
 
 import hljs from 'highlight.js/lib/core'
 import bash from 'highlight.js/lib/languages/bash'
@@ -20,6 +21,8 @@ hljs.registerLanguage('shell', shell)
 
 import 'highlight.js/styles/stackoverflow-light.css'
 import 'leaflet/dist/leaflet.css'
+import 'leaflet.markercluster/dist/MarkerCluster.css'
+import 'leaflet.markercluster/dist/MarkerCluster.Default.css'
 
 import TimeAgo from 'javascript-time-ago'
 import en from 'javascript-time-ago/locale/en'
@@ -238,14 +241,17 @@ Hooks.WorldMap = {
     let mapId = this.el.id;
     this.markers = [];
 
-    var mapOptionsNoZoom = {
+    var mapOptionsZoom = {
       attributionControl: false,
-      zoomControl: false,
-      scrollWheelZoom: false,
+      zoomControl: true,
+      scrollWheelZoom: true,
       boxZoom: false,
       doubleClickZoom: false,
-      dragging: false,
-      keyboard: false
+      dragging: true,
+      keyboard: false,
+      maxZoom: 18,
+      minZoom: 1.4,
+      renderer: L.canvas()
     };
 
     var mapStyle = {
@@ -258,7 +264,8 @@ Hooks.WorldMap = {
     };
 
     // initialize the map
-    this.map = L.map(mapId, mapOptionsNoZoom).setView([40.5, 10], 2);
+    this.map = L.map(mapId, mapOptionsZoom).setView([0, 0], 1);
+    this.map.setMaxBounds(this.map.getBounds());
     this.handleEvent(
       "markers",
       ({ markers }) => {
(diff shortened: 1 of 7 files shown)


mix.lock
```diff
@@ -52,23 +51,23 @@
...
-  "phoenix_test": {:hex, :phoenix_test, "0.3.1", "adf5d67cb152fa1e0220527a9827db7f865a20817c7c0161315ba6fe86a8946c", [:mix], [{:deep_merge, "~> 1.0", [hex: :deep_merge, repo: "hexpm", optional: false]}, {:floki, ">= 0.30.0", [hex: :floki, repo: "hexpm", optional: false]}, {:jason, "~> 1.4", [hex: :jason, repo: "hexpm", optional: false]}, {:phoenix, "~> 1.7.10", [hex: :phoenix, repo: "hexpm", optional: false]}, {:phoenix_live_view, "~> 0.20.1", [hex: :phoenix_live_view, repo: "hexpm", optional: false]}], "hexpm", "03f24332a966673ff40d35c45f427c7671537ea508b5777141dd60e60cdae22a"},
+  "phoenix_test": {:hex, :phoenix_test, "0.3.2", "030d98c401ef95db564da6a15b5077909e11a532986690b757f40c105cda8be4", [:mix], [{:floki, ">= 0.30.0", [hex: :floki, repo: "hexpm", optional: false]}, {:jason, "~> 1.4", [hex: :jason, repo: "hexpm", optional: false]}, {:phoenix, "~> 1.7.10", [hex: :phoenix, repo: "hexpm", optional: false]}, {:phoenix_live_view, "~> 0.20.1", [hex: :phoenix_live_view, repo: "hexpm", optional: false]}], "hexpm", "538be4b25207e250b1939f6374bb1fc9f6346edefcb3f5e1daa51067544a3b76"},
```
joshk, line 62: Updating `phoenix_test` is not possible right now. You will need to revert this update if you want the test suite to pass.

joshk (changes requested):

mix.lock
lawik: Ah, I had to nuke and pave deps for some ungodly state they got stuck in but happy to revert that.

lawik:
@joshk this will be squashed and so on, but I think you should try the functionality, see how you like it :)

lawik:
Should be good to go. Ideally take a look and see how you enjoy it @joshk :)

assets/js/app.js
```diff
@@ -278,87 +298,79 @@ Hooks.WorldMap = {
...
+      fillOpacity: 1,
+      renderer: myRenderer
     };
 
-    // Clear previous defined device layer before adding markers
-    if (this.deviceLayer !== undefined) { this.map.removeLayer(this.deviceLayer); }
-
-    this.deviceLayer = L.geoJson(devices, {
-      pointToLayer: function (feature, latlng) {
+    var clusterLayer = L.markerClusterGroup({ chunkedLoading: true, chunkProgress: this.updateProgressBar });
```
nshoes, line 339: This was difficult for me to read, so I ran through it and made some adjustments:

```js
var defaultOptions = {
  radius: 6,
  weight: 1,
  opacity: 0,
  fillOpacity: 1,
  renderer: myRenderer,
  fillColor: "#4dd54f"
}

var markerOfflineColor = "rgba(196,49,49,1)"
var markerOutdatedColor = "rgba(99,99,99,1)"

var devices = this.markers.reduce(function(acc, marker) {
  let location = marker["l"]
  var latLng = [location["at"], location["ng"]]

  // if no location or we don't care about mode, move on
  if (!location["ng"] && !location["at"]) return acc
  if (!["connected", "updated"].includes(mode)) return acc

  if (marker["s"] == "connected") {
    return L.circleMarker(latLng, defaultOptions)
  } else if (mode == "connected") {
    return L.circleMarker(latLng, Object.assign(defaultOptions, { fillColor: "rgba(196,49,49,1)" }))
  } else if (mode =="updated") {
    return L.circleMarker(latLng, Object.assign(defaultOptions, { fillColor: "rgba(99,99,99,1)" }))
  }
}, [])
```

It's not tested but it's close. Let me know what you think!

lib/mix/tasks/gen.devices.ex
```diff
@@ -0,0 +1,41 @@
+defmodule Mix.Tasks.Gen.Devices do
```
nshoes, line 1: I wrote something similar this morning seeding thousands of devices locally; this is nice!

lib/nerves_hub_web/live/dashboard/index.ex
```diff
@@ -7,7 +7,7 @@ defmodule NervesHubWeb.Live.Dashboard.Index do
 
   alias Phoenix.Socket.Broadcast
 
-  @default_refresh 5000
+  @default_refresh 50000
```
nshoes, line 10: Any reason for picking 50 seconds? My brain wants this to be 45 or 60s 😆, but I don't feel strongly.

lib/nerves_hub_web/live/dashboard/index.ex
```diff
@@ -117,8 +117,8 @@ defmodule NervesHubWeb.Live.Dashboard.Index do
 
   defp generate_map_marker(
          %{
-           id: id,
-           identifier: identifier,
+           id: _id,
+           identifier: _identifier,
```
nshoes, line 121:
```suggestion
```

assets/js/app.js
```diff
@@ -237,15 +240,31 @@ Hooks.WorldMap = {
...
 
-    var mapOptionsNoZoom = {
+        // progressBar.classList.remove("hidden");
+        // progressBar.style.width = Math.round(processed / total * 100) + '%';
+      }
+
+      if (processed === total) {
+        // all markers processed - hide the progress bar:
+        // progress.classList.add = 'hidden';
+      }
```
nshoes, line 254: Did this not work properly with your changes?

assets/js/app.js
lawik: good stuff, reworked a bit but a similar thing will be there :)

lawik:
@nshoes looked at your notes, fixed all of it I believe :)

lib/nerves_hub_web/live/dashboard/index.ex
nshoes: No no no please no I don't feel that strongly 😆😆😆

nshoes (approved):
🚀 🚀 🚀 🚀 🚀

lawik:
Need to fix CI though

nshoes:
@lawik done! :)

PR #1620, opened 2024-10-29
Clear inflight updates for unhealthy devices

### LINKED ISSUE #1617: Devices in `penalty box` take up concurrent slot of deployments
We have deployments which targets several hundred devices. In some cases, it takes hours or days to get through updating devices. While investigating, it was noted that several devices in flight were actually now in the penalty box but were still being counted in the deployment orchestrators concurrent limit. So the effect in our case was that only 1 or 2 devices were being updated at a time

nshoes:
Fixes #1617.

If we know a device is unhealthy, we want to clear inflight updates. This unblocks the orchestrator and allows it to continue updating as many devices as possible.


jjcarstens (approved):
🎉 Thanks!

PR #1627, opened 2024-10-31
Show helpful message when uploading duplicate firmware

### LINKED ISSUE #1473: Uploading duplicate firmware gives unfriendly error message
**Describe the bug**
When uploading a firmware build that has already been uploaded successfully, you are given a generic "Unknown error uploading firmware.
" error message rather than a more specific message saying its a duplicate. The logs show 

**To Reproduce**
1. Go to the /firmware/upload page
2. Click on the box to upload firmware
3. Select a firmware build
4. upload
5. go back to the /firmware/upload page
6. try and re upload the same firmware that you did in the previous step
7. See error `Unknown error uploading firmware.`

**Expected behavior**
A message saying this firmware has already been uploaded, so that I don't think there's something wrong with the firmware build. 

**Screenshots**
Screenshot 2024-08-27 at 9 00 10 AM

**Smartphone (please complete the following information):**

**Additional context**
Error logs observed:
```
16:00:59.167 [error] pid=<0.1763.0> mfa=NervesHub.Firmwares.create_firmware/3 Error while publishing firmware: #Ecto.Changeset<action: :insert, changes: %{size: 32461873, version: "1.xx", description: "xx", author: "xx", org_id: x, product_id: x, uuid: "x", platform: "x", architecture: "armx, delta_updatable: x, org_key_id: x, upload_metadata: %{"s3_key" => "firmware/x/x.fw"}}, errors: [uuid: {"has already been taken", [constraint: :unique, constraint_name: "firmwares_product_id_uuid_index"]}], data: #NervesHub.Firmwares.Firmware<>, valid?: false>
```

nshoes:
Fixes #1473.

This adds an error clause to provide helpful feedback when uploading firmware that has already been uploaded. [There's already a test for this constraint](https://github.com/nerves-hub/nerves_hub_web/blob/b7efd90ff828c0a6ee52b0b40ba3065cd68b6985/test/nerves_hub/firmwares_test.exs#L53), but let me know if a LiveView test is wanted/needed.

CleanShot 2024-11-01 at 08 35 36

### DIFF
--- a/lib/nerves_hub_web/live/firmware.ex
+++ b/lib/nerves_hub_web/live/firmware.ex
@@ -41,6 +41,7 @@ defmodule NervesHubWeb.Live.Firmware do
     socket
     |> page_title("Upload Firmware - #{product.name}")
     |> assign(:error_message, nil)
+    |> assign(:error_code, nil)
     |> allow_upload(:firmware,
       accept: ~w(.fw),
       max_entries: 1,
@@ -140,6 +141,16 @@ defmodule NervesHubWeb.Live.Firmware do
        }} ->
         error_feedback(socket, "No matching product could be found.")
 
+      {:error,
+       %Ecto.Changeset{
+         errors: [
+           uuid:
+             {"has already been taken",
+              [constraint: :unique, constraint_name: "firmwares_product_id_uuid_index"]}
+         ]
+       } = changeset} ->
+        error_feedback(socket, changeset)
+
       {:error, %Ecto.Changeset{}} ->
         error_feedback(socket, "Unknown error uploading firmware.")
 
@@ -149,6 +160,17 @@ defmodule NervesHubWeb.Live.Firmware do
       _ ->
         error_feedback(socket, "Unknown error uploading firmware")
     end
+  end
+
+  defp error_feedback(
+         socket,
+         %Ecto.Changeset{errors: [uuid: {"has already been taken", _}]} = changeset
+       ) do
+    socket
+    |> assign(:error_message, "Duplicate Firmware Is Not Allowed")
+    |> assign(:error_code, :duplicate_firmware)
+    |> assign(:uuid, changeset.changes.uuid)
+    |> noreply()
   end
 
   defp error_feedback(socket, %Ecto.Changeset{} = changeset) do
--- a/lib/nerves_hub_web/live/firmware_templates/upload_firmware_template.html.heex
+++ b/lib/nerves_hub_web/live/firmware_templates/upload_firmware_template.html.heex
@@ -9,7 +9,14 @@
     <.live_file_input upload={@uploads.firmware} required />
 
     <div class="has-error">
-      <span :if={@error_message} class="help-block"><%= @error_message %></span>
+      <span :if={@error_message} class="help-block">
+        <%= case @error_code do %>
+          <% :duplicate_firmware -> %>
+            <%= @error_message %> <a href={"/org/#{@org.name}/#{@product.name}/firmware/#{@uuid}"} target="_blank">View Firmware</a>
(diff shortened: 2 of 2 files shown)


lib/nerves_hub_web/live/firmware.ex
```diff
@@ -140,6 +140,16 @@ defmodule NervesHubWeb.Live.Firmware do
...
 
+      {:error,
+       %Ecto.Changeset{
+         errors: [
+           uuid:
+             {"has already been taken",
+              [constraint: :unique, constraint_name: "firmwares_product_id_uuid_index"]}
+         ]
+       }} ->
+        error_feedback(socket, "Duplicate firmware uploads are not allowed")
```
jjcarstens, line 151: Do we get the uuid back in that changeset somewhere? It would be really cool to say something like `Duplicate firmware - <link-to-firmware>`

nshoes: Oh, good idea! I'll try.

nshoes: Done! Updated the gif in the PR description.

lib/nerves_hub_web/live/firmware.ex
```diff
@@ -140,6 +140,22 @@ defmodule NervesHubWeb.Live.Firmware do
...
+             {"has already been taken",
+              [constraint: :unique, constraint_name: "firmwares_product_id_uuid_index"]}
+         ]
+       } = c} ->
+        route =
+          "/org/#{socket.assigns.org.name}/#{socket.assigns.product.name}/firmware/#{c.changes.uuid}"
+
+        error_feedback(
+          socket,
+          "Duplicate firmware - <a href='#{route}' target='blank'>View Firmware</a>"
```
joshk, line 156: I'd prefer to find a way to not include an `<a href>` in a string error message, and then also use `raw` for displaying the message.

It might require a little refactoring, eg. instead of line 164 adding an error message to the assigns, maybe we add an `error_code`, like `:duplicate_firmware` and then use a helper function in the template which uses a `case` statement to define the long form message.

joshk (changes requested):

lib/nerves_hub_web/live/firmware.ex
nshoes: I like that a lot more, I'll implement!

nshoes: Changes pushed. I think instead of handling this scenario in a pattern-match with `error_feedback`, it could just be handled inline since we have to pattern-match on the UUID error and then pattern-match it again which feels weird. Lmk what you think!

lib/nerves_hub_web/live/firmware.ex
```diff
@@ -151,6 +162,20 @@ defmodule NervesHubWeb.Live.Firmware do
...
+         socket,
+         %Ecto.Changeset{errors: [uuid: {"has already been taken", _}]} = changeset
+       ) do
+    socket
+    |> assign(:error_message, "Duplicate Firmware Is Not Allowed")
+    |> assign(:error_code, :duplicate_firmware)
+    |> assign(
+      :firmware_path,
+      "/org/#{socket.assigns.org.name}/#{socket.assigns.product.name}/firmware/#{changeset.changes.uuid}"
+    )
```
jjcarstens, line 175:
```suggestion
    |> assign(:error_code, {:duplicate_firmware, uuid})
```

Since we have control of this structure, we could just include the duplicate UUID and let the case handling of that view build the URL for it (or whatever else). That way we're not so coupled to 2 independent pieces of data having to exist in the assigns (`@error_code` and `@firmware_path`)

joshk (approved):
This generally looks good to me

PR #1628, opened 2024-11-01
Add support for filtering devices on alarms

elinol:
Adds filtering support for:
- Devices with alarms / without alarms
- Devices with specific alarms (list with current alarm types for product)

Also adds alarm counter on product dashboard with link to filtered devices list.


lib/nerves_hub/devices/alarms.ex
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  @doc """
+  Creates a list with all current alarm types for a product.
+  """
+  def get_current_alarms(product_id) do
+    query_current_alarms(product_id)
+    |> Repo.all()
+    |> Enum.map(fn %{data: data} ->
+      Map.keys(data["alarms"])
```
elinol, line 47: Currently in this format: `NervesHubHealth.HealthCheckFailed`
We could make them more human friendly perhaps?

elinol:
Looks alright in dropdown selector:
image

But not when selected:
image

Worth it to fix now or wait for UI update?

lib/nerves_hub/devices/alarms.ex
jjcarstens: Using `inspect/1` on the alarm name will pretty print without `Elixir.` - We could also add a helper in `DeviceHealth.save/1` to preemptively adjust/cast `:alarms` keys into the prettier format

lib/nerves_hub/devices/alarms.ex
```diff
@@ -0,0 +1,91 @@
...
+  alias NervesHub.Repo
+  alias NervesHub.Devices.Device
+  alias NervesHub.Devices.DeviceHealth
+
+  @doc """
+  Selects device id:s for devices that has alarm(s) in it's latest health record.
+  Used when filtering devices.
+  """
+  def query_devices_with_alarms() do
+    from(
```
joshk, line 12: Could we change this only to use piped queries instead of the mix of the two.

lib/nerves_hub/devices/alarms.ex
```diff
@@ -0,0 +1,91 @@
...
+    |> join(:inner, [lr], d in Device, on: lr.device_id == d.id)
+    |> select([lr, o], o.id)
+  end
+
+  @doc """
+  Selects device id:s for devices that has provided alarm in it's latest health record.
+  Used when filtering devices.
+  """
+  def query_devices_with_alarm(alarm) do
+    from(
```
joshk, line 26: same as above

lib/nerves_hub/devices/alarms.ex
joshk: I'm wondering if we also remove `NervesHubHealth`?

joshk (changes requested):

lib/nerves_hub/devices/alarms.ex
elinol: Oh, it's `Elixir.NervesHubHealth.HealthCheckFailed` actually so solved it by splitting on dots and using the last item.

lib/nerves_hub/devices/alarms.ex
elinol: yes!

lib/nerves_hub/devices/alarms.ex
```diff
@@ -0,0 +1,98 @@
...
+  alias NervesHub.Devices.Device
+  alias NervesHub.Devices.DeviceHealth
+
+  @doc """
+  Selects device id:s for devices that has alarm(s) in it's latest health record.
+  Used when filtering devices.
+  """
+  def query_devices_with_alarms() do
+    (lr in subquery(latest_row_query()))
+    |> from()
```
joshk, line 13: Is this `from()` needed?

lib/nerves_hub/devices/alarms.ex
```diff
@@ -0,0 +1,98 @@
...
+    |> query_current_alarms()
+    |> select([a], count(a))
+    |> Repo.one!()
+  end
+
+  @doc """
+  Selects latest health per device if alarms is populated and device belongs to product.
+  """
+  def query_current_alarms(product_id) do
+    from(
```
joshk, line 68: Can you please switch these to pipe queries.

joshk (changes requested):

lib/nerves_hub/devices/alarms.ex
elinol: yes

lib/nerves_hub/devices/alarms.ex
elinol: I can! I found this way easier to follow somehow

lib/nerves_hub/devices/alarms.ex
jjcarstens: Will there be more alarms in the future? We have lots of alarms not explicitly scoped to `NervesHubHealth`. I'm not sure if that's the only thing expected to be reported here though the name `alarms` seems to suggest it might

elinol: From what I understand the alarm functionality will be built out, and separated into it's own context. This functions will need an update then :)

joshk (approved):
This is a solid step forward!

lib/nerves_hub/devices/alarms.ex
lawik: NervesHubHealth came from the separate library that we won't be shipping. We can go forward with whatever naming we prefer.

lawik: I agree with using `inspect/1` to get it without the Elixir prefix :)

lawik:
So the NervesHubHealth prefix for the alarm is entirely up to the thing making the alarm. We can decide if we want a different convention in the default alarms for NervesHubLink as we won't be shipping the NervesHubHealth library. It got folded into NervesHubLink's features implementation.

elinol:
image

joshk (approved):
Looks great to me. Some of the Ecto bits are bending my brain, but I think we can try this out in prod and improve from there.

elinol:
@joshk this should be ready now!

joshk:
Awesome work @elinol !

PR #1648, opened 2024-11-19
Show current alarms on device page

elinol:
Display any current alarms for device.
Click on alarm to show description if provided.

### DIFF
--- a/lib/nerves_hub/devices/alarms.ex
+++ b/lib/nerves_hub/devices/alarms.ex
@@ -1,6 +1,7 @@
 defmodule NervesHub.Devices.Alarms do
   import Ecto.Query
   alias NervesHub.Repo
+  alias NervesHub.Devices
   alias NervesHub.Devices.Device
   alias NervesHub.Devices.DeviceHealth
 
@@ -59,6 +60,15 @@ defmodule NervesHub.Devices.Alarms do
     |> query_current_alarms()
     |> select([a], count(a))
     |> Repo.one!()
+  end
+
+  def get_current_alarms_for_device(device) do
+    device.id
+    |> Devices.get_latest_health()
+    |> case do
+      %DeviceHealth{data: data} -> data["alarms"]
+      _ -> nil
+    end
   end
 
   @doc """
--- a/lib/nerves_hub/devices/alarms.ex
+++ b/lib/nerves_hub/devices/alarms.ex
@@ -66,8 +66,12 @@ defmodule NervesHub.Devices.Alarms do
     device.id
     |> Devices.get_latest_health()
     |> case do
-      %DeviceHealth{data: data} -> data["alarms"]
-      _ -> nil
+      %DeviceHealth{data: %{"alarms" => alarms}} when is_map(alarms) ->
+        for {alarm, description} <- alarms,
+            do: {String.trim_leading(alarm, "Elixir."), description}
+
+      _ ->
+        nil
     end
   end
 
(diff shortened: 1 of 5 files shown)


lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -362,4 +364,11 @@ defmodule NervesHubWeb.Live.Devices.Show do
       device.connecting_code
     end
   end
+
+  defp format_alarm(alarm), do: String.trim_leading(alarm, "Elixir.")
+
+  defp has_description?(""), do: false
```
nshoes, line 370: I don't know much about alarms yet, but why would we ever set an alarm description to `""`?

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -362,4 +364,11 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
+  defp has_description?(description) when is_binary(description), do: true
+  # To cover for cases where alarm description is provided as an invalid type.
+  defp has_description?(_), do: false
```
jjcarstens, line 373:
```suggestion
  defp has_description?(description) do
    is_binary(description) and byte_size(description) > 0
  end
```
I also think we could reduce overhead here with an all-encompassing check rather than patterns to clearly show what is going on

lib/nerves_hub/devices/alarms.ex
```diff
@@ -61,6 +62,15 @@ defmodule NervesHub.Devices.Alarms do
     |> Repo.one!()
   end
 
+  def get_current_alarms_for_device(device) do
+    device.id
+    |> Devices.get_latest_health()
+    |> case do
+      %DeviceHealth{data: data} -> data["alarms"]
```
jjcarstens, line 69:
```suggestion
      %DeviceHealth{data: %{"alarms" => alarms}} when is_list(alarms) ->
        for {alarm, description} <- alarms, do: {String.trim_leading(alarm, "Elixir."), description}
```
I think trimming here would be useful as this would be needed anywhere in the app and not just the UI

lib/nerves_hub_web/live/devices/show.ex
elinol: I wasn't sure what do expect from descriptions so went with trying to cover all cases. @jjcarstens suggestion does that a lot smarter though :D

lib/nerves_hub/devices/alarms.ex
elinol: I agree

lib/nerves_hub/devices/alarms.ex
```diff
@@ -61,6 +62,19 @@ defmodule NervesHub.Devices.Alarms do
     |> Repo.one!()
   end
 
+  def get_current_alarms_for_device(device) do
+    device.id
+    |> Devices.get_latest_health()
+    |> case do
+      %DeviceHealth{data: %{"alarms" => alarms}} when is_map(alarms) ->
+        for {alarm, description} <- alarms,
+            do: {String.trim_leading(alarm, "Elixir."), description}
```
jjcarstens, line 71:
```suggestion
        for {alarm, description} <- alarms, into: %{},
            do: {String.trim_leading(alarm, "Elixir."), description}
```
I don't know if it matters. I originally thought these were a list, but looks like they are a map. If that's a case, then this would ensure it's placed back into a map

jjcarstens (approved):

nshoes (approved):

lib/nerves_hub/devices/alarms.ex
elinol: Yes, the alarms comes in a map. For current use cases it's fine to pass them on as maps here.

PR #1651, opened 2024-11-20
Change health section to not show boxes for non-reported default values

lawik:
Before:
Screenshot 2024-11-20 at 14 07 48

After:
Screenshot 2024-11-20 at 14 08 06

The CPU percentage report will be nicer-looking as it becomes a default with the extensions updates. This was just a small nuisance. Non-reported values can just not be shown as that can be intentional on the user's end.


lawik:
@elinol feel free to review as well, can't tag you properly :)

elinol:
I think this is great! Was just about to start a conversation regarding exactly this, I'm doing some stuff with metrics. You have my unofficial approval

nshoes (approved):
Good improvement!

PR #1652, opened 2024-11-20
Don't recalculate deployments for devices

nshoes:
This PR aims to remove nearly all of the device deployment recalculations. We only want to recalculate when a device joins and has no `deployment_id`.


jjcarstens (approved):
LGTM! A couple things I noticed which may have been addressed, but just confirming

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -35,7 +35,7 @@ defmodule NervesHubWeb.DeviceChannel do
 
   @decorate with_span("Channels.DeviceChannel.handle_info:after_join")
   def handle_info({:after_join, params}, %{assigns: %{device: device}} = socket) do
-    device = maybe_update_deployment(device)
+    device = if device.deployment_id, do: device, else: Deployments.set_deployment(device)
```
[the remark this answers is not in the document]
nshoes: We haven't discussed this specifically, no.

> besides abundance of caution to prevent firmware starting to download we know we fail

That's more than enough reason for me.

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -387,20 +315,6 @@ defmodule NervesHubWeb.DeviceChannel do
     {:noreply, socket}
   end
 
-  def handle_in("check_update_available", _params, socket) do
```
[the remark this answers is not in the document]
nshoes: Oh duh, that makes sense. I thought this was cruft since there were no internal calls with `check_update_available`. When would a device request this? I don't see any calls to this in `cloud_bridge` or `nerves_hub_link`. I'm down to keep this around if it's used (minus recalc things).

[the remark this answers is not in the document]
nshoes: This code path is linked to a [nerves_hub_link issue that hasn't been fixed](https://github.com/nerves-hub/nerves_hub_link/issues/47) yet, so I'll bring this back.

lib/nerves_hub_web/channels/device_channel.ex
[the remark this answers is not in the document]
joshk: Hmmmmmm, when you say `we know we fail`, could you give an example of how this might happen?

lib/nerves_hub_web/channels/device_channel.ex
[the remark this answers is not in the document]
joshk: I think this is safe to remove. This was added by me a while ago based on the idea that a device could check if there is an update and make sure it has a current and valid signed download url, but after talking with Frank (and Jon) the preference was that Hub should tell the device that it should update, which moves the smarts from the device to hub. I didn't get to complete the changes in Link, so I think we could remove this.

priv/repo/migrations/20241120211102_remove_deployments_recalculation_type.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHub.Repo.Migrations.RemoveDeploymentsRecalculationType do
```
joshk, line 1: We might want to leave this for a slightly later PR, as it came cause some transient errors when deploying to production.

joshk (changes requested):
I love this PR sooooo much!

Removing code and complexity is wonderful!

One thing we should include is changing the tags and version constraint UI bits to be in their own grouping which also explains that its only used for matching when the device is not assigned to a deployment.

lib/nerves_hub_web/channels/device_channel.ex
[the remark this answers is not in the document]
jjcarstens: Digging into it deeper, there are 3 main things which could fail (2 are from [fwup specific check](https://github.com/nerves-project/nerves_system_rpi0/blob/main/fwup.conf#L373) when applying firmware)
* Platform doesn't match.
  * This is typically the Elixir app name and the NervesHub product
  * `fwup` will fail this
* Architecture doesn't match (`arm` vs `x86_64`, etc)
  * For serial numbers from crypto chips, this would never happen. For people managing serial numbers manually view a writeable file or other means, you could inadvertently reuse an old serial on a difference architecture device
  * `fwup` will fail this
* Reported firmware version no longer meets the deployment version requirement
  * i.e. I might assign a never-connected device to this deployment, but if it is <= 1.0, I don't want it to use this deployment firmware

I think the first 2 are mostly avoidable. If you move a device (i.e change the product), [deployment is removed](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/devices.ex#L1022). And different arch is _highly_ unlikely. My first thought with these is I don't remember _how early_ we fail. If it is "download the whole file, then fail", then that's a lot of wasted bytes for devices on cellular.

So maybe it is if we want to consider a version check on device connect. We had previously talked about version checks and it's wishy-washy right now, so maybe we punt and see what we run into.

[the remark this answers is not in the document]
joshk: Isn't the Platform and Arch configured once and once only?

And regarding the version check, has SmartRent ever added a Version condition to an existing Deployment which has Devices associated with it?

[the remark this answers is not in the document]
joshk: Also, regarding the code, I think:
```
device = Deployments.maybe_set_deployment(device)
```
would be nicer. You can move the `if` clause into that function, which feels slightly tidier.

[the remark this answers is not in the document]
jjcarstens: If I re-burn an SD card with a different firmware project, platform is overwritten.

We use version conditions, but need to ask if we've altered the version condition of _existing_ deployment past the initial create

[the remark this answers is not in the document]
joshk: ahhhh, hmmmm, ok, it might be worth validating if a device can still belong in a deployment. I guess this is a fair safety net to keep in place.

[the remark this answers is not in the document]
jjcarstens: Ya. We can do it without extra DB work. But...what is our action if it _doesn't_ belong anymore?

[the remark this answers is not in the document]
joshk: I think the code would become something like:
```elixir
device = 
  device
  |> Deployments.verify_deployment_membership()
  |> Deployments.maybe_set_deployment()
```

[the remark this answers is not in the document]
jjcarstens: Sorry, I meant what do we do with the unmatched deployment? Just remove it, ignore, put in penalty box, some new status?

As far as code, I agree with Nate's original. The conditional of setting the deployment is really only relevant here. We would end up making a single use function and continuing to bloat a context which I already have strong feelings against typical context use anyway 😉 But just my 2¢ and I'll be happy with whatever is needed to progress

priv/repo/migrations/20241120211102_remove_deployments_recalculation_type.exs
nshoes: Done! Good callout.

lib/nerves_hub_web/channels/device_channel.ex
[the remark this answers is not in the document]
nshoes: I just implemented:
- if device firmware doesn't match deploy firmware, remove device from deployment
- look for and set alternate deployment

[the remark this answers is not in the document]
nshoes: After thinking about this and talking to Jon, I say we go a different route: if a device's platform/arch doesn't match, we remove it from the deployment and flag the device properly. I'm not sure _how_ we want to do it though. The `status` is taken and a device is in the penalty box if `updates_blocked_until` is in the future.

Unless there's something I'm missing, I'd like to add `deployment_stalled_reason` to `device` to indicate why a device isn't on the deployment happy path. In the new world we'll track the higher-level deployment status in `deployment_release.status`, but I want to add additional context at the device level. This would also allow us to expand deployment failure reasons in the future and bubble up useful device info when viewing a deployment: "5 devices in penalty box, 2 devices separated from deployment".

Thoughts?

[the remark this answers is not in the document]
jjcarstens: Sounds great to me. Happy with the name, but for the sake of spit-balln' here is a few more:
* `:deployment_conflict`
* `:deployment_problem`
* `deployment_confidence: :valid | :bad_arch | :bad_platform | ...`
* Adding some iteration of the above to the `:status` field (which may be bad cause I'm not 100% sure what that is supposed to be for)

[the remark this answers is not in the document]
nshoes: `status` is being used for Device Connections currently, we should probably name that column better.

I like `deployment_conflict`, thanks for the suggestions.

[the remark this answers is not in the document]
nshoes: We now set `device.deployment_conflict` when deployment doesn't match. Current options are `bad_platform`, `bad_architecture`, and `bad_architecture_and_platform`.

lib/nerves_hub/devices.ex
```diff
@@ -1067,7 +1034,7 @@ defmodule NervesHub.Devices do
     |> Repo.transaction()
     |> case do
       {:ok, %{move: updated}} ->
-        _ = broadcast(updated, "moved")
+        # Deployments.set_deployment(updated)
```
nshoes, line 1037: What should we do here after we move a device? Before it would attempt to find an alternate deployment. Setting `device.deployment_conflict` feels wrong but only semantically; there isn't an issue with the deployment, but flagging it as `:moved` feels right. Maybe we should rename the column?

joshk: I'm wondering if we don't find a deployment for the device and leave that for when it connects again.
I'm also thinking that we keep that broadcast, but instead of finding a new deployment, we instead disconnect the device and have it reconnect since the device is now bound to a new product.
Thoughts?

lib/nerves_hub/devices/device.ex
```diff
@@ -63,6 +63,10 @@ defmodule NervesHub.Devices.Device do
     field(:connecting_code, :string)
     field(:connection_metadata, :map, default: %{})
 
+    field(:deployment_conflict, Ecto.Enum,
```
joshk, line 66: Can you tell me more about why this column was added?

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -34,7 +34,9 @@ defmodule NervesHubWeb.DeviceChannel do
 
   @decorate with_span("Channels.DeviceChannel.handle_info:after_join")
   def handle_info({:after_join, params}, %{assigns: %{device: device}} = socket) do
-    device = maybe_update_deployment(device)
+    device =
+      device
+      |> Deployments.verify_deployment_membership()
```
joshk, line 39:
```suggestion
    device = Deployments.verify_deployment_membership(device)
```
I think this is a bit cleaner.

lib/nerves_hub_web/live/deployments/edit.html.heex
```diff
@@ -66,18 +66,10 @@
   </div>
 
   <h3 class="mb-2">Conditions</h3>
```
joshk, line 68: I think we could change the header to be a bit more specific, and keep the description (deleted below) but change it to provide further information on how this works.

eg. `Device matching conditions`

`These conditions are used for matching devices which don't have a configured deployment. The matching is undertaken when a device connects to the platform.`

joshk (changes requested):
In addition to my inline comments, we should update https://github.com/nerves-hub/nerves_hub_web/pull/1652/files#diff-214ee95769df74267dd420f9042d92be3bc8543f4942bf9049d63577b9aa54edR46 to explain how tags are only used for matching when a device doesn't have a deployment. I think it would be best to move the `Distributed to` and `Version requirement` information to a new section called `Device matching conditions` and place the information before the current `Deployment details` section.

We should also do something similar to the 'New deployment' workflow:
Screenshot 2024-12-04 at 5 57 06 PM

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Agreed! Brain fart for me :P

lib/nerves_hub/devices.ex
nshoes: Ah! I really like that, I'll implement.

lib/nerves_hub/devices/device.ex
nshoes: We set this column when [verifying deployment membership](https://github.com/nerves-hub/nerves_hub_web/pull/1652/files#diff-48cec1d45c7f0dac953ea8a89cecbf1c568e6a9b61b02a863a3ac4c23fb68920R307) if the architecture and/or platform doesn't match. We can then surface why the device isn't in a deployment and make it easy to find and search devices that aren't a part of a deployment.

lib/nerves_hub_web/live/deployments/edit.html.heex
nshoes: Done!

lib/nerves_hub/devices.ex
nshoes: What's the need for disconnecting and then reconnecting? When a device is moved, its `deployment_id` is cleared so there's no need to verify its deployment membership or any other function of `DeviceChannel`'s `after_join`. I could very well be missing something, but this is why I suggested we flag the device as `:moved`.

lib/nerves_hub_web/live/deployments/edit.html.heex
nshoes: CleanShot 2024-12-04 at 14 28 06@2x

nshoes:
@joshk made adjustments to the `show` and `new` views:

#### show

CleanShot 2024-12-04 at 14 31 26@2x

#### new

CleanShot 2024-12-04 at 14 31 48@2x

lib/nerves_hub/devices.ex
joshk: When a device connects it looks for a deployment (if it doesn't have one). So when moving a device from one product to another, clearing the deployment and disconnecting the device would also take care of assigns and data the device should refresh. eg. we send signing keys to the device on connect, so this would make sure the device would be resent the correct signing keys.

lib/nerves_hub/devices/device.ex
joshk: Ok, seems fair. Do we show this information anywhere right now?

joshk:
Love the UI changes.

For the `show` page, could we move the matching conditions to below the `Deployment details` section?

nshoes:
Done!

joshk:
@nshoes can you resolve the PR merge conflicts and then we can merge this in!

lib/nerves_hub/devices.ex
nshoes: Got it, that makes sense!

lib/nerves_hub/devices/device.ex
nshoes: We do not; but since it's added in here, we should. I'll implement.

nshoes: I think this info may need a bigger callout, but I think this is fine for now:
CleanShot 2024-12-05 at 08 34 57@2x

nshoes: I think an audit message makes more sense here :) Good callout.

nshoes: Fixed, let me know if you have feedback on [the audit description](https://github.com/nerves-hub/nerves_hub_web/pull/1652/commits/21ec1f9044aa1c2f060e246c12dac8f9cb0d1fd7#diff-48cec1d45c7f0dac953ea8a89cecbf1c568e6a9b61b02a863a3ac4c23fb68920R337).

nshoes: More changes from our conversations on Slack: https://github.com/nerves-hub/nerves_hub_web/pull/1652/commits/7258ca6d4db16f9cd53b0ff00deecb604efa5916

joshk (approved):
Amazing work!

This change turned out to be bigger than we thought, and you rocked it!

PR #1653, opened 2024-11-20
Managed Deployments: Groups and Releases

nshoes:
This work contains two sets of changes: the creation of `DeploymentRelease`s and the renaming of `deployments` to `deployment_groups`. `DeploymentRelease` integration will be done in the next body of work for managed deploys, this PR has become too big to add any more functionality.

So many things have been renamed that it's easier to list what _hasn't_ been renamed:
- `deployments` table name
- any indexes with `deployments` in the name
- `deployment_id` foreign key in DB and application code
- pub sub topic and events

Database-related renaming will be done later. Pub sub topic and events should and will be adjusted, but this is not the time. It won't be a simple rename and needs more care and thought.


lib/nerves_hub/audit_logs/templates/deployment_group_templates.ex
```diff
@@ -0,0 +1,47 @@
...
+  @moduledoc """
+  Templates for and handling of audit logging for deployment operations.
+  """
+  alias NervesHub.Accounts.User
+  alias NervesHub.AuditLogs
+  alias NervesHub.Devices.Device
+  alias NervesHub.ManagedDeployments.DeploymentGroup
+
+  @spec audit_deployment_created(User.t(), DeploymentGroup.t()) :: :ok
+  def audit_deployment_created(user, deployment_group) do
```
nshoes, line 11: These function names should change, but they shouldn't just have `_group` added to them. I want to make a more wholistic change to device template that removes the resource name from the function names, e.g. `DeploymentGroupTemplates.audit_created`. We'll save that work for later.

lib/nerves_hub/managed_deployments/inflight_deployment_check.ex
```diff
@@ -0,0 +1,15 @@
+defmodule NervesHub.ManagedDeployments.InflightDeploymentCheck do
```
nshoes, line 1: No need to rename as we're planning on keeping the term "deployments" to describe the literal process instead of a resource.

lib/nerves_hub_web.ex
```diff
@@ -106,6 +106,10 @@ defmodule NervesHubWeb do
 
       def page_title(socket, page_title), do: assign(socket, :page_title, page_title)
 
+      @spec sidebar_tab(
```
nshoes, line 109: Sneaking this in :)

joshk (approved):
This is awesome!

Could you please test on QA before merging in?

Issue #1647, opened 2024-11-19
Extensions improvement: Indicate on device details different extension setting


lawik:
If a device has disabled any extensions that make it differ from the main product we can indicate this.

lawik:
Fixed in #1654

PR #1654, opened 2024-11-21
Hide disabled extensions in the UI

### LINKED ISSUE #1646: Extensions improvement: Hide UI if extension is disabled at product level
- Hide filters for metrics, alarms
- Hide sections on device itself

lawik:
We shouldn't give up UI space for functionality that has been disabled.

There also needs to be a way to realize that a specific device has overridden the product-wide settings and won't get health/geo information.

This PR addresses that. Happy to take any notes. Keeping it separate from extensions PR because this is just UI bits.

Extensions disabled at the product level, UI hidden:
Screenshot_20241121_094541

Extensions enabled at the product level and either not set or enabled on device level:
Screenshot_20241121_094520

Extensions disabled at device-level overriding the product-wide setting:
Screenshot_20241121_094459

Fixes #1646 and #1647


lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -363,4 +364,13 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
     end
   end
+
+  defp extension_overrides(device, product) do
+    device.extensions
+    |> Map.from_struct()
+    |> Enum.filter(fn {extension, enabled} ->
+      enabled == false and Map.get(product.extensions, extension) == true
+    end)
+    |> Enum.map(&elem(&1, 0))
```
nshoes, line 374: I think we could save some cycles with this, thoughts?

```suggestion
    for {ext, enabled} <- device.extensions, !enabled and product.extensions[ext], into: [], do: ext
```

lib/nerves_hub_web/live/devices/show.html.heex
```diff
@@ -61,14 +61,19 @@
...
+      </div>
 
       <div class="display-box">
         <div class="help-text mb-1">Tags</div>
         <p :if={is_nil(@device.tags)}>No Tags</p>
         <span :for={tag <- @device.tags || []} class="badge"><%= tag %></span>
       </div>
 
-      <DeviceLocation.render location={@device.connection_metadata["location"]} />
+      <DeviceLocation.render :if={@product.extensions.geo && @device.extensions.geo != false} location={@device.connection_metadata["location"]} />
```
nshoes, line 76: Thoughts?

```suggestion
      <DeviceLocation.render :if={@product.extensions.geo and @device.extensions.geo} location={@device.connection_metadata["location"]} />
```

lib/nerves_hub_web/live/devices/show.html.heex
```diff
@@ -94,7 +99,7 @@
   </div>
 
   <div class="col-lg-6">
-    <div class="device-health-section">
+    <div :if={@product.extensions.health && @device.extensions.health != false} class="device-health-section">
```
nshoes, line 102: Could be handled the same as above.

```suggestion
    <div :if={@product.extensions.health and @device.extensions.health} class="device-health-section">
```

lib/nerves_hub_web/live/devices/show.ex
lawik: Unfortunately we need to model extension settings being unset or we would have to choose a more aggressive default (always enable, always disable). So they can't just be true/false, they can also be nil. This is in line with Jon's original approach although maybe not intuitive.

So `and` will not work unless we do explicit comparisons to true/false which handles the nil cases for us.

lib/nerves_hub_web/live/devices/show.html.heex
lawik: `&&` allows for `nil` values and `!=` handles the `nil` case.

We could make a larger change and have more aggressive defaults as mentioned above. Maybe that's better?

lib/nerves_hub_web/live/devices/show.html.heex
lawik: Same as above :)

lib/nerves_hub_web/live/devices/show.ex
nshoes: Ah, interesting okay. I updated my suggestion.

lib/nerves_hub_web/live/devices/show.html.heex
nshoes: What's the lift on that? When thinking of modeling extensions, especially when we have to compare two places they're set (product/device), I would much rather prefer defaults than having to always think of `nil`. I say let's do it.

lawik: Yeah, just this conversation kind of convinces me to separate the defaults between devices and products. Default off for products (maybe configurable) and default on per device.

Not that heavy. Can do that in this PR.

lawik:
@nshoes I updated it by fixing the defaults behavior and simply adding a setting for people that want to fully opt out of extensions.

I figure SmartRent might want the option of keeping it off until ready to actually try it. Preferable to control it per product but this allows a no-disruption update if you just set the env var.

config/runtime.exs
```diff
@@ -35,6 +35,7 @@ config :nerves_hub,
     String.to_integer(System.get_env("DEPLOYMENT_CALCULATOR_INTERVAL_SECONDS", "3600")),
   mapbox_access_token: System.get_env("MAPBOX_ACCESS_TOKEN"),
   dashboard_enabled: System.get_env("DASHBOARD_ENABLED", "false") == "true",
+  extensions_enabled: System.get_env("EXTENSIONS_ENABLED", "true") == "true",
```
nshoes, line 38: Why should we have this on by default, don't we usually treat new features like this as opt-in?

lib/nerves_hub/extensions/extensions_setting.ex
```diff
@@ -1,15 +1,61 @@
...
+    {val, updated}
+  end
+
+  def put(struct, key, val) do
+    if Map.has_key?(struct, key) do
+      Map.put(struct, key, val)
+    else
+      struct
+    end
+  end
```
nshoes, line 60: Do we need all of these? Is this all to remove the `Map.from_struct()` call?

config/runtime.exs
jjcarstens: I also think we don't even need this. Each extension is disabled by default until explicitly set on the product or device. So nothing will attach.

Generally, unknown messages should be ignored in NervesHubLink. If we're trying to avoid sending the message entirely for the really old NHL versions that could crash with unknown server message, then we could first look that any extension is enabled at all for the product/device before sending

lib/nerves_hub/extensions/extensions_setting.ex
```diff
@@ -1,15 +1,61 @@
 defmodule NervesHub.Extensions.ExtensionsSetting do
   use Ecto.Schema
   import Ecto.Changeset
+  @behaviour Access
```
jjcarstens, line 4:
```suggestion
  @derive [Access]
```
You might be able to just derive Access and avoid all the extra implementation if that is what you're aiming for

config/runtime.exs
lawik: NervesHubLink is fine.

So I added this when I changed the default for ExtensionSetting to `default: true`.

I changed that default to not have the confusing modelling of `true|false|nil` that @nshoes and I discussed.
Ideally I don't want a DevicesExtensionSetting (default: true) and a separate ProductExtensionSetting (default: false) but that would be a way to solve it that would work fine to make products default to off.

For NervesCloud we absolutely want new Products being set up to default to having extensions on. I realize SmartRent wants them off by default.

I could see switching to separate defaults for Device (device with default: false, would just disable extensions unless you switch them on for every device, not desirable) and Product fixing this. Then we could change this option to whether new products are created with extensions enabled or disabled and make that part of the product creation step. That would solve our needs.

Let me know what you think :)

lib/nerves_hub/extensions/extensions_setting.ex
lawik: I looked for whatever the heck the intended way is and could not find it :D

lib/nerves_hub/extensions/extensions_setting.ex
lawik: Will remove, I had completely failed to find `@derive`

lawik: And it was to allow `extensions[key]` instead of `Map.get(extensions, key)`

lib/nerves_hub/extensions/extensions_setting.ex
lawik: Ach, access is not a protocol, it is a behaviour.

lib/nerves_hub/extensions/extensions_setting.ex
lawik: Except `@derive` won't work. Because behaviour, not protocol. Trimming it down to just the `fetch`.

config/runtime.exs
lawik: Josh had a good suggestion which would address our needs and let us leave the product to default to disabled.

We can add the extensions form in the create product form so you choose at the time of creating the product to enable/disable whatever extensions you want.

If everyone likes that, I'm good to drop the config and split the default behavior of Device (default on) and Product (default off).

jjcarstens: Yes, that sounds great to have it in the form. For a brand new experimental feature, I think we all want disabled default. We're talking effort of a few minutes to turn on if needed.

lib/nerves_hub/extensions/extensions_setting.ex
jjcarstens: You know, that makes sense. I was going off this article which is 10 years old....clearly it has changed since then 🤣 https://dashbit.co/blog/writing-assertive-code-with-elixir

config/runtime.exs
lawik: Cool, will do it that way. Off by default on product. Easily available on product creation form.

lawik: That should do it.

lawik:
Also fixed some warnings.

nshoes (approved):
🚀

lawik:
Removed the global runtime/deploy-time config.

PR #1655, opened 2024-11-21
Improve metrics queries

elinol:
Remove separation of default and custom metrics in database queries and make UI handle that instead.

### DIFF
--- a/lib/nerves_hub/devices/metrics.ex
+++ b/lib/nerves_hub/devices/metrics.ex
@@ -107,25 +107,30 @@ defmodule NervesHub.Devices.Metrics do
   end
 
   @doc """
-  Get map with latest values for all metric types. Also includes timestamp.
+  Retrieves the latest metric set, together with the timestamp, for a given device.
+
+  Uses a subquery to filter the `DeviceMetric` table on the most recent `inserted_at` timestamp.
   """
   def get_latest_metric_set(device_id) do
     DeviceMetric
     |> where([dm], dm.device_id == ^device_id)
-    |> where([dm], dm.inserted_at == subquery(time_of_latest_insert(device_id)))
+    |> where(
+      [dm],
+      dm.inserted_at ==
+        subquery(
+          DeviceMetric
+          |> select([:inserted_at])
+          |> where(device_id: ^device_id)
+          |> order_by(desc: :inserted_at)
+          |> limit(1)
+        )
+    )
     |> Repo.all()
     |> Enum.reduce(%{}, fn item, acc ->
-      Map.put(acc, item.key, item.value)
+      acc
+      |> Map.put(item.key, item.value)
       |> Map.put_new("timestamp", item.inserted_at)
     end)
-  end
-
-  defp time_of_latest_insert(device_id) do
-    DeviceMetric
-    |> select([:inserted_at])
-    |> where(device_id: ^device_id)
-    |> order_by(desc: :inserted_at)
-    |> limit(1)
   end
 
   @doc """
--- a/lib/nerves_hub/devices/metrics.ex
+++ b/lib/nerves_hub/devices/metrics.ex
@@ -7,12 +7,15 @@ defmodule NervesHub.Devices.Metrics do
   @default_metrics [
     "cpu_temp",
     "cpu_usage_percent",
+    "disk_available_kb",
+    "disk_total_kb",
+    "disk_used_percentage",
     "load_15min",
     "load_1min",
     "load_5min",
-    "size_mb",
-    "used_mb",
-    "used_percent"
+    "mem_size_mb",
+    "mem_used_mb",
+    "mem_used_percent"
   ]
 
   def default_metrics, do: @default_metrics
--- a/lib/nerves_hub/devices/metrics.ex
+++ b/lib/nerves_hub/devices/metrics.ex
@@ -4,18 +4,18 @@ defmodule NervesHub.Devices.Metrics do
   alias NervesHub.Devices.DeviceMetric
   alias NervesHub.Repo
 
-  @default_metric_types [
-    :cpu_temp,
-    :cpu_usage_percent,
-    :load_15min,
-    :load_1min,
-    :load_5min,
-    :size_mb,
-    :used_mb,
-    :used_percent
+  @default_metrics [
+    "cpu_temp",
+    "cpu_usage_percent",
+    "load_15min",
+    "load_1min",
+    "load_5min",
+    "size_mb",
+    "used_mb",
+    "used_percent"
   ]
 
-  def default_metric_types, do: @default_metric_types
+  def default_metrics, do: @default_metrics
 
   @doc """
   Get all metrics for device
@@ -30,8 +30,8 @@ defmodule NervesHub.Devices.Metrics do
   @doc """
   Get metrics by device within a specified time frame
   """
-  def get_device_metrics(device_id, time_unit, amount) do
-    DeviceMetrics
+  def get_device_metrics(device_id, {time_unit, amount}) do
+    DeviceMetric
     |> where(device_id: ^device_id)
     |> where([d], d.inserted_at > ago(^amount, ^time_unit))
     |> order_by(asc: :inserted_at)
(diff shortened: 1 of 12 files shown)


lib/nerves_hub/devices/metrics.ex
```diff
@@ -135,31 +108,25 @@ defmodule NervesHub.Devices.Metrics do
...
+      Map.put(acc, item.key, item.value)
+      |> Map.put_new("timestamp", item.inserted_at)
     end)
-    |> Map.put(:timestamp, get_latest_timestamp_for_device(device_id))
   end
 
-  def get_latest_custom_metrics(device_id) do
-    default_metrics = Enum.map(@default_metric_types, &Atom.to_string/1)
-
+  defp time_of_latest_insert(device_id) do
```
nshoes, line 122: What do you think about placing this in `get_latest_metric_set`? I feel that it's not worth the extra cognitive step to parse that it's a function call and then read the function when it's a small query and could be in the same scope.

lib/nerves_hub_web/live/devices/device_health.ex
```diff
@@ -160,67 +164,57 @@ defmodule NervesHubWeb.Live.Devices.DeviceHealth do
...
 
-  defp organize_metrics_for_chart(metrics) do
+  defp filter_and_sort(metrics) do
+    metrics
+    |> Enum.reject(fn {type, _} -> type in @no_chart_metrics end)
+    |> Enum.sort_by(fn {type, _} ->
+      # Sorts list by @default_metrics order
+      Enum.find_index(@default_metrics, fn {type_, _} ->
+        type_ == type
+      end)
```
nshoes, line 194: This was a bit confusing for me, I think this naming reads easier. What do you think?

```suggestion
      Enum.find_index(@default_metrics, fn {default_type, _} ->
        type == type
      end)
```

lib/nerves_hub_web/live/devices/device_health.ex
```diff
@@ -237,25 +231,28 @@ defmodule NervesHubWeb.Live.Devices.DeviceHealth do
   defp get_time_unit({"day", 1}), do: "hour"
   defp get_time_unit({"day", _}), do: "day"
 
+  @max_100 ["used_percent", "cpu_temp"]
```
nshoes, line 234: Why pull this list out into an attribute? I might be missing something.

lib/nerves_hub_web/components/device_health/health_header.ex
```diff
@@ -5,7 +5,7 @@ defmodule NervesHubWeb.Components.HealthHeader do
   attr(:product, :any)
   attr(:device, :any)
   attr(:latest_connection, :any)
-  attr(:latest_health, :any)
+  attr(:health_reported_at, :any)
```
nshoes, line 8: I'm just curious why we type these props as `any`. That's usually a code smell for things like dialyzer, but I'm unsure if there's a reason (or maybe we really just don't care).

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -368,4 +367,14 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
   end
+
+  defp format_key(key) do
+    key
+    |> String.replace("_", " ")
+    |> String.capitalize()
+  end
+
+  defp default_metrics() do
+    Metrics.default_metrics()
```
nshoes, line 378: What are your thoughts on moving this call to the template? It looks like we don't have a rule or boundary of not calling functions off of modules in templates, and I feel this doesn't improve readability.

test/nerves_hub_web/live/devices/health_test.exs
```diff
@@ -41,24 +41,7 @@ defmodule NervesHubWeb.Devices.HealthTest do
...
-             Metrics.save_metric(%{device_id: device.id, key: "custom_1", value: 12})
-
-    assert {:ok, _} =
-             Metrics.save_metric(%{device_id: device.id, key: "custom_2", value: 13})
-
-    conn
-    |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}/health")
-    |> assert_has("canvas", count: 2)
+    # Six of the default metric types are shown as charts
+    |> assert_has("canvas", count: 6)
```
nshoes, line 45: Can we assert on anything else here? This feels like a great place to test a lot of code. What do you think?

nshoes (commented):
Lots of tiny nits and questions :)

lib/nerves_hub_web/components/device_health/health_header.ex
elinol: Honestly I've just followed previous implementations. Might be a good idea to use `:map` for all these schema structs. From what I understand only structs that are defined by `defstruct`can be declared as `attr` type.

lib/nerves_hub_web/live/devices/device_health.ex
elinol: No, I was missing the option to do `type when type in ["used_percent", "cpu_temp"]` .

lib/nerves_hub_web/live/devices/device_health.ex
elinol: Agree!

lib/nerves_hub/devices/metrics.ex
elinol: I found breaking it out made it more readable but it's not a strong opinion. I moved it and added a little explanation in the docs instead.

lib/nerves_hub_web/components/device_health/health_header.ex
elinol: timestamp types are not supported though, so I guess this particular `health_reported_at` will remain as `:any`

nshoes: Cool! While we're here, let's do it :)

lib/nerves_hub_web/live/devices/device_health.ex
elinol: Sorry for my bad english! I just meant that I didn't realize I could just define the list directly in the case condition instead of using an attribute.

test/nerves_hub_web/live/devices/health_test.exs
elinol: I've added some tests that covers more of the liveview functionality, and structuring of the data required by the JS hooks that generates the graphs.

lib/nerves_hub_web/components/device_health/health_header.ex
elinol: Done!

elinol:
@nshoes I think this is ready, what do you say? 😅

nshoes (approved):
LGTM! Test additions and fixups are great, 🚢 🚢 🚢

lib/nerves_hub_web/live/devices/show.html.heex
```diff
@@ -114,7 +114,7 @@
         </button>
       </div>
 
-      <%= if (!Enum.any?(Map.values(@latest_metrics)) and !Enum.any?(Map.values(@latest_custom_metrics))) and !@alarms do %>
+      <%= if (!Enum.any?(Map.values(@latest_metrics)) and !@alarms )do %>
```
joshk, line 117: I'd recommend moving this to a function to make it a little clearer and easier to read

lib/nerves_hub_web/live/devices/show.html.heex
```diff
@@ -135,69 +135,59 @@
...
               <div class="help-text label">Load avg</div>
-              <%= @latest_metrics.load_1min %> | <%= @latest_metrics.load_5min %> | <%= @latest_metrics.load_15min %>
+              <%= @latest_metrics["load_1min"] %> | <%= @latest_metrics["load_5min"] %> | <%= @latest_metrics["load_15min"] %>
             </div>
-
-            <%= if @latest_metrics.used_mb do %>
-              <div class={"callout #{Utils.memory_to_status(60)}"} x>
-                <div class="help-text label">Memory used</div>
-                <%= round(@latest_metrics.used_mb) %>MB (<%= round(@latest_metrics.used_percent) %>%)
+            <!--Show CPU and memory usage if reported. Higher values will be displayed in warningly colors. -->
```
joshk, line 143: This is an html comment, I think it would be better to switch this to a heex comment, eg. `<%# comment %>`

joshk (approved):
All in all this is great, I just have two small comments.

lib/nerves_hub_web/live/devices/show.html.heex
elinol: how do you like` <%= if (Enum.empty?(@latest_metrics) and is_nil(@alarms) )do %>`? I find it easier to read without the negations

nshoes: +1 for that Elin!

joshk: Sweet, thanks!

lib/nerves_hub_web/live/devices/show.html.heex
elinol: Changed to `<%!-- comment --%>` which seems to be the latest format for comments.

joshk (approved):
Great work!

PR #1656, opened 2024-11-21
Replace Scrivener with Flop

lawik:
This is pretty much a bug-by-bug compatible implementation as far as I can tell.


lawik:
Takes care of issue #1656 

Flop wants the struct so it is a bit messy. There is a lot of potential tidying up that one could do. But this just rips out the old and provides identical functionality.

lib/nerves_hub/devices.ex
```diff
@@ -70,10 +70,17 @@ defmodule NervesHub.Devices do
...
+    {entries, _pager} = get_devices_by_org_id_and_product_id_with_pager(org_id, product_id, opts)
+    entries
+  end
+
+  def get_devices_by_org_id_and_product_id_with_pager(org_id, product_id, opts) do
     pagination = Map.get(opts, :pagination, %{})
     sorting = Map.get(opts, :sort, {:asc, :identifier})
     filters = Map.get(opts, :filters, %{})
 
+    flop = %Flop{page: pagination[:page] || 1, page_size: pagination[:page_size] || 10}
```
nshoes, line 82:
```suggestion
    pagination = Map.get(opts, :pagination, %{page: 1, page_size: 10})
    sorting = Map.get(opts, :sort, {:asc, :identifier})
    filters = Map.get(opts, :filters, %{})

    flop = %Flop{page: pagination[:page], page_size: pagination[:page_size]}
```

lib/nerves_hub/devices.ex
```diff
@@ -86,21 +93,24 @@ defmodule NervesHub.Devices do
...
+    |> Flop.run(flop)
   end
 
+  @spec filter(integer(), map()) :: {list(Device.t()), Flop.Meta.t()}
   def filter(product_id, opts) do
     pagination = Map.get(opts, :pagination, %{})
     sorting = Map.get(opts, :sort, {:asc, :identifier})
     filters = Map.get(opts, :filters, %{})
 
+    flop = %Flop{page: pagination.page, page_size: pagination.page_size}
```
nshoes, line 105: Do we care that these could be `nil` here? We could adopt my suggestion from above if so.

lib/nerves_hub/devices.ex
```diff
@@ -70,10 +70,17 @@ defmodule NervesHub.Devices do
   end
 
   def get_devices_by_org_id_and_product_id(org_id, product_id, opts) do
+    {entries, _pager} = get_devices_by_org_id_and_product_id_with_pager(org_id, product_id, opts)
```
nshoes, line 73: I was originally off-put by `pager`, but its definition is "one who pages"... I love it 😆

lib/nerves_hub_web/views/layout_view.ex
```diff
@@ -83,40 +83,76 @@ defmodule NervesHubWeb.LayoutView do
...
+        end_range: end_range,
+        distance: distance,
+        anchor: anchor
+      })
+
+    ~H"""
+    <div class="btn-group btn-group-toggle">
+      <div :if={@page_number > 1}>
+        <%= link("&lt;&lt;",
+          to: "?page=#{@page_number - 1}#{@anchor}",
```
nshoes, line 102: Are we resolving an index starting at 1 vs 0 here? Is Flop zero-based?

nshoes (changes requested):
Thank you so much for doing this! Couple issues I found:

#### Cannot get back to page 1 after navigating to other pages
The double carets or clicking `1` doesn't send you back to the first page.

CleanShot 2024-11-21 at 07 43 28

#### Single page shows two page `1` icons
They both link to page 1, but it's confusing and should only display a single tile.
CleanShot 2024-11-21 at 07 46 05@2x

nshoes:
> But this just rips out the old and provides identical functionality.

I didn't see this before making my review, this makes me think that the bugs I listed already existed, but I'd love to take care of them here.

lawik:
@nshoes no worries, it is indeed bug-for-bug-compatible. I hope to take some time next week and make it not terrible. But I think replacing the dep is a good starting point and then improve from there.

lib/nerves_hub_web/views/layout_view.ex
lawik: Offset/Limit mode is zero-based, page-mode is not. Fairly certain :)

lib/nerves_hub/devices.ex
lawik: I didn't even think very hard. But of course intended this very fitting name :P

lib/nerves_hub/devices.ex
lawik: They are only used in one place and will never be nil so I don't think we care.

nshoes (approved):

PR #1663, opened 2024-11-26
Fix device pager not paging back to page 1

### LINKED ISSUE #1657: Devices list: paging back to page 1 does not work
There is something about handling the changes and handle_params where it doesn't accept that going back to the default page (1) is enough of a change to actually fetch new data.

lawik:
Closes #1657


joshk (approved):
🎉

PR #1664, opened 2024-11-26
Improve pager

lawik:
- Add visual gaps
- Make it not hop around when going between 1 and 2 by moving the arrow step buttons
- Make page 1 available even if you are paged far in


nshoes (approved):
Awesome!

PR #1668, opened 2024-12-02
Check device api version before requesting extensions

nshoes:
Related: https://github.com/nerves-hub/nerves_hub_link/pull/252

Older versions of `nerves_hub_link` blow up when receiving an unknown message. This PR ensures that we only request extensions when we're communicating with a device that can handle the message.

Tested locally with `nerves_hub_link`.


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -49,8 +49,11 @@ defmodule NervesHubWeb.DeviceChannel do
 
     send(self(), :device_registation)
 
-    # Request device extension capabilities
-    push(socket, "extensions:get", %{})
+    # Request device extension capabilities if possible
+    # Earlier versions of nerves_hub_link don't have a fallback for unknown messages,
+    # so check version before requesting extensions
+    if is_safe_to_request_extensions?(params["device_api_version"]),
+      do: push(socket, "extensions:get", %{})
```
jjcarstens, line 56:
```suggestion
    if is_safe_to_request_extensions?(socket.assigns.device_api_version),
      do: push(socket, "extensions:get", %{})
```
If we move this to after all the assigns, we can piggy back on the safe default that `assign_api_version` gives us

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -612,4 +615,6 @@ defmodule NervesHubWeb.DeviceChannel do
       0
     end
   end
+
+  defp is_safe_to_request_extensions?(version), do: Version.match?(version, ">= 2.5.2")
```
jjcarstens, line 619:
```suggestion
  defp is_safe_to_request_extensions?(version), do: Version.match?(version, "> 2.5.2")
```
`2.5.2` is the currently released NHL version, but the extensions work has not been released to hex. So we probably want that as non-inclusive greater-than to capture the next version which should have the extensions work

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Gosh dangit lol Thanks for the callout.

jjcarstens (approved):

jjcarstens:
Whoops. Missed the failing tests. Assuming you get those worked out, then this is 👍🏼

PR #1672, opened 2024-12-04
Address specs and code path issues found by dialyzer

joshk:
😘  Thank you dialyzer  😘


lib/nerves_hub/extensions/health.ex
```diff
@@ -26,19 +27,15 @@ defmodule NervesHub.Extensions.Health do
 
     send(self(), {__MODULE__, :check})
 
+    timer =
```
lawik, line 30: This change removes the check that the timer is actually > 0. Something more is needed :D

lawik (dismissed):

nshoes (approved):

Issue #1678, opened 2024-12-05
Pagination links are broken


elinol:
Pagination links are looking weird:

![image](https://github.com/user-attachments/assets/472614d4-c113-4e74-92d8-650736969fd4)

Also they are broken:
1. Click on link for page 2 -> second page is displayed and `?page_number=2` is added to URL params.
2. Click on link for page 1 -> Page number params are removed from URL but nothing happens in device list, page 2 still showing. Same thing when clicking arrows link.

joshk:
Closing this issue as it looks to be fixed.

PR #1679, opened 2024-12-05
Fix broken paginator

elinol:
Solves #1678

- Reverts handling of pagination params in devices list that was broken due to a faulty rebase.
- Remove pagination button that displays total pages
- Only display pagination links if total pages is more than 0.


joshk (approved):
Awesome

Thank you.

PR #1680, opened 2024-12-05
Fix slideover causing horizontal scroll

lawik:


lawik:
I will try to prefix all the rework ui with `ui2-` as there will probably be a bunch of them and ideally on the smaller side.

PR #1685, opened 2024-12-06
Include New UI asset building in Dockerfile

jjcarstens:
The new UI uses Tailwind and ESbuild via the `assets.deploy` task which needs to be run when building the release.

This adjusts the Dockerfile a bit to include the previous Node assets during the build phase for `mix assets.deploy` to also use


joshk (approved):
I'd ashamed I didn't listen to Jons advice from the start.

What took me a day took Jon an hour.

PR #1688, opened 2024-12-09
Some minor JS touchups

joshk:
- Removed unused JS args in the two `app.js` files and `console.js`
- Added an ESLint hint to not warn on an empty catch (try-catch)
- And some minor indenting

### DIFF
--- a/assets/js/console.js
+++ b/assets/js/console.js
@@ -254,7 +254,7 @@ dropzone.addEventListener(
   e => {
     e.preventDefault()
     e.stopPropagation()
-    ;[...e.dataTransfer.items].forEach(item => {
+    e.dataTransfer.items.forEach(item => {
       const file = item.getAsFile()
       const reader = file.stream().getReader()
 
--- a/assets/js/console.js
+++ b/assets/js/console.js
@@ -1,4 +1,5 @@
 /* eslint no-console: ["error", { allow: ["log"] }] */
+/* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
 
 import { Socket } from 'phoenix'
 import { Terminal } from 'xterm'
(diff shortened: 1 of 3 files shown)


assets/js/console.js
```diff
@@ -253,8 +254,7 @@ dropzone.addEventListener(
   e => {
     e.preventDefault()
     e.stopPropagation()
-
-    ;[...e.dataTransfer.items].forEach((item, i) => {
+    ;[...e.dataTransfer.items].forEach(item => {
```
nshoes, line 257: This pains me 😅 There's no need to spread here, so we can just:
```suggestion
    e.dataTransfer.items.forEach(item => {
```

assets/js/console.js
```diff
@@ -1,4 +1,5 @@
 /* eslint no-console: ["error", { allow: ["log"] }] */
+/* eslint no-empty: ["error", { "allowEmptyCatch": true }] */
```
nshoes, line 2: I think this is a good opportunity to start an [ESLint config file](https://eslint.org/docs/latest/use/configure/rules#using-configuration-files). I'd hate to start having these comments at the top of every file and we should enforce rules across files uniformly.

joshk: I'm cool with us adding an ESLint config, could you do this in another PR?

nshoes: I'm happy to oblige, but I thought it made sense to do it here since we're already touching ESLint things. What are your reasons against it?

joshk: Oh, more that I don't know how to do it :D
Could you add to this PR?

nshoes: Oh! Hehe :) Yeah, I'll add to this PR 🫡

nshoes: Nope, I actually won't 😆 It's [a bit more involved](https://github.com/nerves-hub/nerves_hub_web/pull/1690) then just adding the config file like I had hoped so we'll leave it for another PR.

nshoes (approved):

Issue #1681, opened 2024-12-06
Add UI for script runs


elinol:
Currently the scripts can only be executed via API, let's add UI support for this.

Add a section for it on Show Device page.

PR #1689, opened 2024-12-09
Add script section to device UI

elinol:
Solves #1681 

Adds a simple section on device page for listing and running support scripts.

Not too much time spent on the looks since the whole UI is redesigned.

### DIFF
--- a/lib/nerves_hub_web/helpers/authorization.ex
+++ b/lib/nerves_hub_web/helpers/authorization.ex
@@ -58,6 +58,7 @@ defmodule NervesHub.Helpers.Authorization do
   def authorized?(:"support_script:create", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:update", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:delete", %OrgUser{role: role}), do: role_check(:manage, role)
+  def authorized?(:"support_script:run", %OrgUser{role: role}), do: role_check(:manage, role)
 
   defp role_check(required_role, user_role) do
     required_role
--- a/lib/nerves_hub_web/helpers/authorization.ex
+++ b/lib/nerves_hub_web/helpers/authorization.ex
@@ -58,7 +58,7 @@ defmodule NervesHub.Helpers.Authorization do
   def authorized?(:"support_script:create", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:update", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:delete", %OrgUser{role: role}), do: role_check(:manage, role)
-  def authorized?(:"support_script:run", %OrgUser{role: role}), do: role_check(:manage, role)
+  def authorized?(:"support_script:run", %OrgUser{role: role}), do: role_check(:view, role)
 
   defp role_check(required_role, user_role) do
     required_role
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -21,7 +21,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
 
   alias Phoenix.Socket.Broadcast
 
-  @default_script_output %{output: nil, id: 0}
+  @running_script_placeholder "Running Script.."
 
   def mount(%{"device_identifier" => device_identifier}, _session, socket) do
     %{org: org, product: product} = socket.assigns
@@ -47,8 +47,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> assign(:alarms, Alarms.get_current_alarms_for_device(device))
     |> assign(:extension_overrides, extension_overrides(device, product))
     |> assign(:latest_metrics, Metrics.get_latest_metric_set(device.id))
-    |> assign(:scripts, Scripts.all_by_product(product))
-    |> assign(:script_output, @default_script_output)
+    |> assign(:scripts, scripts_with_output(product))
     |> assign_metadata()
     |> schedule_health_check_timer()
     |> assign(:fwup_progress, nil)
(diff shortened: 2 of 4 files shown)


test/nerves_hub_web/live/devices/show_test.exs
```diff
@@ -392,6 +392,35 @@ defmodule NervesHubWeb.Live.Devices.ShowTest do
...
+    } do
+      {:ok, _command} =
+        NervesHub.Scripts.create(product, %{name: "MOTD", text: "NervesMOTD.print()"})
+
+      conn
+      |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}")
+      |> assert_has("h3", text: "Support Scripts")
+      |> assert_has("div", text: "MOTD")
+      |> assert_has("button", text: "Run")
+    end
```
elinol, line 421: I wanted to add a test for clicking Run button, but it seems too involved with device channel and such. Couldn't find any tests for running tests at all, might be better to start there?

elinol:
image

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -316,6 +321,27 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
+        %{"script_id" => script_id},
+        %{assigns: %{device: device, scripts: scripts}} = socket
+      ) do
+    script_id = String.to_integer(script_id)
+    script = Enum.find(scripts, &(&1.id == script_id))
+
+    output = Scripts.Runner.send(device, script)
+
+    socket
+    |> assign(:script_output, %{output: output, id: script_id})
```
nshoes, line 335: This means we can only see a single script's output at a time. If we care about seeing multiple, we could leverage the `order_by(:name)` from `Scripts.all_by_product/1` and build a data structure that's easier to depend on. What are your thoughts?

joshk: there is also another way to do this, instead of using an `assign`, you can use `push_event` and have a js function which updates the UI. This means that you can run two scripts and the second script run won't remove the result from the first script run. I would recommend `push_event` over `assign`

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -316,6 +321,27 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
 
+  def handle_event(
+        "run-script",
+        %{"script_id" => script_id},
+        %{assigns: %{device: device, scripts: scripts}} = socket
+      ) do
+    script_id = String.to_integer(script_id)
+    script = Enum.find(scripts, &(&1.id == script_id))
+
+    output = Scripts.Runner.send(device, script)
```
joshk, line 332: I believe this is a blocking function, which means it would block the LiveView from doing anything while its processing. I think you could use https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#start_async/4 to run the command, and then display a flash to say that the script is running.

Then in `def handle_async(:run_script, {:ok, result}, socket) do` you can update the flash and use `push_event` to update the UI.

joshk: With this change you could also remove lines 24, 50, and 51.

joshk (changes requested):
I really love the direction of this. This is a great first implementation.

Similar to how the API allows you to do run support scripts, you will need to check if the user has the required role (https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/controllers/api/script_controller.ex#L36)

In our liveviews we do this with: `authorized!(:"support_scripts:run", socket.assigns.org_user)` and then adding the definition to https://github.com/nerves-hub/nerves_hub_web/blob/d7215609eb3c8c5584168d0324bb0d0efa5a269d/lib/nerves_hub_web/helpers/authorization.ex#L5

lib/nerves_hub_web/live/devices/show.ex
elinol: @joshk I'm not that found of involving js stuff but I'll give this a go and who knows, I might be convinced otherwise

elinol: oh, do you mean like `JS.hide()` or so? I'm ok with that :D

elinol:
image

lib/nerves_hub_web/live/devices/show.ex
elinol: Why exactly do you recommend `push_event` over assigns @joshk ?

I found it simpler to use the assigns together with `start_async`

lib/nerves_hub_web/helpers/authorization.ex
```diff
@@ -58,6 +58,7 @@ defmodule NervesHub.Helpers.Authorization do
   def authorized?(:"support_script:create", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:update", %OrgUser{role: role}), do: role_check(:manage, role)
   def authorized?(:"support_script:delete", %OrgUser{role: role}), do: role_check(:manage, role)
+  def authorized?(:"support_script:run", %OrgUser{role: role}), do: role_check(:manage, role)
```
joshk, line 61: Looking at the permissions required for `API.ScriptController.send`, you can use `:view` for `role_check`

test/nerves_hub_web/live/devices/show_test.exs
joshk: Agree, this is fine for now.

joshk (changes requested):
One tiny change requested, then we can merge!

lib/nerves_hub_web/live/devices/show.ex
joshk: I think your approach is fine for shipping.

lib/nerves_hub_web/helpers/authorization.ex
elinol: oops missed that. fixed now!

joshk (approved):

PR #1690, opened 2024-12-09
ESLint upgrade and config

nshoes:
This PR updates ESLint and moves our configuration into the 21st century.

- Use ESModules instead of CommonJS
- Use modern ESLint configuration file
- Prefer double quotes to single for Prettier and ESLint

I sanity checked both the old and new UI.


assets/package.json
```diff
@@ -27,7 +28,8 @@
     "xterm": "^5.3.0",
     "xterm-addon-search": "^0.13.0",
     "xterm-addon-web-links": "^0.9.0",
-    "xterm-addon-webgl": "^0.16.0"
+    "xterm-addon-webgl": "^0.16.0",
+    "yarn": "^1.22.22"
```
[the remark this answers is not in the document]
nshoes: No, this is cruft and I have no idea how exactly it got added 😆 I'll remove. I still need to make adjustments to the config in regards to node version and how `package.json` is setup.

assets/package.json
```diff
@@ -24,23 +25,25 @@
     "phoenix_html": "file:../deps/phoenix_html",
     "phoenix_live_view": "file:../deps/phoenix_live_view",
     "popper.js": "^1.16.1",
+    "semver": "^7.3.0",
```
nshoes, line 28: `console.js` will blow up without this explicitly stated as a dependency.

assets/vendor/topbar.js
```diff
@@ -1,165 +1,145 @@
...
-        hide: function() {
-          clearTimeout(delayTimerId)
-          delayTimerId = null
-          if (!showing) return
-          showing = false
-          if (progressTimerId != null) {
-            window.cancelAnimationFrame(progressTimerId)
-            progressTimerId = null
-          }
+// based on https://buunguyen.github.io/topbar
```
nshoes, line 1: No behavior was touched here, just converted to an ESM.

joshk: how did you convert it? if we are to upgrade the topbar version in the future, is there something we will need to do to convert it to an ESM?

nshoes: I just unwrapped the IIFEE's by hand 😬 If we are serious about versioning this, we need to consume it from NPM and not consume it with a copypasta from GitHub.

joshk: ok, we should switch to using it from NPM then (https://www.npmjs.com/package/topbar)

joshk (approved):
❤️

PR #1693, opened 2024-12-11
Update Elixir and Erlang versions

joshk:
Elixir 1.18.0
Erlang 27.1.3


nshoes (approved):

nshoes:
Could we run the new `mix format --migrate` here, too?

joshk:
> Could we run the new `mix format --migrate` here, too?

I might do that in a separate PR so we can have this change just focused on the Elixir version

Issue #1662, opened 2024-11-22
Empty metrics report gets treated as an error


elinol:
When a device is reporting health with an empty metrics map, it gets treated as an error and triggers `Logger.warning("Failed to save metrics")`.

See matching on result of `Repo.insert_all` in `Metrics.save_metrics`:

```
case results do
      {0, _} -> :error
      {count, _} -> {:ok, count}
 end
```

PR #1694, opened 2024-12-11
Ignore empty metrics map in health report

elinol:
Solves #1662.

Also, the deprecated broadcast topic for health checks is replaced with correct one.

### DIFF
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -126,9 +126,12 @@ defmodule NervesHubWeb.Live.Devices.Show do
   def handle_info(:check_health_interval, socket) do
     timer_ref = Process.send_after(self(), :check_health_interval, 65_000)
 
-    socket.endpoint.broadcast("device:#{socket.assigns.device.id}", "check_health", %{})
+    topic = "device:#{socket.assigns.device.id}:extensions"
+    NervesHubWeb.DeviceEndpoint.broadcast(topic, "health:check", %{})
 
-    {:noreply, assign(socket, :health_check_timer, timer_ref)}
+    socket
+    |> assign(:health_check_timer, timer_ref)
+    |> noreply()
   end
 
   def handle_info(%Broadcast{event: "location:updated"}, socket) do
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -127,7 +127,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
     timer_ref = Process.send_after(self(), :check_health_interval, 65_000)
 
     topic = "device:#{socket.assigns.device.id}:extensions"
-    NervesHubWeb.DeviceEndpoint.broadcast(topic, "health:check", %{})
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -126,9 +126,12 @@ defmodule NervesHubWeb.Live.Devices.Show do
   def handle_info(:check_health_interval, socket) do
     timer_ref = Process.send_after(self(), :check_health_interval, 65_000)
 
-    socket.endpoint.broadcast("device:#{socket.assigns.device.id}", "check_health", %{})
+    topic = "device:#{socket.assigns.device.id}:extensions"
+    NervesHubWeb.DeviceEndpoint.broadcast(topic, "health:check", %{})
```
joshk, line 130: Why did you change these lines?

We need to use `socket.endpoint.broadcast` as the web app will fail if `DeviceEndpoint` isn't started (it isn't started in production, they are only started for there respective apps)

joshk (changes requested):

lib/nerves_hub_web/live/devices/show.ex
elinol: It had to be changed to include "extensions", and I just copied how it was done in other files in the app. Is it wise to change to `socket.endpoint.broadcast` in all those files then?

joshk: oh, which other files does this exist?

elinol: In `NervesHub.Devices`, `NervesHub.Products`, `NervesHub.Extensions.Health` etc. In fact it seems like `DeviceEndpoint` is used all over

joshk: It looks like `NervesHubWeb.DeviceEndpoint` was added several times when the extensions work was completed. This has the chance of failing. Changing these in context modules is tricky, as we don't have a socket to use, but I also see some that can be removed. I think these extra ones found will need to be a different PR. If you can update your PR then we can merge it in.

elinol: Cool, I'll try that as soon I get my little device to connect

elinol: PubSub didn't work, it doesn't reach the device. what about `socket.endpoint.broadcast(topic, "health:check", %{})`? Or it that equal to `NervesHubWeb.DeviceEndpoint.broadcast(topic, "health:check", %{})` ?

joshk: Using `socket.endpoint.broadcast(topic, "health:check", %{})` is fine within LiveView controllers, but it won't work in Contexts (eg. `NervesHub.Devices` because we would need to pass the socket in.

For `broadcasts` within Contexts we should use `NervesHub.PubSub.broadcast`, if we can. It would look something like `NervesHub.PubSub.broadcast(NervesHub.PubSub, topic, event, message)`. Happy to pair on this next week.

elinol: Yes, let's!

joshk (approved):

PR #1701, opened 2024-12-16
Log unsupported API version when requesting device extensions

elinol:


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -62,8 +63,15 @@ defmodule NervesHubWeb.DeviceChannel do
...
     # so check version before requesting extensions
-    if is_safe_to_request_extensions?(socket.assigns.device_api_version),
-      do: push(socket, "extensions:get", %{})
+    if is_safe_to_request_extensions?(socket.assigns.device_api_version) do
+      push(socket, "extensions:get", %{})
+    else
+      description =
+        "device #{device.identifier} could not get extensions: Unsupported API version."
+
+      AuditLogs.audit!(device, device, description)
```
joshk, line 72: Could you please move this logic into https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/audit_logs/templates.ex

joshk (changes requested):
Nice work

One small change then we can merge this in

lib/nerves_hub_web/channels/device_channel.ex
elinol: Well I can, but i figured that would require the log message (description) to be defined in two places. Is it because the template might be used in other cases? Is the goal to put all audit-logging in this templates module? Just curious here :D

joshk: You can put the log message in the audit template too, even though thats breaking the boundaries a little.

Yes, the goal is to move all the audit logging to the templates module. It makes it easier to search for what audit logging is being used and cleans up some of the duplicated code.

joshk (approved):

PR #1703, opened 2024-12-17
Encapsulate health check requests into `Extensions.Health`

joshk:


lawik (approved):

lib/nerves_hub/extensions/health.ex
```diff
@@ -95,8 +95,13 @@ defmodule NervesHub.Extensions.Health do
...
 
+  def request_health_check(device) do
+    device_internal_broadcast!(device, "health:check", %{})
+  end
+
   defp device_internal_broadcast!(device, event, payload) do
     topic = "device:#{device.identifier}:extensions"
-    NervesHubWeb.DeviceEndpoint.broadcast_from!(self(), topic, event, payload)
+
+    Phoenix.Channel.Server.broadcast_from(NervesHub.PubSub, self(), topic, event, payload)
```
nshoes, line 105: Are there any behavior differences between this and `NervesHubWeb.DeviceEndpoint.broadcast_from!/4`?

joshk: Nope, `Endpoint`s use `Phoenix.Channel.Server` under the hood. The real issue is that we have two Endpoints, and if you use `DeviceEndpoint` within the web UI (which doesn't start the `DeviceEndpoint`) then it causes an exception.

nshoes: Ah, and this publishes to all nodes so the task running the devices endpoint gets it. Got it, thanks!

joshk: Kinda sorta. The Endpoint doesn't matter on the receiving end; `Endpoint.broadcast` is just a convenience method, really. It's just that the resolving of the `PubSub` to use, which is defined in the config, isn't setup if the Endpoint isn't started. By using `Phoenix.Channel.Server.broadcast` and passing in `NervesHub.PubSub` (the PubSub used in each of the Endpoint configs) we are working around this problem.

PR #1704, opened 2024-12-17
Oban creator recommends using PG notifier where possible

joshk:


lawik:
Do we use Oban for anything high-volume that runs frequently? Just thinking if this would have a high or negligible impact on DB utilization :)

nshoes:
@lawik high-volume _and_ runs frequently? I don't think so. The only one we'd have to worry about now is `NervesHub.Workers.CleanDeviceConnectionStates` which is run every minute, but I think it's fine.

nshoes (approved):

joshk:
If anything, this reduces the impact on the DB. 
This will fix an issue I have seen in the past where Oban has died on a node, and that node was the leader, so jobs weren't being scheduled correctly. From Parkers previous Elixir slack chats, the PG notifier is better at finding a new leader.

Fwiw, this is why I created the PR:

Screenshot 2024-12-19 at 10 36 46 AM

nshoes:
Straight from the source 💯

PR #1709, opened 2024-12-18
Move all audit logging to templates

elinol:
Moves all audit logging to the template module.

One exception is in the function `Devices.update_device_with_audit/4` where the audit log description is passed via the updating action (like `tag_device`for example). I don't know if it's worth it adding this to templates, since the templates themselves (the description) are defined outside the templates module. Any input on this is welcomed!

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -8,7 +8,7 @@ defmodule NervesHub.Devices do
   alias NervesHub.Accounts.OrgKey
   alias NervesHub.Accounts.User
   alias NervesHub.AuditLogs
-  alias NervesHub.AuditLogs.Templates
+  alias NervesHub.AuditLogs.DeviceTemplates
   alias NervesHub.Certificate
   alias NervesHub.Deployments.Deployment
   alias NervesHub.Deployments.Orchestrator
@@ -682,7 +682,7 @@ defmodule NervesHub.Devices do
   end
 
   def update_firmware_metadata(device, metadata) do
-    Templates.audit_firmware_metadata_updated(device)
+    DeviceTemplates.audit_firmware_metadata_updated(device)
     update_device(device, %{firmware_metadata: metadata})
   end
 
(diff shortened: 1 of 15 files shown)


lib/nerves_hub/devices.ex
```diff
@@ -1065,24 +1053,15 @@ defmodule NervesHub.Devices do
...
-    |> Multi.run(:audit_device, fn _, _ ->
-      AuditLogs.audit(user, device, description)
-    end)
-    |> Multi.run(:audit_target, fn _, _ ->
-      AuditLogs.audit(user, product, description)
-    end)
-    |> Multi.run(:audit_source, fn _, _ ->
-      AuditLogs.audit(user, source_product, description)
+    |> Multi.run(:audit, fn _, _ ->
+      Templates.audit_device_moved(user, device, product, source_product)
```
elinol, line 1064: I'm not sure this approach is the best, it gets kind of nested with a `Repo.transaction` inside `Templates.audit_device_moved`, but at least the audits are moved to a centralised place.

lib/nerves_hub/devices.ex
```diff
@@ -1091,6 +1070,7 @@ defmodule NervesHub.Devices do
         {:ok, updated}
 
       err ->
+        Logger.warning("Could not move device: #{inspect(err)}")
```
elinol, line 1073: will this be logged to sentry as well? @joshk

lib/nerves_hub/devices.ex
nshoes: Because all of this logic is encapsulated in `Device.move/3` I'd argue that keeping the original implementation is fine. Centralizing logic is important when it's reused, but a one-off call, especially when all three audit logs use the same description, feels like an appropriate exception.

nshoes (approved):
This is great, thank you!

joshk (approved):
LOVE LOVE LOVE THIS PR

THE CONSISTENCY IS FANTASTIC, THE ENCAPSULATION, THE CLEANING

Thank you!

PR #1712, opened 2024-12-18
Remove Device From Deployment While Viewing It

nshoes:


nshoes:
I kept the styling barebones so we can flesh out what it should like here (in review):
CleanShot 2024-12-18 at 15 54 03@2x

Thoughts? I think it needs a callout, whether that's color + icon or one of the other. I think placement is fine unless we want to reorganize the device deployment section.

joshk:
Maybe place it below the deployment and in red?

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -348,6 +348,16 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> noreply()
   end
 
+  def handle_event(
+        "remove-from-deployment",
+        _,
+        %{assigns: %{device: device}} = socket
+      ) do
+    socket
+    |> assign(:device, Devices.clear_deployment(device))
```
joshk, line 357: on line 45 we also have `|> assign(:deployment, device.deployment)` which would need to be set to nil.

joshk (changes requested):
Love it!

Could you also add a test for this?

lib/nerves_hub_web/live/devices/show.ex
nshoes: Done!

nshoes:
Tests have been written!

CleanShot 2024-12-19 at 09 43 39

joshk (approved):
Amazing work!

PR #1713, opened 2024-12-20
Use the `JSON` module instead of the `Jason` lib

joshk:
This is a bit of an experiment.

I've removed `Jason` from `mix.exs` but it is still being included by PhoenixTest and Esbuild.


lib/nerves_hub/sentry_json_wrapper.ex
```diff
@@ -0,0 +1,20 @@
...
+      {:ok, JSON.encode!(data)}
+    rescue
+      e ->
+        {:error, e}
+    end
+  end
+
+  def decode(data) do
+    JSON.decode(data)
+  end
```
nshoes, line 19: Tiny nit :) 

```suggestion
  defdelegate decode(data), to: JSON
```

mix.exs
```diff
@@ -40,7 +40,6 @@ defmodule NervesHub.MixProject do
       extra_applications: [
         :base62,
         :inets,
-        :jason,
```
nshoes, line 43: Does esbuild still function properly without this?

nshoes (approved):
LGTM!

mix.exs
joshk: I can't find anything that says that `:jason` needs to be added to `extra_applications`.

PR #1714, opened 2024-12-20
Replace endpoint broadcasts

elinol:
See #1703.

Removes all broadcasting to DeviceEndpoint.

### DIFF
--- a/lib/nerves_hub/extensions.ex
+++ b/lib/nerves_hub/extensions.ex
@@ -64,4 +64,23 @@ defmodule NervesHub.Extensions do
   def module(_key, _ver) do
     :unsupported
   end
+
+  def toggle_extension(device, event, extension) do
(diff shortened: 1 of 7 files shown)


elinol:
@joshk here's a recreated PR. don't ask why there's two commit on updating a function argument name...

lib/nerves_hub/extensions.ex
```diff
@@ -64,4 +64,23 @@ defmodule NervesHub.Extensions do
   def module(_key, _ver) do
     :unsupported
   end
+
+  def toggle_extension(target, event, extension) do
```
joshk, line 68: I'd recommend renaming this function to `broadcast_extension_event` as the current name makes it sound like the extension is being enabled/disabled.

joshk (approved):
I'm going to merge in this PR, but there is one suggestion I think we should address in a different PR.

PR #1715, opened 2024-12-23
Fixes related to dialyzer warnings

joshk:


lib/nerves_hub_web/live/dashboard/index.ex
```diff
@@ -145,14 +145,16 @@ defmodule NervesHubWeb.Live.Dashboard.Index do
...
-
+    maybe_cancel_timer(socket.assigns.next_timer)
     timer = Process.send_after(self(), :refresh_device_list, delay)
     assign(socket, :next_timer, timer)
   end
 
+  def maybe_cancel_timer(timer) do
+    _ = if timer, do: Process.cancel_timer(timer)
+    :ok
+  end
```
nshoes, line 156: What was the warning that caused this change? It feels harder to understand now.

.dialyzer_ignore.exs
```diff
@@ -1,4 +1,7 @@
 [
   {"lib/x509/certificate.ex", "Unknown type: X509.ASN1.record/1."},
-  {"lib/x509/certificate/extension.ex", "Unknown type: X509.ASN1.record/1."}
+  {"lib/x509/certificate/extension.ex", "Unknown type: X509.ASN1.record/1."},
+  {"lib/nerves_hub/deployments/orchestrator.ex", :unmatched_return, 1},
+  {"lib/nerves_hub_web/channels/device_channel.ex", :unmatched_return, 1},
+  {"lib/nerves_hub_web/channels/device_socket.ex", :unmatched_return, 1}
```
nshoes, line 6: Why are we ignoring these? Mismatched dialyzer specs in deps?

joshk: These warnings are coming from the use of `open_telemetry_decorator`. I believe its fine to ignore these for now.

lib/nerves_hub_web/live/dashboard/index.ex
joshk: I think its an 'unmatched return'. `Process.cancel_timer` has multiple return types and we aren't checking them, and dialyzer doesn't like that.

nshoes: Ok, that makes sense. In the `refresh_after/2`, what do you think of:

```
_ = if socket.assigns.next_timer, do: Process.cancel_timer(socket.assigns.next_timer)
```

And you could pull the timer out into a var if you think it's too unreadable (I don't mind it).

joshk: my code is a bit verbose, so if you prefer your suggestion you are welcome to add it to the PR

nshoes (approved):

PR #1718, opened 2024-12-27
Bulk move devices to deployment

nshoes:
This is a quality-of-life improvement since we're moving away from devices automatically finding deployments.

CleanShot 2024-12-27 at 11 12 48

### Todos
- [x] Tests


joshk (changes requested):
As discussed with Nate on a Zoom, we need to check if the device(s) meet the requirements of the Deployment before we assign it to them

nshoes:
With info flash:

CleanShot 2025-01-02 at 11 54 24

lib/nerves_hub/devices.ex
```diff
@@ -1156,6 +1156,21 @@ defmodule NervesHub.Devices do
     update_device_with_audit(device, params, user, description)
   end
 
+  @spec move_many_to_deployment([integer()], integer()) :: {integer(), nil}
```
joshk, line 1159: I think it would be good to change the return values to be specific to the task and not Ecto. eg.
`{:ok, {updated: 5, ignored: 0}}`?

lib/nerves_hub/devices.ex
```diff
@@ -1156,6 +1156,21 @@ defmodule NervesHub.Devices do
...
+  def move_many_to_deployment(device_ids, deployment_id) do
+    firmware =
+      Firmware
+      |> join(:inner, [f], d in Deployment, on: d.id == ^deployment_id)
+      |> limit(1)
+      |> Repo.one()
```
joshk, line 1165: My brain is having trouble reading this. Are you fetching the Firmware attached to the Deployment? I think it would read better to fetch the deployment and preload the firmware. We can also move that method to the `Deployments` context (there should we one we can use there)

lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -289,6 +303,60 @@ defmodule NervesHubWeb.Live.Devices.Index do
...
+            )
+
+          [updated_count, not_updated_count] ->
+            put_flash(
+              socket,
+              :info,
+              "#{updated_count} #{maybe_pluralize.(updated_count, "device")} added to deployment #{target_deployment.name}. #{not_updated_count} #{maybe_pluralize.(not_updated_count, "device")} could not be added to deployment because of mismatched firmware"
+            )
+        end
+      end)
```
joshk, line 355: Could you move this to its own method? eg. `update_flash(socket, devices_updated_count, devices_not_updated_count)`

joshk (changes requested):
Really awesome, some minor suggested changes.

lib/nerves_hub/devices.ex
nshoes: I swapped back and forth between what I have now and that, will gladly change.

lib/nerves_hub/devices.ex
nshoes: I wanted to get the firmware from just the deployment ID. I can see how this could be confusing, I'll change.

lib/nerves_hub_web/live/devices/index.ex
nshoes: https://github.com/nerves-hub/nerves_hub_web/pull/1718/commits/eb7ff2424ac9cc638ca804d8e2be0bdcef5d8d8c

lib/nerves_hub/devices.ex
```diff
@@ -1156,6 +1156,27 @@ defmodule NervesHub.Devices do
...
+  @spec move_many_to_deployment([integer()], integer()) ::
+          {:ok, %{updated: non_neg_integer(), ignored: non_neg_integer()}}
+  def move_many_to_deployment(device_ids, deployment_id) do
+    %{firmware: firmware} =
+      Deployment |> where(id: ^deployment_id) |> preload(:firmware) |> Repo.one()
+
+    Firmware
+    |> join(:inner, [f], d in Deployment, on: d.id == ^deployment_id)
+    |> limit(1)
+    |> Repo.one()
```
joshk, line 1168: Did you mean to leave this here?

joshk (approved):
Looking great. One small comment regarding some dead code.

lib/nerves_hub/devices.ex
nshoes: Nope! Removed, thank you.

PR #1719, opened 2024-12-27
Make sure the latest connection is loaded

joshk:
Fixes a UI issue where the connection status and information is removed after a device heartbeat occurs


nshoes (approved):
Great find and fix! 🥇

PR #1721, opened 2024-12-28
Fetch the signing salt for the web socket connections from the config

joshk:
This approach is taken from https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-connect-info

```
Additionally, session_config may be a MFA, such as {MyAppWeb.Auth, :get_session_config, []}, 
to allow loading config in runtime.
```

This allows for the signing salt to not be published and shared between all running NervesHub installs.


nshoes (approved):
🚀

PR #1726, opened 2025-01-02
Set deployment when viewing device

nshoes:
This work adds the ability to set a device's deployment when viewing it. If no eligible deployments are found, the UI is hidden.


nshoes:
CleanShot 2025-01-02 at 14 27 18

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -256,6 +258,28 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
+  def handle_event(
+        "set-deployment",
+        %{"deployment_id" => deployment_id},
+        %{assigns: %{user: user, device: device, eligible_deployments: eligible_deployments}} =
+          socket
+      ) do
+    deployment = Enum.find(eligible_deployments, &(&1.id == String.to_integer(deployment_id)))
+    device = Devices.update_deployment(device, deployment)
+
+    AuditLogs.audit!(
```
joshk, line 270: Could you please move this to `NervesHub.AuditLogs.Templates`

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -256,6 +258,28 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
+      user,
+      device,
+      "#{user.name} set #{device.identifier}'s deployment to #{deployment.name}"
+    )
+
+    socket
+    |> assign(:device, device)
+    |> assign(:deployment, deployment)
+    |> put_flash(:info, "Deployment successfully set")
```
joshk, line 279: Could you please change this to `Deployment successfully updated`

joshk: Also, when a deployment is removed from a device, could you please update the flash to say `Device successfully removed from the deployment`. (I don't love this message, when we move to toasts I think it will work better)

test/nerves_hub_web/live/devices/show_test.exs
```diff
@@ -437,7 +437,8 @@ defmodule NervesHubWeb.Live.Devices.ShowTest do
       conn
       |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}")
       |> click_button("Remove From Deployment")
-      |> assert_has("span", text: "No Assigned Deployment")
```
joshk, line 440: why did you remove this assert?

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -1,4 +1,5 @@
 defmodule NervesHubWeb.Live.Devices.Show do
+  alias NervesHub.Deployments
```
joshk, line 2: minor style nit pick, could you please move the `alias` to after the `use`, along with a newline between them.

joshk (changes requested):
Awesome addition!

Some minor comments. 

For new devices without a known arch or platform, do you think we should show a message about how these are needed before a device can be added to a deployment?

test/nerves_hub_web/live/devices/show_test.exs
nshoes: This text now only displays when there is no assigned deployment _and_ no eligible deployments. In this test case, we remove the device from the deployment but said deployment is still eligible, so I opted to assert on `device.deployment_id` instead.

lib/nerves_hub_web/live/devices/show.ex
nshoes: Oh, a VSCode extension did this lol I would like it to be known that I would never make this change willingly 😆 Fixed https://github.com/nerves-hub/nerves_hub_web/pull/1726/commits/2d2e03835a5a5fa9a9ef66fc574d0b65c01271db

lib/nerves_hub_web/live/devices/show.ex
nshoes: Done. https://github.com/nerves-hub/nerves_hub_web/pull/1726/commits/2d2e03835a5a5fa9a9ef66fc574d0b65c01271db

lib/nerves_hub/audit_logs/templates.ex
```diff
@@ -38,4 +42,13 @@ defmodule NervesHub.AuditLogs.Templates do
     AuditLogs.audit!(device, device, description)
     Logger.info("[DeviceChannel] #{description}")
   end
+
+  @spec audit_device_deployment_update(User.t(), Device.t(), Deployment.t()) :: AuditLog.t()
```
joshk, line 46: love the spec!

test/nerves_hub_web/live/devices/show_test.exs
joshk: Makes sense. Could we also assert on the flash text and the header (`Eligible Deployments`) for the deployments dropdown?

joshk (approved):
I have one small change request, but all in all this is great and can be merged in asap.

test/nerves_hub_web/live/devices/show_test.exs
nshoes: Done! Found a bug in the process as well, thanks for the callout 💯 https://github.com/nerves-hub/nerves_hub_web/pull/1726/commits/589e7ea5e2d9ae0389d186f277a79c328c6b3ae8

PR #1731, opened 2025-01-07
Improve device buttons

joshk:
Add an SVG for the 'reconnect' button, and disable `reboot`, `reconnect`, and `identify` if the device isn't connected.

Screenshot 2025-01-07 at 3 38 57 PM


lawik (approved):
Very nice 💖

PR #1735, opened 2025-01-07
Exclude deleted devices from Deployment device counts

joshk:
Missed this one from an earlier PR


lawik (approved):

joshk:
> Can we get a LiveView test for this?

fiiiiiiiiiiiine

nshoes (approved):

PR #1736, opened 2025-01-07
Relax restrictions when querying for eligible deployments

nshoes:
Eligible deployments now only match against product, platform, and architecture. When there are no eligible deployments, helper text is displayed.


joshk (approved):
😍

PR #1737, opened 2025-01-07
Create audit logs when setting deployment

nshoes:


lib/nerves_hub/audit_logs/templates.ex
```diff
@@ -51,4 +51,22 @@ defmodule NervesHub.AuditLogs.Templates do
...
+          AuditLog.t()
+  def audit_set_deployment(device, deployment, :one_found) do
+    AuditLogs.audit!(
+      device,
+      device,
+      "Updating #{device.identifier}'s deployment to #{deployment.name}"
+    )
+  end
+
+  def audit_set_deployment(device, deployment, :multiple_found) do
```
nshoes, line 65: @joshk do you think it's worth it to differentiate between one or multiple deployments found?

joshk: I think these are great audit messages!

joshk (approved):

PR #1738, opened 2025-01-07
Refresh device-related assigns when receiving connection online message, replace `mox` with `mimic`

nshoes:


lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -43,15 +43,11 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> page_title("Device #{device.identifier} - #{product.name}")
     |> assign(:tab_hint, :devices)
     |> assign(:device, device)
-    |> assign(:device_connection, device_connection(device))
-    |> assign(:console_active?, Tracker.console_active?(device))
     |> assign(:deployment, device.deployment)
```
joshk, line 46: I think this needs to move into `general_assigns` as well

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -77,7 +73,19 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
 
   def handle_info(
-        %Broadcast{event: "connection:status"},
+        %Broadcast{event: "connection:status", payload: %{status: "online"}},
+        %{assigns: %{device: device}} = socket
+      ) do
+    socket =
+      socket
+      |> general_assigns(device)
+      |> assign(:device_connection, Connections.get_latest_for_device(device.id))
```
joshk, line 82: I think we should reload the device, then do the `general_assigns`, and since we will be reloading the device we won't need to do line 82, instead we can move `|> assign(:device_connection, device_connection(device))` from `mount` into `general_assigns`

joshk (changes requested):
Thank you for taking care of this.

Do you think we should add a test to confirm data is updated on the page when a device connects?

nshoes:
> Do you think we should add a test to confirm data is updated on the page when a device connects?

Yes! On it.

lib/nerves_hub_web/live/devices/show.ex
nshoes: When setting `assigns.device_connection`, which should we prefer? `Connections.get_latest_for_device(device.id)` or `device_connection(device)`? I think that in the lifecycle events that we're talking about (`mount`, connection status inline, connection status offline) we should always fetch the latest device connection record. Thoughts?

joshk: If we are fetching the Device, the latest device connection is included, but if we aren't fetching the Device (like when the status changes to "offline") we don't need to fetch the Device, we can just use `Connections.get_latest_for_device(device.id)`, does that make sense?

nshoes: Yep! Thank you 🙇

nshoes:
@joshk test added!

Do you have any opinion on `mox` vs `mimic` for stubbing in tests? We've used `mimic` at SR and really enjoy it. My main issue with `mox` is the ceremony of setting up a behaviour when I just want to mock quickly and move on. I added `mimic` and used it in my test as an example. I checked the usage of `mox` already implemented and it would be very easy to convert to `mimic`, their APIs are very similar.

joshk:
> @joshk test added!
> 
> Do you have any opinion on `mox` vs `mimic` for stubbing in tests? We've used `mimic` at SR and really enjoy it. My main issue with `mox` is the ceremony of setting up a behaviour when I just want to mock quickly and move on. I added `mimic` and used it in my test as an example. I checked the usage of `mox` already implemented and it would be very easy to convert to `mimic`, their APIs are very similar.

Go for gold!

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -43,15 +43,9 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> page_title("Device #{device.identifier} - #{product.name}")
     |> assign(:tab_hint, :devices)
     |> assign(:device, device)
```
joshk, line 45: since this is in `general_assigns`, I think we can remove this line

joshk (approved):
Awesome work. One small comment, then I can merge it in!

lib/nerves_hub_web/live/devices/show.ex
nshoes: Done! https://github.com/nerves-hub/nerves_hub_web/pull/1738/commits/819e303dacdd0e4ba2573a9ccbd5af182078c96c

mix.lock
```diff
@@ -104,7 +102,6 @@
...
   "telemetry_metrics_statsd": {:hex, :telemetry_metrics_statsd, "0.7.1", "3502235bb5b35ce50d608bf0f34369ef76eb92a4dbc8708c7e8780ca0da2d53e", [:mix], [{:nimble_options, "~> 0.4 or ~> 1.0", [hex: :nimble_options, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:telemetry_metrics, "~> 0.6 or ~> 1.0", [hex: :telemetry_metrics, repo: "hexpm", optional: false]}], "hexpm", "06338d9dc3b4a202f11a6e706fd3feba4c46100d0aca23688dea0b8f801c361f"},
   "telemetry_poller": {:hex, :telemetry_poller, "1.1.0", "58fa7c216257291caaf8d05678c8d01bd45f4bdbc1286838a28c4bb62ef32999", [:rebar3], [{:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "9eb9d9cbfd81cbd7cdd24682f8711b6e2b691289a0de6826e58452f28c103c8f"},
-  "telemetry_registry": {:hex, :telemetry_registry, "0.3.2", "701576890320be6428189bff963e865e8f23e0ff3615eade8f78662be0fc003c", [:mix, :rebar3], [{:telemetry, "~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "e7ed191eb1d115a3034af8e1e35e4e63d5348851d556646d46ca3d1b4e16bab9"},
```
nshoes, line 105: This and `opentelemetry_liveview` above are unused deps that weren't unlocked and cleaned, so I removed them.

joshk: Great find!

PR #1740, opened 2025-01-08
The `Credo` experiment

joshk:
This PR adds credo and addresses the most important issues (and some low hanging fruit) raised when using `--strict`.

This also adds it to our CI, but it doesn't fail it for now.

I've also modified some of the settings:

- disabled `{Credo.Check.Readability.ParenthesesOnZeroArityDefs, []}` as I don't know if I agree with this, at the very least we should discuss and decide if we want this enforced. (this check, if enabled, would require us to remove `()` from functions which don't accept any arguments)
- disabled `{Credo.Check.Refactor.CondStatements, []}` as we have two 2-condition `cond` statements in `NervesHub.Extensions` which are useful to keep around. If we want to enforce this then we might need to change these until it makes sense switching back to using `cond` for them.
- customized `{Credo.Check.Refactor.Nesting, [max_nesting: 3]}`, the default is 2 but we have enough 3 level nesting that I didn't feel like going down that path right now.
- customized `Credo.Check.Warning.MissedMetadataKeyInLoggerConfig` to tell it that we do use `:all` in `runtime.exs`
- customized `Credo.Check.Readability.ParenthesesOnZeroArityDefs` to require parentheses (the opposite of its default)
- enabled `Credo.Check.Warning.LeakyEnvironment` to warn us if we are passing the full environment from the parent process when running system commands.
- enabled `Credo.Check.Readability.ImplTrue` as @jjcarstens hates it when `@impl true` is used
- and finally, enabled `Credo.Check.Readability.MultiAlias` as the use of `alias NervesHub.{Deployments, Devices, Products, ...}` is a code smell (@jjcarstens can swear to this)

We have only `readability` and `software design suggestions` issues to address, or ignore. These primarily consist of:
- `Modules should have a @moduledoc tag`
- `Nested modules could be aliased at the top of the invoking module.`
- and `Found a TODO tag in a comment`

I think we should address all of these, but we can instead leave these for another PR.


nshoes:
> disabled {Credo.Check.Readability.ParenthesesOnZeroArityDefs, []}

I abhor no parens 😆 I like the rule that there are _always_ parens, regardless of arity. Aesthetically this makes more sense to me.

> disabled {Credo.Check.Refactor.CondStatements, []} as we have two 2-condition cond statements in NervesHub.Extensions which are useful to keep around

This seems a little silly to me. Why not refactor those two-condition `cond`s?

> changed {Credo.Check.Refactor.Nesting, [max_nesting: 3]}, the default is 2 but we have enough 3 level nesting that I didn't feel like going down that path right now.

Yep, absolutely 😆 3 is fine for now.

joshk:
> > disabled {Credo.Check.Refactor.CondStatements, []} as we have two 2-condition cond statements in NervesHub.Extensions which are useful to keep around
> 
> This seems a little silly to me. Why not refactor those two-condition `cond`s?

`cond` makes a ton more sense in these two places because when another `Extension` version is released we can just slot it in there. So there is part of me that is like "why change this when it will change back in the not so distant future?"

And yet I also see your point, so I'm kinda in two minds.

I'll defer to @jjcarstens for the final decision 😄

.credo.exs
```diff
@@ -0,0 +1,224 @@
...
+          # Priority values are: `low, normal, high, higher`
+          #
+          {Credo.Check.Design.AliasUsage,
+           [priority: :low, if_nested_deeper_than: 2, if_called_more_often_than: 0]},
+          {Credo.Check.Design.TagFIXME, []},
+          # You can also customize the exit_status of each check.
+          # If you don't want TODO comments to cause `mix credo` to fail, just
+          # set this value to 0 (zero).
+          #
+          {Credo.Check.Design.TagTODO, [exit_status: 2]},
```
jjcarstens, line 92:
```suggestion
          {Credo.Check.Design.TagTODO, [exit_status: 0]},
```
I like TODO comments. There are lots of tools to keep up with it. I think having it non-error here is acceptable

.credo.exs
```diff
@@ -0,0 +1,224 @@
...
+          {Credo.Check.Readability.AliasOrder, []},
+          {Credo.Check.Readability.FunctionNames, []},
+          {Credo.Check.Readability.LargeNumbers, []},
+          {Credo.Check.Readability.MaxLineLength, [priority: :low, max_length: 120]},
+          {Credo.Check.Readability.ModuleAttributeNames, []},
+          {Credo.Check.Readability.ModuleDoc, []},
+          {Credo.Check.Readability.ModuleNames, []},
+          {Credo.Check.Readability.ParenthesesInCondition, []},
+          # Disabled as I'm not sure I agree with this one and it requires discussion
+          # {Credo.Check.Readability.ParenthesesOnZeroArityDefs, []},
```
jjcarstens, line 106:
```suggestion
          {Credo.Check.Readability.ParenthesesOnZeroArityDefs, []},
```
Like Lars said, all Nerves libs require this (or end up requiring if we having caught them all by now). The logic is that the language _requires_ parenthesis when calling the function and will warn if not included, so we want it in the definition for consistency and readability (especially when grepping to find places)

.credo.exs
```diff
@@ -0,0 +1,224 @@
...
+          {Credo.Check.Readability.UnnecessaryAliasExpansion, []},
+          {Credo.Check.Readability.VariableNames, []},
+          {Credo.Check.Readability.WithSingleClause, []},
+
+          #
+          ## Refactoring Opportunities
+          #
+          {Credo.Check.Refactor.Apply, []},
+          # Disabling this as we use a two 2-statement conds in `NervesHub.Extensions`
+          # {Credo.Check.Refactor.CondStatements, []},
```
jjcarstens, line 125: 👍🏼 Agreed. I'm unbothered by 2-statement conds. Its typical enough, especially when we think there will be more and serves as a waiting ground

.credo.exs
```diff
@@ -0,0 +1,224 @@
...
+        disabled: [
+          #
+          # Checks scheduled for next check update (opt-in for now)
+          {Credo.Check.Refactor.UtcNowTruncate, []},
+
+          #
+          # Controversial and experimental checks (opt-in, just move the check to `:enabled`
+          #   and be sure to use `mix credo --strict` to see low priority checks)
+          #
+          {Credo.Check.Consistency.MultiAliasImportRequireUse, []},
```
jjcarstens, line 180: I agree this should always be disabled (i.e. `alias Something.{Here, There, Everywhere}` == bad imo)

.credo.exs
joshk: I'm confused. Do you want to enforce `SomeModule.some_function` vs `SomeModule.some_function()`?

The Credo rule you are suggesting we enable would require us to use no parentheses when calling a function which has no arguments. 

I thought Lars would agreeing with us requiring parentheses with functions with no arguments.

sooooo confused.

.credo.exs
joshk: In that case, I think we should move this rule into the `enabled` group as it will warn when we use the nested approach you use as an example. This rule analyses our use of `alias`, `import`, `require`, and `use` and says if we are straying from the most common use, which in our case is not nested.

joshk:
@jjcarstens you might be interested in us enabling `Credo.Check.Readability.ImplTrue`, which warns against using `@impl true`, which I know you hate.

.credo.exs
joshk: Also, you might be interested in https://hexdocs.pm/credo/1.7.11/Credo.Check.Readability.MultiAlias.html

joshk:
I've gone ahead and enabled `MultiAlias`, `ImplTrue`, and `LeakyEnvironment` credo checks, as well as fixing the issues associated with them.

.credo.exs
joshk: As per my comment below, I've enabled `MultiAlias` and fixed the issues found by credo.

joshk:
@lawik @nshoes @jjcarstens this is ready for review

.credo.exs
jjcarstens: O! I think I confused myself and had things backwards. I was meaning requiring parens for zero arity (`def something() do`) - whatever is required for that

joshk: I've just pushed a commit which enforces your desires

joshk:
@nshoes @lawik @jjcarstens please reread the PR description as I've updated it to reflect some recent changes (a few rules enabled, and a few rules customized)

.credo.exs
```diff
@@ -0,0 +1,223 @@
...
+      #
+      # If you create your own checks, you must specify the source files for
+      # them here, so they can be loaded by Credo before running the analysis.
+      #
+      requires: [],
+      #
+      # If you want to enforce a style guide and need a more traditional linting
+      # experience, you can change `strict` to `true` below:
+      #
+      strict: false,
```
jjcarstens, line 49:
```suggestion
      strict: true,
```
I generally leave this on so it's always strict, but I'm also good if we want to just require the `--strict` switch for now and eventually get to this

jjcarstens (approved):

.credo.exs
joshk: Good idea, I've pushed a new commit which changes the credo config and our ci workflow file.

Issue #1742, opened 2025-01-08
Devices upgrading when deployment is set to off


gBillal:
**Describe Bug**
I created a deployment and set it to off. However my devices still upgrade themselves :sweat_smile: 
![image](https://github.com/user-attachments/assets/396be200-a358-4f98-83d5-7dcfc75a3488)

**Expected behavior**
Devices should not attempt upgrade when deplyoment is `OFF`.

**Screenshots**

lawik:
Interesting.
I haven't looked at the on/off behavior of Deployments with regard to the orchestrator. I wonder if that's only used for whether things get added to a Deployment or if it actually stop Deployment of firmwares.

Can take a look.

gBillal:
Sorry I forgot to add the steps:

- create a deployment
- Set it on
- Some devices upgraded
- Set the deployment to OFF
- Update the deployment to use another image
- The devices did downgrade themselves (since they still met the criteria)

gBillal:
Updated the image in the deployment again when it's still OFF and the devices started upgrading

lawik:
Yeah, the flag for ON/OFF is not respected by the orchestrator. Good catch. Will fix :)

joshk:
This fix has been merged in and deployed.

Thank you for reporting this @gBillal

PR #1743, opened 2025-01-08
Make orchestrator and resolve_update respect deployment.is_active

lawik:
Solves issue #1742

### DIFF
--- a/lib/nerves_hub/deployments/orchestrator.ex
+++ b/lib/nerves_hub/deployments/orchestrator.ex
@@ -15,6 +15,8 @@ defmodule NervesHub.Deployments.Orchestrator do
 
   alias NervesHub.Devices
   alias NervesHub.Devices.Device
+  alias NervesHub.Deployments
+  alias NervesHub.Deployments.Deployment
   alias NervesHub.Repo
   alias Phoenix.PubSub
   alias Phoenix.Socket.Broadcast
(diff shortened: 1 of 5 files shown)


lib/nerves_hub/deployments/orchestrator.ex
```diff
@@ -50,6 +52,10 @@ defmodule NervesHub.Deployments.Orchestrator do
   was successful, and the process is repeated.
   """
   @decorate with_span("Deployments.Orchestrator.trigger_update")
```
joshk, line 54: I'm not 100% certain if the `with_span` needs to be added to each function. Maybe, to be safe, could you move the `with_span` to be above the other `trigger_update`? thats the only one we care about.

joshk (approved):
Awesome work!

lib/nerves_hub/deployments/orchestrator.ex
```diff
@@ -49,6 +51,11 @@ defmodule NervesHub.Deployments.Orchestrator do
   As devices update and reconnect, the new orchestrator is told that the update
   was successful, and the process is repeated.
   """
+  @decorate with_span("Deployments.Orchestrator.trigger_update")
```
joshk, line 54:
```suggestion
  @decorate with_span("Deployments.Orchestrator.trigger_update#noop")
```
This will allow us to easily see why one trigger is faster than others. We might want to remove this `with_span` entirely, but for now I think the `#noop` will help.

joshk (approved):

PR #1746, opened 2025-01-08
Allow for the firmware file size limit to be configurable

joshk:
This allows for the firmware file size upload limit to be adjusted via runtime env var.

This also restrictions large file uploads to only the firmwares api route.

I'd also like this to be configurable per org, such that free users can only upload smaller firmware sizes, but thats for a different discussion.


lawik (approved):
Seems smart :)

PR #1747, opened 2025-01-10
Clean stale `DeviceConnection`s

joshk:
I noticed in the DB that we have several old/stale device connections. 

Upon looking into this more, I realised that when we added the `DeviceConnection` schema and relationships, we didn't add a way to clean the stale connections, which might linger due to unforeseen circumstances, like the app crashing.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -128,4 +128,20 @@ defmodule NervesHub.Devices.Connections do
     |> distinct(:device_id)
     |> order_by([:device_id, desc: :last_seen_at])
   end
+
+  def clean_stale_connections() do
+    interval = Application.get_env(:nerves_hub, :device_last_seen_update_interval_minutes)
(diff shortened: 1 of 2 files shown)


lawik (approved):
Looks like a win ✨

nshoes (commented):
LGTM!

1.) Since we're here, should we add a function that deletes old `device_connections` and call it from `CleanDeviceConnectionStates`? 

2.) Could we get a test or two for `CleanDeviceConnectionStates`?

joshk:
Sooooo, I'm going to cheat a little.

> 1.) Since we're here, should we add a function that deletes old `device_connections` and call it from `CleanDeviceConnectionStates`?

https://github.com/nerves-hub/nerves_hub_web/issues/1750

> 2.) Could we get a test or two for `CleanDeviceConnectionStates`?

https://github.com/nerves-hub/nerves_hub_web/issues/1751

I agree on both your points but I'm going to do those in different PRs so I can at least get the current issue of stale device connection rows fixed. I've created issues for these so they aren't forgotten.

Issue #1748, opened 2025-01-10
Bug: Cannot show devices without firmware metadata


elinol:
When trying to open a device from devices list by clicking on it I got this:
```
    ** (BadMapError) expected a map, got: nil
        (elixir 1.18.0) lib/map.ex:535: Map.get(nil, :platform, nil)
        (nerves_hub 2.0.0+42320dc7) lib/nerves_hub/deployments.ex:460: NervesHub.Deployments.eligible_deployments/1
        (nerves_hub 2.0.0+42320dc7) lib/nerves_hub_web/live/devices/show.ex:60: NervesHubWeb.Live.Devices.Show.mount/3
        (phoenix_live_view 0.20.17) lib/phoenix_live_view/utils.ex:343: anonymous fn/6 in Phoenix.LiveView.Utils.maybe_call_live_view_mount!/5
        (telemetry 1.3.0) /Users/elinol/WORK/smartrent/nerves_hub_web/deps/telemetry/src/telemetry.erl:324: :telemetry.span/3
        (phoenix_live_view 0.20.17) lib/phoenix_live_view/channel.ex:1177: Phoenix.LiveView.Channel.verified_mount/8
        (phoenix_live_view 0.20.17) lib/phoenix_live_view/channel.ex:84: Phoenix.LiveView.Channel.handle_info/2
        (stdlib 6.1.2) gen_server.erl:2345: :gen_server.try_handle_info/3
        (stdlib 6.1.2) gen_server.erl:2433: :gen_server.handle_msg/6
        (stdlib 6.1.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3
```

This happens to devices without firmware metadata (like those added via the UI), when trying to find eligible deployments for that device

PR #1749, opened 2025-01-10
Allow devices to have empty firmware when checking for deployments

elinol:
Fixes a bug (#1748 ) where the liveview crashes when trying to show a device without firmware metadata.

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -463,6 +463,8 @@ defmodule NervesHub.Deployments do
   and should only be used when a human is choosing the deployment for a device.
   """
   @spec eligible_deployments(Device.t()) :: [Deployment.t()]
+  def eligible_deployments(%Device{firmware_metadata: nil}), do: []
+
   def eligible_deployments(device) do
     Deployment
     |> join(:inner, [d], assoc(d, :firmware), as: :firmware)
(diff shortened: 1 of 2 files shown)


nshoes (approved):
Good catch! Thank you.

nshoes:
Oops, approved too soon... could we get a test for this? :)

joshk:
Sooooo, I have an idea I would like to run past everyone.

How about we list all deployments, and include in the drop down the aarch and platform of the deployment, allowing for the device to be added to the deployment, and instead add a note under the dropdown or a flash message that the device will be removed from the deployment upon connection if the aarch and platform of the device doesn't match.

This feels like a nicer user experience as a device can be manually added, if it be via api or UI, and placed in a deployment (if tags aren't used for matching). If we don't allow this then we require the device to connect, which can remove users from there flow of learning the system.

joshk (commented):
Sooooo, I have an idea I would like to run past everyone.

How about we list all deployments, and include in the drop down the aarch and platform of the deployment, allowing for the device to be added to the deployment, and instead add a note under the dropdown or a flash message that the device will be removed from the deployment upon connection if the aarch and platform of the device doesn't match.

This feels like a nicer user experience as a device can be manually added, if it be via api or UI, and placed in a deployment (if tags aren't used for matching). If we don't allow this then we require the device to connect, which can remove users from there flow of learning the system.

nshoes:
@joshk I like this! Want to create an issue?

joshk:
https://github.com/nerves-hub/nerves_hub_web/issues/1760

PR #1755, opened 2025-01-11
A small device header simplification

joshk:
'Last seen' wasn't clear in its meaning (based on a user conversation), so I instead moved this within a tooltip and refer to it as a heartbeat.

https://github.com/user-attachments/assets/87e14ee4-2677-44a1-9d86-c9efce77d2d0


nshoes (approved):
Love this! Great change :shipit:

PR #1756, opened 2025-01-12
Mix task best practises

joshk:
Adding a `@moduledoc` to the tasks means they are now listed when using `mix help`, plus it has the added advantage of showing more information when using `mix help nerves_hub.gen.devices` or `mix help nerves_hub.gen.metrics`

Oh, and I have made the task naming consistent, and given them a `@preferred_cli_env` of `:dev`

Screenshot 2025-01-12 at 10 10 15 PM


nshoes (approved):
💯 💯 💯

Issue #1707, opened 2024-12-17
Changeset not capturing constraint on deleting deployment


lawik:
A customer hit this. You can't delete a deployment if it is used by devices. Fine. But the error needs to be graceful :)

```
lib/nerves_hub/deployments.ex in NervesHub.Deployments.delete_deployment/1 at line 116
```

```
Ecto.ConstraintError
constraint error when attempting to delete struct:

    * "devices_deployment_id_fkey" (foreign_key_constraint)

If you would like to stop this constraint violation from raising an
exception and instead add it as an error to your changeset, please
call `foreign_key_constraint/3` on your changeset with the constraint
`:name` as an option.

The changeset has not defined any constraint.
```

joshk:
After some discussions in Slack, I think it would be best to change the `delete_deployment` function to nil the `deployment_id`s stored on the devices. 

This could be done by adding:

`has_many(:devices, Device, on_delete: :nilify_all)`

We would also need to add a confirmation popup when deleting the Deployment to warn that the deployment will be removed from the devices.

PR #1761, opened 2025-01-14
Ensure deployments devices are nilified on deletion

tomielee:
An error occured due to foreign key constraint when deleting a deployment. 

Atm we don't want the devices to be deleted. Check comment in issue.

Closes https://github.com/nerves-hub/nerves_hub_web/issues/1707


joshk:
❤️ ❤️ ❤️

elinol (approved):
Looks great!

PR #1763, opened 2025-01-15
Update changelog for v2.2.0 release

nshoes:
- `v2.1.0` did _not_ have the extensions work (🤦🤦🤦) , so I've moved its descriptions to `v2.2.0`. I updated the `v2.1.0` release notes as well.
- Taking formatting inspiration from [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) so we have some kind of standard
- Next release I'd like to get closer to modeling [Elixir's changelog](https://github.com/elixir-lang/elixir/releases/tag/v1.18.0). Specifically, I want to start writing about general changes always, not just when we add massive changes, and keep the more granular descriptions as a list and not refer to commits or PRs. This work is a good change in that direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,167 @@
 # Changelog
 
-## Unreleased
+Format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+# [v2.2.0] - 2025-01-14
+
+## Extensions
+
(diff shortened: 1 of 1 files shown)


nshoes:
@joshk could I get a blurb from you in the `Deployment Recalculation Changes` and `UI Rework` sections?

Issue #1734, opened 2025-01-07
Remove or explain "Current Alarms" section when empty


elinol:
When Current Alarms list is empty we should remove that section from device health in UI, alternatively tell that list is empty.
It looks a bit odd with a header without any content.

![image](https://github.com/user-attachments/assets/1e6ccdf9-1d47-4f36-aeca-714f31d07bd2)

joshk:
One idea, if there are no active alarms we should change the text to `No active alarms` with a link below it that reads `Learn how to setup Device alarms` that links to some `nerves_hub_link` code/docs/guide?

elinol:
I'm not sure there is any docs updated enough to link to, can only find this: https://github.com/nerves-hub/nerves_hub_link?tab=readme-ov-file#alarms

let's start with `No active alarms`?

lawik:
Handled now :)

PR #1765, opened 2025-01-15
Make it clear device has no active alarms

elinol:
Makes UI for device show "No active alarms" when that's the case.

Also adds test case for this.

In #1734 there's a suggestion of displaying a link to docs or code explaining how to setup alarms on devices, but I couldn't find any examples good enough.

### DIFF
--- a/lib/nerves_hub/devices/alarms.ex
+++ b/lib/nerves_hub/devices/alarms.ex
@@ -68,7 +68,7 @@ defmodule NervesHub.Devices.Alarms do
     device.id
     |> Devices.get_latest_health()
     |> case do
-      %DeviceHealth{data: %{"alarms" => alarms}} when is_map(alarms) ->
+      %DeviceHealth{data: %{"alarms" => alarms}} when is_map(alarms) and map_size(alarms) > 0 ->
         for {alarm, description} <- alarms,
             do: {String.trim_leading(alarm, "Elixir."), description}
 
--- a/lib/nerves_hub_web/live/devices/show.html.heex
+++ b/lib/nerves_hub_web/live/devices/show.html.heex
@@ -143,19 +143,21 @@
         </div>
       <% end %>
       <div>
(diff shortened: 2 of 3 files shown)


elinol:
image

nshoes (approved):
Wonderful! Thank you for the test as well 😇

joshk:
I'm happy to merge this in. Can you please talk to @lawik about adding some information to the link readme on how to use alarms? Then we can link to that.

Longer term, we need to revamp the hexdocs for link so we can explain extensions, alarms, health, etc etc.

lawik:
So this is essentially covered by that rather long README?
https://github.com/nerves-hub/nerves_hub_link?tab=readme-ov-file#configure-alarms

Or did you have something else in mind? Like explaining setting and clearing them?

elinol:
Oh great, I missed that. I stopped at the Alarms section here: https://github.com/nerves-hub/nerves_hub_link?tab=readme-ov-file#alarms

lawik:
No-no. Not the Alarms. The Alarms :D 

Yeah, I stopped there as well but having written in that document was ready for "oh, alarms this library can set/clear, gotcha". I knew I had an extensions section somewhere :D

joshk:
I think the information in the readme is a bit sparse. Having some examples would be great. 
Pointing to that section really isn't much to point to.

lawik:
@elinol I reckon you could write up a bit about how to set and clear alarms. Maybe even show what that looks like in NervesMOTD.print and on NervesHub? Screenshots are underused :)

lawik:
(fine if you don't want to, just lob it back to me)

elinol:
Sure. There's a PR in nerveshublink (https://github.com/nerves-hub/nerves_hub_link/pull/263) regarding using `Alarmist` as default alarm handler when reporting alarms, we'd might want to get that ready before updating the docs though :D

Issue #1760, opened 2025-01-13
Allow devices to be added to a deployment when they have no firmware yet


joshk:
Currently, devices can only be added to a deployment when their aarch and platform are known, which happens when they have connected to Nerves Hub for the first time (and thus send over their firmware information)

I believe it would be a better UI experience to list all deployments, and include in the drop down the aarch and platform of the deployment, allowing for the device to be added to the deployment, and instead add a note under the dropdown or a flash message that the device will be removed from the deployment upon connection if the aarch and platform of the device doesn't match.

This feels like a nicer user experience as a device can be manually added, if it be via api or UI, and placed in a deployment (if tags aren't used for matching). If we don't allow this then we require the device to connect, which can remove users from there flow of learning the system.

elinol:
I cannot assign myself but I'll grab this if someone hasn't started working on it already?

joshk:
Awesome, thanks @elinol !

joshk:
This has been completed

PR #1767, opened 2025-01-16
Allow unprovisioned devices to set deployments

elinol:
Allows for unprovisioned devices to list and set deployments belonging to product. (#1760)

- List all product deployments instead of eligible deployments when device status is `:registered`
- When a deployment is set, an info text tells the user it will be removed from device if firmware and aarch don't match on connection
- Add test for this case
- Set device to `:provisioned` when testing eligible deployments

### DIFF
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -511,14 +511,11 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> assign(:audit_pager, audit_pager)
   end
 
-  defp assign_deployments(%{assigns: %{device: device, product: product}} = socket) do
-    deployments =
-      if device.status == :provisioned,
-        do: Deployments.eligible_deployments(device),
-        else: Deployments.get_deployments_by_product(product)
+  defp assign_deployments(%{assigns: %{device: %{status: :provisioned} = device}} = socket),
+    do: assign(socket, deployments: Deployments.eligible_deployments(device))
 
-    assign(socket, deployments: deployments)
-  end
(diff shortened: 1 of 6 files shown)


elinol:
image

image

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -502,6 +502,15 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
   end
 
+  defp assign_deployments(%{assigns: %{device: device, product: product}} = socket) do
+    deployments =
+      if device.status == :provisioned,
+        do: Deployments.eligible_deployments(device),
+        else: Deployments.get_deployments_by_product(product)
+
+    assign(socket, deployments: deployments)
+  end
```
nshoes, line 512: Tiny nit, don't feel strongly about it, but I do like to burn `if`s to the ground with pattern matching 😈 

```suggestion
  defp assign_deployments(%{assigns: %{device: %{status: :provisioned} = device}} = socket),
    do: assign(socket, deployments: Deployments.eligible_deployments(device))
  
  defp assign_deployments(%{assigns: %{product: product}} = socket),
    do: assign(socket, deployments: Deployments.get_deployments_by_product(product))
```

nshoes (approved):
LGTM! :shipit:

elinol:
@nshoes think this is ready for merge?

elinol:
image

image

elinol:
It would be great to style the select options a bit, but I think that would require a custom component since option tags cannot be styled. For the future perhaps.

nshoes (approved):

joshk (approved):
This is fantastic!

And thank you for porting it to the new UI!

PR #1774, opened 2025-01-17
Add Oban web for server introspection

jjcarstens:
Oban Web was just open sourced! :heart:

This adds the Oban web dashboard into NervesHub to aid in server management and introspection.

It also introduces a `server_role` to the general User schema to allow labelling users with an `admin` or `view` role. The usecase here is that you typically want most internal users to be able to at least see status (dashboard status, job status, failure rates, etc), but don't want all users to be able to do harmful or destructive actions (retry job, cancel jobs, etc). This abstraction provides a way to restrict general users from seeing the server health pages along with allowing view or full admin privileges to some you want to introspect servers.

> [!NOTE]
> UI was excluded on purpose as I know lots of UI changes are happening right now. This also opens up more server management UI which might take extra considerations and be a bigger review
>
> I'm happy to add a simple UI to CRUD this `server_role` at the least, but it could also be accomplished with simple SQL for the time being
> `UPDATE users SET server_role = 'admin' WHERE email = 'your-fav-user@somewhere.com';`


lawik (approved):
Seems rad and reasonable :)

joshk (approved):
niiiiiiiiice

jjcarstens:
Added via b6bb866b (thanks to some fun git 😉 )

PR #1776, opened 2025-01-17
Rework console.js into LiveView Hook for UI 2.0, organize app.js and existing hooks

nshoes:
The device console-related JavaScript is now a LiveVIew Hook, allowing the code and channel connection to be refreshed when switching back and forth between the console tab. I also organized `app.js` by extracting all the hooks, that code took up too much space.


nshoes:
Handling merge conflicts now.

joshk (approved):
I loooooooooooooooove this!

I did have a question regarding window event listener cleanup, but its not urgent.

PR #1785, opened 2025-01-19
New UI: Firmwares - Part 1

joshk:
Firmware list, show, upload, and delete.

This includes some sorting options on the list page, and a count of devices using that firmware.

Some device list empty state tweaks (basic but it works)

Minor sorting and pager improvements

image

image


lawik (approved):
Love it!

PR #1786, opened 2025-01-19
Fix refresh of dynamic templates

lawik:


joshk:
Ahhhh I see how it works now!

Doesn't this mean that changes to the non `new` template will no longer trigger a reload correctly?

lawik:
Good catch. Probably. I had this at compile-time if you recall.

It may be possible to duplicate the module attributes and get both...

lawik:
I am having a hard time reproducing that it _would not_ refresh :D

joshk:
I've give this a whirl tomorrow

lawik:
I tried adding both and that seemed to work. But then I couldn't get it to not refresh when I removed them.

May need a `mix compile --force` to change the behavior but not quite sure.

joshk:
I found https://hexdocs.pm/mix/1.14/Mix.Tasks.Compile.Elixir.html#module-__mix_recompile__-0 (from https://elixirforum.com/t/external-resource-for-folder-how-does-embed-templates-do-it/55823), which shows we can define `@external_resource` twice!

I tested it locally and it fixes it for me!

joshk:
Not to say your fix didn't work, although granted I haven't checked it yet.

lawik:
Double it!

lib/nerves_hub_web/dynamic_template_renderer.ex
```diff
@@ -64,8 +64,8 @@ defmodule NervesHubWeb.DynamicTemplateRenderer do
         new_ast = new_engine.compile(new_template, filename)
 
         quote do
-          @file unquote(template)
-          @external_resource unquote(template)
+          @file unquote(new_template)
+          @external_resource unquote(new_template)
```
joshk, line 68:
```suggestion
          @file unquote(template)
          @external_resource unquote(template)
          @external_resource unquote(new_template)
```

joshk: Incorporating some magic from the Elixir docs 
(use the same `@file`, which is used for backtraces, but use two `@external_resource`, as that seems to work)

joshk (approved):

PR #1793, opened 2025-01-20
Fix issue where filter slide-over blocked device list

lawik:


lib/nerves_hub_web/live/devices/index-new.html.heex
```diff
@@ -282,11 +282,11 @@
   <% end %>
 </div>
 
-<div class={["relative", @show_filters && "z-50", !@show_filters && "-z-10"]} aria-labelledby="slide-over-title" role="dialog" aria-modal="true">
+<div class={["pointer-events-none relative", @show_filters && "z-50", !@show_filters && "-z-10"]} aria-labelledby="slide-over-title" role="dialog" aria-modal="true">
   <!-- Background backdrop, show/hide based on slide-over state. -->
   <div class={["fixed inset-0", !@show_filters && "invisable"]}></div>
 
-  <div class="fixed inset-0 overflow-hidden">
+  <div class="fixed inset-0 overflow-hidden ">
```
joshk, line 289:
```suggestion
  <div class="fixed inset-0 overflow-hidden">
```

joshk:
When you say "blocked device list", do you mean you can't interact with the device list? 

If so, I did that on purpose, do you think its better to allow the device list to be interacted with? 

If you would like to interact with the device list then I think some of the tab UI can be simplified further, similar to the multi action pop over.

lawik:
Oh! It wasn't even dimmed afaict so I assumed it was on accident.

Okay, the filter is on the side and the view is fully visible. It feels very weird to not be able to apply a filter and then click on the device or start checking devices. I 100% thought it was unintentional. Because it felt bad.

To be clear, this state:
Screenshot 2025-01-20 at 16 22 08

Can't click on a device to open it or anything :)

I am not sure I understand what you mean with the tab UI? You are talking about the Device Details view? Maybe having more things live in slide-overs? I don't know about that. But possible I suppose 🤔 

What I'm very confident about is that this feels like a bug and I would not have picked this type of slide-over if it meant it needed to be dismissed to use the listing :)

I simply had the multi-action slide-over on top of the filter one because that's what you want to DO. But both are super nice to have while working with multiple devices.

joshk:
Ah, yeah, I didn't add an blur or different color overlay, so I can see why it looked unintentional.

I'm fine with us allowing devices to be clicked and selected, although we just need to make sure the 'multi action' slide over closes or appears over the filter slide over.

> I am not sure I understand what you mean with the tab UI? You are talking about the Device Details view? Maybe having more things live in slide-overs? I don't know about that. But possible I suppose 🤔

Sorry, I meant to say slide over UI, not tab UI. My bad.

Oh, after reading your comments further I see that the multi action slide over doesn't appear over the filter slide over. Sweet.

I'll get this reviewed and merged in today.

lawik:
> Oh, after reading your comments further I see that the multi action slide over doesn't appear over the filter slide over. Sweet.

Now I'm confused. The way I did it in my branch was that it would put the multi-action on top of the filter. Not sure what happens in the current design.. :) worth trying

lawik:
@joshk did this get lost, I've hit it about 12 times since :D

joshk:
Sorry, not lost, I will get to this today now that I've finished with the majority of the new UI

joshk:
I've fixed the issue in https://github.com/nerves-hub/nerves_hub_web/pull/1813

My PR simplifies the slide over a little, allowing us to eliminate some divs.

PR #1802, opened 2025-01-21
New UI: Product settings

joshk:
Fully working product settings.

Screenshot 2025-01-21 at 4 45 59 PM

Screenshot 2025-01-21 at 4 46 36 PM


elinol:
Looks great as usual! 
Just wondering about the circles around the extensions checkboxes, the checkbox for delta updates in the top does not have a circle, is it intentional?

joshk:
Not intentional, and not sure which I like better. 

I'm going to leave them as they are for now and improve along the way.

lawik (approved):
I prefer without the circles I think but nothing that stops the merge.

PR #1803, opened 2025-01-21
Change use of CPU metrics in health indicator

lawik:
Add CPU percentage as the primary CPU indicator with temperature as secondary.

Screenshot_20250121_092252


lawik:
Also fixes a wrapping issue on smaller displays.

joshk (approved):

Issue #1796, opened 2025-01-20
New UI: Health indicators


lawik:
I just noticed that the health indicators are limited to the three:

- CPU temp
- Memory used
- Load 1|5|15

And those are not all we want. We have more info by default and the values the customer wants to expose should be surfaced front and center (in my opinion : ):

Disk is one other default reported metric that matters.

And then any custom ones sent in should also be included. Among the default ones we can make some special design choices and then custom ones need to be a bit generic.

I would say CPU usage percent is more important than temp.

In my previous implementation of this I used the CPU % as the big number and then put CPU degrees as the smaller (but colored) one. Think that works well.

Reference: https://github.com/lawik/nerves_hub_web/blob/ui2-device-details-old/lib/nerves_hub_web/live/devices/show-new.html.heex#L250

And I made a gray variant, but could just as well be the purple variant, for all custom ones.

Once we have health metrics we can switch the coloring to have that type of meaning as @joshk has suggested.

Current implementation:
https://github.com/nerves-hub/nerves_hub_web/blob/85c898401ac34e97216aa7b9b82e752c8285a6aa/lib/nerves_hub_web/components/device_page/details.ex#L122

lawik:
PR showing what I am going for is on the way.

joshk:
I do have some concerns about showing all metrics in there. 

I'm trying to balance form and function, as well as consistency (of design and layout)

I believe the details page should be the place to call out the most important metrics. Not all metrics are important, because if everything is important, nothing is important.

lawik:
I could see not showing all metrics. But by default it is quite few metrics and the graphs do not surface metrics efficiently.  And any metric the customer adds should probably be considered important-ish?

In line with the discussion on health: #1794 

I think it would make sense to show all that are used as health criteria potentially?

But I agree, not too many. But hard-coded is incredibly limiting for demonstrations and people wanting to try something interesting.

elinol:
Would it be weird to have an `Expand` button which shows the rest of the metrics when clicked?

lawik:
![Image](https://github.com/user-attachments/assets/476150c3-adf4-426b-a3ec-3e7fefc1eba8)

I think they can look alright. Having a way to expand that isn't going into the Health charts view could be good. Or putting these indicator numbers at the top of the chart view as a "full health overview". That could work. And only use the prioritized ones on the Details page. Priority being a bit arbitrary and chosen by use for now and dictated by Health feature later?

[...]

PR #1804, opened 2025-01-21
Custom metrics when viewing device

lawik:
Adds custom metrics for display:
Screenshot_20250121_091250

### DIFF
--- a/lib/nerves_hub_web/components/device_page/details.ex
+++ b/lib/nerves_hub_web/components/device_page/details.ex
@@ -165,9 +165,10 @@ defmodule NervesHubWeb.Components.DevicePage.Details do
               </div>
             </div>
           </div>
-          <div class="px-4 pb-4">
-            <.link class="text-xs font-normal text-zinc-400 hover:text-neutral-50" href="https://github.com/nerves-hub/nerves_hub_link?tab=readme-ov-file#configure-health">
-              Learn more about device health reporting.
+          <div class="px-4 pb-4 text-xs font-normal text-zinc-400 ">
+            Learn more about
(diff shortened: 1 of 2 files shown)


lawik:
This is the more controversial part of #1796 and we don't have to merge it. I think it demos well. I think it is more useful than just showing our preferred ones. But not a hill I'll die on, made the PR as much to save the generic code since we may want it on the Health view :D

lawik:
Yeah, I am on board with some kind of product-wide setting to control what shows up. I'll noodle on something.

nshoes:
> I'd be happy to merge this in if we are to combine it with an opt-in in the Product settings to show all or 'extended' metrics.

+1 for me as well!

lawik:
Want to do the setting thing. But more fiddly to get the database stuff set up. So for now removing the custom ones from the details view but including it in health. So now if you choose to look at health you get an overview of the latest data before all the graphs.

Screenshot 2025-01-22 at 23 26 55

My Mac doesn't report health properly. But I essentially copy+pasted from details and then adjusted a little bit.

We can do the settings bit in a later PR. Let me know what you think :)

joshk:
I love this! I had intended to add this to the health tab, thank you!

joshk:
Ah, sorry, didn't see that it hadn't been pushed to the PR yet 😅 

I know its a little none standard, I'm wondering if we remove the border and title and just have the stats.

We can also move the 'last updated' next to the period select buttons?

lawik:
Did I forget to commit and just pushed air? Maybe a little..

joshk:
I've made a small adjustment to your PR:

Screenshot 2025-01-24 at 9 22 33 AM

I removed the headers and link, and moved the 'last reported at' to the metrics header. I tried this without any border around the stats section, but i felt it wasn't keeping with the design.

I like this a bit better, and would be happy to merge it in.

joshk (approved):

PR #1805, opened 2025-01-21
Very basic health status report for device

elinol:
A very very simple module for getting device health status based on latest metric set, together with tests.

Not sure if it's even worth merging? Might just be a start for further work on health statuses.

Note: default thresholds are currently just placeholders and should probably be updated.


lawik:
A delicious cupcake. I have a head full of directions to go, @joshk you usually have more restraint. What next? :)

joshk:
I love where this is going.

And since we have a UI design for this, I have an idea of how we can get this into the new UI right now! 

Here are my suggestions:

- The `latest_metrics_status/1` function should also return a list of why the status is `:warning` or `:unhealthy`, eg:
```
  @type health_status() ::
          :unknown
          | :healthy
          | {:warning, [{}]}
          | {:unhealthy, [{}]}
```
eg. `{:warning, [{warning: "cpu_usage_percent"}]}` or {:unhealthy, [{unhealthy: "cpu_usage_percent"}, {warning: "mem_used_percent"}]}

This way, we get high-level health and the various factors that make up this reason. We can use these reasons in a tooltip later on.

I also wonder if we should cache the high-level reason as well, as that will allow us to filter on it. Plus, we can query on this like "any devices that were unhealthy anytime in the last hour"

- Change the `latest_metrics_status/1` function to accept a health report, instead of querying for one.

- Add a column to the `Device`s schema called `lastest_health_report_id` (which is technically a `belongs_to` association) where we can cache the latest health report for quick fetching. Alternatively, we could add this to the `DeviceConnection`s schema. @lawik thoughts?

- With the above point done, we can hook into `handle_in("report", params, socket)` in `NervesHub.Extensions.Health` to update `lastest_health_report_id` in `Device` (or `DeviceConnection`) 

- And with ^ done, we can add this to the UI
Screenshot 2025-01-23 at 7 49 04 PM

elinol:
There is a `DeviceHealth` schema, what about using that for storing the latest report?

joshk:
I think that's a good idea. We can store the full report in `DeviceHealth`. We can also store in a new field the calculated status, and in another field the 'reasons' for the status.

As for fetching the latest report, I want to connect with @esvinson and figure out a good way to do this without putting pressure on the db.

lib/nerves_hub/devices/health_status.ex
```diff
@@ -0,0 +1,77 @@
...
+    end)
+    |> report_status()
+  end
+
+  @spec latest_metrics_status(Device.t()) :: health_status()
+  def latest_metrics_status(device) do
+    device.id
+    |> Metrics.get_latest_metric_set()
+    |> calculate_metrics_status()
+  end
```
elinol, line 53: can probably be removed, along with belonging tests.

elinol:
> As for fetching the latest report, I want to connect with @esvinson and figure out a good way to do this without putting pressure on the db.

Any news on this @joshk ? What about using a `device.latest_health_report_id` , similar to the `latest_connection_id` approach?

lib/nerves_hub/devices/health_status.ex
joshk: Agree

joshk:
This is looking solid.

Looking at the tests where the output of `HealthStatus.calculate_metrics_status/1` is shown, one idea I had was to change the output from:

```elixir
{:unhealthy, 
  %{
    warning: ["disk_used_percentage"], 
    unhealthy: ["mem_used_percent"]
  }
}
```

to 

```elixir
{:unhealthy, 
  %{
    warning: %{
      "disk_used_percentage" => %{value: 73, threshold: 70}
    }, 
    unhealthy: %{
      "mem_used_percent" => %{value: 88, threshold: 80}
    }
  }
}
```

This gives useful context for the tooltips without having to inspect or run the evaluations again. 

It also means that as we work towards products customizing this logic, it will be reflected in the reports straight away.

joshk:
> > As for fetching the latest report, I want to connect with @esvinson and figure out a good way to do this without putting pressure on the db.
> 
> Any news on this @joshk ? What about using a `device.latest_health_report_id` , similar to the `latest_connection_id` approach?

Sorry @elinol, I forgot to answer this. I think adding this to the `devices` table should be fine as it won't be indexed, and thus Postgres should be smart with updates to that fields.

joshk (commented):
This is sooooo close to being ready!

lib/nerves_hub/devices.ex
```diff
@@ -226,22 +230,30 @@ defmodule NervesHub.Devices do
...
   end
 
+  defp join_and_preload(query, [assoc]), do: join_and_preload(query, assoc)
+
+  defp join_and_preload(query, [assoc | rest]) do
+    query
+    |> join_and_preload(assoc)
+    |> join_and_preload(rest)
+  end
+
```
elinol, line 256: This is might a bit messy to read?

joshk: I think its both hard to follow and feels like a smell.

Whats calling it? What do you need to load?

lib/nerves_hub/devices.ex
```diff
@@ -1133,10 +1151,21 @@ defmodule NervesHub.Devices do
...
+      {:ok, device_health} ->
+        Device
+        |> where([d], d.id == ^device_id)
+        |> Repo.update_all(set: [latest_health_id: device_health.id])
+
+        {:ok, device_health}
+
+      {:error, _} = error ->
+        error
+    end
```
joshk, line 1168: this might be a good use case for a `Multi`

lib/nerves_hub/devices.ex
joshk: maybe it might be clearer to pattern match on the assoc being a list and then using `Enum` to cycle through the preloads? Although I guess this works too, so maybe its just about leaving some comments to explain what its doing and allowing us to do.

lib/nerves_hub/devices.ex
elinol: Agree, I was just being lazy 🤡

lib/nerves_hub/devices.ex
elinol: Switched to Enum in latest commit, made much more sense :D

joshk:
I've add the health status to the device list, and added a tooltip hover.

Screenshot 2025-02-03 at 3 45 43 PM

Screenshot 2025-02-03 at 3 45 54 PM

Last thing to do is to add this information to the Device show page.

@elinol do you have some time to take a stab at this?

elinol:
image

joshk (approved):
Lets 🚢  !

elinol:
I'm not sure about the default thresholds, are they ok?

joshk:
I think the thresholds can be lowered a lot, but we can do that in another PR 😄

PR #1807, opened 2025-01-22
Remove old `device_connections` in worker, remove connection state stored on `device`

nshoes:
- Introduce `device.latest_connection`, which references the latest `device_connection` record for the device
- Start cleaning up `device_connections` every hour with `NervesHub.Workers.DeleteOldDeviceConnections`
  - This targets rows that have a `last_seen_at` older than two weeks by default (configurable), a `status` of `:disconnected` and does not delete a device's last `device_connection`, even if the previous criteria are met
- Remove `connection_status`, `connection_established_at`, and `connection_last_seen_at` from the `device` schema.
- Slot in `device.latest_connection`, simplifying a lot of queries and logic related to `device_connections`
- Remove `Devices.clean_connection_states` and it's call from a worker since `device.connection_status` was removed and connection logic is moving to `device.latest_connection`
- When a device connects to the socket, we set the created `device_connection` as `device.lastest_connection`
- Create `device.latest_connection` when seeding devices

Todo:
- [x] check filtering UI locally


lib/nerves_hub/devices/connections.ex
```diff
@@ -144,4 +144,13 @@ defmodule NervesHub.Devices.Connections do
       ]
     )
   end
+
+  def delete_old_connections() do
+    interval = Application.get_env(:nerves_hub, :device_connection_max_age_days)
+    days_ago = DateTime.shift(DateTime.utc_now(), day: -interval)
+
+    DeviceConnection
+    |> where([d], d.last_seen_at < ^days_ago)
```
nshoes, line 153: @joshk Since there are no timestamps on `device_connections` is `last_seen_at` the right column to query against?

joshk: you can actually do this by id, as UUID v7s have the timestamp at the front, but I think a few other things pop out at me first.

We shouldn't delete connections which are still active (where `state == connected`). 

And it would also be good to keep at least one device connection row per device. This current query could remove all device connections for a device if it hasn't reconnected recently.

joshk: I think if this is run hourly then we will be fine, its the first one which we might want to run manually with some batching.

joshk: You could argue that batching should be the default, in case truncation is enabled for an installation that has a large dataset already, but I don't think anyone is using it yet, or at least to the extremes as SmartRent.

lib/mix/tasks/gen.devices.ex
```diff
@@ -40,9 +40,6 @@ defmodule Mix.Tasks.NervesHub.Gen.Devices do
         org_id: org.id,
         product_id: product.id,
         identifier: "generated-#{i}",
-        connection_status: :connected,
-        connection_established_at: DateTime.now!("Etc/UTC"),
-        connection_last_seen_at: DateTime.now!("Etc/UTC"),
         connection_metadata: %{
```
joshk, line 43: since we have `metadata` on `DeviceConnection`, I wonder if we remove this field too? (we just need to make sure we save the geo data to the `DeviceConnection`

lib/nerves_hub/devices.ex
```diff
@@ -89,11 +72,22 @@ defmodule NervesHub.Devices do
...
+          order_by(query, [d, o, p, dp, lc], desc_nulls_last: lc.last_seen_at)
+
+        sorting ->
+          sorting
+      end
+    end)
     |> Filtering.build_filters(filters)
-    |> preload([d, o, p, dp, f], org: o, product: p, deployment: {dp, firmware: f})
-    |> Connections.preload_latest_connection()
+    |> preload([d, o, p, dp, f, _lc], org: o, product: p, deployment: {dp, firmware: f})
```
joshk, line 90:
```suggestion
    |> preload([d, o, p, dp, f, lc], org: o, product: p, deployment: {dp, firmware: f}, latest_connection: lc)
```

lib/nerves_hub/devices/connections.ex
```diff
@@ -4,6 +4,7 @@ defmodule NervesHub.Devices.Connections do
   """
   import Ecto.Query
 
+  alias Ecto.Multi
```
joshk, line 7:
```suggestion
```

lib/nerves_hub/devices/connections.ex
```diff
@@ -144,4 +141,16 @@ defmodule NervesHub.Devices.Connections do
...
+
+  def delete_old_connections() do
+    interval = Application.get_env(:nerves_hub, :device_connection_max_age_days)
+    days_ago = DateTime.shift(DateTime.utc_now(), day: -interval)
+
+    DeviceConnection
+    |> join(:inner, [dc], d in Device, on: dc.device_id == d.id)
+    |> where([dc, _d], dc.last_seen_at < ^days_ago)
+    |> where([dc, _d], dc.status != :connected)
+    |> where([dc, d], dc.id != d.latest_connection_id)
```
joshk, line 153: that is sooooo smart!!!!!

lib/nerves_hub/devices/device.ex
```diff
@@ -67,15 +64,16 @@ defmodule NervesHub.Devices.Device do
...
+    # Deprecated fields, remove these any time after 29/1/2025.
+    # Also remove index from NervesHub.Repo.Migrations.AddConnectionStatusIndexToDevices.
+    # field(:connection_status, Ecto.Enum,
+    #   values: [:connected, :disconnected, :not_seen],
+    #   default: :not_seen
+    # )
+    # field(:connection_established_at, :utc_datetime)
+    # field(:connection_disconnected_at, :utc_datetime)
+    # field(:connection_last_seen_at, :utc_datetime)
+    field(:latest_connection_id, :binary_id)
```
joshk, line 76: is this field needed anymore?

lib/nerves_hub/tracker.ex
```diff
@@ -69,7 +70,10 @@ defmodule NervesHub.Tracker do
   Returns `true` if device's latest connections has a status of `:connected`,
   otherwise `false`.
   """
-  def online?(%{device_connections: [%{status: :connected}]}), do: true
+  def online?(%{latest_connection: %Ecto.Association.NotLoaded{}} = device),
+    do: Repo.preload(device, :latest_connection)
```
joshk, line 74: isn't returning a device breaking the expected true/false return type?

lib/nerves_hub/devices/connections.ex
nshoes: Batching by default is right here, I'll implement.

lib/nerves_hub/devices/device.ex
nshoes: Yes! I was surprised, but it makes sense:

```
** (Ecto.QueryError) deps/ecto/lib/ecto/repo/queryable.ex:174: field `latest_connection_id` in `update` does not exist in schema NervesHub.Devices.Device in query:

     from d0 in NervesHub.Devices.Device,
       where: d0.id == ^12229,
       update: [set: [latest_connection_id: ^"0194919f-8c43-73ff-bdcd-e269e5edef88"]]
```

lib/nerves_hub/tracker.ex
nshoes: Yes 😮‍💨 good catch!

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -468,7 +470,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
     end)
   end
 
-  defp device_connection(%{device_connections: [connection]}), do: connection
+  defp device_connection(%{latest_connection: latest_connection}), do: latest_connection
```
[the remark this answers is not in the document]
nshoes: How will it always return something? `device.latest_connection_id` will be `nil` until the device connects to the socket, so there is an in-between period. Should we backfill? Even then, a device could still not have a device_connection, so maybe that's a bad idea.

lib/nerves_hub/devices/device.ex
joshk: I'm confused, shouldn't `has_one` add this field?

Silly us, i just read the Ecto docs in more detail and I think what we want is https://hexdocs.pm/ecto/Ecto.Schema.html#belongs_to/3

> You should use belongs_to in the table that contains the foreign key. Imagine a company <-> employee relationship. If the employee contains the company_id in the underlying database table, we say the employee belongs to company.

lib/nerves_hub_web/live/devices/show.ex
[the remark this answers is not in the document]
joshk: well, what I mean to say is if `device_connection(device)` returns the latest device connection, why not just do `device.lastest_connection`?

lib/nerves_hub/devices/connections.ex
joshk: Whats the API for batching?

lib/nerves_hub/devices/device.ex
```diff
@@ -44,6 +40,7 @@ defmodule NervesHub.Devices.Device do
     embeds_one(:firmware_metadata, FirmwareMetadata, on_replace: :update)
     has_many(:device_certificates, DeviceCertificate, on_delete: :delete_all)
     has_many(:device_connections, DeviceConnection, on_delete: :delete_all)
+    has_one(:latest_connection, DeviceConnection)
```
elinol, line 43: 🙌

lib/nerves_hub_web/live/devices/show.ex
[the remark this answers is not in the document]
nshoes: Ah! I see, agreed.

lib/nerves_hub/devices/connections.ex
nshoes: Generally, it's:
- construct query
- `Repo.stream` with `max_rows` set
- `Stream.chunk_every(max_rows)`
- then whatever `Stream` function you need to process the data

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -121,6 +121,8 @@ defmodule NervesHubWeb.Live.Devices.Show do
     # Get device with its latest connection data preloaded
     {:ok, device} = Devices.get_device_by_identifier(org, device.identifier, :latest_connection)
 
+    dbg(device)
```
jjcarstens, line 124:
```suggestion
```

lib/mix/tasks/gen.devices.ex
nshoes: I'm going to hold off on this and create an issue for it.

nshoes: https://github.com/nerves-hub/nerves_hub_web/issues/1819

lib/nerves_hub/workers/schedule_org_audit_log_truncation.ex
```diff
@@ -1,6 +1,7 @@
 defmodule NervesHub.Workers.ScheduleOrgAuditLogTruncation do
   use Oban.Worker,
-    queue: :truncation
+    queue: :truncation,
+    max_attempts: 1
```
nshoes, line 4: Moved this from `config.exs`.

lib/nerves_hub/devices.ex
```diff
@@ -204,13 +193,13 @@ defmodule NervesHub.Devices do
     end)
   end
 
-  defp sort_devices({:asc, :connection_last_seen_at}),
-    do: {:asc_nulls_first, :connection_last_seen_at}
+  defp sort_devices_by_latest_connection(query, {:asc, :connection_last_seen_at}),
```
joshk, line 196: I would just rename this to `sort_devices`. This function is then just a handy want to overload it if you need a special sorting case included.

lib/nerves_hub/devices.ex
```diff
@@ -298,8 +287,7 @@ defmodule NervesHub.Devices do
   end
 
   defp join_and_preload(query, :latest_connection) do
```
joshk, line 289: technically this isn't doing a join and preload anymore. Either we remove the function and just use `preload` in the places it was used, or we implement the handy `join and then preload from the join`

lib/nerves_hub/devices.ex
```diff
@@ -655,21 +643,6 @@ defmodule NervesHub.Devices do
     |> Repo.all()
   end
 
-  def clean_connection_states() do
```
joshk, line 658: was this removed or moved to somewhere else?

lib/nerves_hub/devices/connections.ex
```diff
@@ -144,4 +106,29 @@ defmodule NervesHub.Devices.Connections do
       ]
     )
   end
+
+  def delete_old_connections() do
```
joshk, line 110: I've been thinking about this, and I don't think this is very optimal.

If you do a standard delete query, postgres doesn't have to transfer any data, and it is smart with how it deletes it. 

If you do this streaming then you are fetching ids, sending them to elixir, then constructing another query which has to be sent with all the ids. Its more work for the DB.

If we wanted to do some form of 'chunking' then I would do it based on date ranges, but the fact of the matter is that this query is run every hour, so it doesn't make much sense to do anything special.

priv/repo/seeds.exs
```diff
@@ -43,7 +43,8 @@ defmodule NervesHub.SeedHelpers do
       conditions: %{"version" => "< 1.0.0", "tags" => ["beta"]}
     })
 
-    Fixtures.device_fixture(org, product, firmwares |> elem(1), %{connection_last_seen_at: DateTime.utc_now()})
+    device = Fixtures.device_fixture(org, product, firmwares |> elem(1))
+    # create device_connection.latest_connection
```
joshk, line 47: comment needed?

joshk (changes requested):

lib/nerves_hub/devices.ex
nshoes: Good call!

lib/nerves_hub/devices.ex
nshoes: Good call, implemented as a join + preload.

lib/nerves_hub/devices.ex
nshoes: Removed because it was updating `devices.connection_status`. After thinking, does this still need to exist, but update `device.latest_connection`?

priv/repo/seeds.exs
nshoes: Nope, was cruft. I have implemented the fixture call.

lib/nerves_hub/devices.ex
joshk: ahhh this is correct to remove, and we have kept around the `Connection` variant, so all is good

joshk (approved):
This looks good to me!

lib/nerves_hub/devices.ex
nshoes: The answer is no, talked about it offline.

lib/nerves_hub/devices/connections.ex
nshoes: Talked offline and landed on a recursive strategy.

lib/nerves_hub/devices/connections.ex
```diff
@@ -144,4 +106,33 @@ defmodule NervesHub.Devices.Connections do
...
+      DeviceConnection
+      |> where([d], d.id in ^ids)
+      |> Repo.delete_all()
+
+    if delete_count == 0 do
+      :ok
+    else
+      # relax stress on Ecto pool and go again
+      Process.sleep(2000)
+      delete_old_connections()
```
joshk, line 135: could we instead create a new oban job?

joshk (approved):
One comment, but lets test this out!

lib/nerves_hub/devices/connections.ex
```diff
@@ -109,26 +109,30 @@ defmodule NervesHub.Devices.Connections do
 
   def delete_old_connections() do
     interval = Application.get_env(:nerves_hub, :device_connection_max_age_days)
+    delete_limit = Application.get_env(:nerves_hub, :device_connection_delete_limit)
     days_ago = DateTime.shift(DateTime.utc_now(), day: -interval)
 
-    query =
+    ids =
```
nshoes, line 115: This seems scary but 100k UUIDS in a list takes up only 4.1MB of memory.

nshoes: JK, subquery to the rescue.

lib/nerves_hub/devices/connections.ex
nshoes: We can at a later time 😎

PR #1813, opened 2025-01-23
New UI: Alignment improvements, and allow entire table cells to be clickable

joshk:


lib/nerves_hub_web/live/devices/index-new.html.heex
```diff
@@ -237,7 +237,7 @@
               </div>
             </td>
 
-            <td class="actions">
+            <%!-- <td class="actions">
```
nshoes, line 240: Cruft? I HATE CRUFT.

joshk: Oh, I'm sorry, I'm gonna leave this here while we decide if we really need these actions or not. Pretty please.

nshoes: I'll allow it this one time.

nshoes (approved):

lib/nerves_hub_web/live/devices/index-new.html.heex
joshk: YYYYAAAAAAAAYYYY

PR #1814, opened 2025-01-23
New UI: Improve the readability of the device list progress bar

joshk:
Before:

https://github.com/user-attachments/assets/6206bb99-05c1-4d64-9c85-f758f056f529

I felt the radial background was too muted, and it wasn't clear which device the solid green bar was for (think of a device that is half way down the list)

Now:

https://github.com/user-attachments/assets/034b7d83-8209-4048-bb28-61e5adf3a702

I've added a progress bar label, removed the top line, and tweaked the gradient a little

### DIFF
--- a/lib/nerves_hub_web/live/devices/index-new.html.heex
+++ b/lib/nerves_hub_web/live/devices/index-new.html.heex
@@ -172,6 +172,9 @@
                 <.link navigate={~p"/org/#{@org.name}/#{@product.name}/devices/#{device.identifier}"} class={"ff-m #{firmware_update_status(device)}"} title={firmware_update_title(device)}>
                   {device.identifier}
                 </.link>
+                <span :if={@progress[device.id]} class="flex items-center gap-1 ml-2 pl-2.5 pr-2.5 py-0.5 border border-zinc-700 rounded-full bg-zinc-800">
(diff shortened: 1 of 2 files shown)


nshoes (approved):
🚀 🚀 🚀

lawik:
This was my first approach. I do really like the line. I guess this is more clear though.
What about this _and_ the line? Did you try that?

This definitely works.

I like the progress label. Makes it very clear that an update is in progress. I think if the numeric part of that has a fixed width it would not jiggle the layout except when it shows up and goes away.

I'll try a variation :D

joshk:
Yep, I did try this and the line: (a slightly earlier variation)

https://github.com/user-attachments/assets/d283d4c0-7bea-4f7c-99ae-60851532707c

What I didn't think worked well was when the device that was updating was half way down the list, the line looked like it might be part of the device above. I'm not sure how to describe it, it felt a little busy.

Just the updating background is my preference.

lawik:
Variant here: https://github.com/nerves-hub/nerves_hub_web/pull/1816

It is essentially the one in Josh's video above. I prefer that one :D 
I also dropped the percentages from the updating label which should reduce table-jiggle a lot.

PR #1816, opened 2025-01-23
Variant of the progress bar tweaks

lawik:
If you can't tell I really like the thin progress bar :D 

Screenshot_20250123_095813

This uses the fix of the background from: https://github.com/nerves-hub/nerves_hub_web/pull/1814

But it also uses the thin stripe.

For simplicity I cut the % display in the label. I think the label is helpful but it was making the whole table jiggle and it would be a bunch of faffing about to make it behave nice.

### DIFF
--- a/lib/nerves_hub_web/live/devices/index-new.html.heex
+++ b/lib/nerves_hub_web/live/devices/index-new.html.heex
@@ -173,7 +173,7 @@
                   {device.identifier}
                 </.link>
                 <span :if={@progress[device.id]} class="flex items-center gap-1 ml-2 pl-2.5 pr-2.5 py-0.5 border border-zinc-700 rounded-full bg-zinc-800">
-                  <span class="text-xs text-zinc-300 tracking-tight">Updating: {@progress[device.id]}%</span>
+                  <span class="text-xs text-zinc-300 tracking-tight">updating</span>
                 </span>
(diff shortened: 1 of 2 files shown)


joshk:
I can live without the percentage in the label, maybe `Updating` instead of `updating` would look better.

As for the thin green line, can you show an image or video of what this looks like with a few extra devices and two devices updating at the same time in the middle of the list?

lawik:
I liked lower-case for the label, but not married to it :D 

I tried it with a couple of devices and a fixed number for the progress. I don't have a good setup for shipping firmware on my local dev env so when I've done video I've had to do weird mocking of progress-increments. Didn't do that this time.

If you are shipping yours to some test-environment, let me know if that's something I can easily do as well :)

I think the background is a big improvement in clarity. I think we can ship this so we can tweak it based on how it works day-to-day. And if we find the line confusing in practice I'll happily drop it :)

I'd like to see it in action and getting that action locally is currently a pain. I think I need to look at an S3 setup or something, not sure what the problem is.

joshk:
I'll deploy it to prod tomorrow and create some videos using the cluster product.

joshk:
Uploading Screen Recording 2025-01-24 at 12.22.23 AM.mov…

I can live with this

PR #1825, opened 2025-01-24
Various fixes found by Dialyzer

joshk:
Mostly specs, some aliases and type defs, and some return types that don't match. And one function no longer used.


elinol (approved):
✨

PR #1829, opened 2025-01-25
New UI: Add support for creating deployments

joshk:
I rejigged the UI a little.

I've had to use a `deployments/newz` for the route, for now, as the liveview and UI were different enough.

Screenshot 2025-01-25 at 3 38 41 PM

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -301,6 +301,11 @@ defmodule NervesHub.Deployments do
     :ok
   end
 
+  @spec new_deployment() :: Changeset.t()
+  def new_deployment() do
+    Ecto.Changeset.change(%Deployment{})
+  end
+
   @spec change_deployment(Deployment.t(), map()) :: Changeset.t()
   def change_deployment(deployment, params) do
     Deployment.changeset(deployment, params)
--- a/lib/nerves_hub_web/live/deployments/newz.ex
+++ b/lib/nerves_hub_web/live/deployments/newz.ex
@@ -8,7 +8,7 @@ defmodule NervesHubWeb.Live.Deployments.Newz do
 
   @impl Phoenix.LiveView
   def mount(_params, _session, %{assigns: %{product: product}} = socket) do
-    if Firmwares.count(socket.assigns.product) == 0 do
+    if Firmwares.count(product) == 0 do
       socket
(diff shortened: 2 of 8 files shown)


lib/nerves_hub_web/live/deployments/newz.ex
```diff
@@ -0,0 +1,163 @@
...
+  use NervesHubWeb, :updated_live_view
+
+  alias NervesHub.AuditLogs.DeploymentTemplates
+  alias NervesHub.Deployments
+  alias NervesHub.Firmwares
+  alias NervesHub.Firmwares.Firmware
+
+  @impl Phoenix.LiveView
+  def mount(_params, _session, %{assigns: %{product: product}} = socket) do
+    if Firmwares.count(socket.assigns.product) == 0 do
```
elinol, line 11: why not use the `product` match in function head?

joshk: ooooh, hahaha, good spotting.

joshk, line 11:
```suggestion
    if Firmwares.count(product) == 0 do
```

lib/nerves_hub/deployments.ex
```diff
@@ -301,6 +301,11 @@ defmodule NervesHub.Deployments do
     :ok
   end
 
+  @spec new_deployment() :: Changeset.t()
+  def new_deployment() do
+    Ecto.Changeset.change(%Deployment{})
+  end
+
```
elinol, line 308: I'm just curious, why is this not a "regular" changeset with casts and so on?

elinol (approved):

lib/nerves_hub/deployments.ex
joshk: This just returns an empty Deployment changeset, which is required by `to_form`. Since we aren't checking any validations, we don't need to add them yet. Its common to have different changesets for different use cases.

elinol: I see. I'm just not used to seeing a changeset this empty :D

joshk: totally understand, here is more information for whats I'm doing : https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#to_form/2-creating-a-form-from-changesets

PR #1831, opened 2025-01-25
Dialyzer and credo fixes

joshk:


lib/nerves_hub/devices.ex
```diff
@@ -784,7 +787,8 @@ defmodule NervesHub.Devices do
           |> DateTime.truncate(:second)
           |> DateTime.add(deployment.penalty_timeout_minutes * 60, :second)
 
-        DeviceTemplates.audit_firmware_upgrade_blocked(deployment, device)
+        _ = DeviceTemplates.audit_firmware_upgrade_blocked(deployment, device)
```
jjcarstens, line 790: I think if we don't plan on ever caring about the result here, then we should adjust template functions to only ever return `:ok`

lib/nerves_hub/devices.ex
```diff
@@ -698,20 +698,22 @@ defmodule NervesHub.Devices do
...
-    |> Ecto.Changeset.change()
-    |> Ecto.Changeset.put_change(:deployment_id, deployment.id)
-    |> Repo.update!()
-    |> Repo.preload(:deployment, force: true)
+    %Device{} =
+      device
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.put_change(:deployment_id, deployment.id)
+      |> Repo.update!()
+      |> Repo.preload(:deployment, force: true)
```
jjcarstens, line 706: Im guessing this is a unique work around to `Repo.preload` allowing returns of list, struct, or nil. But I would vote we learn from dialyzer to rethink the approach here rather than add something that passes but is visually odd and maybe confusing to others.

i.e. it seems what we're really trying to accomplish is ensuring the deployment is in the struct. So why not just put it in there rather than requiring a DB reload?
```suggestion
    %Device{} =
      device
      |> Ecto.Changeset.change()
      |> Ecto.Changeset.put_change(:deployment_id, deployment.id)
      |> Repo.update!()
      |> Map.put(:deployment, deployment)
```

lib/nerves_hub/devices.ex
```diff
@@ -698,20 +698,22 @@ defmodule NervesHub.Devices do
...
-    |> Ecto.Changeset.change()
-    |> Ecto.Changeset.put_change(:deployment_id, nil)
-    |> Repo.update!()
-    |> Repo.preload(:deployment, force: true)
+    %Device{} =
+      device
+      |> Ecto.Changeset.change()
+      |> Ecto.Changeset.put_change(:deployment_id, nil)
+      |> Repo.update!()
+      |> Repo.preload(:deployment, force: true)
```
jjcarstens, line 716: Likewise here. Why do another DB action when we know nothing will load. We can simply do the work and skip the call _and_ it should fix dialyzer
```suggestion
      device
      |> Ecto.Changeset.change()
      |> Ecto.Changeset.put_change(:deployment_id, nil)
      |> Repo.update!()
      |> Map.put(:deployment, nil)
```

lib/nerves_hub/devices.ex
joshk: Good idea, I'm going to change the PR to do that instead.

lib/nerves_hub/devices.ex
joshk: Fair call, I'll merge this in and check it out.

nshoes (approved):

PR #1834, opened 2025-01-26
Fix: New devices connecting shouldn't cause extension errors

joshk:
This relates to an issue found via Sentry.

```
UndefinedFunctionError nil.__struct__/0
lib/nerves_hub_web/channels/extensions_channel.ex in anonymous fn/3 in NervesHubWeb.ExtensionsChannel.parse_extensions/2 at line 34
```

This only occurred when the device was created during connection, eg. shared secrets jitp. The `:extensions` `embeds_one` in Device was nil on creation. The fix was just asking Ecto to set a default.


elinol (approved):
Great!!

PR #1835, opened 2025-01-27
Remove the need to get a connection before updating it

joshk:
This rejigs the `device_heartbeat` function a little, moving the broadcast into the function and out of the socket.

This also removes the `Repo.get!`, instead using `Repo.update_all`

A `join` has been added so we can fetch the device identifier to use for the broadcast.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -77,15 +77,13 @@ defmodule NervesHub.Devices.Connections do
         ]
       )
 
-    Phoenix.Channel.Server.broadcast_from!(
+    Phoenix.Channel.broadcast_from!(
       NervesHub.PubSub,
       self(),
       "device:#{result.identifier}:internal",
       "connection:heartbeat",
       %{}
     )
-
-    :ok
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/devices/connections.ex
```diff
@@ -63,12 +63,29 @@ defmodule NervesHub.Devices.Connections do
...
+
+    Phoenix.Channel.Server.broadcast_from!(
+      NervesHub.PubSub,
+      self(),
+      "device:#{result.identifier}:internal",
+      "connection:heartbeat",
+      %{}
+    )
+
+    :ok
```
jjcarstens, line 88:
```suggestion
    Phoenix.Channel.broadcast_from!(
      NervesHub.PubSub,
      self(),
      "device:#{result.identifier}:internal",
      "connection:heartbeat",
      %{}
    )
```

`Phoenix.Channel.Server.broadcast_from!/5` is speced to return `:ok` (via `Phoenix.PubSub.broadcast_from/5` which it is calling out to) So you can leave this as the return

jjcarstens (approved):

lib/nerves_hub/devices/connections.ex
joshk: Ahhh, nice! Thank you!

PR #1836, opened 2025-01-27
Encapsulate device connection behaviour

joshk:
This PR is a foundational step towards NervesHub supporting different device transport backends.

The logic for device connections has been better encapsulated in a new module, `DeviceLink.Connections`. The `DeviceLink` namespace is intended to be used for further behaviour expansion.

These focused modules are easier to test and clearer to understand.

The next steps are to work through the various channel modules and apply a similar pattern.

### DIFF
--- a/lib/nerves_hub/device_link/connections.ex
+++ b/lib/nerves_hub/device_link/connections.ex
@@ -18,6 +18,9 @@ defmodule NervesHub.DeviceLink.Connections do
   @type auth() :: {:ssl_certs, any()} | {:shared_secrets, list()}
   @type connection_id() :: binary()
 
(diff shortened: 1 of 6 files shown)


nshoes:
👁️ 👃 👁️ 

Is this just organizational work or are we changing behavior as well?

joshk:
> 👁️ 👃 👁️
> 
> Is this just organizational work or are we changing behavior as well?

This isn't changing any behavior. Not one test is changed, all tests are passing.

This is more organizationally focused, leading up to a bigger set of work focused on enabling different device transports (websocket proxy, mqtt, <insert new fandangled message queue>)

PR #1837, opened 2025-01-27
Reduce queries made during device socket connection

joshk:
Optimize some queries used during device connection, with a focus on making sure the product assoc is returned with the device.


elinol (approved):
Looks good to me!

PR #1838, opened 2025-01-27
Track who created or edited script

elinol:
Adds `:created_by` and `:last_updated_by` to `Script` schema.

Note: Does not take account deletion into account. Are user accounts ever deleted at all?

### DIFF
--- a/lib/nerves_hub/scripts.ex
+++ b/lib/nerves_hub/scripts.ex
@@ -57,24 +57,36 @@ defmodule NervesHub.Scripts do
 
   @spec create(Product.t(), User.t(), map()) :: {:ok, Script.t()} | {:error, Changeset.t()}
   def create(product, user, params) do
-    product
-    |> Ecto.build_assoc(:scripts)
-    |> Script.changeset(params)
-    |> Repo.insert()
-    |> case do
+    params =
+      params
+      |> Map.put(:created_by_id, user.id)
+      |> normalize_keys()
+
+    result =
+      product
+      |> Ecto.build_assoc(:scripts)
+      |> Script.create_changeset(params)
+      |> Repo.insert()
+
+    case result do
       {:ok, script} ->
         ProductTemplates.audit_script_created(user, product, script)
         {:ok, script}
 
-      err ->
-        err
+      _ ->
+        result
     end
   end
 
   @spec update(Script.t(), User.t(), map()) :: {:ok, Script.t()} | {:error, Changeset.t()}
   def update(script, user, params) do
+    params =
+      params
+      |> Map.put(:last_updated_by_id, user.id)
+      |> normalize_keys()
+
     script
-    |> Script.changeset(params)
+    |> Script.update_changeset(params)
     |> Repo.update()
     |> case do
       {:ok, script} ->
@@ -87,5 +99,12 @@ defmodule NervesHub.Scripts do
       err ->
         err
     end
+  end
+
+  defp normalize_keys(params) do
+    Enum.reduce(params, %{}, fn {key, value}, acc ->
+      key = if is_atom(key), do: key, else: String.to_existing_atom(key)
+      Map.put(acc, key, value)
+    end)
   end
 end
--- a/lib/nerves_hub/scripts.ex
+++ b/lib/nerves_hub/scripts.ex
@@ -57,36 +57,23 @@ defmodule NervesHub.Scripts do
 
   @spec create(Product.t(), User.t(), map()) :: {:ok, Script.t()} | {:error, Changeset.t()}
   def create(product, user, params) do
-    params =
-      params
-      |> Map.put(:created_by_id, user.id)
-      |> normalize_keys()
-
-    result =
-      product
-      |> Ecto.build_assoc(:scripts)
-      |> Script.create_changeset(params)
-      |> Repo.insert()
-
-    case result do
+    %Script{}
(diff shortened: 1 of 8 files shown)


lib/nerves_hub/scripts/script.ex
```diff
@@ -3,23 +3,36 @@ defmodule NervesHub.Scripts.Script do
...
     struct
-    |> cast(params, [:name, :text])
-    |> validate_required([:name, :text])
+    |> cast(params, @required ++ [:created_by_id])
+    |> validate_required(@required ++ [:created_by_id])
     |> validate_length(:name, lte: 255)
+    |> foreign_key_constraint(:created_by_id)
+  end
+
+  def update_changeset(struct, params) do
```
nshoes, line 31:
```suggestion
  def update_changeset(struct, params \\ %{}) do
```
If we add a sensible default here we don't have to remember to pass in an empty map when calling. Small nit for me, feel free to disregard.

nshoes (commented):
Yes, accounts can get deleted! Check out `NervesHub.Accounts.RemoveAccount`.

lib/nerves_hub/scripts.ex
```diff
@@ -57,24 +57,36 @@ defmodule NervesHub.Scripts do
...
+      |> Map.put(:created_by_id, user.id)
+      |> normalize_keys()
+
+    result =
+      product
+      |> Ecto.build_assoc(:scripts)
+      |> Script.create_changeset(params)
+      |> Repo.insert()
+
+    case result do
```
joshk, line 71: I'd recommend:

```
%Script{}
|> Script.create_changeset(params, product, user)
|> Repo.insert()
|> case do
```
This would also mean changing `create_changeset/2` to:
```
  def create_changeset(struct, product, created_by, params) do
    struct
    |> cast(params, @required)
    |> put_assoc(:product, product)
    |> put_assoc(:created_by, created_by)
    |> validate_required(@required ++ [:created_by])
    |> validate_length(:name, lte: 255)
    |> foreign_key_constraint(:created_by_id)
  end
```

lib/nerves_hub/scripts/script.ex
```diff
@@ -3,23 +3,36 @@ defmodule NervesHub.Scripts.Script do
 
   import Ecto.Changeset
 
+  alias NervesHub.Accounts.User
   alias NervesHub.Products.Product
 
   @type t :: %__MODULE__{}
+  @required [:name, :text]
```
joshk, line 10: `:product` is missing

lib/nerves_hub/scripts.ex
```diff
@@ -57,24 +57,36 @@ defmodule NervesHub.Scripts do
...
   @spec update(Script.t(), User.t(), map()) :: {:ok, Script.t()} | {:error, Changeset.t()}
   def update(script, user, params) do
+    params =
+      params
+      |> Map.put(:last_updated_by_id, user.id)
+      |> normalize_keys()
+
     script
-    |> Script.changeset(params)
+    |> Script.update_changeset(params)
```
joshk, line 89: Similar to my comment above, I think a similar approach here is best.

```
script
|> Script.update_changeset(updated_by, params)
```

This does away with needing lines 83-86, and the `normalize_keys` function.

lib/nerves_hub_web/live/support_scripts/edit.ex
```diff
@@ -8,14 +8,14 @@ defmodule NervesHubWeb.Live.SupportScripts.Edit do
...
+        %{assigns: %{org: org, product: product}} = socket
       ) do
     script = Scripts.get_by_product_and_id!(product, script_id)
 
     socket
-    |> page_title("Edit Support Script - #{socket.assigns.org.name}")
+    |> page_title("Edit Support Script - #{org.name}")
     |> sidebar_tab(:support_scripts)
-    |> assign(:form, to_form(Script.changeset(script, %{})))
+    |> assign(:form, to_form(Script.update_changeset(script, %{})))
```
joshk, line 18: Since we aren't running validations when displaying the form for the first time, `Ecto.Changeset.change/1` is better to use.

lib/nerves_hub_web/live/support_scripts/new.ex
```diff
@@ -9,7 +9,7 @@ defmodule NervesHubWeb.Live.SupportScripts.New do
     socket
     |> page_title("New Support Script - #{socket.assigns.org.name}")
     |> sidebar_tab(:support_scripts)
-    |> assign(:form, to_form(Script.changeset(%Script{}, %{})))
+    |> assign(:form, to_form(Script.create_changeset(%Script{}, %{})))
```
joshk, line 12: Since we aren't running validations when displaying the form for the first time, Ecto.Changeset.change/1 is better to use.

joshk (changes requested):

lib/nerves_hub/scripts/script.ex
elinol: Is it necessary when using `put_assoc` ?

elinol:
> Yes, accounts can get deleted! Check out `NervesHub.Accounts.RemoveAccount`.

Oh, I was using the new UI where account removal is not implemented yet :D

What do we want to do with `created_by` and `last_updated_by` fields when user is deleted then? Nilify?

joshk (approved):
This is fantastic! Great choice to nillify references when a user is deleted.

PR #1843, opened 2025-01-28
Fix an extensions bug related to channels not sharing assigns

joshk:
We don't know of a devices firmware metadata until it connects for the first time. When it does connect, the `DeviceSocket` will fetch the device and the `firmware_metadata` field will be nil. This device is the one shared with other channels.

In `Extensions.Health` we are using `socket.assigns.device.firmware_metadata` which is going to be nil during its first connection to NervesHub, and will continue to exit sending reports and creating Sentry errors.

This test replicates the scenario of a device connecting for the first time,  then its firmware metadata being updated, and then connecting to the health extension, exposing the error.

Sadly, in tests, channels share assigns, so this was much harder to replicate.

### DIFF
--- a/lib/nerves_hub_web/channels/extensions_channel.ex
+++ b/lib/nerves_hub_web/channels/extensions_channel.ex
@@ -3,6 +3,7 @@ defmodule NervesHubWeb.ExtensionsChannel do
 
   alias NervesHub.Extensions
   alias NervesHub.Helpers.Logging
+  alias NervesHub.Repo
   alias Phoenix.PubSub
   alias Phoenix.Socket.Broadcast
 
@@ -10,6 +11,11 @@ defmodule NervesHubWeb.ExtensionsChannel do
 
   @impl Phoenix.Channel
   def join("extensions", extension_versions, socket) do
+    # the assigns are not shared between channels, so if we don't
+    # reload the device we are likely to have incorrect data, especially
+    # after the first connect event
+    socket = reload_device(socket)
+
     extensions = parse_extensions(socket.assigns.device, extension_versions)
     socket = assign(socket, :extensions, extensions)
 
@@ -115,4 +121,13 @@ defmodule NervesHubWeb.ExtensionsChannel do
   end
 
   def handle_info(_msg, socket), do: {:noreply, socket}
+
+  defp reload_device(%{assigns: %{device: device}} = socket) do
+    device =
+      device
+      |> Repo.reload()
+      |> Repo.preload(:product)
+
+    assign(socket, :device, device)
+  end
 end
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/channels/extensions_channel.ex
```diff
@@ -22,6 +25,9 @@ defmodule NervesHubWeb.ExtensionsChannel do
     topic = "device:#{socket.assigns.device.id}:extensions"
     :ok = PubSub.subscribe(NervesHub.PubSub, topic)
 
+    # the assigns are not shared between channels, so if we don't
+    # reload the device we are likely to have incorrect data, especially
+    # after the first connect event
```
nshoes, line 30: This comment should be above the call to `reload_device/1`, right?

test/nerves_hub_web/channels/extensions_channel_test.exs
```diff
@@ -23,6 +25,56 @@ defmodule NervesHubWeb.ExtensionsChannelTest do
     assert_push("extensions:get", _extensions)
   end
 
+  test "joining extensions channel works when the device has connected for the first time" do
+    user = Fixtures.user_fixture()
+    org = Fixtures.org_fixture(user)
+    product = Fixtures.product_fixture(user, org)
+
+    {:ok, device} =
+      Devices.create_device(%{
```
nshoes, line 34: Why use this instead of the fixture?

test/nerves_hub_web/channels/extensions_channel_test.exs
```diff
@@ -23,6 +25,56 @@ defmodule NervesHubWeb.ExtensionsChannelTest do
...
+    # simulate the device channel updating the params
+    params = %{
+      "nerves_fw_uuid" => Ecto.UUID.generate(),
+      "nerves_fw_product" => product.name,
+      "nerves_fw_architecture" => "arm64",
+      "nerves_fw_version" => "0.0.0",
+      "nerves_fw_platform" => "test_host"
+    }
+
+    device = NervesHub.Repo.one(NervesHub.Devices.Device)
```
nshoes, line 54: We can just reload this since we have a reference to it now?

test/nerves_hub_web/channels/extensions_channel_test.exs
joshk: Yep, the fixture was adding everything, including firmware, which we didn't want.

lib/nerves_hub_web/channels/extensions_channel.ex
joshk: good spotting, i originally had the reload in a different area

test/nerves_hub_web/channels/extensions_channel_test.exs
joshk: This isn't needed, good spotting

nshoes (approved):

PR #1844, opened 2025-01-28
Health Report: Only store metadata directly from the device

joshk:
My suggestion is that the health report should be for data directly from the device, and thus, we shouldn't store `socket.device.firmware_metadata` in the report `metadata`

@lawik what are your thoughts?


lawik:
I would be inclined to agree but I suppose we lose the storage of firmware metadata over time? Knowing which version correlates with which changes in health is kind of powerful.

PR #1845, opened 2025-01-28
New UI: Allow deletion of scripts

elinol:
- Delete button (with confirm box) on edit script page.
- Adds `Scripts.delete/3` function.
- Audit log when script is deleted


elinol:
@joshk I was missing the option to delete my test scripts in the new UI, so added it.

PR #1846, opened 2025-01-28
Broadcast that a device has been updated when setting and clearing deployment

nshoes:
The device channel needs to be notified when a device's deployment changes so it can resubscribe to the proper deployment channel and send an update if available.


joshk (approved):
Loooove this! Only thing remaining is the failing test.

PR #1847, opened 2025-01-28
Create audit log when archive is sent to device

nshoes:
If an archive is sent to a device on connect or from an archive update, an audit log is created specifying device, deployment, and archive details.


joshk (changes requested):
can you please return `:ok` from the templates, this makes dialyzer happy

jjcarstens (dismissed):
I don't think we want this. There is no discernment check for archive versions and this happens on every connection, so this has potential to really bloat the logs again (like online/offline/exetension logs that have been removed)

Perhaps a small metric is useful or we wait until proper adjustment and handling of archives?

joshk:
> I don't think we want this. There is no discernment check for archive versions and this happens on every connection, so this has potential to really bloat the logs again (like online/offline/exetension logs that have been removed)
> 
> Perhaps a small metric is useful or we wait until proper adjustment and handling of archives?

Ahhh, true true, there is no annoucement from the device about what archive version has been downloaded and installed (something bigger we should consider improving)

what about moving the logging to `handle_info(%Broadcast{event: "archives/updated"}, socket)`, as this will only be logged when a device is already connected?

nshoes:
> so this has potential to really bloat the logs again (like online/offline/exetension logs that have been removed)

Great callout, thank you.

> what about moving the logging to handle_info(%Broadcast{event: "archives/updated"}, socket), as this will only be logged when a device is already connected?

This is a good option to avoid table bloat over time, but then we don't get an audit log if we do send the archive update on connect (which feels weird). I think if we log 2 out of 3 scenarios that's better than none?

esvinson:
Do we need to store it in the database or can we just log it with the normal logger?

esvinson:
> Perhaps a small metric is useful or we wait until proper adjustment and handling of archives?
We'd want to avoid anything that's hub specific for cost reasons.  Deployment specific might be ok.

joshk:
I would suggest only adding this audit log event to `handle_info(%Broadcast{event: "archives/updated"}, socket)` as it would only occur when the deployment archive is updated and sent to connected devices. It would help devs and support know if an online device received the archive.

nshoes:
Updated to not create audit log if archive is sent on connect.

Can someone sanity check me? I see the device channel responds to a "archives/updated" message from pubsub, but the [only place we publish that is for deployments](https://github.com/nerves-hub/nerves_hub_web/blob/add-audit-logs-when-sending-archives/lib/nerves_hub/deployments.ex#L273) 🤔 Can I remove the `handle_info` for that or am I missing something?

joshk:
> Updated to not create audit log if archive is sent on connect.
> 
> Can someone sanity check me? I see the device channel responds to a "archives/updated" message from pubsub, but the [only place we publish that is for deployments](https://github.com/nerves-hub/nerves_hub_web/blob/add-audit-logs-when-sending-archives/lib/nerves_hub/deployments.ex#L273) 🤔 Can I remove the `handle_info` for that or am I missing something?

I find it odd that we do the broadcast in the audit function. 

But why would you want to remove the `handle_info`?

nshoes:
> But why would you want to remove the handle_info?

TIL that the [device channel is subscribed to `depoyment:id` topics](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/channels/device_channel.ex#L55) 🙃 My mistake 😄

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -197,7 +197,7 @@ defmodule NervesHubWeb.DeviceChannel do
       socket
       |> update_device(device)
       |> maybe_start_penalty_timer()
-      |> maybe_send_archive()
+      |> maybe_send_archive(audit_log: true)
```
joshk, line 200: I'm torn if we should audit in `"devices/updated"`. I fear this event is used waaaaaay too generally. 

Happy to try it though.

joshk (approved):

lib/nerves_hub_web/channels/device_channel.ex
joshk: Should be fine, I'm most likely being too concerned about nothing.

PR #1848, opened 2025-01-29
Add support for a custom defined device location

joshk:
This feature allows for a custom location to be selected on a map for a device. 

Support has been added to override the location being reported by the device, as well as removing a previously selected custom location.

https://github.com/user-attachments/assets/e6669cb6-2892-4b4a-944d-0bb34f00d19f


joshk:
I need some help from @nshoes to make the js beautiful

nshoes (approved):

PR #1849, opened 2025-01-29
Device Cert Auth: Don't fetch data if we don't need it

joshk:
Prefer `Repo.exists?/1` if we are just confirming it is in the db. 
Preload from a `join` where possible.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -547,8 +547,8 @@ defmodule NervesHub.Devices do
     end
   end
 
-  @spec ca_certificate_by_ski_match?(binary) :: boolean()
-  def ca_certificate_by_ski_match?(ski) do
+  @spec known_ca_ski?(binary) :: boolean()
+  def known_ca_ski?(ski) do
     CACertificate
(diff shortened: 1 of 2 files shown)


nshoes (approved):

lib/nerves_hub/devices.ex
```diff
@@ -545,11 +547,22 @@ defmodule NervesHub.Devices do
     end
   end
 
+  @spec ca_certificate_by_ski_match?(binary) :: boolean()
+  def ca_certificate_by_ski_match?(ski) do
```
jjcarstens, line 551:
```suggestion
  @spec known_ca_ski?(binary) :: boolean()
  def known_ca_ski?(ski) do
```

Not set on this, but being more contextually based function name on what it is trying to accomplish vs _how_ it is accomplishing that might be a little more useful. What do you think?

jjcarstens (approved):
LGTM!

lib/nerves_hub/devices.ex
joshk: Fixed!

PR #1850, opened 2025-01-29
Save the device location to `DeviceConnection.metadata`

joshk:
This moves `connection_metadata` and `connection_types` to both use `DeviceConnection.metadata`.

Next week we can create a PR to remove unused fields from the `devices` table.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -90,20 +90,40 @@ defmodule NervesHub.Devices.Connections do
   Updates `status` and relevant timestamps for a device connection record,
   and stores the reason for disconnection if provided.
   """
-  @spec device_disconnected(UUIDv7.t(), String.t() | nil) ::
-          {:ok, DeviceConnection.t()} | {:error, Ecto.Changeset.t()}
+  @spec device_disconnected(UUIDv7.t(), String.t() | nil) :: :ok | :error
   def device_disconnected(ref_id, reason \\ nil) do
     now = DateTime.utc_now()
 
     DeviceConnection
-    |> Repo.get!(ref_id)
-    |> DeviceConnection.update_changeset(%{
-      last_seen_at: now,
-      disconnected_at: now,
-      disconnected_reason: reason,
-      status: :disconnected
-    })
-    |> Repo.update()
+    |> where(id: ^ref_id)
+    |> update([dc],
+      set: [
+        last_seen_at: ^now,
+        disconnected_at: ^now,
+        disconnected_reason: ^reason,
+        status: :disconnected
+      ]
+    )
+    |> Repo.update_all([])
+    |> case do
+      {1, _} -> :ok
+      _ -> :error
+    end
+  end
+
+  @doc """
+  Updates the connection `metadata` by merging in new metadata.
+  """
+  @spec merge_update_metadata(UUIDv7.t(), map()) :: :ok | :error
+  def merge_update_metadata(ref_id, new_metadata) do
+    DeviceConnection
+    |> where(id: ^ref_id)
+    |> update([dc], set: [metadata: fragment("? || ?::jsonb", dc.metadata, ^new_metadata)])
+    |> Repo.update_all([])
+    |> case do
+      {1, _} -> :ok
+      _ -> :error
+    end
   end
 
   def clean_stale_connections() do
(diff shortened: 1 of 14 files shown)


lib/nerves_hub/devices/device.ex
```diff
@@ -27,40 +27,47 @@ defmodule NervesHub.Devices.Device do
...
     :connection_types,
-    :connection_metadata,
     :status,
     :first_seen_at
   ]
   @required_params [:org_id, :product_id, :identifier]
 
   schema "devices" do
-    belongs_to(:org, Org, where: [deleted_at: nil])
-    belongs_to(:product, Product, where: [deleted_at: nil])
```
nshoes, line 38: Why remove these? Because cascading deletes are set and this wouldn't be a real scenario?

lib/nerves_hub/devices/device.ex
```diff
@@ -73,10 +80,9 @@ defmodule NervesHub.Devices.Device do
     # field(:connection_established_at, :utc_datetime)
     # field(:connection_disconnected_at, :utc_datetime)
     # field(:connection_last_seen_at, :utc_datetime)
-    embeds_one(:extensions, DeviceExtensionsSetting,
-      defaults_to_struct: true,
-      on_replace: :update
-    )
+
+    # Deprecated fields, remove these any time after 28/2/2025.
+    # field(:connection_metadata, :map, default: %{})
```
nshoes, line 82:
```suggestion
    # Deprecated fields, remove these any time after 28/2/2025.
    # field(:connection_metadata, :map, default: %{})
    # field(:connection_established_at, :utc_datetime)
    # field(:connection_disconnected_at, :utc_datetime)
    # field(:connection_last_seen_at, :utc_datetime)
```

lib/nerves_hub/devices/connections.ex
```diff
@@ -90,20 +90,40 @@ defmodule NervesHub.Devices.Connections do
...
+  end
+
+  @doc """
+  Updates the connection `metadata` by merging in new metadata.
+  """
+  @spec merge_update_metadata(UUIDv7.t(), map()) :: :ok | :error
+  def merge_update_metadata(ref_id, new_metadata) do
+    DeviceConnection
+    |> where(id: ^ref_id)
+    |> update([dc], set: [metadata: fragment("? || ?::jsonb", dc.metadata, ^new_metadata)])
```
nshoes, line 121: TIL how to merge jsonb in postgres! Cool.

lib/nerves_hub/devices/connections.ex
```diff
@@ -90,20 +90,40 @@ defmodule NervesHub.Devices.Connections do
...
+    |> where(id: ^ref_id)
+    |> update([dc],
+      set: [
+        last_seen_at: ^now,
+        disconnected_at: ^now,
+        disconnected_reason: ^reason,
+        status: :disconnected
+      ]
+    )
+    |> Repo.update_all([])
```
nshoes, line 107: Mild nit: I'd prefer to `set` the update in `update_all` instead of `update`. It's an extra function call and we're not using anything of the row we're updating.

lib/nerves_hub/devices/device.ex
joshk: I haven't removed them, I just reorganized the fields and associations. I did remove the `where` clauses, as they don't make sense for cases where you fetched the Device and try to use `device.org`.

lib/nerves_hub/devices/device.ex
joshk: You want to remove all these fields together? I'm thinking we can remove them next week if you are able to get a deploy out this week?

lib/nerves_hub/devices/device.ex
nshoes: Sorry, was referring to the `where` clause. Sounds good!

lib/nerves_hub/devices/device.ex
nshoes: Yes, these four fields should be removed after we deploy this PR. My comment was suggesting to group them all back together under the comment header.

lib/nerves_hub/devices/connections.ex
nshoes: Yes 😎

nshoes (approved):

PR #1860, opened 2025-01-30
Remove the need to store the devices deployment in the `DeviceChannel`

joshk:
This is all about reducing the information stored in the assigns for the `DeviceChannel`. 

It's also allowed us to remove a bunch of preloads which were being done often and fairly aggressively.

### DIFF
--- a/lib/nerves_hub/archives.ex
+++ b/lib/nerves_hub/archives.ex
@@ -71,18 +71,10 @@ defmodule NervesHub.Archives do
   def archive_for_deployment(nil), do: nil
 
   def archive_for_deployment(deployment_id) do
-    deployment =
-      Deployment
-      |> where(id: ^deployment_id)
-      |> join(:inner, [d], a in assoc(d, :archive))
-      |> preload([_, a], archive: a)
-      |> Repo.one()
-
-    if deployment do
-      deployment.archive
-    else
-      nil
-    end
+    Archive
+    |> join(:inner, [a], d in Deployment, on: d.archive_id == a.id)
+    |> where([a, d], d.id == ^deployment_id)
+    |> Repo.one()
   end
 
   # TODO: check on other return signatures
--- a/lib/nerves_hub/archives.ex
+++ b/lib/nerves_hub/archives.ex
@@ -7,6 +7,7 @@ defmodule NervesHub.Archives do
 
   require Logger
 
+  alias NervesHub.Deployments.Deployment
   alias NervesHub.Archives.Archive
   alias NervesHub.Fwup
   alias NervesHub.Products.Product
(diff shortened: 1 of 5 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -41,6 +41,7 @@ defmodule NervesHubWeb.DeviceChannel do
       device
       |> Deployments.verify_deployment_membership()
       |> Deployments.set_deployment()
+      |> Map.put(:deployment, nil)
```
nshoes, line 44: Is there a functional reason for this?

nshoes (commented):
I like this direction a lot! Just one question.

lib/nerves_hub_web/channels/device_channel.ex
joshk: I don't love this either, but i wanted to just have a quick improvement to start with as there are more memory/data changes to come.

The idea is that the Device schema we store in assigns should be slim. Since we don't need the deployment information we should not store it.

lib/nerves_hub/archives.ex
```diff
@@ -66,6 +67,24 @@ defmodule NervesHub.Archives do
...
+      |> where(id: ^deployment_id)
+      |> join(:inner, [d], a in assoc(d, :archive))
+      |> preload([_, a], archive: a)
+      |> Repo.one()
+
+    if deployment do
+      deployment.archive
+    else
+      nil
+    end
```
jjcarstens, line 85: Would it make sense to query from the Archive so that is the return instead of having to do a check?
```suggestion
    Archive
    |> join(:inner, [a], d in Deployment, on: d.archive_id == a.id)
    |> where([a, d], d.id == ^deployment_id)
    |> Repo.one()
```

lib/nerves_hub/deployments.ex
```diff
@@ -385,7 +401,8 @@ defmodule NervesHub.Deployments do
   @spec verify_deployment_membership(Device.t()) :: Device.t()
   def verify_deployment_membership(%Device{deployment_id: deployment_id} = device)
       when not is_nil(deployment_id) do
-    %{deployment: deployment} = device = Repo.preload(device, deployment: :firmware)
+    {:ok, deployment} = get_deployment_for_device(device)
```
jjcarstens, line 404: What is the motivation for manually fetching here rather than preload? I feel like there is a reason I just can't remember

lib/nerves_hub_web/channels/device_channel.ex
jjcarstens: Just to callout this way will restrict preloading this struct later on without a `force: true`. But I _believe_ this will be short lived and probably deleted soon anyway?

```elixir
iex(6)> device.deployment_id
7
iex(7)> device.deployment
#Ecto.Association.NotLoaded<association :deployment is not loaded>
iex(8)> device2 = NervesHub.Repo.preload(%{device | deployment: nil}, :deployment)
iex(9)> device2.deployment
nil
```

jjcarstens (approved):
LGTM - Just a couple Jon thoughts

lib/nerves_hub_web/channels/device_channel.ex
joshk: I think I'm fine with this, as I using `preload` generally means this information is stored in the device and taking up more memory when we aren't going to use the struct. Plus, we generally want to load extra information with the deployment, like the firmware and archive, so doing the preload will usually result in several db calls instead of one by using joins + preload.

lib/nerves_hub/deployments.ex
joshk: `join` + `preload` == 1 query, vs `Repo.preload(device, deployment: :firmware)` == 2 queries

PR #1864, opened 2025-01-30
Add support for clear and update deployment broadcasts

joshk:
More focused events, and use the information from the event instead of reloading data.

### DIFF
--- a/test/nerves_hub_web/channels/device_channel_test.exs
+++ b/test/nerves_hub_web/channels/device_channel_test.exs
@@ -4,6 +4,7 @@ defmodule NervesHubWeb.DeviceChannelTest do
 
   import TrackerHelper
 
+  alias NervesHub.Devices
   alias NervesHub.Fixtures
   alias NervesHubWeb.DeviceChannel
   alias NervesHubWeb.DeviceSocket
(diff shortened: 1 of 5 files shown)


test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -130,6 +131,63 @@ defmodule NervesHubWeb.DeviceChannelTest do
...
+    {:ok, _join_reply, socket} =
+      subscribe_and_join(socket, DeviceChannel, "device")
+
+    refute is_nil(socket.assigns.device.deployment_id)
+    refute is_nil(socket.assigns.deployment_channel)
+
+    Devices.clear_deployment(device)
+
+    # we need to let the channel process all messages before we can
+    # check the state of the device's connection types
```
nshoes, line 153: How does `get_state` achieve this?! Will it wait until the mailbox is empty or?

nshoes (approved):
LGTM, one small question.

test/nerves_hub_web/channels/device_channel_test.exs
jjcarstens: `:sys.get_state` is just like a `GenServer.call` - Under the hood it is still a message sent to the process mailbox and blocks until receiving a response. Visually its something like this:

```
[
  <message from Devices.clear_deployment/1>,
  <maybe another one from Devices.clear_deployment/1>,
  <request from :sys.get_state/1
]
```

We're banking on the messages sent by `Devices.clear_deployment/1` getting sent to the process _before_ our request for the state. The process handles them FIFO

nshoes: Very clever! Love it

PR #1865, opened 2025-01-31
Run one deployment orchestrator per deployment per cluster

joshk:
Currently we run one `Deployments.Orchestrator` per deployment per node, it doesn't matter if the deployment is active or not, it will still be run on each node.

If we have 100 deployments, and 10 nodes, then we have 100 orchestrators per node, and 1000 overalls.

This architecture, specifically having an orchestrator for a deployment run on every node, makes it very hard to ensure only a certain number of concurrent updates are running at the same time. And it also makes it hard to build out deployment workflows.

This PR uses `ProcessHub` to ensure one orchestrator is running per deployment across the cluster. Additionally, it only runs orchestrators that are set as active. 

This ensures centralised management of a deployment, while using `ProcessHub` to redistribute orchestrators to other nodes if a node is to go offline.

The current implementation allows for us to switch between strategies.

If we commit to this path (after testing in the field), we can remove the device registry, simplifying device connections. Plus, it will also enable us to add support for other device transports and transport setups, like a socket proxy or mqtt.

To run this locally, open up two terminals and run:
```elixir
make iex-server-clustered
```
and
```elixir
make iex-server-clustered num=2
```

The new distributed orchestrators are set to only run on **web** nodes, which also builds up to us supporting a 'device proxy' setup.

**Still required from this PR:**
- [x] tests
- [x] testing in an isolated environment
- [x] specs and docs


lawik:
Very cool.

Running orchestrators to do nothing does seem like a waste, though I guess orchestrators do have ad-hoc work as new devices join the deployment.

I guess throughput for highly concurrent rollouts (1000-2000 at a time) of hundreds of thousands of total devices is a bit of a question. How fast is it to have a single node doing that work?

The way orchestrators work now is unusual and we definitely want all this under test regardless of strategy. I think having multiple nodes share the work of pushing out updates has some upside in terms of efficiency and reliably getting updates out but it does have drawbacks for concurrency control and monitoring the process. I guess focusing it back into one node makes it a lot easier to check the results of the fleet health as the deployment rolls out as well.

I am overall in favor. I haven't worked with Horde before so don't instinctively trust it but have heard good things :D

lib/nerves_hub/devices.ex
```diff
@@ -665,6 +667,43 @@ defmodule NervesHub.Devices do
...
+      on: [product_id: ^deployment.product_id],
+      as: :firmware
+    )
+    |> join(:left, [d], ifu in InflightUpdate, on: d.id == ifu.device_id, as: :inflight_update)
+    |> where(deployment_id: ^deployment.id)
+    |> where(updates_enabled: true)
+    |> where([latest_connection: lc], lc.status == :connected)
+    |> where([d], not is_nil(d.firmware_metadata))
+    |> where([d, firmware: f], fragment("(? #>> '{\"uuid\"}') != ?", d.firmware_metadata, f.uuid))
+    |> where([inflight_update: ifu], is_nil(ifu.id))
```
nshoes, line 700: Should this be: (?)
```suggestion
    |> where([inflight_update: ifu], is_nil(ifu))
```

lib/nerves_hub/deployments/distributed/orchestrator_registration.ex
```diff
@@ -0,0 +1,43 @@
...
+    }
+  end
+
+  def start_link() do
+    GenServer.start_link(__MODULE__, [])
+  end
+
+  @impl GenServer
+  def init(_) do
+    Process.send_after(self(), :start_orchestrators, :timer.seconds(3))
```
nshoes, line 27: Why are we doing this 🙃

lib/nerves_hub/deployments.ex
```diff
@@ -258,6 +284,14 @@ defmodule NervesHub.Deployments do
         _ = maybe_trigger_delta_generation(deployment, changeset)
         :ok = broadcast(deployment, "deployments/update")
 
+        if Map.has_key?(changeset.changes, :is_active) do
+          if deployment.is_active do
+            deployment_activated_event(deployment)
+          else
+            deployment_deactivated_event(deployment)
+          end
+        end
```
nshoes, line 293: Can we get a test for this?

lib/nerves_hub/devices.ex
```diff
@@ -894,6 +934,68 @@ defmodule NervesHub.Devices do
...
   end
 
+  def deployment_device_online(%Device{deployment_id: nil}) do
+    :ok
+  end
+
+  def deployment_device_online(device) do
+    case Application.get_env(:nerves_hub, :deployments_orchestrator) do
+      "multi" ->
+        :ok
```
nshoes, line 944: Shouldn't we tell the local orchestrator that the device has come online so we can (maybe) trigger an update?

lib/nerves_hub/devices.ex
```diff
@@ -665,6 +667,43 @@ defmodule NervesHub.Devices do
...
+  - have automatic updates enabled
+  - not currently updating
+  - not be running the same firmware version associated with the deployment
+  - not in the penalty box (based on `updates_blocked_until`)
+
+  The list is ordered by current connection age. Devices that have been online longer
+  are updated first.
+  """
+  @spec available_for_update(Deployment.t(), non_neg_integer()) :: [Device.t()]
+  def available_for_update(deployment, count) do
```
nshoes, line 684: This is real nice 💯

lib/nerves_hub/devices.ex
```diff
@@ -1276,11 +1403,19 @@ defmodule NervesHub.Devices do
...
-    |> Ecto.Changeset.change()
-    |> Ecto.Changeset.put_change(:status, "updating")
-    |> Repo.update!()
+  def update_started!(inflight_update, device, deployment, reference_id) do
+    Repo.transaction(fn ->
+      DeviceTemplates.audit_device_deployment_update_triggered(
+        device,
+        deployment,
+        reference_id
+      )
```
nshoes, line 1412: Is there a reason why we're creating an audit log inside of a transaction?

joshk: It was me being extra safe, in case one of the inserts failed.

lib/nerves_hub/devices.ex
joshk: possibly, although that would need to happen in the device channel as that it would need to use `send`

I'd suggest punting this as the issue already exists for the current orchestrator, or that we at least leave it for a different PR

lib/nerves_hub/deployments.ex
joshk: What would you like to test from this?

lib/nerves_hub/deployments/distributed/orchestrator_registration.ex
joshk: I want to give `ProcessHub` a little time to start and a little time for libcluster to connect to other nodes.

I'm likely being over protective.

lib/nerves_hub/devices.ex
joshk: either work. I can certainly change it.

lib/nerves_hub/devices.ex
nshoes: That makes sense, thanks.

lib/nerves_hub/deployments/distributed/orchestrator_registration.ex
nshoes: I believe we are here; [`libcluster` is started in the application tree before the new orchestrators](https://github.com/nerves-hub/nerves_hub_web/pull/1865/files#diff-1a572c6a4aff9c22d62b287dd611e9677136374b5c7b1f5c0be5f16b22be7a43R41). I feel pretty strongly against defensive code like this, but I appreciate the careful thought and see what you were thinking. Could we remove this unless we find documentation or proof that we need to wait?

lib/nerves_hub/deployments.ex
nshoes: FYI I'm working on these tests, no worries! Just want to make sure we're not duplicating work :)

nshoes:
@joshk did some things [here](https://github.com/nerves-hub/nerves_hub_web/pull/1865/commits/d426dd135228c6cc048a1048626ff67cb5e5fe15) 😬 
- Default to "clustered" in test env, we should be testing like we fly IMO
- The distributed orchestrator test was created with the `.ex` extension so it hasn't been running in CI
- Remove Mimic global mode in the distributed orchestrator test. It was causing issues while running with other tests; not 100% certain why since we're `async: false`. I got it working with `allow`.
- Add tests for broadcasts when a deployment goes from `active` -> `inactive` and vice versa

I am looking into CI failures now.

nshoes:
@joshk Ah, there seems to be flaky distributed orchestrator tests. I saw it a few times locally. [CI caught one](https://github.com/nerves-hub/nerves_hub_web/actions/runs/13142522913/job/36672999487), I'm going to take a stab at it.

nshoes:
@joshk I haven't actually fixed the flaky test, I've just inversely flaked it 😆 [This test](https://github.com/nerves-hub/nerves_hub_web/pull/1865/files#diff-99da744e863ca68497d2310c17d570a7a9db240d087c3ffdc8b155de1fa53c75R188-R190) asserts on messages and their order, the happy path after the fix being:
```
- deployment/device-online
- devices/deployment-updated
- update-scheduled
```
But sometimes we get:
```
- deployment/device-online
- update-scheduled
- devices/deployment-updated
```

I'm not sure what the right fix is since we can only pattern match with `assert_receive`.

nshoes:
After some debugging with `receive` I believe I have fixed it

nshoes (approved):
:shipit: :shipit: :shipit: :shipit:

PR #1868, opened 2025-02-03
Improvements to the 'updating' header on the deployments summary tab

joshk:
All devices up to date:

Screenshot 2025-02-03 at 5 59 49 PM

Updating:

Screenshot 2025-02-03 at 6 00 52 PM

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -901,6 +901,14 @@ defmodule NervesHub.Devices do
     |> Repo.aggregate(:count)
   end
 
+  @spec updating_count(Deployment.t()) :: term() | nil
+  def updating_count(%Deployment{id: id}) do
+    InflightUpdate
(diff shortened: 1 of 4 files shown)


lawik:
Great to get the broken logic fixed (which I assume this does from skimming the diff : )

I think we should use the top-of-page progress bar to match the device details view.

If you definitely want it kept inside the box I would try and make it flush up against the upper border. Now it has a weird offset :) I'm okay with in the box but the offset bothers me.

Should be a great improvement ✨

joshk:
Please feel free to play with this. If you can find a way for it to look good connected to the top bar with some of that info, then lets do it.

joshk:
Do you want me to hold off merging?

joshk:
@lawik ^

lawik (approved):
Don't have the bandwidth to tweak it. Can definitely ship for now.

nshoes (approved):

PR #1869, opened 2025-02-03
Handle cases where, upon connection, we know the device is already updating

joshk:
This fixes issues where socket reconnections cause us to mark a device as having updated several times.

Thankfully we have been receiving this information from Link for around 4 years (thanks to @jjcarstens), we just haven't been using it.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -27,7 +27,13 @@ defmodule NervesHubWeb.DeviceChannel do
       {:ok, device} ->
         send(self(), {:after_join, params})
 
-        {:ok, assign(socket, :device, device)}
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -27,7 +27,13 @@ defmodule NervesHubWeb.DeviceChannel do
       {:ok, device} ->
         send(self(), {:after_join, params})
 
-        {:ok, assign(socket, :device, device)}
+        socket =
+          socket
+          |> assign(:currently_downloading_uuid, params["currently_downloading_uuid"])
+          |> assign(:update_started?, !!params["currently_downloading_uuid"])
```
lawik, line 33: spicy `!!`

PR #1870, opened 2025-02-03
Remove a remnant of the old `DeviceLink`

joshk:


nshoes (approved):
Curious what the original intentions were here. Maybe to hook into `handle_in`s without changes web-side?

joshk:
> Curious what the original intentions were here. Maybe to hook into `handle_in`s without changes web-side?

🤷

PR #1878, opened 2025-02-04
Make sure all health reports are deleted when a device is destroyed

joshk:
This fixes a DB constraint error


nshoes (approved):
AND A TEST!!!!!!!! :shipit:

PR #1881, opened 2025-02-05
Support restricting device connections to the web endpoint

joshk:
This allows installations to disable shared secret device connections to web endpoints.

This is ideal for cases where you are running a devices endpoint and want to make sure all connections go there.

### DIFF
--- a/lib/nerves_hub_web/channels/device_socket.ex
+++ b/lib/nerves_hub_web/channels/device_socket.ex
@@ -92,11 +92,12 @@ defmodule NervesHubWeb.DeviceSocket do
 
   # Used by Devices connecting with HMAC Shared Secrets
   @decorate with_span("Channels.DeviceSocket.connect")
-  def connect(_params, socket, %{x_headers: x_headers})
+  def connect(_params, socket, %{x_headers: x_headers} = connect_info)
       when is_list(x_headers) and length(x_headers) > 0 do
     headers = Map.new(x_headers)
 
-    with :ok <- check_shared_secret_enabled(),
+    with :ok <- check_source_enabled(connect_info[:source]),
+         :ok <- check_shared_secret_enabled(),
          {:ok, key, salt, verification_opts} <- decode_from_headers(headers),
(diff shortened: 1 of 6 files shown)


lib/nerves_hub_web/helpers/websocket_connection_error.ex
```diff
@@ -1,12 +1,19 @@
 defmodule NervesHub.Helpers.WebsocketConnectionError do
   import Plug.Conn
 
-  @message "no certificate pair or shared secrets connection settings were provided"
+  @no_auth_message "no certificate pair or shared secrets connection settings were provided"
+  @check_uri_message "incorrect uri used, please contact support"
```
lawik, line 5: can we get some configured URL info into this message so we can actually tell them the fix?

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -188,6 +198,14 @@ defmodule NervesHubWeb.DeviceSocket do
     end
   end
 
+  defp check_source_enabled(source) do
```
lawik, line 201: I wonder if we should have an in-between mode for the transition that just throws a "wrong-endpoint" key in the device connection metadata or similar. And then we could show it to people.

Though we have few enough people that will be hit by it that we can essentially tell them and give them a deadline?

lib/nerves_hub_web/helpers/websocket_connection_error.ex
joshk: Sure. Although, just so you know, slipstream doesn't show this message by default, which is a pity

lib/nerves_hub_web/channels/device_socket.ex
joshk: Yeah, I wouldn't want to add too much extra complexity to this. Instead we can just send out a warning to current users, check our logs to make sure people are off, and then turn it off. I have no intention to turn this off yet.

nshoes:
@joshk why would anyone be trying to connect devices to the web endpoint in the first place? I'm confused 😖

PR #1895, opened 2025-02-09
Cleanly close sockets in tests

joshk:
This PR covers a few test related bits I discovered:

- The SSL tests should use `tmp_dir` for cert related fixtures

- Fixed a websocket SSL cert test by making sure it attempted to connect 
Once I added the `wait_connect` I confirmed the failure was happening. 

- Fix an issue where `on_disconnect` was being called twice
I saw this when debugging the tests, but I've also been seen in production. I believe it's related to channels being told to disconnect, and if a socket has two channels, we see two disconnection messages. Or it could be Phoenix/Bandit sending it twice for some reason? Either way, this tracks when it was first called and storing it in the assigns.

- and the main issue I was working towards.... cleanly close sockets in the websocket tests
This closes sockets, and then checks for connection change events and that the DB has been updated accordingly.
I've also added a check to ensure the device has fully come online. I've seen some DB error messages where connections were closed while Ecto was in the device registration process.
This fixes all the db messages related to Ecto parent processes we see when running tests.

This PR should eliminate output like https://github.com/nerves-hub/nerves_hub_web/actions/runs/13210884330/job/36883932735#step:15:17

### DIFF
--- a/lib/nerves_hub_web/channels/device_socket.ex
+++ b/lib/nerves_hub_web/channels/device_socket.ex
@@ -30,9 +30,9 @@ defmodule NervesHubWeb.DeviceSocket do
 
   @impl Phoenix.Socket.Transport
   @decorate with_span("Channels.DeviceSocket.terminate")
-  def terminate(reason, {_channels_info, socket} = state) do
-    on_disconnect(reason, socket)
-    super(reason, state)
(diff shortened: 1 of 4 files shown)


nshoes (approved):
Great changes, thank you! 💯

PR #1897, opened 2025-02-11
Pinned devices

elinol:
Implements Pinned Devices feature.

- Migration, schema and context for pinned devices
- Listing of pinned devices in "dashboard" page (orgs)
- Pin/unpin event in show device liveview
- Automatic sorting on pinned devices in device list view (pinned devices first)

Pinned devices are cleared when:
- device is moved to other organization which user don't have access to
- device is soft deleted
- user is removed from organisation
- user account is soft deleted


elinol:
image

image

image

elinol:
Not that good.. :S
image

elinol:
Expandable listing:
image

elinol:
@joshk I've added pinned devices as default ordering (always on top, if in filtered list), but you preferred it as a filter if I remember correctly?

otherwise I think this is ready for review.
@nshoes and @lawik ping :D

joshk (changes requested):
Hey @elinol 

This is really interesting.

Here are some thoughts after talking this through with @nshoes.

- I'm not in favour of pinned devices changing the order of devices. While I can understand the reasoning, for me it would be annoying to always have pinned devices at the top even though I wouldn't be searching for a pinned device. I would remove this for now.
- I also don't see any filter for just showing pinned devices, wasn't this something we agreed on?
- If we are going to include the pinned list on the 'Dashboard' (Org and Product list) page then I would like to see it as a tab instead of at the top of at the bottom of the page. It would also mean a little think about the 'empty state', and what happens if there are hundreds of pinned devices. 

More than happy to talk this through on a Zoom

lawik:
I am fine with dropping the sorting thing. I think that idea came from me, I find it interesting and hopefully useful but don't feel strongly that it should stay. It does muddle clarity a bit so I'm fine to not do it.

Filter for Pinned would be good to have, yeah :)

I really feel like a tab is worse. Right now the starting page is kind of pointless, aside from some basic navigation. This would give you a shortcut back to whatever you are working on. Needing to go one tab over just feels like shoving it in a drawer to me. People wouldn't discover it and it loses the convenience of being on the landing.

Even when that view becomes something more like a dashboard/overview we'll probably want to guide the user back to their work in a fast way. Pinned devices for a developer would probably mean their development device while for a support person it would likely mean the device I'm currently investigating a ticket for.

I would suggest hiding the box if there are no devices, you could show a "No pinned devices yet.." type of message but I'd say just hide it.

The list shows a few devices and expands on request to show more already, so it should deal decently well with a large number of devices. We could cap it at max 50 or something since that's already sort of absurd. I don't think the experience of pinning a hundred devices needs to be fully baked for this to be really good.

joshk:
Had a good chat with Lars.

Lets go with the pinned devices list at the top of the page, and hide it if you have no pinned devices.

We'll aim to return to this in the coming month or so as we might want to move it to a tabbed view, we can see how feedback goes first.

joshk:
One thing I forgot to note down, I noticed that in the pinned devices list we don't subscribe to pubsub updates for the devices. The would be a good thing to add, which would have this behave similar to the product device list.

lib/nerves_hub_web/live/orgs/index-new.html.heex
```diff
@@ -12,6 +12,15 @@
...
   <div class="w-[800px]">
+    <PinnedDevices.render
+      :if={!Enum.empty?(@pinned_devices)}
+      devices={if @show_all_pinned?, do: @pinned_devices, else: limit_devices(@pinned_devices)}
+      statuses={@device_statuses}
+      show_all?={@show_all_pinned?}
+      device_limit={@device_limit}
+      total_count={length(@pinned_devices)}
+    />
+    <%!-- <.live_component module={PinnedDevices} id="pinned_devices" devices={@pinned_devices} /> --%>
```
joshk, line 23: Can we remove this commented line?

joshk (approved):
Looks good. Happy to merge this and deploy it this evening. 
I've left one small comment.

lib/nerves_hub_web/live/orgs/index-new.html.heex
elinol: Done!

PR #1898, opened 2025-02-12
Switch to showing (and sorting) on a devices last connection `established_at` datetime

joshk:
This was some feedback from a NervesCloud user, and makes a lot of sense to me.

There is some code duplication which will be resolved when we start removing the old UI code.


elinol (approved):
🙌

nshoes (approved):

PR #1912, opened 2025-02-20
Ensure available deployments match selected devices architecture and platform when batch-updating

nshoes:
Previously, we displayed all deployments as options when batch-updating devices. This is confusing because you could see deployments that shouldn't be options. This PR now requires you to filter by platform first before having the ability to bulk move devices into a deployment. 

This change does create an extra step when moving devices to a deployment but it's important to note this is an interim feature before #1653 is finished.

### DIFF
--- a/lib/nerves_hub/deployments.ex
+++ b/lib/nerves_hub/deployments.ex
@@ -182,6 +182,16 @@ defmodule NervesHub.Deployments do
     |> Repo.one!()
   end
 
+  @spec get_by_product_and_platform(Product.t(), binary()) :: [Deployment.t()]
+  def get_by_product_and_platform(product, platform) do
+    Deployment
+    |> where(product_id: ^product.id)
+    |> join(:left, [d], f in assoc(d, :firmware))
+    |> where([_d, f], f.platform == ^platform)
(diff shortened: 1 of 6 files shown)


nshoes:
My only concern is when a device with no set firmware metadata is selected. Currently it's treated like a no match, no deployments will populate when it's selected. Instead, should it be allowed to go into any deployment?

jjcarstens (commented):
I think this works, but it seems a little complicated. What if we only show the deployment dropdown if the `platform` filter is show and a "please select platform to view available deployments" text or something if it is not picked? That might simplify how you find deployments?

nshoes:
@jjcarstens welp, that's a lot simpler 😅 I like that a lot more and will implement it.

nshoes:
@jjcarstens I'm not sure this is the change we want here. We go from being able to select any devices and move them to any products in the deployment to now ensuring you filter by platform first

nshoes:
@jjcarstens I implemented the change! You now have to filter by platform before moving devices to a deployment.

joshk (approved):
Looks good to me

PR #1914, opened 2025-02-21
Add deployment filter for devices

nshoes:


nshoes:
This change displays the deployment id in the URL when filtering a device's deployment, e.g. `/devices?deployment_id=2`. We normally other identifiers as slugs throughout the app, should we do the same here? Filter on `deployment.name` instead?

jjcarstens (approved):
If we are okay with deployment_id being a param in the path only filled by code, then we're good. I don't have strong opinions. But I also don't think we expose the deployment_id anywhere in the UI

PR #1918, opened 2025-02-21
Reinstate deployment version check when device connects

nshoes:
This behavior was removed in #1652 and was relied on more than expected.


joshk (approved):
❤️

PR #1923, opened 2025-02-26
Capture support script timeouts

joshk:
Timeouts are to be expected and shouldn't raise an error in Sentry.

This also adds a friendlier message to the UI

Screenshot 2025-02-26 at 2 32 31 PM

I also increased the gap between items.


lawik (approved):
Love it :)

PR #1937, opened 2025-03-03
Fix starting scripts from API endpoint

jjcarstens:
https://github.com/nerves-hub/nerves_hub_web/pull/1923 introduced a change to handle when scripts timeout waiting for data back. However, it missed adjusting for the change in `NervesHubWeb.API.ScriptController` so using the API is broken.

This fixes to match that new return value and adds a `504` response if it times out waiting for the script reply.

### DIFF
--- a/lib/nerves_hub_web/controllers/api/script_controller.ex
+++ b/lib/nerves_hub_web/controllers/api/script_controller.ex
@@ -31,30 +31,27 @@ defmodule NervesHubWeb.API.ScriptController do
   def send(conn, %{"identifier" => identifier, "id" => id}) do
     %{user: user} = conn.assigns
 
-    case Devices.get_by_identifier(identifier) do
-      {:ok, device} ->
-        if Accounts.has_org_role?(device.org, user, :view) do
-          case Scripts.get(device.product, id) do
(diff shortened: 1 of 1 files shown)


lib/nerves_hub_web/controllers/api/script_controller.ex
```diff
@@ -31,30 +31,27 @@ defmodule NervesHubWeb.API.ScriptController do
...
-          conn
-          |> put_resp_header("content-type", "application/json")
-          |> send_resp(403, Jason.encode!(%{status: "missing required role: read"}))
-        end
-
+    with {:ok, device} <- Devices.get_by_identifier(identifier),
+         true <- Accounts.has_org_role?(device.org, user, :view) || :not_allowed,
+         {:ok, command} <- Scripts.get(device.product, id),
+         {:ok, io} <- Scripts.Runner.send(device, command) do
+      text(conn, io)
```
jjcarstens, line 38: This is the magic sauce to fix things. The rest is refactor since the nesting was starting to get very deep

joshk:
Thank you! (and sorry)

joshk (approved):
❤️

PR #1946, opened 2025-03-06
Fix console drag-and-drop uploads

### LINKED ISSUE #1930: Drag-and-Drop files onto console no longer works
The functionality to be able to drop a file on the console and have it get transferred to the device doesn't seem to work in new or old UI.

Have't investigated yet. Frank reported and I confirmed that I had the same problem.

joshk:


assets/js/console.js
```diff
@@ -251,40 +251,42 @@ dropzone.addEventListener(
   e => {
     e.preventDefault()
     e.stopPropagation()
-    e.dataTransfer.items.forEach(item => {
-      const file = item.getAsFile()
-      const reader = file.stream().getReader()
+    if (e.dataTransfer.items) {
```
nshoes, line 254: Any ideas on what caused this to break?

nshoes (approved):

assets/js/console.js
joshk: From what I could tell, its that `e.dataTransfer.items.forEach` is the culprit. I don't think `forEach` is a method on `items`.

nshoes: Ah, makes sense! Thanks.

PR #1947, opened 2025-03-06
Show users currently watching device page

elinol:
Adds Phoenix Presence, to application in general and device page specifically.

Tracks user on mount, subscribes to events for device's presence and populates user list on presence diff.

Todo:
- Better styling on user circles

### DIFF
--- a/lib/nerves_hub/application.ex
+++ b/lib/nerves_hub/application.ex
@@ -40,7 +40,8 @@ defmodule NervesHub.Application do
           {Phoenix.PubSub, name: NervesHub.PubSub},
           {Cluster.Supervisor, [topologies]},
           {Task.Supervisor, name: NervesHub.TaskSupervisor},
-          {Oban, Application.fetch_env!(:nerves_hub, Oban)}
+          {Oban, Application.fetch_env!(:nerves_hub, Oban)},
+          NervesHubWeb.Presence
         ] ++
         deployments_orchestrator(deploy_env()) ++
         endpoints(deploy_env())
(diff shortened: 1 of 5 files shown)


elinol:
image

joshk:
This is looking very cool!

I was reading up on `Phoenix.Presence` myself, to better understand all the bits that were going on, and came across https://hexdocs.pm/phoenix/presence.html#usage-with-liveview

Your approach appears a bit more in line with the channels implementation recommendations. The thing that might be of interested is:

> One thing to keep in mind when dealing with LiveView, is that each LiveView is a stateful process, so if we keep the presence state in the LiveView, each LiveView process will contain the full list of online users in memory. Instead, we can keep track of the online users within the Presence process, and pass separate events to the LiveView, which can use a stream to update the online list.

elinol:
I read that too and that was my first approach but I didn't got it working alright. Might be worth giving it another try though!

elinol:
@joshk Streams is working fine, but I haven't yet find at way to get the list position of each presence, to be able to set z-index dynamically. `Enum.with_index` cannot be used on streams. Is it worth spending more time or better to just go with the assigns solution, WDYT?

joshk:
Hey @elinol , sorry for taking so long to help out with this.

I've tweaked the css so that we don't need to use z indexes.

If we can add a test or two for this I'd say its ready to be merged in.

joshk:
I had some spare time so added a test or two.

nshoes (approved):
LGTM! Great work Elin, and thank you for the tests Josh :)

PR #1951, opened 2025-03-10
Filter devices without deployment

nshoes:
I think it'd be useful to filter devices without deployment.

### DIFF
--- a/lib/nerves_hub_web/live/devices/index-new.html.heex
+++ b/lib/nerves_hub_web/live/devices/index-new.html.heex
@@ -335,6 +335,7 @@
             <label class="sidebar-label" for="deployment_id">Deployment</label>
             <select class="sidebar-select" name="deployment_id" id="deployment_id">
               <option {selected?(@current_filters, :deployment_id, "")} value="">All</option>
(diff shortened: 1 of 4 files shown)


lib/nerves_hub_web/live/devices/index-new.html.heex
```diff
@@ -335,6 +335,7 @@
             <label class="sidebar-label" for="deployment_id">Deployment</label>
             <select class="sidebar-select" name="deployment_id" id="deployment_id">
               <option {selected?(@current_filters, :deployment_id, "")} value="">All</option>
+              <option {selected?(@current_filters, :deployment_id, nil)} value="-1">No Deployment</option>
```
nshoes, line 338: Note the `value` field; this implementation is strange at first glance. We use Ecto to cast the filters before calling `Devices.Filtering`, but I needed a type that could help in representing three states: ignore `deployment_id`, a `nil` `deployment_id`, and search by `deployment_id`. I'd love to hear any suggestions.

joshk (approved):
Awesome!

joshk:
Y U NO MERGE?

nshoes:
It was a busy morning and I wanted to get tests in 😈

PR #1952, opened 2025-03-11
Correctly show device updates status

joshk:
Fixes an issue reported via Slack where the penalty box status was incorrect on the device list page.

I also added a tooltip to show how long the device is in the penalty box, aaaaaand I've also made the entire row clickable.

Oh, and I also made the device identifier and firmware version use `font-mono`

Screenshot 2025-03-11 at 7 17 40 PM
Screenshot 2025-03-11 at 7 17 00 PM

### DIFF
--- a/lib/nerves_hub_web/components/device_update_status.ex
+++ b/lib/nerves_hub_web/components/device_update_status.ex
@@ -7,19 +7,25 @@ defmodule NervesHubWeb.Components.DeviceUpdateStatus do
     cond do
       Devices.device_in_penalty_box?(device) ->
         ~H"""
-        <svg title="Device in the penalty box" class="size-4 stroke-amber-500" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none">
-          <path
-            d="M19 14V5C17.5 5.16667 14 5 12 3C11.4286 3.57143 10.7347 3.9932 10 4.30029M5 5V14C5 18 12 21 12 21C12 21 15.2039 19.6269 17.2766 17.5M3 3L21 21"
-            stroke-width="1.2"
-            stroke-linecap="round"
-            stroke-linejoin="round"
-          />
-        </svg>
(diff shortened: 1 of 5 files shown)


nshoes (approved):

lib/nerves_hub_web/components/device_update_status.ex
```diff
@@ -0,0 +1,91 @@
...
+          <div class="tooltip-content hidden w-max absolute top-0 left-0 z-20 text-xs px-2 py-1.5 rounded border border-[#3F3F46] bg-base-900 flex">
+            Updates disabled
+            <div class="tooltip-arrow absolute w-2 h-2 border-[#3F3F46] bg-base-900 origin-center rotate-45"></div>
+          </div>
+        </div>
+        """
+    end
+  end
+
+  def friendly_blocked_until(blocked_until) do
```
nshoes, line 58: This is a bit scary :) Any reason you went a custom route over `Timex`?
```
iex(3)> date = DateTime.utc_now() |> DateTime.add(-12, :minute)
~U[2025-03-11 15:44:50.343102Z]
iex(4)> Timex.from_now(date)
"12 minutes ago"
```

joshk: I wanted to use `Timex.from_now` but I found that with a date in the future, you got something like:
```elixir
Timex.from_now(blocked_until) #=> "in 12 hours"
```
and I can't for the life of me figure out how to drop the `in`

joshk: Additionally, I think the results can be a bit padded and vague.  For a date an hour and a half away:

```elixir
Timex.from_now(blocked_until) #=> "in 1 hour"
```

I'm interested in switching back, I just want to get rid of the `in`

nshoes: My code snippet may just hold the answer you're looking for :)

joshk:
After thinking about it, I've decided to keep this custom logic for now as it does provide flexibility on how we want to display 'blocked until' dates and times, especially when the blocked until is over 24 hours.

PR #1953, opened 2025-03-11
Update Elixir, OTP, Ubuntu, and other CI bits

jjcarstens:
The primary objective was to get to `noble-20250127` which has some security improvements. This required also bumping Elixir and OTP.

`fwup` was updated to the latest and `actions/cache@v4` was needed because the [older versions are deprecated](https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/)


nshoes (approved):
Sick! Thanks for this.

joshk (approved):
<3

PR #1962, opened 2025-03-11
Add the deployment version constraint to the deployments list

### LINKED ISSUE #1955: New UI: While viewing deployments, version condition isn't displayed as a column
It was in the old UI and is a useful bit of info, it should be added back.

joshk:
Closes #1955


nshoes (approved):
:shipit:

PR #1964, opened 2025-03-12
Backport device_connection indexes, sort when deleting them so index is used

nshoes:
This PR backports two indexes on `device_connections` we've been using in production. I'm also adding an `order_by` when deleting `device_connections` so an index is used.


nshoes:
@esvinson 👀

joshk (approved):
👍

PR #1965, opened 2025-03-13
Improve firmware update button on the Device show page

joshk:
If the Device isn't connected to a Deployment:
image

And use toggle buttons for the update status:
https://github.com/user-attachments/assets/73ae63fa-0b69-4539-a072-23c16c72d363

Also, I've updated the logic when the penalty box is cleared so that the `updates_enabled` field is set to `true`


nshoes (approved):
It's beautiful 😻

PR #1969, opened 2025-03-17
Migrate old Deployment audit logs to reference DeploymentGroup

nshoes:
Audit logs persist the Elixir module name of the resource it is auditing. In https://github.com/nerves-hub/nerves_hub_web/pull/1653, I renamed `Deployment` -> `DeploymentGroup` and attempting to load audit logs for a deployment [will blow up](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/types.ex#L70) because that module doesn't exist anymore.

This PR migrates old references in `audit_logs.resource_type` to reference the new module. Tested locally.


joshk:
Awesome!

Can you also add a similar update all for the `actor_type` column

joshk (approved):
Awesome, thank you!

Issue #1958, opened 2025-03-11
Manual Deployment Recollect


nshoes:
Removing deployment device recalculations hurt this specific workflow:

- Devices ready for deployment get tags set
- Deployment's tags are updated to device's tags
- Devices in deployment are recalculated

I believe this is a reasonable workflow and we should provide the user to _recollect_ devices for a deployment manually. When viewing a deployment, there should be a button that says "Recollect Devices". This will query for devices with:

- no deployment
- matching platform
- matching architecture
- exact matching tags

If any devices are found, we should show the user a confirm dialog that displays how many devices will be affected before updating.

joshk:
From a Zoom call with @nshoes 

Add to the Deployment show page, in the 'device matching conditions' section, along with links to view these devices in the devices list page:

- A count of how many devices in the deployment **do** match these conditions 
- A count of how many devices **don't** match these conditions
- A count of how many devices **not in the deployment** match these conditions

Provide quick actions to:

- Remove devices which no longer match these conditions
- Add devices which aren't in the deployment which do match these conditions

joshk:
🎉 Completed! 🎉

PR #1970, opened 2025-03-18
Manual Deployment Recollect

nshoes:
This work adds the ability to add and remove devices from a deployment group. All devices considered are first filtered by product and the deployment group's architecture and platform requirements. When a deployment has tags and no version, devices with at least one matching tag or considered. When a deployment has a version and no tags, only the version is considered when matching devices. When a deployment has both tags and a version, we combine the logic.

I'm content with the UI design as a first pass, but there could certainly be some improvements around alignment and mobile-readiness. 

### Empty state
CleanShot 2025-03-31 at 07 01 29@2x

### All devices in deployment group match
CleanShot 2025-03-31 at 06 44 37@2x

### Devices outside deployment group match
CleanShot 2025-03-31 at 06 43 32@2x

### Devices inside deployment group don't match
CleanShot 2025-03-31 at 06 58 01@2x

### All UI bits at once
CleanShot 2025-03-31 at 06 58 45@2x

### Moving tooltip
CleanShot 2025-03-31 at 11 30 26@2x

### Removing tooltip
CleanShot 2025-03-31 at 11 29 17@2x

### Moving 200k devices
CleanShot 2025-03-31 at 08 10 24

### Removing 200k devices
CleanShot 2025-03-31 at 11 21 44

## Other
My only concern is that the query to move 200k devices in SmartRent's QA takes ~14 seconds. I'm considering batching the update by 100k so we don't have to worry about timeouts. Looking for reviewers' thoughts.

Solves #1958.


lib/nerves_hub/devices.ex
```diff
@@ -798,14 +798,6 @@ defmodule NervesHub.Devices do
     Map.put(device, :deployment_group, deployment_group)
   end
 
-  @spec update_many_deployment_groups([non_neg_integer()], DeploymentGroup.t()) ::
-          {integer(), nil | [term()]}
-  def update_many_deployment_groups(device_ids, deployment_group) do
-    Device
-    |> where([d], d.id in ^device_ids)
-    |> Repo.update_all(set: [deployment_id: deployment_group.id])
-  end
```
nshoes, line 807: Unused function.

lib/nerves_hub_web/live/deployment_groups/show.html.heex
```diff
@@ -19,7 +19,7 @@
       aria-label="Delete"
       data-confirm={[
         "Are you sure you want to delete this deployment?",
-        @deployment_group.device_count > 0 && " All devices assigned to this deployment will be assigned a new deployment when they reconnect. ",
+        (@deployment_group.device_count > 0 && " All devices assigned to this deployment will be assigned a new deployment when they reconnect. ") || [],
```
nshoes, line 22: `Phoenix.HTML` will blow up if this list has a boolean.

lib/nerves_hub_web/components/deployment_group_page/settings.ex
```diff
@@ -284,7 +284,7 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Settings do
               aria-label="Delete"
               data-confirm={[
                 "Are you sure you want to delete this deployment group?",
-                @deployment_group.device_count > 0 && " All devices assigned to this deployment group will be assigned a new deployment when they reconnect. ",
+                (@deployment_group.device_count > 0 && " All devices assigned to this deployment group will be assigned a new deployment when they reconnect. ") || [],
```
nshoes, line 287: `Phoenix.HTML` will blow up if this list has a boolean.

priv/repo/migrations/20250326023630_index_device_firmware_metadata_architecture_and_platform.exs
```diff
@@ -0,0 +1,10 @@
+defmodule NervesHub.Repo.Migrations.IndexDeviceFirmwareMetadataArchitectureAndPlatform do
```
nshoes, line 1: @esvinson 👀

joshk:
Really awesome work on this feature!

I love how the UI has turned out, contextual, clear, and like it was always designed to be there.

I wouldn't be worried about the timeout right now, if anything I think we could do a follow up PR to change the toast message to a full screen overlay with a "please wait" message, as any navigation away from this page will cause the async to be cancelled (from my understanding).

I'll review the PR today.

lib/nerves_hub/helpers/logging.ex
joshk, line 1: Love the changes to the sentry logging bits!

lib/nerves_hub_web/components/deployment_group_page/settings.ex
joshk: Great catch!

lib/nerves_hub_web/components/deployment_group_page/summary.ex
```diff
@@ -156,17 +156,104 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Summary do
...
+                >
+                  <svg class="size-5 mr-1" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+                    <path
+                      d="M15 7H9M15 7H18M15 7C15 5.34315 13.6569 4 12 4C10.3431 4 9 5.34315 9 7M9 7H6M4 7H6M6 7V18C6 19.1046 6.89543 20 8 20H16C17.1046 20 18 19.1046 18 18V7M18 7H20M10 11V16M14 16V11"
+                      stroke-width="1.2"
+                      stroke-linecap="round"
+                      stroke-linejoin="round"
+                      stroke="#A1A1AA"
+                    />
+                  </svg>
```
joshk, line 195: When possible, its nice to add these icons to the icons component. This comment shouldn't hold up a merge.

lib/nerves_hub_web/live/deployment_groups/show-new.html.heex
```diff
@@ -101,6 +101,9 @@
     org={@org}
     org_user={@org_user}
     user={@user}
+    matched_device_count={@matched_device_count}
+    unmatched_device_count={@unmatched_device_count}
+    matched_devices_outside_deployment_group_count={@matched_devices_outside_deployment_group_count}
```
joshk, line 106: I'd suggest we don't pass the assigns along but instead move this logic into the live component, unless there was a reason you preferred this way?

lib/nerves_hub_web/live/deployment_groups/show.ex
```diff
@@ -1,12 +1,17 @@
 defmodule NervesHubWeb.Live.DeploymentGroups.Show do
   use NervesHubWeb, :updated_live_view
 
+  import Ecto.Query
```
joshk, line 4: I'm not a fan of importing `Ecto.Query` into liveview controllers.

lib/nerves_hub_web/live/deployment_groups/show.html.heex
joshk: Great catch

priv/repo/migrations/20250326023630_index_device_firmware_metadata_architecture_and_platform.exs
joshk: Looooove these

joshk (approved):
Loooooove this PR.

Two things which could warrant some changes, but also fine with us following up on them later:
1. `import Ecto.Query` in the live view controller
2. the match queries in the show controller instead of the live component

joshk:
Also, love the attention to the test suite!

priv/repo/migrations/20250326023630_index_device_firmware_metadata_architecture_and_platform.exs
esvinson: These are low cardinality fields, so there's a good chance these indexes won't actually be used for queries.

joshk:
> > as any navigation away from this page will cause the async to be cancelled (from my understanding)
> 
> Oh? I need to test this. I assumed it would still do the work and just the result wouldn't get back to the client.

I believe that isn't the case as its a process attached to the live view, and if the live view is shut down (navigated away from) then the async process will also be killed.

lib/nerves_hub_web/live/deployment_groups/show-new.html.heex
nshoes: There wasn't, I just wasn't thinking through the options and grabbed the first one that came to mind :) However, since we're moving away from LiveComponents, I should probably keep this implementation now?

joshk: Good idea!

lib/nerves_hub_web/components/deployment_group_page/summary.ex
nshoes: Done!

lib/nerves_hub_web/live/deployment_groups/show.ex
nshoes: Me neither, I moved work to the Devices context.

priv/repo/migrations/20250326023630_index_device_firmware_metadata_architecture_and_platform.exs
nshoes: Hm, I checked against QA, matching the cardinality of these two fields in production as well as the same number of devices. I swore this index worked when `EXPLAIN`ing... I'll test again in the morning to double-check.

joshk: I think these are worth having for now and seeing if they are effective based on real-world datasets.

PR #1973, opened 2025-03-22
Cache the `product_id` in the `DeviceConnection`

joshk:
This is required for improved query perf when polling for the count of connected devices grouped by `product_id`.

The way to do such a query, before these changes, required a join, and that itself just added more for the query planner to take care of.

This work is a required part of an Insights dashboard I'm working on.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -56,6 +56,7 @@ config :nerves_hub, Oban,
   notifier: Oban.Notifiers.PG,
   log: false,
   queues: [
+    analytics: 3,
     delete_archive: 1,
     delete_firmware: 1,
     device: 1,
@@ -73,7 +74,8 @@ config :nerves_hub, Oban,
        {"*/1 * * * *", NervesHub.Workers.CleanStaleDeviceConnections},
        {"1,16,31,46 * * * *", NervesHub.Workers.DeleteOldDeviceConnections},
        {"*/5 * * * *", NervesHub.Workers.ExpireInflightUpdates},
-       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation}
+       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation},
+       {"*/1 * * * *", NervesHub.Workers.ConnectedDeviceCount}
      ]}
   ]
 
--- a/config/config.exs
+++ b/config/config.exs
@@ -56,7 +56,6 @@ config :nerves_hub, Oban,
   notifier: Oban.Notifiers.PG,
   log: false,
   queues: [
-    analytics: 3,
     delete_archive: 1,
     delete_firmware: 1,
     device: 1,
@@ -74,8 +73,7 @@ config :nerves_hub, Oban,
        {"*/1 * * * *", NervesHub.Workers.CleanStaleDeviceConnections},
        {"1,16,31,46 * * * *", NervesHub.Workers.DeleteOldDeviceConnections},
        {"*/5 * * * *", NervesHub.Workers.ExpireInflightUpdates},
-       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation},
-       {"*/1 * * * *", NervesHub.Workers.ConnectedDeviceCount}
+       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation}
      ]}
   ]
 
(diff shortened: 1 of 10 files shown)


priv/repo/migrations/20250322003835_add_product_id_to_device_connections_table.exs
```diff
@@ -0,0 +1,11 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:device_connections) do
+      add(:product_id, :integer)
+    end
+
+    create(index(:device_connections, [:product_id]))
+  end
+end
```
nshoes, line 11:
```suggestion
defmodule NervesHub.Repo.Migrations.AddProductIdToDeviceConnectionsTable do
  use Ecto.Migration
  
  @disable_ddl_transaction true
  @disable_migration_lock true

  def change do
    alter table(:device_connections) do
      add(:product_id, :integer)
    end

    create(index(:device_connections, [:product_id], concurrently: true))
  end
end
```
@esvinson anything else to add here?

config/config.exs
```diff
@@ -73,7 +74,8 @@ config :nerves_hub, Oban,
        {"*/1 * * * *", NervesHub.Workers.CleanStaleDeviceConnections},
        {"1,16,31,46 * * * *", NervesHub.Workers.DeleteOldDeviceConnections},
        {"*/5 * * * *", NervesHub.Workers.ExpireInflightUpdates},
-       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation}
+       {"*/15 * * * *", NervesHub.Workers.DeviceHealthTruncation},
+       {"*/1 * * * *", NervesHub.Workers.ConnectedDeviceCount}
```
nshoes, line 78: Cruft? I don't see this worker defined anywhere else but here.

priv/repo/migrations/20250322003835_add_product_id_to_device_connections_table.exs
esvinson: I'd recommend doing the index in a separate migration with the `disable_ddl_transaction/disable_migration_lock` in that migration so we aren't disabling the transaction for the table schema modification.

priv/repo/migrations/20250322003835_add_product_id_to_device_connections_table.exs
```diff
@@ -0,0 +1,11 @@
+defmodule NervesHub.Repo.Migrations.AddProductIdToDeviceConnectionsTable do
+  use Ecto.Migration
+
+  def change do
+    alter table(:device_connections) do
+      add(:product_id, :integer)
```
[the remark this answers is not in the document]
nshoes: Done!

priv/repo/migrations/20250322003835_add_product_id_to_device_connections_table.exs
nshoes: Done!

config/config.exs
esvinson: Also as a side note, we already get connected counts from statsd metrics.
https://github.com/nerves-hub/nerves_hub_web/blob/4fdb5722cb880bffaa6139612f34a98baffbe72a/lib/nerves_hub/statsd_metrics_reporter.ex#L28

Unless that's going to be for some reconciliation with database vs the nodes themselves.

priv/repo/migrations/20250324172254_index_device_connections_product_id.exs
```diff
@@ -0,0 +1,10 @@
+defmodule NervesHub.Repo.Migrations.IndexDeviceConnectionsProductId do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change do
+    create(index(:device_connections, [:product_id], concurrently: true))
```
esvinson, line 8: Migrations look good now 👍  Only other potential optimization might be adding a conditional for `WHERE product_id IS NOT NULL` on the index if we're going to have a lot of rows with a null product_id, if that's not the case then it won't really matter in 30 days (or whatever the retention period on this table is)

config/config.exs
joshk: @nshoes Oops, that is a worker I'm working on, but I let the config slip in.

@esvinson thats for pointing this out. The `Devices.Registry` is likely to disappear soon, as we won't need it for deployments, and by removing it we enable the ability to support a 'device proxy' (new platform updates won't affect device connections) or MQTT. Also, this work is a precursor to an 'insights dashboard'.

priv/repo/migrations/20250324172254_index_device_connections_product_id.exs
joshk: All devices have a product, so I think we are fine with not having this where clause.

joshk:
@nshoes Thank you for the code contributions!

nshoes (approved):
🚀

config/config.exs
esvinson: This is an important observability metric, so we want to make sure it doesn't get broken.

PR #1977, opened 2025-03-24
Add a theater mode, pseudo-fullscreen to console

lawik:
Un-theatered:
Screenshot_20250324_133304

Theatered:
Screenshot_20250324_132858

It could be good to have product name or some more breadcrumbs in there. But we'll see :)


lib/nerves_hub_web/components/device_page/console.ex
```diff
@@ -28,6 +29,30 @@ defmodule NervesHubWeb.Components.DevicePage.Console do
     |> ok()
   end
 
+  def toggle_fullscreen(js \\ %JS{}) do
+    js
+    |> JS.toggle_class("relative", to: "#dropzone")
```
nshoes, line 34: TIL you can do this, very cool.

nshoes (approved):
I love this, thanks Lars 😺

lib/nerves_hub_web/components/device_page/console.ex
```diff
@@ -28,6 +29,30 @@ defmodule NervesHubWeb.Components.DevicePage.Console do
...
+    |> JS.toggle_class("fixed", to: "#dropzone")
+    |> JS.toggle_class("top-0", to: "#dropzone")
+    |> JS.toggle_class("left-0", to: "#dropzone")
+    |> JS.toggle_class("right-0", to: "#dropzone")
+    |> JS.toggle_class("bottom-0", to: "#dropzone")
+    |> JS.toggle_class("w-screen", to: "#dropzone")
+    |> JS.toggle_class("h-screen", to: "#dropzone")
+    |> JS.toggle_class("p-6", to: "#dropzone")
```
elinol, line 42: Is there a reason you don't use `|> JS.toggle_class("fixed top-0 left-0 etc", to: "#dropzone")`?

lib/nerves_hub_web/components/device_page/console.ex
lawik: Ah, you hadn't see LiveView.JS yet? It is quite nice and doesn't require a roundtrip to the server.

lib/nerves_hub_web/components/device_page/console.ex
lawik: It wasn't obvious to me from the docs if I could do that. Also, I don't know what happens to the various things. Some are supposed to toggle on and off. Can try it :D

joshk:
This is very cool. I love its direction.

I'd suggest we add the 'full screen' icon to the top right, and keep it there for when you want to minimize. 

It would also be good to have a bar across the top, add the device id to the left (instead of the right), and we can add the bubbles of users also in the console to the right, next to the toggle button.

I'm also cool with us merging this in and addressing UI tweaks as we go.

lawik:
Good ideas all. I will tweak the locations, then we can add things in a new iteration.

PR #1978, opened 2025-03-24
Filter deployment groups by name

elinol:
Adds a search input that enables filtering on deployment group name (New UI only).
Added filtering mechanisms which makes it easy to add more filters in the future, much like on devices view.

### DIFF
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -10,6 +10,7 @@ defmodule NervesHub.ManagedDeployments do
   alias NervesHub.Devices.Device
   alias NervesHub.ManagedDeployments.DeploymentGroup
   alias NervesHub.ManagedDeployments.Distributed.Orchestrator, as: DistributedOrchestrator
+  alias NervesHub.ManagedDeployments.Filtering
   alias NervesHub.ManagedDeployments.InflightDeploymentCheck
   alias NervesHub.Products.Product
   alias NervesHub.Workers.FirmwareDeltaBuilder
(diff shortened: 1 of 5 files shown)


joshk:
Could you please add a screenshot to the PR?

lib/nerves_hub/managed_deployments.ex
```diff
@@ -58,12 +59,31 @@ defmodule NervesHub.ManagedDeployments do
...
     |> where([d], d.product_id == ^product.id)
+    |> build_filters(filters)
     |> sort_deployment_groups(sort_opts)
     |> preload([_d, _dev, f], firmware: f)
     |> select_merge([_f, dev], %{device_count: dev.device_count})
     |> Flop.run(flop)
   end
 
+  # If filter amount increases the filtering functions can be moved to their own module,
+  # like device filtering.
```
nshoes, line 70: I'm for doing this now; I think it's more confusing that we would have a module for device filtering and not deployment filtering, regardless of how many lines of code it takes. What do you think?

test/nerves_hub_web/live/deployment_groups/index_test.exs
```diff
@@ -55,4 +55,26 @@ defmodule NervesHubWeb.Live.DeploymentGroups.IndexTest do
     |> assert_has("a", text: deployment_group.name)
     |> assert_has("td div", text: "0")
   end
+
+  # Disabled until new UI is the default template.
```
nshoes, line 59: What's stopping us from testing the new UI now? I meant to ask Josh this recently but I keep forgetting 🤣 Since we don't have a clear date on when it will be the default I'd rather start testing it now. Note we don't necessarily have to implement the ability here, just want to discuss.

cc @joshk

joshk: This test doesn't need to be disabled, we can still test the new UI.

joshk: Sorry, that might of come off a little terse.

I'm in full support of us testing the new UI. And would also support focusing our testing attention on it as well.

elinol: It was easy peasy to test it with ` conn |> put_session("new_ui", true) |> visit..` etc. Will add it in a upcoming commit

elinol:
image

nshoes (approved):
Awesome! Thank you 🚀

joshk:
Awesome work @elinol

PR #1985, opened 2025-03-25
Make sure 'skip queue' firmware update message are sent to devices

joshk:
I introduced some poor logic which impacted 'skip queue' calls.

The message was only sent to devices if they were using the new distributed orchestrator.

This fixes the issue by leaving the message sending to `told_to_update`, and making sure the pid of the device channel is passed through if the old message passing pattern is required.

Oh, also, the message that was being sent wasn't actually doing anything, so this cleans that up.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -1420,13 +1420,19 @@ defmodule NervesHub.Devices do
   @doc """
   Deployment orchestrator told a device to update
   """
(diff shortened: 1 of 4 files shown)


joshk:
@fhunleth this is the fix for the 'skip the queue' bug you found the other day. We should be able to merge and deploy it this week.

nshoes (approved):
🚀 🚀 🚀 🚀

PR #1990, opened 2025-03-27
Reduce some of the work the heartbeat needs to do

joshk:
We don't need the join and return, we have the data readily available.


nshoes:
Awesome! Thank you 🫡

PR #1992, opened 2025-03-27
Incorporate a random jitter into the device heartbeat

joshk:
If a horde of devices connects at the same time, this will help distribute the heartbeat updates, and the jitter will be incorporated on every check, further randomizing the spread.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -28,7 +28,9 @@ config :nerves_hub,
   device_deployment_change_jitter_seconds:
     String.to_integer(System.get_env("DEVICE_DEPLOYMENT_CHANGE_JITTER_SECONDS", "10")),
   device_last_seen_update_interval_minutes:
-    String.to_integer(System.get_env("DEVICE_LAST_SEEN_UPDATE_INTERVAL_MINUTES", "5")),
+    String.to_integer(System.get_env("DEVICE_LAST_SEEN_UPDATE_INTERVAL_MINUTES", "15")),
+  device_last_seen_update_interval_jitter:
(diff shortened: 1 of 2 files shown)


nshoes (approved):
🚀 🚀 🚀 🚀

fhunleth:
When the jitter is measured in minutes, does this have much of a smoothing effect?

joshk:
for SmartRent, I would be suggesting an update interval of 60 mins with a jitter of 15. That jitter is -15..15, so its a 30min distribution. Plus, this is going to best help hordes of devices which connect at the same time, eg. after a deploy. This should both reduce the load of the db for tracking when a device last checked in, and spread it more effectively.
At least that is the intention.

fhunleth:
What I'm talking about is that the random number generator used here only picks the numbers -5, -4, -3, -2, -1, 0, 1, 2, 3, 4, 5. Picking one out of 11 numbers just didn't seem like much smoothing to me. 

I don't know what point it matters on the granularity, but since most things are specified in milliseconds, picking a random number between -300,000 and 300,000 ms (+/- 5 minutes) would do more.

joshk:
Ahhhh well, the thing is that we aren't running a timer, we are instead looking into the Phoenix heartbeat messages and checking how long ago the last one was. Both the Phoenix js and Slipstream send heartbeats every 30 seconds (default), unless there has been other activity during that period. 

From my understanding, switching to a millisecond config wouldn't give us anything extra. We could look at switching to a seconds based config instead, but all of this is built around the websocket heartbeat messages.

PR #1993, opened 2025-03-28
Use seconds for the heartbeat interval jitter

joshk:


nshoes (approved):
Good catch! No one wants to set jitter in minutes 😆

PR #1996, opened 2025-03-28
Improve the support script output

joshk:
image

### DIFF
--- a/assets/ui-rework/hooks/supportScriptOutput.js
+++ b/assets/ui-rework/hooks/supportScriptOutput.js
@@ -1,0 +1,55 @@
+import { Terminal } from "xterm"
+import { WebglAddon } from "xterm-addon-webgl"
+import { WebLinksAddon } from "xterm-addon-web-links"
+import { FitAddon } from "xterm-addon-fit"
+
+const defaultTermOptions = {
+  disableStdin: true,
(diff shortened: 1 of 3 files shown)


fhunleth:
Yay!!!

fhunleth (approved):

assets/ui-rework/hooks/supportScriptOutput.js
```diff
@@ -0,0 +1,74 @@
...
+    term.writeln(content.split("\n").join("\n\r"))
+
+    fitAddon.fit()
+
+    // resize terminal on window resize
+    window.addEventListener(
+      "resize",
+      debounce(() => {
+        resizeContent.apply(null, [term, fitAddon])
+      }, 300)
```
nshoes, line 71:
```suggestion
      debounce(resizeContent.bind(null, term, fitAddon), 300)
```
No need to create an anonymous function and call `resizeContent`, this saves a call.

nshoes (approved):
One tiny nit :)

assets/ui-rework/hooks/supportScriptOutput.js
joshk: removed this entirely and added some simplier row setting logic

PR #2000, opened 2025-03-29
Fix how device health is updated in the UI

joshk:
A bug was discovered where support script output was being cleared without the clear button being pressed.

A deep dive found two core issues:
1. the latest health report subscriptions were wrong and weren't being sent through to the details live component
2. anytime the device updated, which would happen with heartbeats or health updates, the details page `update/2` would be called, which would rerender the live component, clearing the support script.
 
I also found a follow-up issue: the health live component was, well, built incorrectly (by me). Instead of the parent liveview sending data through, it was trying to subscribe to broadcasts, and thus not getting any updates. The reason it would update is because the device was updated, which propagated to the health live component, and whenever that updated it would call `update/2` on the live component, refreshing all the data.

This PR addresses these issues, and implements some best practices in regards to assigns for live components.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -41,6 +41,26 @@ defmodule NervesHub.Devices do
     Repo.get(Device, device_id)
   end
 
+  def get_complete_device(device_id) do
+    Device
+    |> where([d], d.id == ^device_id)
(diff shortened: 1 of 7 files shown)


lawik (approved):

nshoes (approved):
Lovely! Thank you for taking the time to fix and polish this.

PR #2001, opened 2025-03-31
Fix date comparison when checking certificate status.

CJRChang:
Currently certificate_status displays as expired on valid certificates.


joshk:
Thanks for the fix!

joshk:
✨

PR #2012, opened 2025-04-01
Generate a dynamic livebook based on encoded params

joshk:
This is an experiment towards making it easier to connect to a device using Livebook.

I could see this being its own app, but since its just one dynamic template I'd prefer to keep it here, at least to start with.

This uses a `livebook.` subdomain for the route, and the node and cookie for the script are base64 encoded in the url.

eg. https://livebook.nervescloud.com/eyJub2RlIjoibmVydmVzQDEwMC4xMDguMTA0Ljg4IiwiY29va2llIjoiOHQ2UU9JNjNiNk0yc1E3Tm5sdzdJSzBoZ294a2RubUFST21MWURuSGpSdyJ9/nerves.livemd

### DIFF
--- a/lib/nerves_hub_web/controllers/livebook_template_controller.ex
+++ b/lib/nerves_hub_web/controllers/livebook_template_controller.ex
@@ -1,0 +1,70 @@
+defmodule NervesHubWeb.LivebookTemplateController do
+  use NervesHubWeb, :controller
+
+  def generate(conn, params) do
+    decoded =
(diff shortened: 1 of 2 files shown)


lawik:
I am good with it but will let @nshoes and @jjcarstens weigh in :)

joshk:
I think I'm going to pull this out and put it into its own app, at least for now :)

joshk:
I'm going to close this for the time being.

PR #2013, opened 2025-04-02
Remove the old orchestrator

joshk:
The new distributed orchestrator is the approved path forward, its time to remove the old and unused.


lib/nerves_hub/application.ex
```diff
@@ -28,7 +28,6 @@ defmodule NervesHub.Application do
         {Ecto.Migrator,
          repos: Application.fetch_env!(:nerves_hub, :ecto_repos),
          skip: Application.get_env(:nerves_hub, :database_auto_migrator) != true},
-        {Registry, keys: :unique, name: NervesHub.Devices.Registry},
```
nshoes, line 31: Goodbye old friend 🫡

priv/repo/migrations/20250213215925_set_default_deployment_orchestrator_strategy.exs
```diff
@@ -5,6 +5,6 @@ defmodule NervesHub.Repo.Migrations.SetDefaultDeploymentOrchestratorStrategy do
   alias NervesHub.Repo
 
   def change do
-    Repo.update_all(DeploymentGroup, set: [orchestrator_strategy: "distributed"])
+    repo().query!("UPDATE deployments SET orchestrator_strategy = 'distributed'")
```
nshoes, line 8: We just ran this in prod this morning :)

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -58,17 +58,15 @@ defmodule NervesHubWeb.DeviceChannel do
...
 
-    send(self(), :device_registration)
-
     socket =
       socket
       |> assign(:device, device)
       |> assign(:deployment_channel, deployment_channel)
       |> assign_api_version(params)
-      |> assign(:penalty_timer, nil)
-      |> maybe_start_penalty_timer()
```
nshoes, line 69: We're removing this because it only has to do with keeping `device_payload.updates_enabled` up to date in the registry, correct? I want to make sure I understand.

joshk: correct, although I think there is some touch up work we can do around the penalty box, like having a cleanup oban worker.

PR #2014, opened 2025-04-02
Improved function naming for toggling automatic updates

joshk:
This is a follow on from #2010 where the `phx-click` naming wasn't clear, as well as having duplicates.


nshoes (approved):
Great, thanks for following up with this.

PR #2015, opened 2025-04-02
Only include specified device in presence stream

elinol:
Changes topic for presence stream, from general "online-users" to "device-IDENTIFIER".

Also includes a fix for z-index on stacked user avatars.


joshk (approved):
Great work!

PR #2018, opened 2025-04-02
Expand new UI device tests

nshoes:


lib/nerves_hub_web/components/device_page/details.ex
```diff
@@ -328,7 +328,9 @@ defmodule NervesHubWeb.Components.DevicePage.Details do
           <div :if={is_nil(@device.deployment_group) && Enum.any?(@deployment_groups)} class="flex p-4 gap-4 items-center border-t border-zinc-700">
             <form phx-target={@myself} phx-submit="set-deployment-group" class="flex gap-2 items-center w-full">
               <div class="grow grid grid-cols-1">
+                <label for="deployment_group" class="hidden">Deployment Group</label>
```
nshoes, line 331: I'm vehemnetly against tests forcing us to write application code to support them, but:

1. This is technically invalid markup without a `label` and
2. I could not figure out how to make this work with `phoenix_test`'s `select/2`

lib/nerves_hub_web/components/device_page/details.ex
```diff
@@ -373,8 +375,9 @@ defmodule NervesHubWeb.Components.DevicePage.Details do
           </div>
 
           <div class="flex p-4 gap-4 items-center border-t border-zinc-700">
-            <form :if={Enum.any?(@firmwares)} phx-target={@myself} phx-submit="push-update" class="flex gap-2 items-center w-full">
+            <form :if={Enum.any?(@firmwares)} id="push-update-form" phx-target={@myself} phx-submit="push-update" class="flex gap-2 items-center w-full">
```
nshoes, line 378: In `phoenix_test`'s current state you can't have two forms in the same document without an id, it'll throw an error. I dove into the code enough to figure that out, but don't feel pressed to add the feature or ask if it's possible.

I'm ok with strengthening our semantic markup, even if the reason is our test suite.

lib/nerves_hub_web/components/device_page/details.ex
joshk: funny, and annoying, and good catch

lib/nerves_hub_web/components/device_page/details.ex
joshk: 👍

joshk (approved):
❤️ ❤️ ❤️

joshk:
@nshoes merge whenever you are ready

PR #2021, opened 2025-04-02
New UI: Display first connecting code properly

nshoes:
The key `:first_connect_code` doesn't exist on the device-changeset-to-form map. The name of the device struct's column, `:connecting_code`, must be used.


joshk (approved):
Ooooooops

Great catch!

PR #2022, opened 2025-04-03
Default to sorting deployment groups by name while viewing DeploymentGroups.Index

nshoes:
@esvinson pointed this out: defaulting to deployment group name descending feels unnatural as a default sort order. `name asc` feels much better.

### DIFF
--- a/lib/nerves_hub_web/live/deployment_groups/index.ex
+++ b/lib/nerves_hub_web/live/deployment_groups/index.ex
@@ -115,7 +115,7 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Index do
       page: pagination_opts["page_number"],
       page_size: pagination_opts["page_size"],
       sort: pagination_opts["sort"] || "name",
-      sort_direction: pagination_opts["sort_direction"],
+      sort_direction: pagination_opts["sort_direction"] || "asc",
       filters: filters
     }
 


lib/nerves_hub_web/live/deployment_groups/index.ex
```diff
@@ -22,13 +22,6 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Index do
     deployment_groups = ManagedDeployments.get_deployment_groups_by_product(product)
     counts = ManagedDeployments.get_device_counts_by_product(product)
 
-    deployment_groups =
-      deployment_groups
-      |> Enum.sort_by(& &1.name)
-      |> Enum.group_by(fn deployment_group ->
-        deployment_group.firmware.platform
-      end)
```
nshoes, line 30: This isn't needed because deployment group order is handled in `assign_deployment_groups_with_pagination/1`, called in `handle_params/3`. The template then uses `socket.assigns.entries` as the collection of deployment groups to iterate over. No need to do any work on ordering before we hand it to `flop`.

nshoes: Actually, this is how the old UI renders deployment groups 😅 We still need this.

joshk (approved):
❤️

PR #2024, opened 2025-04-04
A revamp of our user token storage and management

joshk:
Heavily inspired by the Phoenix auth generators, including the API token guide. This improves many aspects of how we store and use tokens, as well as generating session tokens so we can delete them if a user's account is compromised.

Key improvements:

- session and api tokens are entirely random and don't include any components which can be guessed
- store session tokens, allowing for sessions to be cancelled/deleted by a user or admin
- add a 'context' to the user token so we can track different types of tokens issues

While there are some functions which I have included but aren't used, these will be used in the coming weeks as I update the login UI.


priv/repo/migrations/20250404022742_change_user_token_column_type.exs
jjcarstens, line 1: This migration does many things at once which don't allow a smooth migration path. I think the delete/rotate should be manual to allow coordination of new and old tokens without service interruption.

I also don't fully understand the benefit of moving from string to binary? I'm not necessarily opposed except that this current migration is forcing me to delete and recreate everything at once. The user tokens table is so small that it might make sense to rename the column, recreate it as expected, and then copy all the existing data over and drop the renamed column.

lib/nerves_hub_web/plugs/api/user.ex
```diff
@@ -39,17 +39,20 @@ defmodule NervesHubWeb.API.Plugs.User do
     end
   end
 
-  defp token_auth(<<"nh", _u, "_", hmac::30-bytes, crc_bin::6-bytes>> = token) do
```
jjcarstens, line 42: I think we should keep the crc check. It is a simple enough filter against garbage.

priv/repo/migrations/20250204100906_create_pinned_devices.exs
```diff
@@ -2,11 +2,11 @@ defmodule NervesHub.Repo.Migrations.CreatePinnedDevices do
   use Ecto.Migration
 
   def change do
-      create table(:pinned_devices) do
-        add :user_id, references(:users, on_delete: :delete_all)
-        add :device_id, references(:devices, on_delete: :delete_all)
```
jjcarstens, line 7: I am guessing you have some formatter tool which fixed formatting here?

lib/nerves_hub/accounts/user_token.ex
```diff
@@ -1,43 +1,137 @@
...
+  For example, the non-hashed token can be sent to the user's email while the
+  hashed part is stored in the database. The original token cannot be reconstructed,
+  which means anyone with read-only access to the database cannot directly use
+  the token in the application to gain access.
+  """
+  def build_hashed_token(user, context, note) do
+    token = :crypto.strong_rand_bytes(@rand_size)
+    hashed_token = :crypto.hash(@hash_algorithm, token)
+
+    friendly_token = "nhu_" <> Base.url_encode64(token, padding: false)
```
jjcarstens, line 60: Why have we switched to base64? That adds more character sets which may need URL encoded and/or not easily copiable?

jjcarstens (commented):
Added a few thoughts. There is a lot here and I think I need help differentiating between what is actually used/important here or not, and what has been brought in from these auth suggestions and/or guides. In fact, I would vote the unused stuff not be included and come next when you intend to use it

lib/nerves_hub/accounts/user_token.ex
joshk: `url_encode64` is [safe for URLs:](https://hexdocs.pm/elixir/1.18.2/Base.html#url_encode64/2)

> Encodes a binary string into a base 64 encoded string with URL and filename safe alphabet.

joshk: After discussing this with Jon, and reading the GitHub blog a few extra times, I've switched over to Base 62

priv/repo/migrations/20250204100906_create_pinned_devices.exs
joshk: Oh, sorry, that was me, I didn't mean to include this, I can remove it again if you want?

lib/nerves_hub_web/plugs/api/user.ex
joshk: CRC check added! 🎉

lib/nerves_hub/accounts/user_token.ex
```diff
@@ -71,6 +71,16 @@ defmodule NervesHub.Accounts.UserToken do
...
 
+  defp base62_safe_token() do
+    case :crypto.strong_rand_bytes(@rand_size) do
+      # Base62 is an integer based calculation and cannot
+      # deal with leading null bytes since they are ignored
+      # so we generate another one to avoid that problem
+      <<0, _::binary>> -> base62_safe_token()
+      token -> token
+    end
+  end
```
joshk, line 82: This is so simple and beautiful

nshoes:
Ok, I've come up with a workable solution to support both token types. It's not the prettiest, but I think it's okay for the interim.

I'm unsure on the migration, though. I think all we need is to add the `context` column with a default of `"api"` and the index. Looking for others' thoughts.

priv/repo/migrations/20250404022742_change_user_token_column_type.exs
```diff
@@ -0,0 +1,12 @@
+defmodule NervesHub.Repo.Migrations.ChangeUserTokenColumnType do
+  use Ecto.Migration
+
+  def change do
+    alter table(:user_tokens) do
+      modify(:token, :binary)
```
joshk, line 6: I don't think this is going to work.

https://github.com/nerves-hub/nerves_hub_web/actions/runs/14371081171/job/40294136739?pr=2024#step:14:1291

After thinking about this more, I don't think this can be reversible because if we are going from binary to string we will need to use some form of base encoding (from my understanding).

nshoes: Idea has been implemented. We'll just rename `token` -> `old_token` and add `token` as a binary type. The query for the validate old token path now references `old_token`. Downside is that the new `token` column can't have a `null: false` validation until we remove the code and column referencing the old tokens.

priv/repo/migrations/20250404022742_change_user_token_column_type.exs
```diff
@@ -2,8 +2,10 @@ defmodule NervesHub.Repo.Migrations.ChangeUserTokenColumnType do
   use Ecto.Migration
 
   def change do
+    rename table(:user_tokens), :token, to: :old_token
+
     alter table(:user_tokens) do
-      modify(:token, :binary)
+      add(:token, :binary)
       add(:context, :string, null: false, default: "api")
```
joshk, line 9: does this apply the default to existing rows? or will we also need an 'update all'?

nshoes: Good catch! Thank you and fixed.

priv/repo/migrations/20250404022742_change_user_token_column_type.exs
```diff
@@ -1,14 +1,22 @@
...
     rename table(:user_tokens), :token, to: :old_token
 
     alter table(:user_tokens) do
       add(:token, :binary)
       add(:context, :string, null: false, default: "api")
     end
 
+    flush()
+
+    NervesHub.Repo.update_all("user_tokens", [set: [context: "api"]])
```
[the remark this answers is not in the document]
nshoes: Oh that's right, thanks for reminding me of that syntax, I'll update.

What do you mean by "confirm"? I was assuming the `flush()` was enough.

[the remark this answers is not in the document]
joshk: Sorry, this isn't related to the flush. (from my understanding) If in the future the context field is removed from the schema then this migration will fail for new full runs of the migration as `update_all` will check if `context` is part of the `UserToken` schema definition.

nshoes (approved):

Issue #1971, opened 2025-03-20
Cannot click on anything on the new UI devices


gBillal:
**Describe the bug**
So when I land in the devices page in the new UI. I don't get any focus to click the menu at the left or any of the top buttons (Export, Add Device, Filters). If I click anywhere on the page, I land on one of the devices details page. From there I can click on the left menu.
Btw scrolling also doesn't work.

*This only happens with Safari*

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.3.2
 - Browser Safari
 - Version 18.3.1

**Additional context**
This happened either after my software update or something changed on your side.

joshk:
Thank you for reporting this @gBillal 

I'll start diving in and see where we introduced these bugs.

nshoes:
@gBillal the fix has been merged! Thank you for reporting this.

PR #2025, opened 2025-04-05
Support device index table row links in webkit

nshoes:
Webkit doesn't support `display: relative` on `<tr />`'s. That means that our absolute-positioned table row mask, which allows the entire table row to be clickable, takes up the entire page 😖 Using a mix of `clip-path` and `transform: translate()` is enough to tell Webkit to keep the bounds in the table row.

#### Webkit pre-fix
CleanShot 2025-04-05 at 07 35 56

#### Webkit post-fix
CleanShot 2025-04-05 at 07 33 06

#### Chromium post-fix
CleanShot 2025-04-05 at 07 31 26

### DIFF
--- a/assets/ui-rework/app.css
+++ b/assets/ui-rework/app.css
@@ -253,6 +253,19 @@
     font-weight: 500;
     line-height: 20px;
   }
+
+  .clickable-table-row {
+    position: relative;
+    transform: translate(0);
+    clip-path: inset(0);
+  }
+
+  .clickable-table-row-mask {
+    position: absolute;
+    inset: 0;
+    z-index: 10;
+    max-height: 52px;
+  }
 }
 
 html {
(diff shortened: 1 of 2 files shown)


nshoes:
Fixes #1971

assets/ui-rework/app.css
```diff
@@ -253,6 +253,19 @@
...
+    position: relative;
+    transform: translate(0);
+    clip-path: inset(0);
+  }
+
+  .clickable-table-row-mask {
+    position: absolute;
+    inset: 0;
+    z-index: 10;
+    max-height: 52px;
```
nshoes, line 267: This is the worst part about this fix, but I think it's workable for a table row: their height isn't dynamic.

joshk (approved):
This. Is. Amazing.

Thank you @nshoes !

I had been meaning to get to this but kept putting it aside as I couldn't find a solutions. 

You are a legend!

PR #2026, opened 2025-04-07
Use hooks instead of LiveComponents

joshk:
When I built the tabs for the Device page, and then used a similar pattern for the Deployment page, I quickly learnt that LiveComponents have a lot of quicks which lead to confusing code and duplicated state between the parent and child component.

For example, LiveComponents can't subscribe to topic, listen for pubsub messages, or respond to timer events.

I recently discovered a better path forward which uses LiveView hooks : https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#attach_hook/4

I've created some reusable code for implementing the same pattern with the Deployment page, and the upcoming Org pages.


lib/nerves_hub_web.ex
```diff
@@ -262,6 +297,94 @@ defmodule NervesHubWeb do
...
+          allow_upload: 3,
+          attach_hook: 4,
+          detach_hook: 3,
+          push_patch: 2,
+          push_event: 3,
+          push_navigate: 2,
+          start_async: 3,
+          connected?: 1,
+          consume_uploaded_entry: 3
+        ]
```
nshoes, line 322: Why the heavy use of the `only` option here?

lib/nerves_hub_web.ex
```diff
@@ -126,6 +126,41 @@ defmodule NervesHubWeb do
         |> Enum.filter(fn x -> !is_nil(params[to_string(x)]) end)
         |> Enum.into(%{}, fn x -> {x, params[to_string(x)]} end)
       end
+
+      defp setup_tab_components(socket, tabs \\ []) do
+        if socket.assigns[:new_ui] do
+          Enum.reduce(tabs, socket, fn component, socket ->
+            component.connect(socket)
+          end)
+          |> put_private(:tabs, tabs)
```
nshoes, line 135: Why not in `assigns`?

lib/nerves_hub_web.ex
```diff
@@ -126,6 +126,41 @@ defmodule NervesHubWeb do
...
+        else
+          socket
+        end
+      end
+
+      defp update_tab_component_hooks(socket) do
+        if socket.assigns[:new_ui] do
+          socket
+          |> detach_hooks()
+          |> attach_hooks()
```
nshoes, line 145: Why attach and detach every time? Initial thought is to attach hooks only for the components that have rendered, then detach once we navigate away from the parent LV. Thoughts?

nshoes (commented):
LGTM, a couple of thoughts

lib/nerves_hub_web.ex
```diff
@@ -262,6 +297,94 @@ defmodule NervesHubWeb do
...
+        |> assign(:sidebar_tab, tab)
+        |> assign(:tab_hint, tab)
+      end
+
+      def send_toast(socket, kind, msg) do
+        NervesHubWeb.LiveToast.send_toast(kind, msg)
+        socket
+      end
+
+      @tab_id unquote(tab_id)
```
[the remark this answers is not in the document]
joshk: I was. And after thinking about this more, I have an idea that I'll try out this morning.

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -76,16 +48,19 @@ defmodule NervesHubWeb.Components.DevicePage.Details do
...
-  defp assign_deployment_groups(%{assigns: %{product: product}} = socket),
-    do:
-      assign(socket,
-        deployment_groups: ManagedDeployments.get_deployment_groups_by_product(product)
-      )
+  defp assign_deployment_groups(%{assigns: %{product: product}} = socket) do
+    assign(socket,
+      deployment_groups: ManagedDeployments.get_deployment_groups_by_product(product)
+    )
+  end
```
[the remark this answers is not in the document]
joshk: Hahaha, well, it wasn't a one liner in the end 😛
I think I can shave a line off my version, which also got formatted to be the same number of lines.

lib/nerves_hub_web.ex
```diff
@@ -262,6 +297,94 @@ defmodule NervesHubWeb do
...
+
+      def connect(socket) do
+        attach_hook(socket, tab_hook_id(), :handle_params, &__MODULE__.hooked_params/3)
+      end
+
+      def attach_hooks(%{assigns: %{tab: tab}} = socket) when tab == @tab_id do
+        socket
+        |> attach_hook(tab_hook_id(), :handle_async, &__MODULE__.hooked_async/3)
+        |> attach_hook(tab_hook_id(), :handle_event, &__MODULE__.hooked_event/3)
+        |> attach_hook(tab_hook_id(), :handle_info, &__MODULE__.hooked_info/2)
```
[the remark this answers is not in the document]
joshk: We are using `handle_info` in some of the tabs. You are right that `handle_info` isn't supported in LiveComponents, but this isn't LiveComponents 😄

lib/nerves_hub_web.ex
joshk: You are right that we might not need to attach and detach each time. I went this route because there are one or two instances of tab components listening for the same events. I liked the cleanliness of knowing that a tab was only attached when it was active. Happy to discuss more.

lib/nerves_hub_web.ex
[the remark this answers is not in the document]
nshoes: Right! Good point.

lib/nerves_hub_web.ex
nshoes: Talked over Zoom, all good here.

nshoes (approved):

PR #2027, opened 2025-04-07
Fix moving devices to deployment group while viewing devices index

nshoes:
I missed this naming mistake in recent work: `target-deployment` vs `target-deployment-group`. The old UI was tested around this, the new UI was not. I added a test for the new UI.


test/nerves_hub_web/live/devices/index_test.exs
```diff
@@ -339,11 +339,10 @@ defmodule NervesHubWeb.Live.Devices.IndexTest do
...
       |> assert_has("span", text: "2 selected")
-      |> unwrap(fn view ->
-        render_change(view, "target-deployment-group", %{
-          "deployment_group" => to_string(deployment_group.id)
-        })
-      end)
+      |> select("Move device(s) to deployment group:",
+        option: deployment_group.name,
+        exact_option: false
+      )
```
nshoes, line 345: Although this wasn't the core issue, this is a good time to point out how dangerous `unwrap` is when testing HTML -> LV interactions. We should lean towards not using it unless we have to.

lib/nerves_hub_web/live/devices/index-new.html.heex
```diff
@@ -152,7 +152,7 @@
                 </span>
                 <.link navigate={~p"/org/#{@org.name}/#{@product.name}/devices/#{device.identifier}"} class="ff-m font-mono">
                   {device.identifier}
-                  <span class="clickable-table-row-mask" />
+                  <span class="absolute inset-0 z-10"></span>
```
joshk, line 155: didn't you only just add this in recent PR?

test/nerves_hub_web/live/devices/index_test.exs
joshk: Tell me more.

lib/nerves_hub_web/live/devices/index-new.html.heex
nshoes: yep :) fixed

test/nerves_hub_web/live/devices/index_test.exs
nshoes: Talked on Zoom about this. For posterity:
- `unwrap` does not interact with any markup
- if there are mistakes in your markup, tests will not catch it
- we should favor moving to more semantic markup vs using an escape hatch to test LV code

joshk (approved):

PR #2029, opened 2025-04-08
Remove `live_toast` and just use standard flashes

joshk:
`LiveToast` wasn't giving us anything extra, and instead added more complexity and indirection.

This PR moves us back to standard flashes, but with some customizations around auto closing and ease in transitions.


lib/nerves_hub_web/components/deployment_group_page/settings.ex
```diff
@@ -29,7 +29,7 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Settings do
   def render(assigns) do
     ~H"""
     <div class="flex flex-col items-start justify-between gap-4 p-6">
-      <.form for={@form} class="w-full flex flex-col gap-4" phx-submit="update-deployment-group" phx-target={@myself}>
+      <.form for={@form} class="w-full flex flex-col gap-4" phx-submit="update-deployment-group">
```
nshoes, line 32: Removing this broke updating deployment groups in the new UI. I'm leaning towards not including any of these `phx-target` changes here and handling them in a separate PR. That way we can isolate the changes as well as write tests.

If you want to remove the changes here I'm happy to follow-up with a PR that implements the changes + tests.

joshk: Ahhhhhh, this should not have slipped in. I'll make sure the deployment group tabs aren't changed. Thanks for spotting this.

joshk: Fixed! Thanks for noticing this 🎉

nshoes (approved):

PR #2030, opened 2025-04-09
Remove `phoenix_view` elements from the API

joshk:
This could almost be classed as bike-shedding.

I was reviewing the API code and had trouble following along with some of its structure. This is mostly because of how Phoenix has changed its recommendations, including the use of `phoenix_view`.

This PR includes the following changes:

- Remove the use of `phoenix_view` from the API
- Assert error rendering in the tests
- Prefer the use of the `FallbackController` and `ErrorJSON` flow instead of custom responses in controllers
- Use the validation plug instead of custom checks in each controller action
- Use 422 status codes (not 500) for validation errors
- Clean up preloads (more to be done in this area)
- Removed the API health controller, the use of `/status/alive` is the recommended approach

**Breaking change:**

The Device `/move` API used to require the params `product_name` and `org_name`, but these conflicted with the nested URL params. They are now `new_org_name` and `new_product_name`

~~There were a bunch of device related API routes which weren't using the standard nesting.~~
~~I've removed this as we had a double-up of the possible actions, as it made the controller code more confusing and oddly built.~~
~~I don't personally love how we do URLs, but I do think its best to keep what we have consistent.~~

After talking with @jjcarstens , I've added these device routes back, but also added tests for them and improved how fetching and authorization works.


nshoes (commented):
~~This generally LGTM. @joshk I rebased, can you look into the failing test?~~

Never mind, it was my fault, fixed.

nshoes (approved):

PR #2033, opened 2025-04-10
Empty state UI tweaks (plus a button fix)

joshk:
- Improve the Device Health tab empty state
- Hide the entire update section on the Details tab if there are no firmware
- Add an empty state for the Device activity tab
- Prefer `:if` over `<%= if ... %>`
- Fix the "Add Support Script" button
- Some minor piping improvements

### DIFF
--- a/lib/nerves_hub_web/components/device_page/activity_tab.ex
+++ b/lib/nerves_hub_web/components/device_page/activity_tab.ex
@@ -51,7 +51,12 @@ defmodule NervesHubWeb.Components.DevicePage.ActivityTab do
               </.link>
             </div>
           </div>
-          <div class="py-2 px-4 flex flex-col gap-1">
+          <div :if={Enum.empty?(@activity)} class="py-2 px-4 flex flex-col gap-1">
+            <div class="flex items-center justify-center p-14">
+              <span class="text-zinc-500 font-extralight">No audit logs found for the device.</span>
+            </div>
+          </div>
+          <div :if={Enum.any?(@activity)} class="py-2 px-4 flex flex-col gap-1">
             <div :for={entry <- @activity} class="flex items-center gap-6 h-16 p-2">
               <div class="flex items-center h-8 py-1 px-2 bg-zinc-800 border border-zinc-700 rounded-full">
                 <svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
--- a/lib/nerves_hub_web/components/device_page/health_tab.ex
+++ b/lib/nerves_hub_web/components/device_page/health_tab.ex
@@ -374,9 +374,7 @@ defmodule NervesHubWeb.Components.DevicePage.HealthTab do
   end
 
   defp custom_metrics(metrics) do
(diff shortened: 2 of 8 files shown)


lib/nerves_hub_web/components/device_page/health_tab.ex
```diff
@@ -378,8 +374,7 @@ defmodule NervesHubWeb.Components.DevicePage.HealthTab do
   end
 
   defp custom_metrics(metrics) do
-    metrics
-    |> Enum.reject(fn {key, _value} ->
+    Enum.reject(metrics, fn {key, _value} ->
```
nshoes, line 377: I'd do it, but you don't have to :P
```suggestion
    Enum.reject(metrics, & elem(&1, 1) in @manual_metrics)
```

lib/nerves_hub_web/live/support_scripts/index-new.html.heex
```diff
@@ -19,46 +19,43 @@
...
+
+          <td>
+            {Calendar.strftime(script.inserted_at, "%Y-%m-%d at %I:%M %p UTC")}
+          </td>
+        </tr>
+      </tbody>
+    </table>
+  </div>
+
+  <div :if={Enum.empty?(@scripts)} class="h-full pb-16 flex items-center justify-center">
```
nshoes, line 56: Can we add tests for these two `:if`s please? :)

lib/nerves_hub_web/components/device_page/activity_tab.ex
```diff
@@ -51,7 +51,12 @@ defmodule NervesHubWeb.Components.DevicePage.ActivityTab do
               </.link>
             </div>
           </div>
-          <div class="py-2 px-4 flex flex-col gap-1">
+          <div :if={Enum.empty?(@activity)} class="py-2 px-4 flex flex-col gap-1">
+            <div class="flex items-center justify-center p-14">
+              <span class="text-zinc-500 font-extralight">No audit logs found for the device.</span>
+            </div>
+          </div>
+          <div :if={Enum.any?(@activity)} class="py-2 px-4 flex flex-col gap-1">
```
nshoes, line 59: Tests for both of these `:if`s, sir? 🥺🥣

nshoes (approved):
LGTM! One tiny nit, two mildly large nits :)

lib/nerves_hub_web/components/device_page/health_tab.ex
joshk: Love it! I would commit your suggestion but you didn't include the two lines below.

lib/nerves_hub_web/live/support_scripts/index-new.html.heex
joshk: Sooooo there isn't anything to test here, this is just changing `<%= if`s to `:if={`

lib/nerves_hub_web/components/device_page/activity_tab.ex
joshk: I can add a test for this one, in fact, I'm adding updated New UI tests for this tab and the Support Scripts page.

PR #2035, opened 2025-04-10
Fixes found after a deploy

joshk:
After deploying to NervesCloud I noticed a bunch of Sentry errors (and one web inspector console error) related to some recent PRs.


nshoes (approved):
Thanks for this!

PR #2039, opened 2025-04-12
Remove unneeded :base62 lib 

jjcarstens:
This slightly adjusts the implementation so we no longer need to keep around the old `Base62` library that we no longer wish to use. The handling of the incoming token needed to be adjusted to go from string -> binary -> integer to get the CRC value needed for `:erlang.crc32/2` comparison

I've also included a suggestion commit which refactors this logic a bit to reduce the number of calls on this incoming binary preventing unnecessary work and making a clean breakpoint when fully deprecating later


joshk (approved):
This is 🔥 !

nshoes:
Thank you!!

PR #2040, opened 2025-04-13
Ensure User Token V1 continues to work for now

jjcarstens:
The previous adjustment to updating UserTokens (#2024) was broken for V1 tokens. Specifically, the code to mark the `last_used` field when used in an API request would only consider the format of the new tokens and request would fail even though the old token was successfully verified.

This changes to remove extra verification of the token format before marking last_used. Instead, it returns the `UserToken` along with the user in the verification query and then updates that record separately.

It also adds an API integration test to confirm requests with the old tokens will continue to work for now


joshk (approved):
Your code is ✨

PR #2042, opened 2025-04-15
Run dialyzer in GH actions, fix dialyzer warnings

nshoes:
This adds dialyzer as a step in CI. PLTs are cached by elixir and erlang version.

### DIFF
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -523,8 +523,8 @@ defmodule NervesHubWeb.Live.Devices.Show do
     socket = stream(socket, :presences, [])
 
     if connected?(socket) do
-      Presence.track_user(topic, user.id, %{name: user.name})
-      Presence.subscribe(topic)
+      _ = Presence.track_user(topic, user.id, %{name: user.name})
+      _ = Presence.subscribe(topic)
       stream(socket, :presences, Presence.list_online_users(topic))
     else
       socket
--- a/lib/nerves_hub_web/live/devices/show.ex
+++ b/lib/nerves_hub_web/live/devices/show.ex
@@ -523,6 +523,7 @@ defmodule NervesHubWeb.Live.Devices.Show do
     socket = stream(socket, :presences, [])
 
     if connected?(socket) do
+      # TODO: handle the unhappy-path for both calls to Presence
       _ = Presence.track_user(topic, user.id, %{name: user.name})
       _ = Presence.subscribe(topic)
       stream(socket, :presences, Presence.list_online_users(topic))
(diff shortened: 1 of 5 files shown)


nshoes:
It's working:
CleanShot 2025-04-15 at 09 45 27@2x

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -523,8 +523,8 @@ defmodule NervesHubWeb.Live.Devices.Show do
     socket = stream(socket, :presences, [])
 
     if connected?(socket) do
-      Presence.track_user(topic, user.id, %{name: user.name})
-      Presence.subscribe(topic)
+      _ = Presence.track_user(topic, user.id, %{name: user.name})
+      _ = Presence.subscribe(topic)
```
fhunleth, line 527: Isn't it kind of bad if these return errors?

nshoes: Absolutely. This code heavily relies on the happy path 😅 I don't think this should be fixed here, though. I'll create an issue.

fhunleth: How about putting a TODO comment? I'm cool with not fixing everything at once, but this doesn't really feel like fixing the Dialyzer warning. It's more about just ignoring the warning for now.

nshoes: Oops, I created an issue before you commented this :) https://github.com/nerves-hub/nerves_hub_web/issues/2045 I don't feel strongly about having a TODO comment in addition to the issue.

fhunleth: I feel strongly about having a comment justifying every `_ = ` since they're indications that it's ok to silently error. I find this code hard to review since it's not obvious that it's ok to silently error and writing a unit test to verify when that call fails seems hard. I usually think there's an API issue, like when Elixir's `Logger.*` functions could return error and literally no sane person would check for error on every Logger call. The called function probably should have raised a nice exception that would be easy to figure out from a Sentry report.

I haven't traced through the code here to see what the ramifications of this would be, but if this code only survives happy paths anyway, then `:ok = ` seems like a more appropriate fix. I could imagine that you don't want to figure that out so `_ = ` is the easiest solution. I really feel it needs a comment, though, to indicate whether the return value is being ignored on purpose (like you spent time and that's actually the right behavior), or it's being ignored to satisfy Dialyzer and sometime in the future it could be traced out to see what the correct failure behavior is.

nshoes: Sounds good. It is absolutely being ignored here to make dialyzer happy. Both of those functions call out to Phoenix functions that provide good error returns that can be easily handled.  I'll add the comment TODO here, thanks for expanding!

fhunleth (approved):

Issue #2043, opened 2025-04-15
Filter by platform not working


gBillal:
As the title says, filtering by `platform` is not working.

nshoes:
Confirmed this isn't working, thanks for reporting! Working on a fix.

nshoes:
Fixed!

PR #2044, opened 2025-04-15
Fix filtering by platform in new UI

nshoes:
Solves #2043. I changed the `name` of the platform select in #2027, which broke filtering from the filter drawer.


joshk (approved):
Great catch!

PR #2049, opened 2025-04-16
Fix incorrect Dialyzer caching

joshk:
Using the `_build` directory was causing errors with beam files being redefined.

The common practice is to use `/priv`

An example of one of the errors I was seeing: https://github.com/nerves-hub/nerves_hub_web/actions/runs/14485042748/job/40629027836#step:10:12


nshoes (approved):
Good catch, thanks for this!

PR #2050, opened 2025-04-17
Use `Phoenix.Param` for cleaner URL generation

joshk:
https://hexdocs.pm/phoenix/1.7.21/Phoenix.Param.html

This PR cleans up our URL generation by implementing `Phoenix.Param` for schemas that does use `id`s in the path segments.


nshoes (approved):
This is cool! LGTM

PR #2055, opened 2025-04-18
Show name of user when hovering over initials while viewing device

nshoes:
- Display the user's name while hovering over user initial discs
- Force cursor to default on hover; it was defaulting to the text cursor which felt terrible


joshk:
@nshoes did you try using the tooltips hooks we use in other places?

nshoes:
@elinol YOU ROCK!!! Thanks so much, I appreciate it.

CleanShot 2025-04-22 at 07 37 35

PR #2059, opened 2025-04-21
Add spell check

emkguts:
Add spell check to CI and correct typos.
Note: Spell check locally does not work with the current version of nodejs (Must be >=18). Not sure if we want to do anything about that.


config/runtime.exs
```diff
@@ -452,7 +452,7 @@ if host = System.get_env("STATSD_HOST") do
 end
 
 config :nerves_hub, :audit_logs,
-  enabled: System.get_env("TRUNATE_AUDIT_LOGS_ENABLED", "false") == "true",
+  enabled: System.get_env("TRUNCATE_AUDIT_LOGS_ENABLED", "false") == "true",
```
nshoes, line 455: It's wild how many times I've read this and it was wrong 😆

lib/nerves_hub_web/live/product/settings.ex
```diff
@@ -71,7 +71,7 @@ defmodule NervesHubWeb.Live.Product.Settings do
     |> noreply()
   end
 
-  def handle_event("delete-product", _parmas, socket) do
```
nshoes, line 74: 👩‍🍳

nshoes (approved):
Thanks for this! 🚀  I'll poke around a Node update in a separate PR.

PR #2062, opened 2025-04-23
Add filtering sidebar component

elinol:
Moves sidebar filtering into component.


lib/nerves_hub_web/live/deployment_groups/index-new.html.heex
```diff
@@ -146,4 +149,30 @@
...
+      architecture: %{
+        label: "Architecture",
+        type: :select,
+        values: [{"All", ""} | Enum.map(@architectures, &{if(&1, do: &1, else: "Unknown"), &1})]
+      }
+    }
+  }
+  on_toggle="toggle-filters"
+  on_update="update-filters"
+  on_reset="reset-filters"
```
joshk, line 175: These three options seem to be the same between usage. I'd suggest we set a default value for them and thus they don't need to be specified each time.

lib/nerves_hub_web/live/deployment_groups/index-new.html.heex
```diff
@@ -146,4 +149,30 @@
...
+        type: :select,
+        values: [{"All", ""} | Enum.map(@platforms, &{if(&1, do: &1, else: "Unknown"), &1})]
+      },
+      architecture: %{
+        label: "Architecture",
+        type: :select,
+        values: [{"All", ""} | Enum.map(@architectures, &{if(&1, do: &1, else: "Unknown"), &1})]
+      }
+    }
+  }
```
joshk, line 172: While these filter options aren't extreme, the one used on the device list is massive and complex and an eye sore.

Instead, I'd recommend we use slots, which would be cleaner and provide better documentation.

- https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-slots
- https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#module-slot-attributes
- https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/components/core_components.ex#L559-L561
- https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/components/core_components.ex#L576

This is what it would possibly look like, which I think is a nicer API.

```elixir
<FilterSidebar.render show={@show_filters} current_filters={@current_filters}
  <:filter attr="name" label="Search" type={:text} />
  <:filter attr="platform" label="Platform" type={:select} values={[{"All", ""} | Enum.map(@platforms, &{if(&1, do: &1, else: "Unknown"), &1})]} />
  <:filter attr="architecture" label="Architecture" type={:select} values={[{"All", ""} | Enum.map(@architectures, &{if(&1, do: &1, else: "Unknown"), &1})]} />
</FilterSidebar.render >
```

joshk (changes requested):
This is a solid direction.

If we want to take this one step further you could use https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#attach_hook/4 and move the implementation of the `handle_event`s to the function module.  We can also add this later if preferred.

lib/nerves_hub_web/live/deployment_groups/index-new.html.heex
nshoes: 🤤 🤤 🤤 🤤

elinol: oh I like that!

elinol:
> Nice work, this is dope. Some thoughts after testing locally:
> 
> * I don't think having a search field inside and outside the filter drawer makes sense. I'd vote to remove the one outside, happy to hear others input.
> * In the filter drawer, we should be explicit with _what_ the search is... searching 😆 . Right now it's `name` for both devices and deployment groups, and the input label should reflect that.
> * Thoughts on adding an event listener for closing the filter drawer with the escape key? I believe it's a common UI pattern and I enjoy it.

I agree on the search input! If we keep it outside the filters I think it should be a search for everything and not just the name/identifier, typing rpi4 should give you all items on that platform for example. And inside the filters it should be labeled name or identifier like you said @nshoes 

I'm all for closing the drawer with esc or clicking outside or something, it frustrating to have to scroll the way back up when filtering devices.

elinol:
I added a more general search filter query to the "outside" search input, if you like it I'll write tests for it, otherwise I'll just revert the commit.
Note that the search input inside filter drawer is labeled `name` or `identifier` and will only filter on those.
@nshoes @joshk

lib/nerves_hub_web/live/devices/index-new.html.heex
```diff
@@ -323,175 +323,75 @@
...
-      </div>
-    </div>
-
-    <div class="flex shrink-0 justify-end h-16 px-4 py-4 mb-14 border-t border-zinc-700">
-      <button class="sidebar-button" type="button" phx-click="reset-filters">Reset Filters</button>
-    </div>
-  </div>
-</div>
+<FilterSidebar.render show={@show_filters} current_filters={@current_filters}>
+  <:filter attr="identifier" label="Device Identifier" type={:text} />
```
nshoes, line 327:
```suggestion
  <:filter attr="identifier" label="Identifier" type={:text} />
```

nshoes (approved):
Awesome! I love the general search, that's extremely helpful. One tiny nit from me and this is good to :shipit:

PR #2063, opened 2025-04-23
OrgUser API improvements

joshk:
I saw a Sentry error pop through the NervesCloud account late last week which was caused by the user trying to add a user to an org via the CLI.

The issue was that the CLI was asking for a username, but the API expects a users ID.

We don't expose the ID of the users in other related API/CLI calls, plus IDs aren't really friendly.

This PR:

- favors a users email for adding, removing, and updating their org membership
- added a new `/invite` endpoint
- and complete OpenAPI specs

This is a breaking change. In my opinion, I think the change is fine to merge, although thats just my opinion.

I have a CLI PR in the works which will use these updated routes.


lib/nerves_hub_web/controllers/api/fallback_controller.ex
```diff
@@ -28,6 +28,26 @@ defmodule NervesHubWeb.API.FallbackController do
...
+
+  def call(conn, {:error, :org_user_exists}) do
+    conn
+    |> put_status(422)
+    |> put_view(NervesHubWeb.API.ErrorJSON)
+    |> render(:"422", %{
+      reason:
+        "A user with that email address already exists, please use the add user api endpoint."
+    })
+  end
```
nshoes, line 49: I love how easy it is to add to this controller and move on. Nice and organized 💯

nshoes (approved):
LGTM. I'm fine with the breaking change as well :shipit:

lib/nerves_hub_web/controllers/api/fallback_controller.ex
joshk: Yeah, love this pattern!

PR #2066, opened 2025-04-23
Add new templates and LiveViews to build out the product and org UI

lawik:


lawik:
This is a start on Create Product and listing products under an org. Some of the missing UI.
It had a weird action-based structure so I made it work more like everything else.

The UI is very basic and I'm fine with nuking any and all of it.
I think the best path is to merge this and then make better layouts and UI on top of this.

A way to get the ball rolling :)

nshoes (approved):
Awesome, thanks for this Lars!

PR #2070, opened 2025-04-30
Handle user presence tracking failing

elinol:
Adds error handling to Presence functions.

Also adds docs and specs.

### DIFF
--- a/lib/nerves_hub_web/channels/presence.ex
+++ b/lib/nerves_hub_web/channels/presence.ex
@@ -11,6 +11,9 @@ defmodule NervesHubWeb.Presence do
 
   require Logger
 
+  alias NervesHub.Devices.Device
+  alias NervesHub.Helpers.Logging
+
   @impl Phoenix.Presence
   def init(_opts) do
     {:ok, %{}}
@@ -58,14 +61,16 @@ defmodule NervesHubWeb.Presence do
 
   Returns an :ok tuple with reference on success, {:error, reason} on failure.
   """
-  @spec track_user(String.t(), String.t(), map()) :: {:ok, binary()} | {:error, term()}
-  def track_user(topic, id, params) do
+  @spec track_user(String.t(), String.t(), map(), Device.t()) ::
+          {:ok, binary()} | {:error, term()}
+  def track_user(topic, id, params, resource) do
     case track(self(), topic, id, params) do
       {:ok, ref} ->
         {:ok, ref}
 
       {:error, reason} ->
         Logger.error("Failed to track user #{id} in topic #{topic}: #{inspect(reason)}")
+        Logging.log_to_sentry(resource, reason)
         {:error, reason}
     end
   end
@@ -75,14 +80,15 @@ defmodule NervesHubWeb.Presence do
 
   Returns :ok on success, :error on failure.
   """
-  @spec subscribe(String.t()) :: :ok | :error
-  def subscribe(topic) do
+  @spec subscribe(String.t(), Device.t()) :: :ok | :error
+  def subscribe(topic, resource) do
     case Phoenix.PubSub.subscribe(NervesHub.PubSub, "proxy:#{topic}") do
       :ok ->
         :ok
 
       {:error, reason} ->
         Logger.error("Failed to subscribe to presence topic #{topic}: #{inspect(reason)}")
+        Logging.log_to_sentry(resource, reason)
         :error
     end
   end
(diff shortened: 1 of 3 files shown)


lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -524,13 +524,13 @@ defmodule NervesHubWeb.Live.Devices.Show do
...
-      _ = Presence.subscribe(topic)
-      stream(socket, :presences, Presence.list_online_users(topic))
+    with true <- connected?(socket),
+         {:ok, _} <- Presence.track_user(topic, user.id, %{name: user.name}),
+         :ok <- Presence.subscribe(topic) do
+      stream(socket, :presences, Presence.list_present_users(topic))
     else
-      socket
+      _ ->
+        socket
```
elinol, line 533: Since a failing presence tracking doesn't affect other parts of the liveview I figured no error flash is necessary.

lib/nerves_hub_web/channels/presence.ex
```diff
@@ -41,10 +46,44 @@ defmodule NervesHubWeb.Presence do
...
+  """
+  @spec track_user(String.t(), String.t(), map()) :: {:ok, binary()} | {:error, term()}
+  def track_user(topic, id, params) do
+    case track(self(), topic, id, params) do
+      {:ok, ref} ->
+        {:ok, ref}
+
+      {:error, reason} ->
+        Logger.error("Failed to track user #{id} in topic #{topic}: #{inspect(reason)}")
+        {:error, reason}
```
elinol, line 69: Do we want this to be logged to Sentry as well?

joshk: Yes please

lib/nerves_hub_web/channels/presence.ex
```diff
@@ -41,10 +46,44 @@ defmodule NervesHubWeb.Presence do
...
+  """
+  @spec subscribe(String.t()) :: :ok | :error
+  def subscribe(topic) do
+    case Phoenix.PubSub.subscribe(NervesHub.PubSub, "proxy:#{topic}") do
+      :ok ->
+        :ok
 
-  def subscribe(topic), do: Phoenix.PubSub.subscribe(NervesHub.PubSub, "proxy:#{topic}")
+      {:error, reason} ->
+        Logger.error("Failed to subscribe to presence topic #{topic}: #{inspect(reason)}")
```
joshk, line 85: And log to Sentry here too please

joshk (changes requested):
Awesome work!

lib/nerves_hub_web/live/devices/show.ex
nshoes: Agreed!

joshk:
@nshoes @elinol I'm wondering, instead of doing logging and sentry error reporting in each spot, should we maybe use telemetry and have a handler that does this?

nshoes:
I don't feel strongly either way. What sways you towards telemetry here?

joshk:
> I don't feel strongly either way. What sways you towards telemetry here?

I'll defer this idea and discussion for a later PR.

It's mostly about encapsulation, as logging + sentry usually are implemented hand in hand

lib/nerves_hub_web/channels/presence.ex
```diff
@@ -41,10 +49,47 @@ defmodule NervesHubWeb.Presence do
...
-  def track_user(topic, id, params), do: track(self(), topic, id, params)
+  @doc """
+  Tracks a user's presence in a topic.
 
-  def subscribe(topic), do: Phoenix.PubSub.subscribe(NervesHub.PubSub, "proxy:#{topic}")
+  Returns an :ok tuple with reference on success, {:error, reason} on failure.
+  """
+  @spec track_user(String.t(), String.t(), map(), Device.t()) ::
+          {:ok, binary()} | {:error, term()}
+  def track_user(topic, id, params, resource) do
```
elinol, line 66: I'm not too happy about having to pass the device here just for logging needs, I'm not sure why, it's just a feeling. Maybe there is a better way to do the Sentry logging? I'm not that familiar with Sentry.

nshoes: Yeah I hear that, I think your feeling is right :) I made some adjustments in e0dbcdd, it would have taken longer to write up an explanation 😆 Because we store the device identifier in the topic, I think it's acceptable to only log a message to Sentry. Let me know what you think!

joshk (approved):
I think this is great.

I see some future improvement options around the metadata we send to Sentry, but that can be a later improvement.

PR #2071, opened 2025-04-30
Scroll pinned-device tags when there are too many

nshoes:
Too many tags for a pinned device breaks the layout. The text should not wrap and it should be scrollable.

Before:
CleanShot 2025-04-30 at 12 25 23@2x

After:
CleanShot 2025-04-30 at 12 24 54

### DIFF
--- a/assets/ui-rework/app.css
+++ b/assets/ui-rework/app.css
@@ -265,6 +265,11 @@
     z-index: 10;
     max-height: 52px;
   }
+
+  .scrollable-table-cell {
+    overflow-y: scroll;
+    scrollbar-width: none;
+  }
 }
 
 html {
--- a/assets/ui-rework/app.css
+++ b/assets/ui-rework/app.css
@@ -266,9 +266,36 @@
     max-height: 52px;
   }
 
-  .scrollable-table-cell {
+  .scrollable-inner {
     overflow-y: scroll;
     scrollbar-width: none;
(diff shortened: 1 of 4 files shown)


joshk:
OMG I love this!

joshk (approved):
It could also be cool to add a little fade overlay on the right so you can see if there are more tags to scroll

nshoes:
Now THAT is a great idea.

joshk:
Oh oh, @nshoes, can you also add this to the Devices list view, and show view?

nshoes:
Good call, will do.

nshoes:
Devices show page looks great:
CleanShot 2025-05-05 at 10 42 21

However, I couldn't get it working for the devices index. We have the clickable mask over the entire row and that prevents elements under it from being scrolled 😢 I poked at it for a bit but couldn't find a satisfactory solution.

nshoes:
Pinned devices for posterity:

https://github.com/user-attachments/assets/e04113a5-7324-4f5d-81a0-53f1e5ce25d3

Note that it's a bit tough for the gradient resolution to come through because of compression.


The hover background bugged me at first but I think it's fine. I tried hiding the gradient mask on hover but then it looks worse because the tags look cut off. Don't have strong opinions here.

joshk:
This is 🔥

PR #2073, opened 2025-05-02
Fix pure text emails

joshk:
My recent updates to the emails missed a crucial `IO.iodata_to_binary()` when processing the content of the pure text emails.

This PR fixes that, removes some email duplication, and fixes a test related to the new oauth login pages.


lawik (approved):

nshoes:
Hm, how did we not catch this? Previewing text emails locally seemed to render properly.

PR #2077, opened 2025-05-03
Use the `citext` column type for a users email

joshk:
This takes care of case sensitivity issues for us, which we aren't currently handling correctly.


joshk:
@nshoes can you please triple check your DB to make sure there aren't any duplicated emails? (mix of upper and lower case)

PR #2080, opened 2025-05-04
Support Devices sending log messages via a new Logging extension

joshk:
Companion to https://github.com/nerves-hub/nerves_hub_link/pull/303

New log lines are streamed to the UI, with only 25 max shown.

This has been implemented using **[ClickHouse](https://clickhouse.com/)**.

I've also used [hammer](https://hexdocs.pm/hammer) for rate limiting log messages (3 per second, with a burst capacity of 10).

**Pros:**
- built-in row truncation (defaults to 3 days)
- isolate analytics from core platform functionality
- the analytics system isn't required, and if turned off, the logging extension can't be attached

**Cons:**
- we don't get transaction isolation in tests
- we don't get nice association helpers as these are separate ecto repos

**Future possibilities:**
- "Show more"
- Filtering on level 
- Graph of number of log lines over X period
- Search log messages (not sure how well this would work with ClickHouse)

**Things to note:**
- using `make iex-server` will try to start the analytics subsystem
- to skip the analytics subsystem, use `ANALYTICS_ENABLED=false make iex-server`
- the docker compose file has been updated to include ClickHouse, its recommended to use this for starting supplementary systems like Postgres and ClickHouse.

---

**Simple UI** (to start with)

https://github.com/user-attachments/assets/b4ccd5e8-2411-4423-809e-8b48d19ebbff

---

**To finish:**

- [x] `device_log_lines` truncation (customizable time period)
- [x] tests


lawik:
Comment over in the nerves_hub_link one applies pretty much to both: https://github.com/nerves-hub/nerves_hub_link/pull/303#issuecomment-2849055825

lib/nerves_hub_web/components/device_page/logs_tab.ex
```diff
@@ -0,0 +1,99 @@
...
+    |> halt()
+  end
+
+  def hooked_info(_event, socket), do: {:cont, socket}
+
+  def render(%{product: %{extensions: %{logging: logging}}} = assigns) when logging == false do
+    ~H"""
+    <div class="size-full p-12">
+      <div class="size-full flex justify-center items-center p-6 gap-6 text-medium font-mono">
+        <div>Device logs aren't enabled for this product.</div>
```
nshoes, line 54: Can we edit these in the product settings? We should call it out here if so.

nshoes (approved):
This is sick! LGTM. Tests are 💯

joshk:
After discussion with @nshoes , I'm going to change this feature from using Postgres for the backend to ClickHouse.

This is a good article to get some inspiration from: https://clickhouse.com/blog/storing-log-data-in-clickhouse-fluent-bit-vector-open-telemetry

The general idea is that features like logging and metrics should not impact the core stability of the platform, so if we do get gigs of data from one device it shouldn't impact the ability to use the core features, like firmware updates.

Additionally, I'll look into using https://hex.pm/packages/hammer for some rate limiting goodness.

lib/nerves_hub_web/components/device_page/logs_tab.ex
joshk: Good call!

Aaaaaand done!

lib/nerves_hub_web.ex
```diff
@@ -125,6 +125,10 @@ defmodule NervesHubWeb do
         |> Enum.into(%{}, fn x -> {x, params[to_string(x)]} end)
       end
 
+      def analytics_enabled?() do
```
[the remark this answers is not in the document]
joshk: I hear you, and I do agree in part. I don't like adding more stuff to the assigns as its change tracked, and this is something that doesn't need to be tracked. I'm essentially defining a helper across all liveviews, but I can accept that this should be extracted to its own module and included in the liveviews that need it. I'll address that in a different PR.

PR #2085, opened 2025-05-06
Monitor deployment processes in worker and report to Sentry

nshoes:
Every 15min we check the counts of how many should be running and how many are running. If the count is off we create a Sentry report with the counts. Alongside putting together a timeline from logs I think this will aid in debugging why our deployment orchestration processes aren't starting properly on deploys.


joshk (approved):
Should we also start deployment orchestrators that should be running?

nshoes:
Yes!

joshk (approved):
I love this!

Is it worth adding a test for this?

nshoes:
Done! It's stubbed to hell, but it's better than nothing.

joshk (approved):
I love the stubs!

PR #2086, opened 2025-05-06
Fix firmware install count column when viewing firmwares

nshoes:
The `:distinct` on the `count` would always return 0 or 1 when it should be counting all the device rows.

CleanShot 2025-05-06 at 12 48 35@2x


joshk (approved):
Thank you!

PR #2089, opened 2025-05-07
Add confirming online status with the tracker on device join

lawik:


lawik:
Let me know if the Tracker.confirm_online was dropped for some reason. I assume it got lost in the shuffle between new connection tracking and that.

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -80,6 +80,8 @@ defmodule NervesHubWeb.DeviceChannel do
   def handle_info(:announce_online, socket) do
     # Update the connection to say that we are fully up and running
     Connections.device_connected(socket.assigns.reference_id)
+    # Tell the tracker that we are online
+    NervesHub.Tracker.online(socket.assigns.device)
```
joshk, line 84: maybe we should move this into `Connections.device_connected`?

lawik: No particular opinion. That would tie the tracker into the Connections stuff. Not sure if we want that 🤷🏻

joshk: Seems more consistent, as in, instead of remembering to call the tracker along with the function `device_connected`, `device_connected` does it for us. I'll look into it later.

joshk (approved):

PR #2090, opened 2025-05-07
Enable firmware delta delivery

lawik:


lawik:
This still has some potential problems. I need to test more as the only device I tried this on would not successfully reboot ;)

But now there is a test that indicates that the deltas we generate produce a correct update when applied.

lawik:
Current version generates diffs on ALL the files in the firmware archive. It should consult the metadata file and only generate diffs for the ones that have the diffing markers.

lib/nerves_hub/devices.ex
```diff
@@ -765,11 +765,20 @@ defmodule NervesHub.Devices do
...
+  def do_resolve_update(device, deployment_group) do
     case verify_update_eligibility(device, deployment_group) do
       {:ok, _device} ->
-        {:ok, url} = Firmwares.get_firmware_url(deployment_group.firmware)
+        {:ok, url} =
+          if deployment_group.product.delta_updatable do
+            get_delta_or_firmware_url(device.firmware_metadata.uuid, deployment_group.firmware)
+          else
+            Firmwares.get_firmware_url(deployment_group.firmware)
+          end
```
joshk: Could we move this conditional to the function conditions, so we can have one definition for cases where the product has deltas enabled, and one for where its not enabled?

eg. 

```elixir
  def get_delta_or_firmware_url(device_firmware_uuid, %{product: %{delta_updatable: false}}) do
    Firmwares.get_firmware_url(deployment_group.firmware)
  end

  def get_delta_or_firmware_url(device_firmware_uuid, %{product: %{delta_updatable: true}}) do
    ...
  end
```

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -594,7 +594,22 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
     %{product: product, device: device, user: user} = socket.assigns
 
     {:ok, firmware} = Firmwares.get_firmware_by_product_and_uuid(product, uuid)
-    {:ok, url} = Firmwares.get_firmware_url(firmware)
+
+    {:ok, url} =
+      if device.firmware_metadata && device.firmware_metadata.uuid do
+        Logger.info(
+          "Manually sending firmware delta for updating from #{device.firmware_metadata.uuid} to #{firmware.uuid} to #{device.identifier}..."
```
joshk, line 603: I've noticed a few new logger messages. I'd recommend using logger metadata for things like firmware uuids and device identifiers as this will make it easier for searching.

lib/nerves_hub_web/live/devices/show.ex
```diff
@@ -378,38 +377,6 @@ defmodule NervesHubWeb.Live.Devices.Show do
     |> noreply()
   end
 
-  def handle_event("push-update", %{"uuid" => ""}, socket) do
```
joshk: why were these removed?

joshk (changes requested):
This is generally looking good.

lib/nerves_hub/firmwares.ex
```diff
@@ -379,40 +379,56 @@ defmodule NervesHub.Firmwares do
...
+            fn ->
+              with upload_metadata <-
+                     firmware_upload_config().metadata(org.id, firmware_delta_filename),
+                   {:ok, firmware_delta} <-
+                     insert_firmware_delta(%{
+                       source_id: source_firmware.id,
+                       target_id: target_firmware.id,
+                       upload_metadata: upload_metadata
+                     }),
+                   {:ok, firmware_delta} <- get_firmware_delta(firmware_delta.id),
```
elinol, line 416: Does this differ from the result from the previous step? (the return from `insert_firmware_delta`)

lib/nerves_hub_web/live/devices/show.ex
lawik: Because I found that the version that exists and runs is in `details_tab.ex`. Ran into it while I made those updates. Could be moved to a different PR. Probably just fixed it so I knew which code actually runs.

joshk: Ah, thanks for explaining. These functions are still needed until we remove the old UI. The old UI doesn't use `details_tab.ex`, only the new UI does.

lawik: Ah, that makes sense. I just figured it was an oversight :D

lawik: Should be back now :)

lib/nerves_hub/firmwares.ex
lawik: It shouldn't...
But I don't feel an urgent need to change it.

lawik: Good catch

lib/nerves_hub_web/components/device_page/details_tab.ex
lawik: Added Logger.metadata where it made sense and passed explicitly where that seemed more prudent.

lib/nerves_hub/devices.ex
```diff
@@ -1716,4 +1715,58 @@ defmodule NervesHub.Devices do
...
+  def get_delta_or_firmware_url(device_firmware_uuid, %DeploymentGroup{
+        product: %{delta_updatable: true},
+        firmware: target_firmware
+      }) do
+    # Get firmware delta URL if available but otherwise deliver full firmware
+    with %Firmware{} = device_firmware <- Firmwares.get_firmware_by_uuid(device_firmware_uuid),
+         {:ok, firmware_delta} <-
+           Firmwares.get_firmware_delta_by_source_and_target(device_firmware, target_firmware) do
+      Logger.info(
+        "Delivering firmware delta between #{device_firmware_uuid} and #{target_firmware.uuid}...",
```
joshk, line 1733: the `...` in the various logger info message are noise.
I also think `between #{device_firmware_uuid} and #{target_firmware.uuid}` isn't needed as this is covered by the metadata.
It would also be good to add the deployment id and device id (if easily available) to the metadata.

joshk (approved):
All in all I'm 👍 on this.

I would like to see less noisey log messages. 

If those could be tweaked then you can merge in as you see fit.

lib/nerves_hub/devices.ex
lawik: To me those messages show the progression of which firmware gets delivered and why. Might be worth cutting down when we have seen this in action for longer. Logging a bit more to make sure we can see the steps while this feature is young seems like the right tradeoff for me.

I believe I added the Logger.metadata call on this LiveView where we know the device. We should also add it on the DeviceChannel. Means as long as we don't cross a process boundary we can have device_id and product_id.

lawik: Correct me if I am wrong but the metadata doesn't show up in a regular log line during development and so on right?

PR #2057, opened 2025-04-20
OpenAPI spec improvements

joshk:
Document the entire API, with the aid of https://hexdocs.pm/open_api_spex

This gives us a route which serves the OpenAPI spec (`http://localhost:4000/api/openapi`) and a Swagger UI for interacting with it (`http://localhost:4000/api/swaggerui`)

I'm not super in love with the integration code, but I do love what this gives us, so I'm willing to give it a punt.

Some other items addressed during this integration:

- Use `send_resp(conn, :no_content, "")` for Device actions that don't return anything
- Move the `/scripts` list from the Device to Product (`/api/orgs/{org_name}/products/{product_name}/scripts`)

I would prefer to merge in this unit of work before continuing further.

---

**Swagger UI Screenshot**:

Screenshot 2025-04-20 at 1 22 58 PM

---

**Next Steps**:

- [x] Review tags (categories)
- [x] Look into splitting the two possible device urls into two sections (might require another controller)

PR #2096, opened 2025-05-08
Reinstate script controller API endpoint

nshoes:
This was removed in #2057 but is needed.


nshoes:
@joshk approved this remotely :) 
CleanShot 2025-05-08 at 15 48 11@2x

PR #2099, opened 2025-05-10
Use the `OrchestratorRegistration` GenServer for monitoring

joshk:
This rework came about after coming across two issues:

1. The current monitoring wasn't starting the orchestrators, which were missing (constant Sentry errors)
2. The case statement in `application.ex` which should only start the Orchestrators on web nodes wasn't matching correctly, causing Orchestrators to run across all nodes.

I addressed issue 1. by using `await` when starting Orchestrators. 

But when I fixed issue number 2. it meant we would no longer use Oban for cron monitoring. 

This PR changes the `OrchestratorRegistrator` to be a permanent `GenServer` which checks the running Orchestrators and starts any that aren't running, based on a random interval every 13-17 mins.

I've tested this on NervesCloud and it works well.


.dialyzer_ignore.exs
```diff
@@ -3,5 +3,6 @@
   {"lib/x509/certificate/extension.ex", "Unknown type: X509.ASN1.record/1."},
   {"lib/nerves_hub_web/channels/device_channel.ex", :unmatched_return, 1},
   {"lib/nerves_hub_web/channels/device_socket.ex", :unmatched_return, 1},
-  {"lib/nerves_hub/analytics_repo.ex", :contract_supertype, 1}
+  {"lib/nerves_hub/analytics_repo.ex", :contract_supertype, 1},
+  {"lib/nerves_hub/managed_deployments/distributed/orchestrator_registration.ex", :pattern_match}
```
nshoes, line 7: Why are we ignoring this entire module?

lib/nerves_hub/managed_deployments/distributed/orchestrator_registration.ex
```diff
@@ -29,23 +31,80 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorRegistration do
...
+        # :already_started is an ok (good) result
+        {:ok, info}
+
+      other ->
+        other
+    end
+  end
+
+  defp report_errors(results) do
+    errors = Enum.filter(results, fn result -> elem(result, 0) == :error end)
```
nshoes, line 89:
```suggestion
    errors = Enum.filter(results, fn {status, _} -> status == :error end)
```

.dialyzer_ignore.exs
joshk: Only because I couldn't figure out how to just ignore one pattern match error

joshk: the specs for ProcessHub don't match the actual results

nshoes: Then ignoring time it is!

nshoes (approved):

PR #2100, opened 2025-05-11
Fixes to device logging processing

joshk:
- Use the time in the meta payload if no timestamp is included
- Do some message formatting
- Correct the `handle_in` event signature
- Use ClickHouses `async_insert`
- Use `Task.Supervisor.start_child` (instead of `async`) since we don't care for the result, at least we don't for now
- Tweak the UI, including adding the ability to stop live updating of the Logs UI

Screenshot 2025-05-12 at 11 07 55 PM


lib/nerves_hub/devices/log_line.ex
```diff
@@ -19,14 +18,37 @@ defmodule NervesHub.Devices.LogLine do
...
+
+  defp format_message(%{"message" => message} = params) when is_binary(message), do: params
+
+  defp format_message(%{"message" => message} = params) when is_list(message) do
+    Map.put(params, "message", List.to_string(message))
+  end
+
+  defp format_message(%{"message" => message} = params) do
+    Map.put(params, "message", inspect(message))
+  end
```
nshoes, line 53: Is this fallback utilized at the moment or are we covering our bases for the future?

lib/nerves_hub_web/components/device_page/logs_tab.ex
```diff
@@ -99,23 +124,55 @@ defmodule NervesHubWeb.Components.DevicePage.LogsTab do
...
-          <div id={"#{DateTime.to_unix(line.timestamp, :microsecond)}-log-line-localtime"} phx-hook="LogLineLocalTime">
-            {line.timestamp}
+      <div :if={@has_logs} class="size-full">
+        <div class="flex flex-row items-center justify-between h-11 border-b border-zinc-700 px-12">
+          <div>
+            <span class="text-sm text-zinc-400">Live log streaming :</span>
+            <button
+              id="toggle-log-streaming"
+              type="button"
+              phx-click="toggle_log_streaming"
```
nshoes, line 138: We tend to use hyphens for `phx-click` attribute values.

```suggestion
              phx-click="toggle-log-streaming"
```

lib/nerves_hub/extensions/logging.ex
```diff
@@ -55,9 +55,9 @@ defmodule NervesHub.Extensions.Logging do
 
   defp schedule_create(device, log_line) do
     _ =
-      Task.Supervisor.async(
+      Task.Supervisor.start_child(
```
nshoes, line 58: Since we're async inserting ClickHouse and broadcasting, does this need to be a Task anymore? Also, what about error-handling?

lib/nerves_hub/devices/log_line.ex
```diff
@@ -19,14 +18,37 @@ defmodule NervesHub.Devices.LogLine do
...
+      |> format_message()
 
     %__MODULE__{}
     |> cast(params, @required ++ @optional)
     |> validate_required(@required)
   end
+
+  defp maybe_set_timestamp(%{"timestamp" => _} = params), do: params
+
+  defp maybe_set_timestamp(params) do
```
nshoes, line 36: Could we get a specific test for this?

test/nerves_hub_web/live/new_ui/devices/logs_tab_test.exs
```diff
@@ -147,14 +183,14 @@ defmodule NervesHubWeb.Live.NewUI.Devices.LogsTabTest do
...
+        "level" => "info",
+        "timestamp" => DateTime.utc_now(),
+        "message" => "something wicked this way comes : #{n}"
       }
 
-      LogLines.create!(device, attrs)
+      {:ok, _} = LogLines.async_create(device, attrs)
 
-      Process.sleep(50)
+      Process.sleep(5)
```
nshoes, line 193: Why is this needed? Timestamp overlap?

test/nerves_hub/devices/log_lines_test.exs
```diff
@@ -35,7 +35,12 @@ defmodule NervesHub.Devices.LogLinesTest do
     device_id = device.id
     product_id = device.product_id
 
-    log = LogLines.create!(device, %{timestamp: logged_at, level: level, message: message})
+    {:ok, log} =
+      LogLines.async_create(device, %{
+        "timestamp" => logged_at,
+        "level" => level,
+        "message" => message
+      })
```
nshoes, line 43: Why are we preferring string keys here?

lib/nerves_hub/devices/log_line.ex
joshk: its primarily for covering bases, as I found out during testing that log messages could be lists, and from some light investigation I believe its possible for messages to be other types too

lib/nerves_hub_web/components/device_page/logs_tab.ex
joshk: true true. I used underscores initially because I had it as an atom, but then realized that it should be a string, but forgot to switch it back around

lib/nerves_hub/extensions/logging.ex
joshk: Soooooo, since we are using http to talk to the DB, and we can't be certain how long an insert will take, i thought this fire-and-forget strategy would be worth using while we build up our experience with ClickHouse. Also, by moving logging to its own task we clear the queue more quickly, freeing it up for other messages being sent.

As for error handling, as this is 'early release', I had punted that for a little later.

test/nerves_hub/devices/log_lines_test.exs
joshk: they come as string keys from the websocket (link), so I'm using string keys here too, otherwise we would get an error about mixed atom / string keys.

lib/nerves_hub/devices/log_line.ex
joshk: Done!

lib/nerves_hub/devices/log_line.ex
nshoes: Good enough for me 🫡

nshoes (approved):

PR #2103, opened 2025-05-15
Device first in line for updates

nshoes:
## Todo
- [ ] Test devices joining channel with no firmware metadata have `first_in_line_for_udpates` set to `true`

### DIFF
--- a/lib/nerves_hub_web/controllers/api/schemas/device_schemas.ex
+++ b/lib/nerves_hub_web/controllers/api/schemas/device_schemas.ex
@@ -45,6 +45,10 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
           description: "Device penalty box expiration timestamp",
           format: :"date-time"
         },
+        first_in_line_for_updates: %Schema{
+          type: :boolean,
+          description: "Prioritizes this device for updates when part of a deployment group"
+        },
         org_name: %Schema{type: :string},
         product_name: %Schema{type: :string},
         last_communication: %Schema{
@@ -81,6 +85,7 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
         },
         "updates_enabled" => true,
         "updates_blocked_until" => "2050-04-20T00:33:09Z",
+        "first_in_line_for_updates" => true,
         "org_name" => "BigCompany",
         "product_name" => "AmazingProduct",
         "last_communication" => "2050-04-20T00:33:09Z"
@@ -129,6 +134,7 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
             },
             "updates_enabled" => true,
             "updates_blocked_until" => "2050-04-20T00:33:09Z",
+            "first_in_line_for_updates" => true,
             "org_name" => "BigCompany",
             "product_name" => "AmazingProduct",
             "last_communication" => "2050-04-20T00:33:09Z"
@@ -161,6 +167,7 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
             },
             "updates_enabled" => true,
             "updates_blocked_until" => "2050-04-20T00:33:09Z",
+            "first_in_line_for_updates" => true,
             "org_name" => "BigCompany",
             "product_name" => "AmazingProduct",
             "last_communication" => "2050-04-20T00:33:09Z"
@@ -181,7 +188,8 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
             description: %Schema{type: :string},
             tags: %Schema{type: :string},
             deployment_group_id: %Schema{type: :integer},
-            updates_enabled: %Schema{type: :boolean}
+            updates_enabled: %Schema{type: :boolean},
+            first_in_line_for_updates: %Schema{type: :boolean}
           },
           required: [:identifier]
         }
@@ -193,7 +201,8 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
           "description" => "Example Device",
           "tags" => "prod, customerJNK",
           "deployment_group_id" => 1,
-          "updates_enabled" => false
+          "updates_enabled" => false,
+          "first_in_line_for_updates" => true
         }
       }
     })
@@ -209,7 +218,8 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
             description: %Schema{type: :string},
             tags: %Schema{type: :string},
             deployment_group_id: %Schema{type: :integer},
-            updates_enabled: %Schema{type: :boolean}
+            updates_enabled: %Schema{type: :boolean},
+            first_in_line_for_updates: %Schema{type: :boolean}
           }
         }
       },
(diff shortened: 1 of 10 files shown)


priv/repo/migrations/20250515021542_add_priority_updates_to_devices.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHub.Repo.Migrations.AddPriorityUpdatesToDevices do
+  use Ecto.Migration
+
+  def change do
+    alter table(:devices) do
+      add(:priority_updates, :boolean, default: false, null: false)
```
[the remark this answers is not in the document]
nshoes: I'd rather keep as is unless I get more pushback. I've been viewing this from the deployment group's POV: "which devices are eligible for priority updates" or "devices marked for priority updates", for example.

priv/repo/migrations/20250519142403_add_enable_priority_updates_to_deployment_groups.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHub.Repo.Migrations.AddEnablePriorityUpdatesToDeploymentGroups do
+  use Ecto.Migration
+
+  def change do
+    alter table(:deployments) do
+      add(:enable_priority_updates, :boolean, default: false, null: false)
```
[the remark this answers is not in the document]
nshoes: Good idea! I dislike how devices marked for priority updates aren't considered in the default query. It would be a nice change all around if they did. I'll make it so.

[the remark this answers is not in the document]
nshoes: Done!

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -273,19 +273,43 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
...
+                  name="device[priority_updates]"
+                  checked={@device.priority_updates}
+                  value="on"
+                  class="rounded border-zinc-700 text-zinc-400 focus:ring-0 checked:bg-indigo-500"
+                />
+                <label for="device[priority_updates]" class="pl-2">Priority Updates</label>
+                <div id="priority-update-info" class="inline-block align-middle relative pl-1" phx-hook="ToolTip" data-placement="right">
+                  <.icon name="info" class="stroke-zinc-400" />
+                  <div class="tooltip-content hidden w-max absolute top-0 left-0 text-xs px-2 py-1.5 rounded border border-[#3F3F46] bg-base-900 flex">
+                    When enabled, this device will be prioritized for updates if it is assigned to a deployment group with priority updates enabled. <br />
```
jjcarstens, line 292:
```suggestion
                    When enabled, this device will be prioritized for updates and ignores the deployment group's queue management. <br />
```

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -273,19 +273,43 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
...
+                    The priority setting will automatically revert after the device completes its next successful update. <br />
```
jjcarstens, line 293:
```suggestion
                    The priority setting will automatically revert after the next successful update. <br />
```

lib/nerves_hub_web/controllers/api/schemas/device_schemas.ex
```diff
@@ -81,6 +85,7 @@ defmodule NervesHubWeb.API.Schemas.DeviceSchemas do
         },
         "updates_enabled" => true,
         "updates_blocked_until" => "2050-04-20T00:33:09Z",
+        "priority_updates" => true,
```
jjcarstens, line 88: I believe we may need to also update the `device_json.ex` file to ensure this field is returned?

jjcarstens (commented):
❤️ Loving it! Just a couple more things I noticed

lib/nerves_hub_web/components/device_page/details_tab.ex
nshoes: Hm, wait a moment, this isn't true. Queue management is always respected, and priority devices are always taken into account: https://github.com/nerves-hub/nerves_hub_web/pull/2103/files#diff-03454c312d5c891b1ae7d916a1eb171cdf023ca00f9aa5155ea5633f7c5cbe81R764-R776.

Maybe we miscommunicated? Do you think a separate deployment group setting for taking devices marked for priority updates into account as well as queue management? Give 'em _all_ the knobs.

lib/nerves_hub_web/controllers/api/schemas/device_schemas.ex
nshoes: Yes, good call! Done.

lib/nerves_hub_web/components/deployment_group_page/settings.ex
```diff
@@ -35,10 +35,17 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Settings do
...
+            <div class="w-1/2">
               <.input field={@form[:name]} label="Name" placeholder="Production" />
             </div>
+            <%!-- <div class="w-1/2"> --%>
+            <%!-- <.input field={@form[:enable_priority_updates]} type="checkbox" label="Enable priority updates">
+                <:rich_hint>
+                  When enabled, devices that have connected for the first most recently are prioritized, followed by devices that are marked for priority updates.
+                </:rich_hint>
+              </.input>
+            </div> --%>
```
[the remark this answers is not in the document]
nshoes: Sure was, removed!

lib/nerves_hub_web/components/device_page/details_tab.ex
jjcarstens: My brain could be broke and could be reading it wrong. This is what I was thinking would happen given FIFO concurrency of 2 ?
```
# CURRENT TIME 8:00
device1@8:00
device3@9:00 (PRIORITY)
----------------------------> (concurrency limit. All below are waiting)
device2@8:30
```
In this case, `device2` should technically be the next slot having connected prior to device3, but it loses its slot since device3 is priority. So queuing is "ignored"

nshoes: Ah, that makes more sense. We'll need to order by `device.priority_updates` first to satisfy that requirement. I believe that's more correct than what I originally had as well. Fixed!

nshoes: But to be clear, it's now:

```
# CURRENT TIME 8:00
device3@9:00 (PRIORITY)
device1@8:00
----------------------------> (concurrency limit. All below are waiting)
device2@8:30
```

jjcarstens: Ooo…okay. That actually seems fine. I feel like either of these cases, the queue scheme is ignore since a priority device always gets bumped to top of queue regardless of the time. Maybe “ignored” is the wrong term?

nshoes: Good point, I think it should be "bypasses". Objections?

PR #2105, opened 2025-05-19
Update pre-generated account

elliotjon:
Update pre-generated account with new password to pass validation of minimum 12 characters.
Noticed this during setup.


joshk (approved):
Thank you!

PR #2107, opened 2025-05-20
Don't require tags when updating deployment group, changeset refactor

nshoes:
This started as an adjustment to deployment group tag validations and ended up being a refactor of deployment group changesets. The core issue is that we handled tag validation differently in the create and update changesets.

When updating a deployment group, tags are required, even though empty tags is a valid state; you can create a deployment group with no tags. This requirement has been removed. In application code, we treat `%{"version" => ""}` as the empty state for `version`. Nothing has stopped us in the changeset or database layer from saving `conditions` to anything but a map with those two keys. Luckily, in SR prod, there are no deployment group rows with one of these keys; it's always both. I added the ability to partially update conditions so you don't have to always provide both keys. This isn't used in application code but is in tests and is a quality-of-life update. 

Shared validations are now reconciled into a base changeset. The create and update changesets then handle their own unique validations. This is all tested and laid out nicely in `NervesHub.ManagedDeployments.DeploymentGroupTest`. I tried a new approach to organizing the tests, looking for feedback.

I then removed the ancient, recursive spaghetti code that  would magically add the `product_id` when creating a deployment group. This has been simplified to requiring the `product_id` to be passed in explicitly. This change avoids a database query when preloading firmware to get `product_id`. The code in question handled a lot of different use cases but preloading was the only one application code used :( In all use cases, the firmware was already loaded making it trivial to pass in `firmware.product_id`.

I also caught a route mistake when creating deployment groups in the new UI.

Some changes I made were already covered by tests; I added tests for the rest. 

I had to update a two-year-old script in `priv/repo`, but I'm thinking the whole thing can be removed, as well as the other script in the same directory. Fine to pull the trigger if someone else double-checks me.

### DIFF
--- a/lib/nerves_hub/managed_deployments/deployment_group.ex
+++ b/lib/nerves_hub/managed_deployments/deployment_group.ex
@@ -87,13 +87,6 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
 
   def create_changeset(%DeploymentGroup{} = deployment, params) do
     base_changeset(deployment, params)
-    |> validate_change(:is_active, fn :is_active, is_active ->
-      if is_active do
-        [is_active: "cannot be true on creation"]
-      else
-        []
-      end
-    end)
   end
 
   def update_changeset(%DeploymentGroup{} = deployment, params) do
--- a/lib/nerves_hub/managed_deployments/deployment_group.ex
+++ b/lib/nerves_hub/managed_deployments/deployment_group.ex
@@ -85,52 +85,19 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
     timestamps()
   end
 
-  def creation_changeset(%DeploymentGroup{} = deployment, params) do
-    # set product_id by getting it from firmware
-    with_product_id = handle_product_id(deployment, params)
-
-    deployment
-    |> cast(with_product_id, @required_fields ++ @optional_fields)
-    |> validate_required(@required_fields)
-    |> unique_constraint(:name, name: :deployments_product_id_name_index)
+  def create_changeset(%DeploymentGroup{} = deployment, params) do
+    base_changeset(deployment, params)
     |> validate_change(:is_active, fn :is_active, is_active ->
-      creation_errors(:is_active, is_active)
+      if is_active do
+        [is_active: "cannot be true on creation"]
+      else
+        []
+      end
     end)
   end
 
-  defp creation_errors(:is_active, is_active) do
-    if is_active do
-      [is_active: "cannot be true on creation"]
-    else
-      []
-    end
-  end
-
-  defp handle_product_id(%DeploymentGroup{}, %{firmware: %Firmware{product_id: p_id}} = params) do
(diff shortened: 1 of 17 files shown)


lib/nerves_hub/managed_deployments/deployment_group.ex
```diff
@@ -85,59 +85,33 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
...
-    |> unique_constraint(:name, name: :deployments_product_id_name_index)
+  def create_changeset(%DeploymentGroup{} = deployment, params) do
+    base_changeset(deployment, params)
     |> validate_change(:is_active, fn :is_active, is_active ->
-      creation_errors(:is_active, is_active)
+      if is_active do
+        [is_active: "cannot be true on creation"]
+      else
+        []
+      end
```
joshk, line 95: Soooooo, why can't deployments be active on creation? Seems annoying to create a deployment only to them activate it.

lib/nerves_hub/managed_deployments/deployment_group.ex
```diff
@@ -85,59 +85,33 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
...
         put_change(changeset, :current_updated_devices, 0)
       else
         changeset
       end
     end)
+  end
+
+  defp base_changeset(%DeploymentGroup{} = deployment, params) do
+    deployment
+    |> cast(params, @required_fields ++ @optional_fields)
```
joshk, line 112: I've come to view this pattern as pretty dangerous. Its possible for form submissions to be maliciously constructed such that the `org_id` or `product_id` are changed to an org or product not accessible by the user. In fact, this is true for the `firmware_id` as well. 

And the use of `@whitelist_fields [:name, :org_id, :firmware_id, :conditions, :is_active, :product_id]`, like in the API controller, is essentially bypassing the strength of changesets, which should be doing this for us.

I think it might be worth us rethinking how we are using `@required_fields ++ @optional_fields` more generally, possibly favoring something like `@creation_fields` etc.

And for special-purpose schemas, like Product or Org, they should be passed in as a param which is then added to the changeset or params map. 

Happy to zoom and talk more about this.

lib/nerves_hub/managed_deployments/deployment_group.ex
nshoes: True! I have no valid reason why they can't be. This must be legacy logic that doesn't apply anymore. I'll give it the axe 🪓

nshoes: Done.

lib/nerves_hub/managed_deployments/deployment_group.ex
nshoes: Good callout. I agree that this is an issue, and we should infer foreign keys instead of expecting or allowing them.

This is an issue across the app, and I don't think we should start solving it in this work (if you were implying that). Let's definitely hop on a zoom to get our thoughts down in a GH issue.

jjcarstens (approved):

Issue #2108, opened 2025-05-20
Downloading device certificate fails


lawik:
Quote from user: "Downloading the Cert for a device doesn’t work in the new UI.  The download.txt “Site wasn’t available” was from the new UI, the old UI worked with a proper filename."

PR #2109, opened 2025-05-21
New UI: Fix broken certificate download

elinol:
Fixes a bug where device certificates could not be downloaded by changing the target link to use certificates serial number.
Also adds a test for downloading certificates in new UI.


lawik (approved):
✨

elinol:
#2108

joshk:
There is a better solution for this. Instead of adding `.serial` to the url, instead add `@derive {Phoenix.Param, key: :serial}` to `device_certificate.ex`.

This is in line with what we do for device, product, org, and deployment.

joshk:
https://hexdocs.pm/phoenix/1.8.0-rc.0/Phoenix.Param.html provides more detail.

elinol:
That's neat!
https://github.com/nerves-hub/nerves_hub_web/pull/2115

Issue #2093, opened 2025-05-07
Device Details view, polling refresh will reset selections in drop-downs


lawik:
This goes for both selecting a Deployment and a Version for sending manually. They'll reset to the default non-selected option whenever the poll hits, sometimes only if not focused, maybe. Bit unclear.

lawik:
Fixed by #2111 ("Ignore update on dropdowns in device detail view") — commit 8dfcd054.

PR #2111, opened 2025-05-21
Ignore update on dropdowns in device detail view

elliotjon:
This change will ignore update on dropdowns in device detail view which prevents clearing selections every time ui polling happens.

Issue https://github.com/nerves-hub/nerves_hub_web/issues/2093


nshoes (commented):
Sweet! Thank you! 🚢

joshk:
I'm not sure this is the best solution. This means we can't update the lists with new deployments or firmware if new options become available.

Issue #2004, opened 2025-03-31
Tags for Support Scripts


jjcarstens:
Internally, we use Support Scripts in a few different ways via the UI and also the API from other services. However, the current implementation shows all of the scripts and we have situations where we might not want some users to run a script and others to be allowed (primarily via API)

Our first thought to help solve this is to add tags to Support Scripts which are useful bits of freeform metadata then allow the API request to filter based on tags. We can organize scripts that makes sense for our situations and our internal services can adjust their requests to filter only the ones we want allowed for that service.

PR #2117, opened 2025-05-22
Add tags to scripts

elinol:
Enables tagging support scripts (as requested in #2004) in the same manor as device and deployment group tagging.

 - Migration and schema updates with a`tags` field for scripts.
 - Add a script filtering module
 - Update api script controller to accept filtering
 - Include script tags in UI
 - Fixes broken pagination for scripts
 - Bonus: Break out resource filtering into a common module

### DIFF
--- a/lib/nerves_hub/filtering.ex
+++ b/lib/nerves_hub/filtering.ex
@@ -2,7 +2,10 @@ defmodule NervesHub.Filtering do
   @moduledoc """
   Common filtering functionality for NervesHub resources.
   """
+  alias NervesHub.Devices.Device
+  alias NervesHub.ManagedDeployments.DeploymentGroup
   alias NervesHub.Products.Product
+  alias NervesHub.Scripts.Script
 
   import Ecto.Query
 
@@ -12,21 +15,21 @@ defmodule NervesHub.Filtering do
   ## Parameters
     - base_query: The initial Ecto query to build upon
     - product: The product to filter by
-    - opts: Map of options including:
+    - filter_builder: Function that takes (query, filters) and returns modified query
+    - sorter: Function that takes (query, sort) and returns modified query
+    - opts: Map of options maybe including:
       - sort: Tuple of {direction, field} for sorting
       - filters: Map of filters to apply
       - pagination: Map of pagination options
-    - filter_builder: Function that takes (query, filters) and returns modified query
-    - sorter: Function that takes (query, sort) and returns modified query
 
   ## Returns
     A tuple containing:
     - List of entries matching the query
     - Flop metadata containing pagination information
   """
-  @spec filter(Ecto.Query.t(), Product.t(), map(), function(), function()) ::
-          {[any()], Flop.Meta.t()}
-  def filter(base_query, product, opts \\ %{}, filter_builder, sorter) do
+  @spec filter(Ecto.Query.t(), Product.t(), function(), function(), map()) ::
+          {[%Device{}] | [%DeploymentGroup{}] | [%Script{}], Flop.Meta.t()}
+  def filter(base_query, product, filter_builder, sorter, opts \\ %{}) do
     opts = Map.reject(opts, fn {_key, val} -> is_nil(val) end)
 
     sorting = Map.get(opts, :sort, {:asc, :name})
--- a/lib/nerves_hub/filtering.ex
+++ b/lib/nerves_hub/filtering.ex
@@ -15,8 +15,7 @@ defmodule NervesHub.Filtering do
     - opts: Map of options including:
       - sort: Tuple of {direction, field} for sorting
       - filters: Map of filters to apply
-      - page: Page number for pagination
-      - page_size: Number of items per page
+      - pagination: Map of pagination options
     - filter_builder: Function that takes (query, filters) and returns modified query
     - sorter: Function that takes (query, sort) and returns modified query
 
--- a/lib/nerves_hub/filtering.ex
+++ b/lib/nerves_hub/filtering.ex
@@ -2,9 +2,12 @@ defmodule NervesHub.Filtering do
   @moduledoc """
   Common filtering functionality for NervesHub resources.
   """
+  alias NervesHub.Devices
   alias NervesHub.Devices.Device
+  alias NervesHub.ManagedDeployments
   alias NervesHub.ManagedDeployments.DeploymentGroup
   alias NervesHub.Products.Product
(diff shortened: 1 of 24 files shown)


elinol:
@nshoes @joshk I'm aware of the dialyzer warnings, this is not 100% ready for review, but before proceeding I'd like your opinion on moving the common filtering into `NervesHub.Filtering`

lib/nerves_hub_web/live/support_scripts/edit-new.html.heex
```diff
@@ -38,6 +38,7 @@
         <div class="flex p-6 gap-6">
           <div class="w-1/2 flex flex-col gap-6">
             <.input field={@form[:name]} label="Name" placeholder="eg. System Debugging" />
+            <.input field={@form[:tags]} value={Utils.tags_to_string(@form[:tags])} label="Tags" placeholder="eg. info,action" />
```
nshoes:
```suggestion
            <.input field={@form[:tags]} value={Utils.tags_to_string(@form[:tags])} label="Tags" placeholder="info, action" />
```

lib/nerves_hub_web/live/support_scripts/new-new.html.heex
```diff
@@ -38,6 +38,7 @@
         <div class="flex p-6 gap-6">
           <div class="w-1/2 flex flex-col gap-6">
             <.input field={@form[:name]} label="Name" placeholder="eg. System Debugging" />
+            <.input field={@form[:tags]} label="Tags" placeholder="eg. info,action" />
```
nshoes:
```suggestion
            <.input field={@form[:tags]} label="Tags" placeholder="info, action" />
```

nshoes: I'd also suggest removing the `eg.` from the placeholder for `name` as well. It's already a placeholder, there's no need to state "for example..."

test/nerves_hub/scripts_test.exs
```diff
@@ -82,4 +96,31 @@ defmodule NervesHub.ScriptsTest do
       assert script.last_updated_by_id == nil
     end
   end
+
+  describe "filter" do
+    test "filter on tags", %{product: product, user: user} do
```
nshoes, line 101: Could we add an `assert` for filtering on multiple tags as well?

lib/nerves_hub/managed_deployments.ex
```diff
@@ -8,6 +8,7 @@ defmodule NervesHub.ManagedDeployments do
   alias NervesHub.Deployments.InflightDeploymentCheck
   alias NervesHub.Devices
   alias NervesHub.Devices.Device
+  alias NervesHub.Filtering, as: CommonFiltering
```
nshoes, line 11: I can't suggest on a line that hasn't changed, so I'll do it here :) I'd prefer to also alias the individual filtering modules as `as: ResourceNameFiltering`. Seeing `Filtering` used by itself is painful across other modules, but I do agree with their current nested structure (`NervesHub. ResourceName.Filtering`).

lib/nerves_hub/filtering.ex
```diff
@@ -0,0 +1,50 @@
...
+      - pagination: Map of pagination options
+
+  ## Returns
+    A tuple containing:
+    - List of entries matching the query
+    - Flop metadata containing pagination information
+  """
+  @spec filter(Ecto.Query.t(), Product.t(), function(), function(), map()) ::
+          {[%Device{}] | [%DeploymentGroup{}] | [%Script{}], Flop.Meta.t()}
+  def filter(base_query, product, filter_builder, sorter, opts \\ %{}) do
```
nshoes: I don't love passing in `filter_builder` and `sorter` functions here. This feels like something that could be configured here locally instead of the caller, cleaning up a bit of code. We could key off the table from introspecting the base query:
```
> query = from NervesHub.Devices.Device
> query.from
%Ecto.Query.FromExpr{
  source: {"devices", NervesHub.Devices.Device},
  file: "iex",
  line: 3,
  as: nil,
  prefix: nil,
  params: [],
  hints: []
}
```

and then have functions that know how to build the config like:

```
defp filter_config("devices"), do: %{filter_builder: NervesHub.Devices.Filtering/2, sorter: NervesHub.Devices.sort_devices/2}
```

What do you think?

nshoes (commented):
Great work, Elin! A few nits. Tested and everything worked locally.

lib/nerves_hub_web/live/support_scripts/index-new.html.heex
```diff
@@ -35,6 +35,9 @@
           <th phx-click="sort" phx-value-sort="inserted_at" class="cursor-pointer">
             <Sorting.sort_icon text="Added on" field="inserted_at" selected_field={@current_sort} selected_direction={@sort_direction} />
           </th>
+          <th>
+            Script Tags
```
joshk: Just `Tags` is fine.

joshk (dismissed):
Love this!

I agree with Nates comments.

Great work @elinol !

lib/nerves_hub/filtering.ex
elinol: I think that makes sense. I've adressed this and your other inputs in the latest commits.

lib/nerves_hub/managed_deployments.ex
```diff
@@ -53,34 +42,37 @@ defmodule NervesHub.ManagedDeployments do
...
+
+    CommonFiltering.filter(
+      base_query,
+      product,
+      opts
+    )
   end
 
-  defp sort_deployment_groups(query, {direction, :platform}) do
+  def sort_deployment_groups(query, {direction, :platform}) do
```
elinol: I'm thinking that sorting could also be moved to a separate module, or the `Filtering` module for each resource. What do you think @nshoes and @joshk ?

nshoes: Yes! Great idea, I'm for it, for all resources.

nshoes (commented):
Changes look great! After moving the sort functions to the filtering modules this is good to go.

nshoes (approved):
🚢 🚢 🚢

nshoes:
@elinol Jon gave me some good feedback here, so I implemented [3a7cbc0](https://github.com/nerves-hub/nerves_hub_web/pull/2117/commits/3a7cbc0e26d535ab30f61beb9b74a8722f1bdc0f). Thanks for all your work here!

PR #2130, opened 2025-06-02
Ignore DB timeout when mass moving devices in and out of deployment groups

nshoes:
We use `Repo.update_all` to move and remove mass devices from deployment groups. These will timeout given enough devices, around ~80,000. I had a few different plans on how to batch this work, but it's simpler to let the DB do the work over time. This uses a single connection, and the feature isn't used often, so backing up the database isn't probable.

Looking for reasons why this is a bad idea :)


lib/nerves_hub/devices.ex
```diff
@@ -1248,7 +1248,7 @@ defmodule NervesHub.Devices do
       |> where([d], d.id in ^device_ids)
       |> where([d], d.firmware_metadata["platform"] == ^firmware.platform)
       |> where([d], d.firmware_metadata["architecture"] == ^firmware.architecture)
-      |> Repo.update_all(set: [deployment_id: id])
+      |> Repo.update_all([set: [deployment_id: id]], timeout: :infinity)
```
jjcarstens: I don't think I'd pick infinity, but maybe a couple minutes? Whatever bound is on the line of "extreme" yet "reasonable"

nshoes: Good point, I think two minutes is more than enough. Unless we ever want to move over 500k devices 😜

jjcarstens (approved):

PR #2133, opened 2025-06-04
Increase script rnner timeout to 30s

nshoes:
- Increase script run timeout to 30s
- Allow 500s to passback error messages
  - Before we'd get the Phoenix fallback "Interal Server Error" for everything
- Add helpful error message when script times out


lib/nerves_hub/scripts/runner.ex
```diff
@@ -22,9 +22,9 @@ defmodule NervesHub.Scripts.Runner do
 
   def send(device, command) do
     {:ok, pid} = start_link(device)
-    {:ok, GenServer.call(pid, {:send, command.text}, 10_000)}
+    {:ok, GenServer.call(pid, {:send, command.text}, 30_000)}
```
jjcarstens: Do we want to allow a timeout option that can also come from the API?

```elixir
def send(device, command, timeout \\ to_timeout(second: 30) do
  {:ok, pid} = start_link(device)
  {:ok, GenServer.call(pid, {:send, command.text}, timeout)}

nshoes: I like that! Will do.

nshoes: Done.

PR #2137, opened 2025-06-09
Remove need for node and NPM using spellweaver

lawik:


joshk (approved):
loooooooove this change!

does a mix call need to be added to run the spell check? or is that run during `mix test`?

nshoes (approved):
Love it! Thank you.

PR #2138, opened 2025-06-09
Set consistent order for products on landing page

lawik:
Prevents flickering between first render and live mount.


joshk (approved):
wonderful! thank you!

PR #2139, opened 2025-06-09
Add information about available deltas to deployment summary

lawik:
This should let users see that deltas exist.

Also included the size of the delta file as well as the size of the full target firmware. Until we collect stats on how much transfer we save this is at least an indication.

It only shows up for products with deltas enabled. Older generated deltas may not have the size information so they won't show it.


lawik:
If we like the concept this should have a few tests to prevent regressions. But wanted the test up for discussion first.

Image attached:
Screenshot_20250609_112111

lawik:
Would also be good to make it self-updating as deltas generate.

nshoes:
I like it!

nshoes (approved):

PR #2144, opened 2025-06-10
Fix device details deployment groups layout

nshoes:
I broke this layout in #2103; I forgot to spot-check a device with a deployment group.


nshoes:
Before
CleanShot 2025-06-10 at 08 07 53@2x


After
CleanShot 2025-06-10 at 08 06 54@2x

jjcarstens (approved):

PR #2145, opened 2025-06-11
Move delta_updatable setting to deployment group

elinol:
Adds `delta_updatable` field to `DeploymentGroup` and removes it from `Product`.

All product update functions are also deleted since Product has no more fields that can be updated (name cannot be changed, and extensions are embedded).

### DIFF
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -251,6 +251,14 @@ defmodule NervesHub.ManagedDeployments do
     end
   end
 
+  @spec toggle_delta_updates(DeploymentGroup.t()) ::
+          {:ok, DeploymentGroup.t()} | {:error, Changeset.t()}
+  def toggle_delta_updates(deployment_group),
+    do:
+      update_deployment_group(deployment_group, %{
+        delta_updatable: !deployment_group.delta_updatable
+      })
+
   @doc """
   Update a deployment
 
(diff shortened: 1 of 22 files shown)


priv/repo/migrations/20250611082642_move_delta_updatable_to_deployment_group.exs
```diff
@@ -0,0 +1,14 @@
+defmodule NervesHub.Repo.Migrations.MoveDeltaUpdatableToDeploymentGroup do
+  use Ecto.Migration
+
+  def change do
+    alter table(:deployments) do
+      add :delta_updatable, :boolean, default: false
```
nshoes: To further increase database integrity
```suggestion
      add :delta_updatable, :boolean, default: false, null: false
```

lib/nerves_hub/managed_deployments.ex
```diff
@@ -251,6 +251,14 @@ defmodule NervesHub.ManagedDeployments do
     end
   end
 
+  @spec toggle_delta_updates(DeploymentGroup.t()) ::
+          {:ok, DeploymentGroup.t()} | {:error, Changeset.t()}
+  def toggle_delta_updates(deployment_group),
```
nshoes, line 256: We should be generating deltas here with `FirmwareDeltaBuilder.start/2`, right? It's in the previous update product workflow.

elinol: Yes! Missed to add the case to `maybe_trigger_delta_generation`, done in latest commit, if I understand the flow correctly.

nshoes: NIce! Looks good.

nshoes (approved):

PR #2146, opened 2025-06-12
Fix metrics time frame not updating when clicking buttons

### LINKED ISSUE #2104: Health over time reports/graphs are not available for offline devices
**Describe the bug**
The Health over time reports/graphs are not available for devices which are unavailable even if they'd have sent data in the past for when they were available.

**To Reproduce**
1. Collect health metrics data from a device over a period of time
2. Check that the health graphs are available.
3. Make the device go offline or become unavailable. 
4. Check that the Health over time graphs won't be available.

**Expected behavior**
The health over time data should be available even if device is offline. It's not possible to collect metrics when the device was offline but the the previously store metrics should be shown.

Currently nothing happens when I update the time period by clicking on  "1 day" or "7 days"

**Screenshots**

Image

Image

 - OS: MacOS 15.3.2
 - Browser: Firefox
 - Version: 136.0.3

**Additional context**
This is from a self-hosted version of nerves_hub_web.

elinol:
Fixes #2104 where time frame was not updating when clicking buttons in device health tab, because `
@default_time_frame` was always assigned on `push_patch`.

When adding tests for this I found out that charts did not update when health is reported.
I don't know if that was intentional for some reason, I'll revert the fix included in this PR if so.

### DIFF
--- a/test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
+++ b/test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
@@ -1,0 +1,194 @@
+defmodule NervesHubWeb.Live.NewUi.Devices.HealthTabTest do
+  use NervesHubWeb.ConnCase.Browser, async: false
+  use Phoenix.HTML
+
+  alias NervesHub.Devices.DeviceMetric
+
+  alias NervesHub.Repo
+  alias NervesHubWeb.Endpoint
+  alias Phoenix.Socket.Broadcast
+
+  @metrics %{
+    "cpu_temp" => 41.381,
+    "load_15min" => 0.06,
+    "load_1min" => 0.55,
+    "load_5min" => 0.15,
+    "mem_size_mb" => 7892,
+    "mem_used_mb" => 172,
+    "mem_used_percent" => 2
+  }
+
+  setup %{conn: conn, device: device} do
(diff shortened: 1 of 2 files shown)


test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
```diff
@@ -0,0 +1,194 @@
...
+  test "assert time frame shows correct metrics - within 7 days", %{
+    conn: conn,
+    org: org,
+    product: product,
+    device: device
+  } do
+    timestamp =
+      DateTime.now!("Etc/UTC")
+      # Just outside span for 7 days
+      |> DateTime.add(-7, :day)
```
nshoes, line 98: Do you think this opens up the possibility of a flaky test?

nshoes (approved):
This LGTM! The test additions are great.

I believe your fix to update charts is valid.

test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
elinol: We have been using the exact same test for the old UI and it has been consistent so I think it's ok. Would tou rather like to add more time?

nshoes: No, not stressed about it now :)

PR #2150, opened 2025-06-13
Fix deleting and destroying devices

nshoes:
Destroying devices in the UI is impossible with the addition of filtering out soft-deleted devices in `Devices. get_device_by_identifier!/3`. I just wanted to unblock the delete -> destroy device workflow in the new UI, but this function is used in a few places and I think we need to double-check what we want to do:

- `NervesHubWeb.Live.Devices.Settings` on mount:
  - Okay if a soft-deleted device is returned here, obviously 
- `NervesHubWeb.Live.Devices.DeviceHealth` on mount:
  - This is a tab when viewing a device in the new UI; maybe we should add a redirect to the settings page? It makes sense to only allow the Settings tab to be viewable on a soft-deleted device.
- `NervesHubWeb.Live.Devices.Show` on mount:
  - Same as above
- `NervesHubWeb.API.Plugs.Device`, scoped to `/devices/:identifier` routes:
  - You can soft-delete through the `/delete` route, but there's no route to destroy. I vote that we either 1.) convert this route to a destroy action and don't allow soft-deleted devices to be queryable or 2.) add a destroy action and allow soft-deleted devices to be queryable.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -235,7 +235,6 @@ defmodule NervesHub.Devices do
   def get_device_by_identifier!(org, identifier, preload_assoc \\ nil)
       when is_binary(identifier) do
     get_device_by_identifier_query(org, identifier, preload_assoc)
-    |> Repo.exclude_deleted()
     |> Repo.one!()
   end
 
--- a/test/nerves_hub_web/live/new_ui/devices/settings_tab_test.exs
+++ b/test/nerves_hub_web/live/new_ui/devices/settings_tab_test.exs
@@ -3,32 +3,60 @@ defmodule NervesHubWeb.NewUi.Devices.SettingsTabTest do
 
   alias NervesHub.Devices
   alias NervesHubWeb.Components.Utils
+  alias NervesHub.Repo
 
   setup %{conn: conn} do
     [conn: init_test_session(conn, %{"new_ui" => true})]
   end
 
-  describe "certificates" do
-    test "download certificate", %{
-      conn: conn,
-      org: org,
-      product: product,
-      device: device
-    } do
-      [certificate] = Devices.get_device_certificates(device)
+  test "download certificate", %{
+    conn: conn,
+    org: org,
+    product: product,
+    device: device
+  } do
+    [certificate] = Devices.get_device_certificates(device)
 
-      result =
-        conn
-        |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}/settingz")
-        |> assert_has("div", text: "Certificates")
-        |> assert_has("div", text: "Serial: #{Utils.format_serial(certificate.serial)}")
-        |> unwrap(fn view ->
-          view
-          |> element("a[download]")
-          |> render_click()
-        end)
+    result =
+      conn
+      |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}/settingz")
+      |> assert_has("div", text: "Certificates")
+      |> assert_has("div", text: "Serial: #{Utils.format_serial(certificate.serial)}")
+      |> unwrap(fn view ->
+        view
+        |> element("a[download]")
+        |> render_click()
+      end)
 
-      assert result.conn.resp_body =~ "-----BEGIN CERTIFICATE-----"
-    end
+    assert result.conn.resp_body =~ "-----BEGIN CERTIFICATE-----"
+  end
+
+  test "deleting device", %{
+    conn: conn,
+    org: org,
+    product: product,
+    device: device
+  } do
+    conn
+    |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}/settingz")
+    |> click_button("Delete device")
+    |> assert_has("div", text: "Device is deleted and must be restored to use.")
+
+    assert Repo.reload(device) |> Map.get(:deleted_at)
+  end
+
+  test "destroying device", %{
+    conn: conn,
+    org: org,
+    product: product,
+    device: device
+  } do
+    conn
+    |> visit("/org/#{org.name}/#{product.name}/devices/#{device.identifier}/settingz")
+    |> click_button("Delete device")
+    |> click_button("Permanently delete device")
+    |> assert_has("div", text: "Device permanently destroyed successfully.")
+
+    refute Repo.reload(device)
   end
 end


nshoes:
image

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -749,8 +749,15 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
       {:error, changeset} ->
         Logger.info("Couldn't update device.priority_updates: #{inspect(changeset)}")
 
+        error =
+          if Keyword.has_key?(changeset.errors, :deleted_at) do
+            "Device cannot be updated because it has been deleted. Please restore the device to make changes."
+          else
+            "There was an issue updating the device, please check the logs."
```
joshk:
```suggestion
            "There was an issue updating the device. Please contact support if this happens again."
```

lib/nerves_hub_web/components/device_page/settings_tab.ex
```diff
@@ -311,8 +307,15 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
         |> halt()
 
       {:error, :update_with_audit, changeset, _} ->
+        error =
+          if Keyword.has_key?(changeset.errors, :deleted_at) do
+            "Device cannot be updated because it has been deleted. Please restore the device to make changes."
+          else
+            "We couldn't save your changes."
```
joshk:
```suggestion
            "We couldn't save your changes. Please contact support if this happens again."
```

lib/nerves_hub/devices.ex
```diff
@@ -233,7 +233,6 @@ defmodule NervesHub.Devices do
   def get_device_by_identifier!(org, identifier, preload_assoc \\ nil)
```
joshk, line 232: this function is also used by `NervesHubWeb.API.Plugs.Device`, do we need to account for that?

nshoes: No, I don't think so. A device can no longer be updated when it's soft-deleted so I don't think we'll have any unsavory side-effects from allowing endpoints to retrieve the device.

joshk: Do we at least show in the Device payloads that it has been soft deleted?

joshk:
> > Do we also want to add a filter option to the device list so soft deleted devices can be found?
> 
> Good idea, I'll implement.

Awesome, thank you!

joshk:
> @joshk done!
> 
> CleanShot 2025-07-23 at 14 44 02 [ CleanShot 2025-07-23 at 14 44 02 ](https://private-user-images.githubusercontent.com/4118389/470008371-a33126d6-ef01-40f9-b286-c5c6dfadbdef.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTMzMDk4OTgsIm5iZiI6MTc1MzMwOTU5OCwicGF0aCI6Ii80MTE4Mzg5LzQ3MDAwODM3MS1hMzMxMjZkNi1lZjAxLTQwZjktYjI4Ni1jNWM2ZGZhZGJkZWYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDcyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA3MjNUMjIyNjM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MjFhMDAxNDI1ZDQyZGRkNmM1MGIxZDQ5MjBjYWY4NmE2YzNhNWE4Njk1YmNmMzkyMmRjZTE0MTI3M2Y1ZTg4MCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.QP6E7EFGYEVCZXxD78hJpEAqMbCs_TnjknL-yZIStV0) [ ](https://private-user-images.githubusercontent.com/4118389/470008371-a33126d6-ef01-40f9-b286-c5c6dfadbdef.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NTMzMDk4OTgsIm5iZiI6MTc1MzMwOTU5OCwicGF0aCI6Ii80MTE4Mzg5LzQ3MDAwODM3MS1hMzMxMjZkNi1lZjAxLTQwZjktYjI4Ni1jNWM2ZGZhZGJkZWYuZ2lmP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDcyMyUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA3MjNUMjIyNjM4WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MjFhMDAxNDI1ZDQyZGRkNmM1MGIxZDQ5MjBjYWY4NmE2YzNhNWE4Njk1YmNmMzkyMmRjZTE0MTI3M2Y1ZTg4MCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.QP6E7EFGYEVCZXxD78hJpEAqMbCs_TnjknL-yZIStV0)
> 
> I opted to display the soft-deleted devices first when filtering. I like that more than just showing deleted or not-deleted devices. Thoughts?

Love your work!

One idea (just an idea), how about "Include soft deleted devices?" : "yes", "no", "only soft deleted devices"?

lib/nerves_hub/devices.ex
nshoes: Good point, we do not, I'll add.

PR #2157, opened 2025-06-17
Allow firmware uploads from products named with spaces

drteeth:
Trying to upload some firmware to NH 2.x, I was seeing this:

```shell
$ nh firmware publish some.fw 
NervesHub Host: example.com
Organization:   ACME
NervesHub product: Big Anvil
------------
Organization: ACME
  product:      Big Anvil
  version:      0.1.18
  platform:     bbb
  architecture: arm
  uuid:         229764bc-938a-5509-d311-060497eb3a55
Publish Firmware? yN y  
|==================================================| 100% (23 / 23) MB
Unhandled error: {:error, nil}
```

Looking into the server logs:
```elixir
(Phoenix.Template.UndefinedError) Could not render "413.html" for NervesHubWeb.ErrorView...
*snip*
%{
  reason: %Plug.Parsers.RequestTooLargeError{
    message: "the request is too large...",
    plug_status: 413
  },
  status: 413
...
}
```

Tracked the problem here:
https://github.com/nerves-hub/nerves_hub_web/blob/f1aa6f9db407aaa2d7e15a6b44429ee4a6697e4a/lib/nerves_hub_web/dynamic_config_multipart.ex#L30

If org or product name has non-word characters (and not '-'), then they don't get the larger limit for firmware uploads.

This PR fixes that by pattern matching on `path_info`. If this isn't safe somehow, we can go back and adjust the regex solution.

### DIFF
--- a/lib/nerves_hub_web/dynamic_config_multipart.ex
+++ b/lib/nerves_hub_web/dynamic_config_multipart.ex
@@ -27,10 +27,12 @@ defmodule NervesHubWeb.DynamicConfigMultipart do
   end
 
   defp max_file_size(conn) do
-    if String.match?(conn.request_path, ~r/^\/api\/orgs\/[-\w]+\/products\/[-\w]+\/firmwares$/) do
-      Application.get_env(:nerves_hub, NervesHub.Firmwares.Upload, [])[:max_size]
-    else
-      1_000_000
+    case conn.path_info do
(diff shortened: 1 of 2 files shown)


joshk (approved):
Great find!

Love the `conn.path_info` check!

joshk:
Thanks for the PR @drteeth 

And big thanks for the test suite additions!

PR #2163, opened 2025-06-18
Check for delta updates based on deployment group, not product

nshoes:


jjcarstens (approved):
Neat! Frank and I were just talking about this today!

I think we also need a way to turn that setting on at the deployment group level. I'm not seeing that currently in our setup? (Could be this or next PR)

joshk (approved):
❤️

joshk:
> Neat! Frank and I were just talking about this today!
> 
> I think we also need a way to turn that setting on at the deployment group level. I'm not seeing that currently in our setup? (Could be this or next PR)

https://github.com/nerves-hub/nerves_hub_web/pull/2145

Issue #2136, opened 2025-06-06
Make delta delivery really safe


lawik:
To be sure that a delta update is safe to do we need to line up a few things:

- Device fwup version
- source firmware encryption usage
- target firmware delta encryption support

The delta encryption support should land as fwup v1.13.

What we already do?

- Check if deployment group had delta updates enabled.
- Check that the target firmware has delta markers for some of the file resources by parsing `meta.conf`. If no markers, no point in making a delta.

Things we should do, in order of importance:

- [x] Skip providing deltas to devices with fwup >= 1.13.0. This will make it safe for SmartRent (and everyone else) to use as it has delta + encrypted filesystem support. But it will exclude usage by some users that could see the delta benefit for unencrypted devices.
- [x] Check for encryption in source firmware, if present, require fwup >= 1.13.0 for deltas.
- [x] Check that for resources where source firmware has encryption, target firmware also has encryption configuration, otherwise don't provide delta.
- [x] Check for FAT delta markers in target firmware, if present, require fwup >= 1.10.0 for deltas.
- [x] Check for raw delta markers in target firmware, if present, require fwup >= 1.6.0 for deltas.

PR #2164, opened 2025-06-19
Default deployment_group.delta_updates_enabled to false in tests to reflect the default changeset value

nshoes:
This allows the option to hide delta updates while editing a deployment group's settings via an environment variable. I'd like the ability to keep it off for SmartRent until https://github.com/nerves-hub/nerves_hub_web/issues/2136 is complete. I also changed the default to `deployment_group.delta_updates` to `false` in `Fixtures` to properly reflect their default changeset value.


jjcarstens (approved):
Fine by me. But I would also be totally happy leaving this out. I think we can manage not enabling for now if this gets to us before the other improvements

nshoes:
> I think this might be superfluous as we didn't have this for delta updates when the setting was on the product level.

Hm, good point, I hadn't thought of that. I no longer like this 😆

PR #2178, opened 2025-06-24
Fix device status updates while viewing device index

nshoes:
When we moved to a LiveView async result for device statuses, the `handle_info`s for `connection:status` and `connection:update` broadcasts weren't updated. This caused them to no-op and never update the online status.

This work checks the proper key on `socket.assigns.device_statuses`, updates the state with a proper `AsyncResult`, and adds a test for each event.


lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -543,9 +543,16 @@ defmodule NervesHubWeb.Live.Devices.Index do
 
   defp update_device_statuses(socket, payload) do
     # Only sync devices currently on display
-    if Map.has_key?(socket.assigns.device_statuses, payload.device_id) do
-      device_statuses = Map.put(socket.assigns.device_statuses, payload.device_id, payload.status)
-      {:noreply, assign(socket, :device_statuses, device_statuses)}
+    if Map.has_key?(socket.assigns.device_statuses.result, payload.device_id) do
+      updated_device_statuses =
+        Map.put(socket.assigns.device_statuses.result, payload.device_id, payload.status)
```
lawik: Think you can get rid of the conditional entirely by using `Map.replace/3` :)
https://hexdocs.pm/elixir/Map.html#replace/3

lawik: Unless this is expected to receive a lot of irrelevant updates.

lawik (approved):
Looks really good. I have one suggested change but not important.

lib/nerves_hub_web/live/devices/index.ex
nshoes: Ooooh good callout! And TIL! Sweet.

joshk:
Thank you for catching and fixing this!

PR #2179, opened 2025-06-25
Safer delta updates

lawik:
Changes to make delta updates a lot safer for anyone running older versions of fwup.

- Adds detailed feature and version checking for deltas, FAT-support and encryption
- Adds storage of update tool (fwup) metadata
- Rework the plumbing of when deltas and full updates get sent

NervesHub now uses the Confuse library to pull information about what features of fwup are used and Confuse knows what fwup versions are needed for those features. Essentially we needed to know "given the fwup version on this device currently, can it support a delta update?" we can also figure out if there could be a problem doing a full update but that's not particularly in scope here.

This means fwup minimum version numbers mostly show up in Confuse rather than NervesHub.

As we want to save the information from Confuse while we have the firmware file ready for upload I added `tool_metadata` to both `Firmware` and `FirmwareDelta` schemas. I did not call it fwup_metadata as there are plans to add support for other tools and since I was touching this code I might as well save us some work for the future.

The `DeltaUpdater` grew some more responsibilities, such as figuring out if a delta update is allowed for this particular tool (fwup) and it is now named `UpdateTool` to reflect the wider responsibilites. Some callbacks were added for figuring out when delta or full update is appropriate. Most of the callbacks got updated signatures to allow more data to be sent back for storage. We had a messy way of doing this before to smuggle some metadata into upload_metadata I believe. This makes most of those fields on `Firmware` and `FirmwareDelta`.

Some of this metadata would be very nice to backfill. That is not covered in this work but we should have safe defaults and we can do that in a separate PR.

Since the on/off switch for deltas is already moved to the DeploymentGroup we have a way to control Delta vs Full updating of many devices in bulk.

This also adds a button to allow choosing to send a delta or a full update from the device details view.

Unzip is added as a dependency to perform some unpacking of `meta.conf`. I expect we'll use it more in the future to slurp meta.conf from uploaded archives without needing to download the whole thing.


lawik:
To address this: https://github.com/nerves-hub/nerves_hub_web/issues/2136

lawik:
Will squash but in case the commits are helpful in understanding what I've been up to I'll keep them for a minute.

lawik:
Sorry about this being on the kind of big side. But as I needed to store information captured specifically from `fwup` and its config file I wanted to separate that out into some general fields that most firmware update tools will care about (eg. `tool_delta_required_version` and `tool_full_required_version` should be broadly applicable) and then have a more flexible part `tool_metadata` to store more specific things in.

This prepares the way for implementing support for future other update tools but doesn't do any of the implementation.

DeltaUpdater or whatever the behaviour was called is now UpdateTool to not be all about deltas. Some of the fwup-specific stuff moved and treated as generic but no real change.

Will update description to properly outline the purpose and the changes.

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+        |> Unzip.new()
+
+      unzip
+      |> Unzip.file_stream!("meta.conf")
+      |> Enum.into(stream, &IO.iodata_to_binary/1)
+
+      {:ok, path}
+    rescue
+      e ->
+        Logger.error("Extracting meta.conf failed due to: #{inspect(e)}")
```
nshoes: Could we log `firmware.uuid` here as well?

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+
+  defp dl_archive(firmware) do
+    try do
+      {:ok, url} = firmware_upload_config().download_file(firmware)
+      archive_path = Path.join(System.tmp_dir(), "nh-#{firmware.uuid}.fw")
+      dl!(url, archive_path)
+      {:ok, archive_path}
+    rescue
+      e ->
+        Logger.error("Downloading firmware failed due to: #{inspect(e)}")
```
nshoes: Could we log firmware.uuid here as well?

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+       }}
+    else
+      err ->
+        err
+    end
+  end
+
+  @impl NervesHub.Firmwares.UpdateTool
+  def get_firmware_metadata_from_upload(firmware) do
+    case dl_archive(firmware) do
```
nshoes:
```suggestion
    case download_firmware(firmware) do
```

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+    rescue
+      e ->
+        Logger.error("Downloading firmware failed due to: #{inspect(e)}")
+        {:error, :download_firmware_failed}
+    end
+  end
+
+  defp firmware_upload_config(), do: Application.fetch_env!(:nerves_hub, :firmware_upload)
+
+  defp dl!(url, filepath) do
```
nshoes, line 341:
```suggestion
  defp download!(url, filepath) do
```

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+    {_, 0} = System.cmd("unzip", ["-qq", source_path, "-d", source_work_dir], env: [])
+    {_, 0} = System.cmd("unzip", ["-qq", target_path, "-d", target_work_dir], env: [])
+
+    {:ok, deltas} = Confuse.Fwup.get_delta_files(Path.join(target_work_dir, "meta.conf"))
+
+    all_delta_files =
+      Enum.flat_map(deltas, fn {_k, files} ->
+        files
+      end)
+      |> Enum.uniq()
```
nshoes:
```suggestion
      deltas
      |> Enum.flat_map(fn {_k, files} ->
        files
      end)
      |> Enum.uniq()
```

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -0,0 +1,289 @@
...
+      |> Path.join("meta.conf")
+
+    File.write!(path, meta)
+    {:ok, feature_usage} = Confuse.Fwup.get_feature_usage(path)
+
+    feature_usage.raw_deltas? or feature_usage.fat_deltas?
+  end
+
+  @very_safe_version "1.13.0"
+  @oldest_version Version.parse!("0.2.0")
```
nshoes: Can we move these to the top of the module?

test/nerves_hub/firmwares/update_tool_test.exs
```diff
@@ -1,14 +1,19 @@
...
   @fat File.read!("test/fixtures/fwup/fat.conf")
   @mixed File.read!("test/fixtures/fwup/mixed.conf")
   @mixed_no_deltas File.read!("test/fixtures/fwup/mixed-no-deltas.conf")
   @pi_style File.read!("test/fixtures/fwup/pi-style.conf")
   @pi_style_delta File.read!("test/fixtures/fwup/pi-style-delta.conf")
+  # TODO: Add fwup conf for encrypted disk with deltas, and tests to avoid generating
+  #       deltas if encryption doesn't match both sides
+
+  # TODO: Add test for UpdateTool.device_update_type default implementation in Fwup
+  #       replaces a test removed with Devices.delta_updatable? becoming private
```
nshoes: I feel strongly that this is the time and place to write these tests. Could I offer myself as tribute? :)

nshoes (commented):
First pass looks generally good to me! I'm going to run through this locally.

lib/nerves_hub/firmwares/update_tool/fwup.ex
joshk: If this is an error, should we send it to Sentry?

config/config.exs
```diff
@@ -112,5 +112,7 @@ config :ueberauth, Ueberauth,
     google: {Ueberauth.Strategy.Google, [default_scope: "email profile openid"]}
   ]
 
+config :bun, :version, "1.2.2"
```
joshk: Can you elaborate a little on what this is for? Are we installing bun?

lib/nerves_hub/devices.ex
```diff
@@ -1701,54 +1706,93 @@ defmodule NervesHub.Devices do
...
-           Firmwares.get_firmware_delta_by_source_and_target(device_firmware, target_firmware) do
+    with %Firmware{} = source <- Firmwares.get_firmware_by_uuid(source_uuid),
+         true <- delta_updatable?(device, target),
+         {:ok, delta} <- Firmwares.get_firmware_delta_by_source_and_target(source, target) do
       Logger.info(
-        "Delivering firmware delta between #{device_firmware_uuid} and #{target_firmware.uuid}...",
-        from_firmware: device_firmware.uuid,
-        to_firmware: target_firmware.uuid,
-        delta: firmware_delta.id
+        "Delivering firmware delta between #{source_uuid} and #{target.uuid}...",
```
joshk: When it comes to log messages, since we have `source_uuid` and `target.uuid` in the metadata, we don't need to include this info in the log message. This would shorten the message to just `Delivering firmware delta` which is much clearer when combined with the metadata.

lib/nerves_hub/devices.ex
```diff
@@ -1701,54 +1706,93 @@ defmodule NervesHub.Devices do
...
 
-      Firmwares.get_firmware_url(firmware_delta)
+      Firmwares.get_firmware_url(delta)
     else
-      _ ->
+      false ->
+        Logger.info(
+          "Delta not supported. Delivering full firmware between #{source_uuid} and #{target.uuid}...",
+          source_firmware: source_uuid,
+          _firmware: target.uuid
```
joshk: is there a particular reason for the metadata key of `_firmware`? is this supposed to be `target_firmware`?

Same comment as above, we can shorten the log message to `Delta not supported. Delivering full firmware.`

lib/nerves_hub/devices.ex
```diff
@@ -1701,54 +1706,93 @@ defmodule NervesHub.Devices do
...
+        )
+
+        Firmwares.get_firmware_url(target)
+
+      {:error, :not_found} ->
+        Logger.info(
+          "No delta found. Delivering full firmware between #{source_uuid} and #{target.uuid}...",
+          source_firmware: source_uuid,
+          _firmware: target.uuid
```
joshk: Same comment as above.

lib/nerves_hub/devices.ex
```diff
@@ -1701,54 +1706,93 @@ defmodule NervesHub.Devices do
...
         # it is best to deliver a firmware even if we can't get a delta.
         # This could be typical for some manual deployments.
-        Logger.info(
-          "Delivering full firmware between #{device_firmware_uuid} and #{target_firmware.uuid}...",
-          source_firmware: device_firmware_uuid,
-          _firmware: target_firmware.uuid
+        Logger.warning(
+          "Delivering full firmware for unusual reason #{inspect(err)}. Between #{source_uuid} and #{target.uuid}...",
+          source_firmware: source_uuid,
+          _firmware: target.uuid
```
joshk: See above. 

I see there are a few more log messages that follow this pattern. Please accept this comment as a blanket 'See above'.

config/config.exs
nshoes: Gah, I meant to ask this too then got distracted by all the code 😈

lawik: It is part of the spellweaver library. Unfortunately it needs a bit of config to be set or it prints a warning.

lib/nerves_hub/devices.ex
lawik: I tend to look at the logs as printed in various consoles, do they actually include the metadata there?
I am okay with the suggestion but generally prefer something I can read straight :)

lib/nerves_hub/devices.ex
lawik: `_firmware` already existed I believe so I didn't want to invent a new separate one that means people need to do lookups on more fields. Happy to normalize that in some way.

lib/nerves_hub/firmwares/update_tool/fwup.ex
lawik: So logged errors don't go to sentry by default? Fair enough, haven't really thought about that.

And we need to pin down the practice of whether we include UUIDs in the log message or just the metadata I suppose.

test/nerves_hub/firmwares/update_tool_test.exs
lawik: I can do it. Less coordination :)

lib/nerves_hub/firmwares/update_tool/fwup.ex
joshk: I don't believe `Logger.error` messages go to Sentry, instead we need to send to Sentry (which we have a helper for)

config/config.exs
joshk: Ahhhhh, cool, thats for the explanation.

lib/nerves_hub/devices.ex
joshk: Our logs do include metadata. The metadata makes searching easier. I'm a big fan of key/value logging, and keeping messages short.

lawik:
I'll try Josh's way of shorter log messages with metadata and see how that treats us :)

joshk:
> Images attached.

Thanks for sharing the images. Is the "Send delta update" disabled or hidden if a Delta doesn't exist? Will this update depending on the firmware version you select?

As for the Deltas list for the Deployment Group, does this show every delta applicable to the updated firmware version?

lawik:
Yeah, it just shows "Send update" if deltas aren't involved and shows the two buttons with explicit labels if a delta is available: https://github.com/nerves-hub/nerves_hub_web/pull/2179/files#diff-63faf641e60b9135383e17fd371a6c94961f9edbfe30a5643e1a7921b57ff677R385-R413

For the deltas list. Yeah, so it shows all deltas we've generated for getting to the target firmware version. And we only generate deltas for source firmwares that are currently active in the fleet. So it should show the relevant source versions to indicate that there are deltas for them.

We still have a fundamental race-condition between delta generation and firmware delivery but solving that relates to the Releases ideas and wasn't in scope for this PR. You can trick it by pausing the deployment group, change the firmware, wait for deltas to show up and only after you see the deltas you want resume it. So there is a way. Or you let it ship a couple of full versions if it is too fast 🤷🏻

joshk:
Ah cool, thanks for the explanations.

All in all, I'm happy with the PR and think we can improve and iterate.

We can also deploy this to NC for testing.

lib/nerves_hub/firmwares/update_tool/fwup.ex
lawik: Added :)

test/nerves_hub/firmwares/update_tool_test.exs
lawik: Found a mistake in my Confuse library when writing this up so will deal with that to get some better firmware-to-firmware compatibility checking.

lawik: This ground has been covered in those updates. The Devices.delta_updatable? logic is mostly in Confuse at this point which is heavily covered. 

I also added fundamental tests to cover the updatable and non-updatable functionality, showing that it works. But it does not yet cover all the permutations of how a device's state (fwup version, current firmware) and the target firmware can affect the possibility of a delta update. Lots of it is covered. Not all of it.

lawik:
Squashed and rebased. I think I've covered most of the feedback now :)

test/nerves_hub/firmwares/update_tool_test.exs
nshoes: Awesome! Thank you.

lawik:
> Could you make an issue detailing this?

Sure :)

https://github.com/nerves-hub/nerves_hub_web/issues/2206

nshoes (approved):
This LGTM, thanks for the effort Lars! 

@joshk let us know how NervesCloud deployment and usage goes.

joshk (approved):
Lets go!

PR #2185, opened 2025-06-26
Don't close batch-update selects when LiveView updates

nshoes:


nshoes:
Before:
CleanShot 2025-06-26 at 11 47 28


After:
CleanShot 2025-06-26 at 11 46 17

jjcarstens (approved):

joshk:
Thank you for this fix!

PR #2187, opened 2025-07-02
Stream devices query when creating CSV

nshoes:
Without streaming, this query caused an OOM crash when loading hundreds of thousands of devices into memory.

Note that there are already tests for `NervesHub.Products.devices_csv/1` in `NervesHub.ProductsTest`.

### DIFF
--- a/lib/nerves_hub/products.ex
+++ b/lib/nerves_hub/products.ex
@@ -203,7 +203,9 @@ defmodule NervesHub.Products do
         Device
         |> where([d], d.product_id == ^product.id)
         |> Repo.exclude_deleted()
-        |> Repo.stream(max_rows: 10_000)
+        |> Repo.stream(max_rows: 100)
+        |> Stream.chunk_every(100)
+        |> Stream.flat_map(&Repo.preload(&1, :device_certificates))
         |> Stream.map(&device_csv_line(&1, product))
         |> Enum.to_list()
       end)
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/products.ex
```diff
@@ -215,6 +225,8 @@ defmodule NervesHub.Products do
   end
 
   defp format_device_certificates(device) do
+    device = Repo.preload(device, :device_certificates)
```
jjcarstens: Why not preload this during the stream of device lookup?

joshk:
Great fix! Same comments as Jon.

lib/nerves_hub/products.ex
nshoes: Can't preload with `Repo.stream` 😭 https://github.com/elixir-ecto/ecto/issues/2424#issuecomment-365985511

joshk: Ohhhhh, Interesting!

I came across https://github.com/danielberkompas/elasticsearch-elixir/issues/54#issuecomment-537691898 which would allow you to reduce the number of times we call preload. 

It might be best to use a `max_rows` of 100, possibly even less.

nshoes: Dope, good find. Implemented!

joshk (approved):
Awesome!

Love it!

PR #2189, opened 2025-07-03
Log savings from delta transfers with ClickHouse

lawik:
Screenshot_20250808_104733


lawik:
Found a missing thing and added that, now need to finish fixing the tests. But it can be reviewed.

lawik:
Adding some UI that is currently in-progress.

lawik:
Rebased and tidy. All ready for you @joshk :)

lib/nerves_hub/devices/update_stats.ex
```diff
@@ -0,0 +1,184 @@
...
+        firmware_delta.size
+      else
+        target_size
+      end
+
+    saved = target_size - delta_size
+    log_stat(device, deployment_group, target, :fwup_delta, delta_size, saved)
+  end
+
+  defp log_stat(
```
nshoes, line 135: Can we add a spec to this?

lib/nerves_hub/devices.ex
```diff
@@ -736,21 +737,23 @@ defmodule NervesHub.Devices do
   @doc """
   Resolve an update for the device's deployment
   """
-  @spec resolve_update(Device.t()) :: UpdatePayload.t()
-  def resolve_update(%Device{status: :registered}), do: %UpdatePayload{update_available: false}
+  @spec resolve_update(Device.t(), keyword()) :: UpdatePayload.t()
```
nshoes, line 741: Can we define what we respond to in `opts`? I find it helps readability, especially around function calls that are three or more deep.

nshoes: Or in `doc`, like you did below :)

nshoes (commented):
This generally LGTM. Can you post some UI screenshots? :)

lib/nerves_hub/devices.ex
lawik: done

lib/nerves_hub/devices/update_stats.ex
lawik: done

lawik:
Screenshotted added, notes addressed ✨

lawik:
Removed spec again, really awkward dialyzer thing.

lib/nerves_hub/devices/update_stats.ex
lawik: undone again, dialyzer was throwing a very opaque fit

nshoes: if I had a nickel...

lawik:
No reason it couldn't but between an column store and a transactional DB I'd put this data in the column store.

You can switch it to Postgres for now. I don't feel strongly about it. It shouldn't be a ton of volume. Feels a bit technically silly but pragmatism is probably the better choice 🤷🏻

nshoes:
Closing in favor of #2295

PR #2199, opened 2025-07-08
Set a default table engine suited for hosted Clickhouse installs

joshk:
Otherwise the creation of the schema table fails, causing the application to fail to start.

### DIFF
--- a/lib/nerves_hub/application.ex
+++ b/lib/nerves_hub/application.ex
@@ -79,15 +79,13 @@ defmodule NervesHub.Application do
       Supervisor.child_spec(
         {Ecto.Migrator,
          repos: [NervesHub.Repo],
-         skip: Application.get_env(:nerves_hub, :database_auto_migrator) != true,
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/application.ex
```diff
@@ -86,7 +86,8 @@ defmodule NervesHub.Application do
       Supervisor.child_spec(
         {Ecto.Migrator,
          repos: [NervesHub.AnalyticsRepo],
-         skip: Application.get_env(:nerves_hub, :analytics_enabled) != true},
+         skip: Application.get_env(:nerves_hub, :analytics_auto_migrator) != true,
+         id: NervesHub.AnalyticsRepoMigrator},
```
nshoes: What is this `id` key needed?

joshk: I was just copying the same `Ecto.Migrator` definition used for the other Migrator.

joshk: Fixed!

PR #2203, opened 2025-07-08
Fix rendering tags with invalid deployment group params so errors display properly

nshoes:
I spent way too much time trying to figure out why this was happening, but I've found the solution. When creating a deployment group with invalid params, the formatting call to `tags_to_string/1` doesn't work as expected. It will simply return the `deployment_group.conditions` map, thus blowing up `NervesHubWeb.CoreComponents.input/1` downstream. 

If you feel inclined, please feel free to nerd snipe me 🫡


joshk (approved):
Love it!

PR #2212, opened 2025-07-10
Added support for Google Cloud Storage

CJRChang:
This PR adds Google Cloud Storage support as an option for firmware storage.

Includes support for setting credentials via `GOOGLE_APPLICATION_CREDENTIALS` and `GOOGLE_APPLICATION_CREDENTIALS_JSON`.


joshk:
Hi @CJRChang

Thanks for creating the PR. 

I just wanted to check, have you looked into using GCS via its AWS Signature support?

Our file storage code works with AWS, Tigris, Cloudflare, and any provider that supports the S3 API.

I haven't tested this with GCS, but I have previously used the GCS S3 API at another company.

It would be preferable not to add another storage provider, as that code is a bit of a mess and needs some love (there is duplication that needs to be addressed).

CJRChang:
Thanks for the feedback @joshk
I haven't tried the GCS S3 API since the auth mechanisms available to me are more GCP specific.

No issues if this isn't upstreamed if there's an impending refactor, I'll just have to deal with it downstream later.

PR #2216, opened 2025-07-11
Increase test coverage for ManagedDeployments, add excoveralls

nshoes:
This PR:

- Adds `excoveralls`. I was having a hard time with `mix test.coverage` always failing with `emfile` errors. I couldn't find the root issue nor a solution, so I switched to a more dependable tool.
- Adds tests coverage to `ManagedDeployments`
- Removes a few unused functions that `mix_unused` missed (it's not perfect)


joshk (approved):
Love it

PR #2217, opened 2025-07-14
Add an external device channel for device/firmware updates

kybishop:
Adds a channel for device updates that external services can listen on. Currently this just adds support for firmware update progress, but we can add whatever else we want to expose as well.

The socket can be listened to on `/api-socket/websocket`

### DIFF
--- a/lib/nerves_hub_web/channels/api_socket.ex
+++ b/lib/nerves_hub_web/channels/api_socket.ex
@@ -1,0 +1,32 @@
+defmodule NervesHubWeb.APISocket do
+  use Phoenix.Socket
+
+  alias NervesHub.Accounts
+
+  channel("device:*", NervesHubWeb.ExternalDeviceListenerChannel)
+
+  @impl Phoenix.Socket
+  def connect(params, socket, _connect_info) do
+    IO.inspect(params, label: "APISocket params")
+
+    case Map.get(params, "token") do
+      token when is_binary(token) and byte_size(token) > 0 ->
+        # Authenticate user and store in socket assigns
+        case Accounts.fetch_user_by_api_token(token) do
+          {:ok, user, user_token} ->
+            :ok = Accounts.mark_last_used(user_token)
+
+            {:ok, assign(socket, :user, user)}
+
+          _ ->
+            {:error, :invalid_token}
+        end
+
+      _ ->
+        {:error, :no_token}
+    end
+  end
+
+  @impl Phoenix.Socket
+  def id(_socket), do: nil
+end
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -19,6 +19,7 @@ defmodule NervesHubWeb.DeviceChannel do
   alias NervesHub.Helpers.Logging
   alias NervesHub.ManagedDeployments
   alias NervesHub.Repo
+  alias NervesHubWeb.ExternalDeviceListenerChannel
   alias Phoenix.Socket.Broadcast
 
   @decorate with_span("Channels.DeviceChannel.join")
@@ -237,6 +238,15 @@ defmodule NervesHubWeb.DeviceChannel do
       device_id: device.id,
       percent: percent
     })
+
+    # Broadcast to external firmware update subscribers
+    case ExternalDeviceListenerChannel.broadcast_firmware_update(device, percent) do
+      :ok ->
+        :ok
+
+      {:error, reason} ->
+        Logger.error("[DeviceChannel] Failed to broadcast firmware update: #{inspect(reason)}")
+    end
 
     # if we know the update has already started, we can move on
     if socket.assigns.update_started? do
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -19,7 +19,6 @@ defmodule NervesHubWeb.DeviceChannel do
   alias NervesHub.Helpers.Logging
   alias NervesHub.ManagedDeployments
   alias NervesHub.Repo
-  alias NervesHubWeb.ExternalDeviceListenerChannel
   alias Phoenix.Socket.Broadcast
 
   @decorate with_span("Channels.DeviceChannel.join")
@@ -238,15 +237,6 @@ defmodule NervesHubWeb.DeviceChannel do
       device_id: device.id,
       percent: percent
     })
-
-    # Broadcast to external firmware update subscribers
-    case ExternalDeviceListenerChannel.broadcast_firmware_update(device, percent) do
-      :ok ->
-        :ok
-
-      {:error, reason} ->
-        Logger.error("[DeviceChannel] Failed to broadcast firmware update: #{inspect(reason)}")
-    end
 
     # if we know the update has already started, we can move on
     if socket.assigns.update_started? do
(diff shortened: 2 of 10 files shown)


joshk:
Hey @kybishop, is this PR ready for review? If its still under active development I'd suggest switching it to a draft PR.

kybishop:
Hey @joshk 👋 It is ready! The two CI issues (now fixed and squashed into the initial commit) were some very small papercuts.

lib/nerves_hub_web/endpoint.ex
```diff
@@ -20,6 +20,8 @@ defmodule NervesHubWeb.Endpoint do
     websocket: [connect_info: [session: @session_options]]
   )
 
+  socket("/api-socket", NervesHubWeb.APISocket, websocket: true)
```
joshk: I'd suggest calling this `events-stream`, and the module name for the socket being `EventStreamSocket`

This isn't an API socket but a socket focused on topic based event streams.

lib/nerves_hub_web/channels/api_socket.ex
```diff
@@ -0,0 +1,32 @@
+defmodule NervesHubWeb.APISocket do
+  use Phoenix.Socket
+
+  alias NervesHub.Accounts
+
+  channel("device:*", NervesHubWeb.ExternalDeviceListenerChannel)
+
+  @impl Phoenix.Socket
+  def connect(params, socket, _connect_info) do
+    IO.inspect(params, label: "APISocket params")
```
joshk:
```suggestion
```

We don't need the inspect.

lib/nerves_hub_web/channels/api_socket.ex
```diff
@@ -0,0 +1,32 @@
...
+
+    case Map.get(params, "token") do
```
joshk: Instead of double nesting case statements, how about adding the `is_binary` and `byte_size` checks to `Accounts.fetch_user_by_api_token`

lib/nerves_hub_web/channels/external_device_listener_channel.ex
```diff
@@ -0,0 +1,69 @@
+defmodule NervesHubWeb.ExternalDeviceListenerChannel do
```
joshk: I think we should call this `DeviceEventsStreamChannel`

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -238,6 +239,15 @@ defmodule NervesHubWeb.DeviceChannel do
       percent: percent
     })
 
+    # Broadcast to external firmware update subscribers
+    case ExternalDeviceListenerChannel.broadcast_firmware_update(device, percent) do
```
joshk: I believe we don't need this, since `device_internal_broadcast!` is already performing a broadcast we can hook into, which also means you can remove the `broadcast_firmware_update` function

lib/nerves_hub_web/channels/external_device_listener_channel.ex
```diff
@@ -0,0 +1,69 @@
...
+  ##########################
+  # Channel implementation #
+  ##########################
+
+  @impl Phoenix.Channel
+  def join("device:" <> device_identifier, _params, socket) do
+    # Socket already has authenticated user, just validate device access
+    case authorize_device_access(socket.assigns.user, device_identifier) do
+      {:ok, _device} ->
+        {:ok, socket}
```
joshk: You should add `Phoenix.PubSub.subscribe(NervesHub.PubSub, "device:#{device.identifier}:internal")`, which will allow you to hook into the `fwup_progress` broadcasts we already send around. (you will need to adjust the `handle_info`s to use `event: "fwup_progress"`

joshk (changes requested):
All in all this is a good first step.

There is a big question of how to approach auth, which I'd be happy with us separating into a different PR.

While using User tokens works, I think this is a good opportunity to add Product tokens. It would mean adding a new schema and supporting table, plus UI, but it also means its more resilient to staffing and user changes. (eg. team member sets this up, and 3 months later leaves, causing confusion about why a token stopped working)

I would also love to see support for a Product events stream channel, but that is 100% a different PR and discussion.

lib/nerves_hub_web/channels/api_socket.ex
jjcarstens: I actually don't think we even need the inner logic. `Accounts.fetch_user_by_api_token/1` already handles non-binary arguments and will return `:error` - It is best security practice to be ambiguous in error return on connect messages anyway, so I don't think we need to be granular in the reason here

lib/nerves_hub_web/endpoint.ex
kybishop: changed ✅

lib/nerves_hub_web/channels/api_socket.ex
kybishop: Artifact from debugging 🤦 Removed!

lib/nerves_hub_web/channels/api_socket.ex
kybishop: I think I'd offer for some pushback on these:

1. While it's true that security by obscurity is standard, this _is_ an open source project where anyone can look at the implementation 😅 Having some very simple feedback to the user that they're simply missing a token param is more a kindness to integrators than anything else.
2. While `Accounts.fetch_user_by_api_token/1` will indeed eventually return an error more by its internal implementation rather than its explicit contract, it's not something I get warm fuzzy feelings about relying on. If the implementation ever changes (e.g. starts raising in some contexts instead of returning an error tuple) we're gambling with our future selves. For the purposes of locality of behavior it seems prudent to handle the error case where we first care about it rather than leaving it up to chance that somewhere down the stack we'll be doing so.

If either of you feel strongly about the change I will make it, but I definitely lean towards handling the requirement for the token to exist before calling into a stack that may or may not handle the case as we expect.

lib/nerves_hub_web/channels/external_device_listener_channel.ex
kybishop: changed ✅

lib/nerves_hub_web/channels/device_channel.ex
kybishop: I didn't know one could hook into other broadcasts like this. Done ✅

lib/nerves_hub_web/channels/external_device_listener_channel.ex
kybishop: Done ✅

lib/nerves_hub_web/channels/api_socket.ex
kybishop: I also made a change to address your concerns about the nested case; switched it to a `with` statement that hopefully improves readability :bow:

joshk: I do agree that useful error messages are helpful for devs and integrators.

In this case, from what I can see, an error message about a missing token would come from the socket, and an error message about invalid auth for a device would come from the channel.

If you look at `UserConsoleChannel` you will see the `authorized?` function, which is only calling `Accounts.find_org_user_with_device(user, device_id)` (and another function, but we don't need to worry about that). 

I think following this pattern is more than enough; it also provides contextual error messages due to the structure of the socket and channel auths.

It's also fair to say that this setup will change soon (removing user token auth in favour of product or org tokens), and the only users of this for now will be SmartRent. Which is to say that this code isn't a hill I'm going to die on since it will likely be changed very soon.

kybishop: Hey @joshk , let me know if [this](https://github.com/nerves-hub/nerves_hub_web/pull/2217/commits/8ea28fb97969684c80f77666d430a9f8e94842c4) accomplished what you're looking for :bow:

joshk (approved):
Thanks for your work on this @kybishop , and your prompt changes during the PR review process.

PR #2218, opened 2025-07-14
Remove sentry report on harmless missing source firmware

lawik:
When a dev firmware is put on a device it is frequently not findable for delta generation on the server side. That's fine and should not generate a Sentry report which it does currently. This commit fixes that.


joshk:
Is it useful to have a test for this?

joshk (approved):
Seems fine without a test, since we are just changing from sending an error to not sending an error.

PR #2220, opened 2025-07-16
Reinstate setting deployment groups on device connect, tests

nshoes:
Code that handled setting deployment groups automatically for devices was recently removed by me during two refactor PRs 😬 This PR reinstates the missing code as well as adding integration tests so this can't happen again.

The only difference here is that the fallback clause for `set_deployment_group/1` doesn't [preload firmware and archive anymore](https://github.com/nerves-hub/nerves_hub_web/pull/2211/files#diff-851c49161e0f787800747af8c8994cd8d2cade6ac0fc8ccf149176b3a01c80acL547). The preloads aren't used in the device channel and was only hit when a device was a part of a deployment group. If there were issues that stemmed from somtimes preloading these associations we would have known by now.


test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -277,6 +278,50 @@ defmodule NervesHubWeb.DeviceChannelTest do
...
+
+    {:ok, _join_reply, device_channel} =
+      subscribe_and_join(socket, DeviceChannel, "device")
+
+    assert device_channel.assigns.device.deployment_id == deployment_group.id
+
+    close_cleanly(device_channel)
+  end
+
+  test "deployment group is removed when device joins" do
```
jjcarstens:
```suggestion
  test "deployment group is removed on join when conditions no longer match" do
```

test/nerves_hub_web/channels/device_channel_test.exs
```diff
@@ -277,6 +278,50 @@ defmodule NervesHubWeb.DeviceChannelTest do
     close_cleanly(device_channel)
   end
 
+  test "deployment group is set when device joins" do
```
jjcarstens:
```suggestion
  test "matching deployment group is set on join when device has no deployment group" do
```

nshoes:
@jjcarstens Yes! Good idea.

jjcarstens (approved):

PR #2228, opened 2025-07-22
Restore scripts list endpoint scoped by product

nshoes:
We have a route defined for `api/orgs/:org/products/:product/scripts`, but there are no actions in our scripts controller that match it. This PR adds an action for it as well as tests.


joshk (approved):
Love the implementation and cleanup!

PR #2229, opened 2025-07-24
Add support for Multi Factor Authentication using TOTP

pulkit110:
This PR allows users to configure multi factor authentication using the TOTP method using [NimbleTOTP](https://hexdocs.pm/nimble_totp/NimbleTOTP.html) under the hood. 

Here are some screenshots: 

User Account: 
Screenshot 2025-07-24 at 11 30 27

Login: 
Screenshot 2025-07-24 at 12 43 55


pulkit110:
Hi @joshk, thanks for the quick look! Let me know if you need anything from me to help get this ready.

lib/nerves_hub/mfa/user_totp.ex
```diff
@@ -0,0 +1,86 @@
...
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "user_totps" do
+    field(:secret, :binary)
+    field(:code, :string, virtual: true)
+
+    belongs_to(:user, NervesHub.Accounts.User)
```
joshk: could you please `alias NervesHub.Accounts.User` and then change this to just `User`

lib/nerves_hub/mfa/user_totp.ex
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def changeset(totp, attrs) do
+    changeset =
+      totp
+      |> cast(attrs, [:code])
+      |> validate_required([:code])
+      |> validate_format(:code, ~r/^\d{6}$/, message: "should be a 6 digit number")
+
+    code = Ecto.Changeset.get_field(changeset, :code)
```
joshk: since you have imported `Ecto.Changeset`, you can use `get_field` without the module

lib/nerves_hub/mfa/user_totp.ex
```diff
@@ -0,0 +1,86 @@
...
+
+    if changeset.valid? and not valid_totp?(totp, code) do
+      Ecto.Changeset.add_error(changeset, :code, "invalid code")
```
joshk: same as above

lib/nerves_hub/mfa.ex
```diff
@@ -0,0 +1,96 @@
+defmodule NervesHub.MFA do
```
joshk: I think we should move this to NervesHub.Accounts.MFA

lib/nerves_hub/mfa.ex
```diff
@@ -0,0 +1,96 @@
...
+  @moduledoc """
+  The MFA context.
+  """
+
+  import Ecto.Query, warn: false
```
joshk: I don't think the `warn: false` is needed

lib/nerves_hub/mfa.ex
```diff
@@ -0,0 +1,96 @@
...
+  def upsert_user_totp(totp, attrs) do
+    totp_changeset =
+      totp
+      |> UserTOTP.changeset(attrs)
+      |> UserTOTP.ensure_backup_codes()
+      # If we are updating, let's make sure the secret
+      # in the struct propagates to the changeset.
+      |> Ecto.Changeset.force_change(:secret, totp.secret)
+
+    Ecto.Multi.new()
```
joshk: We don't need to use a multi here, we can use https://hexdocs.pm/ecto/Ecto.Repo.html#c:transact/2 and https://hexdocs.pm/ecto/constraints-and-upserts.html#upserts (`Repo.insert` with `on_conflict` and `conflict_target`)

lib/nerves_hub/mfa/user_totp.ex
```diff
@@ -0,0 +1,86 @@
...
+
+    embeds_many :backup_codes, BackupCode, on_replace: :delete do
+      field(:code, :string)
+      field(:used_at, :utc_datetime_usec)
+    end
+
+    timestamps()
+  end
+
+  def changeset(totp, attrs) do
```
joshk: is this function use for creation? or when a code is used? or when a backup code is used?

lib/nerves_hub/mfa.ex
```diff
@@ -0,0 +1,96 @@
...
+      {:ok, %Ecto.Changeset{data: %UserTOTP{}}}
+  """
+  def upsert_user_totp(totp, attrs) do
+    totp_changeset =
+      totp
+      |> UserTOTP.changeset(attrs)
+      |> UserTOTP.ensure_backup_codes()
+      # If we are updating, let's make sure the secret
+      # in the struct propagates to the changeset.
+      |> Ecto.Changeset.force_change(:secret, totp.secret)
```
joshk: I think these should all be encapsulated in an `update_secret_changeset` function

lib/nerves_hub/mfa.ex
```diff
@@ -0,0 +1,96 @@
...
+
+  @doc """
+  Regenerates the user backup codes for totp.
+  ## Examples
+      iex> regenerate_user_totp_backup_codes(%UserTOTP{})
+      %UserTOTP{backup_codes: [...]}
+  """
+  def regenerate_user_totp_backup_codes(totp) do
+    {:ok, updated_totp} =
+      Repo.transaction(fn ->
```
joshk: `transaction` is no longer recommended for use, `transact` is the preferred option. 
(I also realize that we use `transaction` everywhere)

lib/nerves_hub_web/controllers/mfa_controller.ex
```diff
@@ -0,0 +1,57 @@
...
+  use NervesHubWeb, :controller
+
+  alias NervesHub.MFA
+  alias NervesHubWeb.Auth
+
+  def new(conn, _params) do
+    user_id = get_session(conn, :mfa_user_id)
+
+    if user_id do
+      form = Phoenix.Component.to_form(%{}, as: "mfa")
```
joshk, line 11: I believe you can just use `to_form`

lib/nerves_hub_web/live/account_mfa.ex
```diff
@@ -0,0 +1,137 @@
...
+    |> assign(:current_totp, totp || MFA.get_user_totp(socket.assigns.user))
+    |> assign(:backup_codes_visible, false)
+  end
+
+  defp assign_totp_qr(socket) do
+    user = socket.assigns.user
+    editing_totp = %MFA.UserTOTP{user_id: user.id}
+    app = "NervesHub"
+    secret = NimbleTOTP.secret()
+    qrcode_uri = NimbleTOTP.otpauth_uri("#{app}:#{user.email}", secret, issuer: app)
```
joshk, line 108: is using `user.email` secure? this is semi-ish guessable, i think using `user.id` would be better

priv/repo/migrations/20250723103517_create_user_totps.exs
```diff
@@ -0,0 +1,15 @@
...
+  def change do
+    create table(:user_totps) do
+      add :secret, :binary, null: false
+      add :backup_codes, :map
+      add :user_id, references(:users, on_delete: :delete_all), null: false
+
+      timestamps()
+    end
+
+    create index(:user_totps, [:user_id])
```
joshk: should this be a unique index?

test/nerves_hub_web/controllers/mfa_controller_test.exs
```diff
@@ -0,0 +1,95 @@
+defmodule NervesHubWeb.MFAControllerTest do
```
joshk, line 1: We use https://hexdocs.pm/phoenix_test/PhoenixTest.html where possible. Could you please look at our other controller tests, like `OAuthControllerTest`, switch to this setup.

joshk: I just noticed that some of your view tests do use `PhoenixTest`, is there a reason you aren't using it here?

joshk (changes requested):
This is very cool, and heading in a good direction.

I still need to try this out locally for a full review.

One question I have is how this will impact API auth?

lib/nerves_hub_web/live/account_mfa.ex
```diff
@@ -0,0 +1,137 @@
...
+    |> render_with(&account_mfa_form_template/1)
+  end
+
+  @impl Phoenix.LiveView
+  def handle_event("toggle-mfa", _params, socket) do
+    user = Accounts.get_user!(socket.assigns.user.id)
+    user_totp = MFA.get_user_totp(user)
+
+    if user_totp do
+      MFA.delete_user_totp(user_totp)
```
joshk: Since `delete_user_totp` uses `Repo.delete!` under the hood, if there is an error it will bubble up and cause the LiveView to crash, and not displaying any sort of error message to the user.

I would suggest not using the `!` variety of `delete`, and handle this use case.

joshk (changes requested):

joshk:
Oh, and final review comment, could we update the MFA UI to use a more PIN style UI, eg. https://preline.co/docs/pin-input.html

test/nerves_hub_web/controllers/mfa_controller_test.exs
pulkit110: We noticed you were using `PhoenixTest` for the view tests, but the controller tests had already been prepared earlier. We'll update them.

lib/nerves_hub_web/controllers/mfa_controller.ex
pulkit110: Thanks, but it doesn't seem to work without it - controllers don't have a `Phoenix.Component` import. It is also being used here: 
https://github.com/nerves-hub/nerves_hub_web/blob/4c6d828f032a128e80ee78db13785dfd1722f83f/lib/nerves_hub_web/controllers/session_controller.ex#L12

lib/nerves_hub_web/live/account_mfa.ex
pulkit110: This is the label which some authentication apps use to show which account an OTP belongs to. It should be easily recognizable by the user and [email is the generally accepted value](https://docs.yubico.com/yesdk/users-manual/application-oath/uri-string-format.html#label).

`secret` is the important bit that grants security and `NimbleTOTP.secret` already takes care of security on that side.

lib/nerves_hub_web/controllers/mfa_controller.ex
joshk: Ah, ok, thanks for checking.

lib/nerves_hub_web/live/account_mfa.ex
joshk: Cool, thanks for the explanation

test/nerves_hub_web/controllers/mfa_controller_test.exs
joshk: Awesome, thank you!

pulkit110:
Thanks for your thoughtful feedback and kind words @joshk.  
We’ve updated the PR to use a PIN input without adding any dependencies.  

Here are some screenshots:  
Screenshot 2025-08-19 at 12 50 42  
Screenshot 2025-08-19 at 12 51 58

pulkit110:
Ah, yes. Forgot to answer that. MFA is not enforced on API side. I think that's the standard practice as well when token based with is used. Please feel free to let us know if you'd like to change something on this side.

joshk:
Hi @pulkit110 

I'm sorry if we didn't help in getting this over the line.

I'm going to close this PR for the time being as a lot has changed in the NervesHub codebase.

Sing out if you are interested in looking into this again.

Issue #2206, opened 2025-07-09
Deployments should "wait" until deltas have been generated


lawik:
I believe @joshk has plans around this for _Releases_. Some kinda of staging step.

But essentially when we switch target version for a deployment we immediately kick off the delta generation. But if you are doing 1000 concurrents that may well trigger 1000 deployments before the deltas are even available.

You can manually operate this by pausing the deployment while you switch and then wait a little bit for deltas to populate, then resume. And it should be all ready to go.

It is not terrible in the current way it works but it is a race between generating deltas and passing out firmware. Definitely not optimal. It would also nice to have an indication of how many delta jobs were fired off so you can watch them complete (whether success or failure). Ideally also report what the failures are for the versions that failed. Lots of potential improvements there.

On the flip side. The current thing is quite simple.

lawik:
So the handler for "deployments/update" could do this it seems. Or rather `Orchestrator.trigger_update`.

The orchestrator state can hold whether it has completed deltas. If deltas have already been generated, attempting them is really cheap, it is essentially just checking if they already exist. So on orchestrator restart or something it might re-check but it wouldn't do very much of the expensive work again.

lawik:
I think this avoids getting state on the DeploymentGroup or Firmware that risks getting out of phase.

nshoes:
@lawik I vote Oban for cleaning up things, it's a pattern we already heavily follow.

lawik:
Addressed by #2262 (preparing/ready statuses on DeploymentGroup).

PR #2235, opened 2025-07-28
Track firmware delta generation in more detail

lawik:
Add status field to firmware deltas:

- Starting the creation of a firmware delta always creates an entry with status "processing"
- Successfully creating the delta set status "completed"
- Failure sets status "failed"

Add firmware delta generation timeout worker

Fetching firmware deltas can now fall back to full because a delta is not done.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -1734,7 +1734,7 @@ defmodule NervesHub.Devices do
            {:delta_updatable, delta_updatable?(device, target)},
          {:firmware, {:ok, source}} <-
            {:firmware, Firmwares.get_firmware_by_product_id_and_uuid(product_id, source_uuid)},
-         {:delta, {:ok, delta}} <-
+         {:delta, {:ok, %{status: :completed} = delta}} <-
            {:delta, Firmwares.get_firmware_delta_by_source_and_target(source, target)} do
       Logger.info(
         "Delivering firmware delta",
@@ -1762,6 +1762,17 @@ defmodule NervesHub.Devices do
           device_id: device.id,
           source_firmware: source_uuid,
           target_firmware: target.uuid
+        )
+
+        Firmwares.get_firmware_url(target)
+
+      {:delta, {:ok, %{status: status} = delta}} ->
+        Logger.info(
+          "Delivering full firmware as delta had status #{status}",
+          device_id: device.id,
+          source_firmware: source_uuid,
+          target_firmware: target.uuid,
+          delta: delta.id
         )
 
         Firmwares.get_firmware_url(target)
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -13,6 +13,7 @@ defmodule NervesHub.Firmwares do
   alias NervesHub.Products
   alias NervesHub.Products.Product
   alias NervesHub.Workers.DeleteFirmware
+  alias NervesHub.Workers.FirmwareDeltaBuilder
 
   alias NervesHub.Repo
 
@@ -391,10 +392,12 @@ defmodule NervesHub.Firmwares do
     FirmwareDelta
     |> where([fd], source_id: ^source_id)
     |> where([fd], target_id: ^target_id)
-    |> Repo.one()
+    |> order_by(desc: :inserted_at)
+    |> limit(1)
+    |> Repo.all()
     |> case do
-      nil -> {:error, :not_found}
-      firmware_delta -> {:ok, firmware_delta}
+      [] -> {:error, :not_found}
+      [firmware_delta] -> {:ok, firmware_delta}
     end
   end
 
@@ -405,11 +408,11 @@ defmodule NervesHub.Firmwares do
     firmware_upload_config().download_file(fw_or_delta)
   end
 
-  @spec create_firmware_delta(Firmware.t(), Firmware.t()) ::
+  @spec generate_firmware_delta(FirmwareDelta.t(), Firmware.t(), Firmware.t()) ::
           :ok
           | {:error, Changeset.t()}
 
-  def create_firmware_delta(source_firmware, target_firmware) do
+  def generate_firmware_delta(firmware_delta, source_firmware, target_firmware) do
     Logger.info(
       "Creating firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}."
     )
@@ -428,16 +431,15 @@ defmodule NervesHub.Firmwares do
               with upload_metadata <-
                      firmware_upload_config().metadata(org.id, firmware_delta_filename),
                    {:ok, firmware_delta} <-
-                     insert_firmware_delta(%{
-                       source_id: source_firmware.id,
-                       target_id: target_firmware.id,
-                       tool: created.tool,
-                       tool_metadata: created.tool_metadata,
-                       size: created.size,
-                       source_size: created.source_size,
-                       target_size: created.target_size,
-                       upload_metadata: upload_metadata
-                     }),
+                     complete_firmware_delta(
+                       firmware_delta,
+                       created.tool,
+                       created.size,
+                       created.source_size,
+                       created.target_size,
+                       created.tool_metadata,
+                       upload_metadata
+                     ),
                    {:ok, firmware_delta} <- get_firmware_delta(firmware_delta.id),
                    :ok <-
                      firmware_upload_config().upload_file(created.filepath, upload_metadata),
@@ -468,8 +470,12 @@ defmodule NervesHub.Firmwares do
           )
 
         case result do
-          {:ok, _delta} -> :ok
-          {:error, err} -> {:error, err}
+          {:ok, _delta} ->
+            :ok
+
+          {:error, err} ->
+            _ = fail_firmware_delta(firmware_delta)
+            {:error, err}
         end
 
       {:error, :no_delta_support_in_firmware} ->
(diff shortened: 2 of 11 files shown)


lawik:
First part of #2206

lib/nerves_hub/firmwares.ex
```diff
@@ -468,8 +477,12 @@ defmodule NervesHub.Firmwares do
           )
 
         case result do
-          {:ok, _delta} -> :ok
-          {:error, err} -> {:error, err}
+          {:ok, _delta} ->
+            :ok
+
+          {:error, err} ->
+            fail_firmware_delta(firmware_delta)
```
nshoes: Since we're updating a record in an error case, I'd suggest we use `Repo.update!/2` instead of having to handle the return states of `Repo.update/2`

lib/nerves_hub/firmwares.ex
```diff
@@ -391,10 +399,12 @@ defmodule NervesHub.Firmwares do
     FirmwareDelta
     |> where([fd], source_id: ^source_id)
     |> where([fd], target_id: ^target_id)
-    |> Repo.one()
+    |> order_by(desc: :inserted_at)
+    |> limit(1)
```
nshoes, line 396: Thank you 🙇

test/nerves_hub/firmwares_test.exs
```diff
@@ -313,11 +319,77 @@ defmodule NervesHub.FirmwaresTest do
...
+      {:ok, %{id: d3}} = Firmwares.start_firmware_delta(source, t3)
+      {:ok, %{id: d4}} = Firmwares.start_firmware_delta(source, t4)
+      Firmwares.time_out_firmware_delta_generations(1000, :millisecond)
+
+      assert [
+               %{id: ^d1, status: :timed_out},
+               %{id: ^d2, status: :timed_out},
+               %{id: ^d3, status: :processing},
+               %{id: ^d4, status: :processing}
+             ] = Firmwares.get_deltas_by_source_firmware(source)
```
nshoes: I've noticed this pattern more frequently as I spend time with this codebase: using context functions in tests but not application code. It feels like this should be a local private function in this test module, but then one could argue that other tests could utilize it as well.

Organization-wise, this doesn't make sense to me, but it's not a hill I'm willing to die on. Thoughts? cc @joshk

lib/nerves_hub/workers/firmware_delta_builder.ex
```diff
@@ -46,11 +43,20 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilder do
 
   defp maybe_create_firmware_delta(source, target) do
```
nshoes, line 44: Thoughts on returning helpful logging and/or a return for the Oban worker so we have some paper trail if we get off the happy path?

lib/nerves_hub/devices.ex
```diff
@@ -1768,6 +1768,17 @@ defmodule NervesHub.Devices do
...
         Firmwares.get_firmware_url(target)
 
+      {:delta, {:ok, %{status: status} = delta}} ->
+        Logger.info(
+          "Delivering full firmware as delta had status #{status}",
+          device_id: device.id,
+          source_firmware: source_uuid,
+          target_firmware: target.uuid,
+          delta: delta.id
+        )
```
nshoes, line 1776: 💯

lib/nerves_hub/firmwares/firmware_delta.ex
```diff
@@ -35,6 +37,77 @@ defmodule NervesHub.Firmwares.FirmwareDelta do
...
+      %{
+        status: :completed,
+        tool: tool,
+        size: size,
+        source_size: source_size,
+        target_size: target_size,
+        tool_metadata: tool_metadata,
+        upload_metadata: upload_metadata
+      },
+      @required_params ++ @optional_params
```
joshk: I find the concept of "required" and "optional" params a bit of a smell. Different params might be required at different stages of the lifecycle of the schema, so should we just list the params? 

Also, for the complete function, shouldn't all that info be included in `validate_required`?

lib/nerves_hub/firmwares/firmware_delta.ex
```diff
@@ -35,6 +37,77 @@ defmodule NervesHub.Firmwares.FirmwareDelta do
...
+    |> validate_required(@required_params)
+    |> unique_constraint(:unique_firmware_delta, name: :source_id_target_id_unique_index)
+    |> foreign_key_constraint(:source_id, name: :firmware_deltas_source_id_fkey)
+    |> foreign_key_constraint(:target_id, name: :firmware_deltas_target_id_fkey)
+  end
+
+  @spec fail(FirmwareDelta.t()) :: Ecto.Changeset.t()
+  def fail(%FirmwareDelta{} = firmware_delta) do
+    firmware_delta
+    |> cast(%{status: :failed}, @required_params ++ @optional_params)
```
joshk: if we are just casting one field, maybe `put_change` is a better fit?

lib/nerves_hub/firmwares/firmware_delta.ex
```diff
@@ -35,6 +37,77 @@ defmodule NervesHub.Firmwares.FirmwareDelta do
...
+    |> validate_required(@required_params)
+  end
+
+  @spec time_out(FirmwareDelta.t()) :: Ecto.Changeset.t()
+  def time_out(%FirmwareDelta{} = firmware_delta) do
+    firmware_delta
+    |> cast(%{status: :timed_out}, @required_params ++ @optional_params)
```
joshk: same as above

lib/nerves_hub/firmwares/firmware_delta.ex
```diff
@@ -35,6 +37,77 @@ defmodule NervesHub.Firmwares.FirmwareDelta do
...
+    |> validate_required(@required_params)
+  end
+
+  @spec changeset(FirmwareDelta.t(), map()) :: Ecto.Changeset.t()
```
joshk: should we rename `changeset` to match where we use it? eg. `create_changeset`?

lib/nerves_hub/workers/firmware_delta_builder.ex
```diff
@@ -46,11 +43,20 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilder do
...
     case Firmwares.get_firmware_delta_by_source_and_target(source, target) do
-      {:ok, _} ->
+      {:ok, %FirmwareDelta{status: :processing} = delta} ->
+        Logger.info(
+          "Attempting firmware delta build for #{source.platform} #{source.version} to #{target.version}..."
+        )
```
joshk, line 49:
```suggestion
        Logger.info(
          "Attempting firmware delta build",
           platform: source.platform,
           source_version: source.version
           target_version: target.version
        )
```

lib/nerves_hub/firmwares.ex
joshk: Why `limit(1)` over `Repo.one()`?

priv/repo/migrations/20250716192310_firmware_deltas_add_status_field.exs
```diff
@@ -0,0 +1,9 @@
+defmodule NervesHub.Repo.Migrations.FirmwareDeltasAddStatusField do
+  use Ecto.Migration
+
+  def change do
+    alter table(:firmware_deltas) do
+      add :status, :string, default: "completed"
```
joshk: I wouldn't add a default of completed, but instead update the existing rows in the table.

The downside of the default like this is if a row gets created and then gets that status when it shouldn't.

test/nerves_hub/firmwares_test.exs
joshk: Really interesting point.

I agree that we should extract these to a test support module.

joshk (commented):
In general, love it.

Some questions, some comments.

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: We should check if still use it at all. The specific ones I know we use. I am okay with either naming.

lib/nerves_hub/firmwares.ex
lawik: `Repo.one` breaks if there have been multiple deltas generated that match. I believe this exists in old data already.

lib/nerves_hub/firmwares.ex
lawik: Well this produces another valid outcome with the same signature as the successful transaction. A failed delta can be because the delta generation is impossible. The return values of `fail_firmware_delta/1` should be valid.

Or maybe I'm missing something :)

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: I am okay with just listing the params but this pattern is pretty common in this codebase right?

And yeah, you are right about the validate_required :)

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: I think people are more used to reading a cast. But don't feel strongly about it :)

lib/nerves_hub/firmwares/firmware_delta.ex
nshoes: Slightly hijacking this, but after pondering, I vote for `create_changeset` and all the other functions in this module to append `_changeset` to their name for clarity.

priv/repo/migrations/20250716192310_firmware_deltas_add_status_field.exs
lawik: I see your point. This seemed simpler and less error prone in terms of running a migration.

test/nerves_hub/firmwares_test.exs
lawik: I suppose. In this case the function mirrors another function we have which _is_ used. Felt straightforward to me. But okay to move it elsewhere.

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: Seems perfectly topical. Wrong but topical :P  

Nah, I kid. I don't feel strongly about it but since we don't have any other functions than changeset functions in Schema modules, is it really meaningful to have the changeset suffix?

lib/nerves_hub/workers/firmware_delta_builder.ex
lawik: I think they log pretty well during generation. We could log the `not_found` I guess but I think there would be prior logging about failing to create the delta.

lib/nerves_hub/firmwares/firmware_delta.ex
nshoes: 😆 

It is meaningful. I tend to lean towards explicitness in naming, and if I'm in a schema module, all of my functions will return a changeset and end in `_changeset`. No one has to guess or dig around to know what `FirmwareDelta.start_changeset` will return or do. We follow this pattern for `DeploymentGroup` and `User` as well.

nshoes: Now that I think about it,  I've worked with groups where your assumption _wasn't_ assumed, so the naming helped.

test/nerves_hub/firmwares_test.exs
lawik: Changed

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: Addressed :)

lib/nerves_hub/firmwares/firmware_delta.ex
lawik: Addressed :)

priv/repo/migrations/20250716192310_firmware_deltas_add_status_field.exs
lawik: Addressed :)

lib/nerves_hub/firmwares.ex
nshoes: I misspoke when originally reporting this issue in SmartRent's Slack 😬 The issue was with `firmwares`, not `firmware_deltas`. I checked our production, and we don't have any rows where `target_id = source_id`. We can revert this change; my apologies.

lib/nerves_hub/workers/firmware_delta_builder.ex
nshoes: Logging seems enough; there's a trail if we need to go dig.

test/nerves_hub/firmwares_test.exs
nshoes: > In this case the function mirrors another function we have which is used

I saw the value here as well, but I believe the organization is more important. Thanks for changing!

lib/nerves_hub/firmwares.ex
nshoes: I see and agree. Don't think you're missing anything :)

lib/nerves_hub/firmwares.ex
lawik: I believe I hit something with this in testing, so wasn't just for that :)

nshoes: Oh! Sounds good.

nshoes:
Resolved the conflicts and added a sort to a new test, allowing it to pass both locally and in CI.

joshk:
❤️

Issue #2234, opened 2025-07-26
Can't delete a device certificate in the new UI


amclain:
**Describe the bug**
When using the new UI, a generic error appears when deleting a device certificate. The certificate cannot be deleted. Deleting the certificate still works from the old UI.

**To Reproduce**
1. Switch to the new UI
2. Navigate to a device with a certificate
3. Try to delete its certificate

**Expected behavior**
A device certificate should be able to be deleted.

**Screenshots**

<img width="1629" height="266" alt="Image" src="https://github.com/user-attachments/assets/4ac9058b-8320-4a62-a435-3856b096d98b" />

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome

nshoes:
Thanks for the report! The fix will be merged in soon.

PR #2238, opened 2025-07-28
Fix device certificate management in new UI

nshoes:
This PR fixes device certificate uploading and deleting in the new UI and adds a test.

Solves #2234.

### DIFF
--- a/lib/nerves_hub_web/components/device_page/settings_tab.ex
+++ b/lib/nerves_hub_web/components/device_page/settings_tab.ex
@@ -119,7 +119,7 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
         <div class="flex justify-between items-center h-14 px-4 border-b border-zinc-700">
           <div class="text-base text-neutral-50 font-medium">Certificates</div>
           <div>
-            <form phx-change="validate-cert" phx-drop-target={@uploads.certificate.ref}>
+            <form id="upload-certificate" phx-change="validate-cert" phx-drop-target={@uploads.certificate.ref}>
               <div class="flex px-3 py-1.5 gap-2 rounded bg-zinc-800 border border-zinc-600 hover:cursor-pointer">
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/components/device_page/settings_tab.ex
```diff
@@ -173,17 +173,17 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
                   <span>Not after:</span>
                   <span>{Calendar.strftime(certificate.not_after, "%Y-%m-%d")}</span>
                 </div>
-                <div class="text-xs text-zinc-400 tracking-wide">
-                  <%!-- <%= Timex.from_now(entry.inserted_at) %> --%>
+                <%!-- <div class="text-xs text-zinc-400 tracking-wide">
```
joshk: should we just remove the commented out code? I can't remember why its there.

lib/nerves_hub_web/components/device_page/settings_tab.ex
```diff
@@ -429,9 +429,9 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
   def hooked_async(_name, _async_fun_result, socket), do: {:cont, socket}
 
   def handle_progress(:certificate, %{done?: true} = entry, socket) do
-    socket
-    |> consume_uploaded_entry(entry, &import_cert(socket, &1.path))
-    |> halt()
```
joshk, line 434: ahhhhh, this makes total sense, the `halt()` is needed as the `handle_progress` function is passed in during setup. Silly me.

lib/nerves_hub_web/components/device_page/settings_tab.ex
nshoes: Yeah, don't think it's required for the future. Blasted it shall be.

lib/nerves_hub_web/components/device_page/settings_tab.ex
nshoes: Bingo! So silly.

joshk (approved):
❤️

PR #2244, opened 2025-08-01
Match deployments to devices with nil tags

lawik:


lib/nerves_hub/managed_deployments.ex
```diff
@@ -506,7 +506,16 @@ defmodule NervesHub.ManagedDeployments do
...
+    |> where(
+      [d],
+      fragment(
+        "?->'tags' <@ to_jsonb(?::text[]) OR (jsonb_array_length(?->'tags') = 0 and ?::text[] is null)",
+        d.conditions,
+        ^device.tags,
+        d.conditions,
+        ^device.tags
+      )
+    )
```
nshoes, line 518: This adjustment covers `device.tags` being `nil`, correct? What about other queries that are structured similarly?

lib/nerves_hub/devices/device.ex
```diff
@@ -95,7 +95,6 @@ defmodule NervesHub.Devices.Device do
     |> cast_embed(:firmware_metadata)
     |> cast_embed(:extensions)
     |> validate_required(@required_params)
-    |> validate_length(:tags, min: 1)
```
joshk, line 98: ❤️

lib/nerves_hub/managed_deployments.ex
joshk: I won't lie, this broke my brain a little as I decompiled it in my head. 

One small suggestion would be to prefer the use of `[dg]` vs `[d]` in the `where` (although it might be best to update all of them) as its easy to first misread that as `[d]` == device, and since this function is using device as well, it was a little confusing. But this is a nit pick, so take it as you like.

joshk: Also, this looks correct to me. Its a pity there isn't a nicer way to handle these use cases in ecto.

PR #2260, opened 2025-08-08
Make device list refresh in a live but safe way

lawik:
For the purpose of making it possible to watch live updates more information should refresh on connection changes. But it should be throttled in a responsible way.

We can even let fwup_progress events indicate that things should refresh if the pace is safe.

One big addition to making this safe is to determine which tabs are not being watched and if they are not watche we don't update them.

This improves on the 10 second timer updater.

### DIFF
--- a/assets/ui-rework/hooks/pageVisible.js
+++ b/assets/ui-rework/hooks/pageVisible.js
@@ -4,7 +4,6 @@ export default {
   },
   init() {
     document.addEventListener("visibilitychange", () => {
-      console.log("visibility changed", document.visibilityState)
       this.pushEvent("page_visibility_change", {
         visible: document.visibilityState === "visible"
       })


nshoes:
I looked into the failing tests (since I wrote them hehe) and I uncovered a race condition, but one that is unique to the test env and doesn't reflect a real-world issue.

However, it does uncover that we're now doubling the work on a `connection:change` or `connection:status` with:
- LV receives `connection:*` message
- `safe_refresh` is called
- `assign_display_devices` is called, starting an async update
- `update_device_statuses` is called, updating the status
- `assign_display_devices` async work is finished, updating the status

Or:
- `Tracker` tells us that a device connection changes
- We ask the `Tracker` the conneciton status of a device

The tests didn't mock `Tracker`, so it would default to offline, failing the tests, which would check if the green dot is visible when a device comes online.

I implemented a fix, but I don't love it. I think it's fine as a temporary; @joshk you've talked about getting rid of Tracker or changing the implementation, but I'm unsure what that entails or if it will happen soon.

lawik:
@nshoes if you can rebase this one I'm happy to give it a review and get it through :)

lawik:
Awesome!

PR #2261, opened 2025-08-08
Device filter inflight

lawik:
This should allow watching updates as they happen. It'll be a lot better if we land the changes to live updating. But this adds the filter and a link on the summary page whenever devices are actively updating.

### DIFF
--- a/lib/nerves_hub/devices/device_filtering.ex
+++ b/lib/nerves_hub/devices/device_filtering.ex
@@ -7,6 +7,7 @@ defmodule NervesHub.Devices.DeviceFiltering do
 
   alias NervesHub.Devices.Alarms
   alias NervesHub.Devices.DeviceMetric
+  alias NervesHub.Devices.InflightUpdate
   alias NervesHub.Types.Tag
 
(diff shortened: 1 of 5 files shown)


joshk (approved):

lib/nerves_hub/devices/device_filtering.ex
```diff
@@ -124,6 +125,12 @@ defmodule NervesHub.Devices.DeviceFiltering do
   def filter(query, _filters, :display_deleted, "only"),
     do: where(query, [d], not is_nil(d.deleted_at))
 
+  def filter(query, _filters, :only_updating, false),
+    do: query
+
+  def filter(query, _filters, :only_updating, true),
+    do: build_inflight_filter(query)
```
nshoes: Can we build the filter here?

```ex
join(query, :inner, [d], iu in InflightUpdate, on: d.id == iu.device_id, as: :inflight_update)
```

nshoes (approved):
LGTM!

lib/nerves_hub/devices/device_filtering.ex
joshk: Agree. Is there a reason for it to be split into another function?

PR #2262, opened 2025-08-08
Add preparation step for orchestrator doing firmware deltas

lawik:


lawik:
I need to do a bunch of testing this with devices on my desk but I'd love a review of the code and conceptually solving it like this. I think it lines up with discussion in #2206.

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
```diff
@@ -26,7 +27,8 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
     defstruct deployment_group: nil,
               rate_limit?: true,
               timer_ref: nil,
-              should_run?: false
+              should_run?: false,
+              subscribed_to: nil
```
nshoes: I'd love if this name reflected the value is a `firmware_id`, like `firmware_subscribed_to`. It was confusing on first read.

lib/nerves_hub/devices.ex
```diff
@@ -645,6 +645,7 @@ defmodule NervesHub.Devices do
     # Exclude the current firmware, we don't need to generate that one
     |> where([dep, dev, f], f.id != dep.firmware_id)
     |> select([dep, dev, f], {f.id, dep.firmware_id})
+    |> distinct(true)
```
nshoes, line 626: Do we need this in the function above, `get_device_firmware_for_delta_generation_by_product/1`, too?

lib/nerves_hub/firmwares.ex
```diff
@@ -573,6 +580,34 @@ defmodule NervesHub.Firmwares do
...
+  def unsubscribe_firmware_delta_target(target_id) do
+    _ = NervesHubWeb.Endpoint.unsubscribe("firmware_delta_target:#{target_id}")
+    :ok
+  end
+
+  defp notify_firmware_delta_target({:ok, %FirmwareDelta{} = firmware_delta}) do
+    _ =
+      NervesHubWeb.Endpoint.broadcast(
+        "firmware_delta_target:#{firmware_delta.target_id}",
+        "firmware_delta_#{to_string(firmware_delta.status)}",
```
nshoes:
```suggestion
        "firmware_delta_#{firmware_delta.status}",
```

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
nshoes: `subscribed_firmware_id`?

lib/nerves_hub/managed_deployments.ex
```diff
@@ -318,6 +318,13 @@ defmodule NervesHub.ManagedDeployments do
     end)
   end
 
+  def get_delta_generation_status(deployment_group) do
+    deltas = Firmwares.get_deltas_by_target_firmware(deployment_group.firmware_id)
+    total = Enum.count(deltas)
+    completed = Enum.count(deltas, &(&1.status != :processing))
```
nshoes: I don't think we should we count `failed` and `timed_out` as "completed"; thoughts?

lib/nerves_hub_web/components/deployment_group_page/summary.ex
```diff
@@ -12,31 +14,42 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Summary do
...
+  def handle_info(
+        %Broadcast{topic: "firmware_delta" <> _},
+        %{assigns: %{deployment_group: deployment_group}} = socket
+      ) do
+    socket
+    |> assign(:delta_status, ManagedDeployments.get_delta_generation_status(deployment_group))
+    |> noreply()
+  end
+
+  def handle_info(_, socket) do
```
nshoes: We're handling all broadcasts that can be subscribed to in the `handle_info/2` above, can we remove this catch-all?

test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -690,7 +690,7 @@ defmodule NervesHubWeb.WebsocketTest do
         })
 
       # This is what the orchestrator process will do
-      Orchestrator.trigger_update(deployment_group)
+      Orchestrator.trigger_update(%Orchestrator.State{deployment_group: deployment_group})
```
nshoes: What was the thought process behind requiring `Orchestrator.State` as an arg to `trigger_update/1`? I liked it's previous implementation because we _didn't_ have to pass that struct.

lawik: Needed the prepare step to be able  to update the state. State goes in, new state comes out.

lib/nerves_hub/managed_deployments/deployment_group.ex
```diff
@@ -69,8 +70,11 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
     field(:current_updated_devices, :integer, default: 0)
     field(:inflight_update_expiration_minutes, :integer, default: 60)
     field(:queue_management, Ecto.Enum, values: [:FIFO, :LIFO], default: :FIFO)
+
     field(:delta_updatable, :boolean, default: true)
 
+    field(:status, Ecto.Enum, values: [:ok, :preparing], default: :ok)
```
joshk: could we use `:ready` instead of `:ok`? `:ok` is use heavily within Elixir for function results, and I don't think it is as descriptive as `:ready` for a `:status`

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
```diff
@@ -104,27 +111,53 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
...
+    source_ids =
+      id
+      |> Devices.get_device_firmware_for_delta_generation_by_deployment_group()
+      |> Enum.map(fn {source_id, _target_id} -> source_id end)
+
+    query =
+      FirmwareDelta
+      |> where([fd], fd.source_id in ^source_ids)
+      |> where([fd], fd.target_id == ^firmware_id)
+      |> where([fd], fd.status != :completed)
```
joshk: Lets move this into a function on `Firmwares`, that way we don't need to import Ecto query bits.

joshk (commented):
Super excited for this!

lib/nerves_hub/managed_deployments/deployment_group.ex
nshoes: Done!

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
nshoes: Is this really needed? I'd be inclined to agree, but this query is hyper-specific to the orchestrator (note querying for `source_ids` based on devices in deployment group), and I don't see why importing Ecto is a bigger red flag. Thoughts?

nshoes: A public function in a context that's not re-used anywhere is a bigger rub for me 😅

joshk: I hear ya, I get ya, and yet (to me) this feels like a mixing of concerns.

I do prefer to "strongly prefer encapsulating database queries in context modules", because as soon as you veer outside of this you can start arguing for importing ecto into LiveViews.

In this particular case, I would create a function called `Firmwares.deltas_processing?(target_id)` which encapsulates the logic from line 128 to 139.

lib/nerves_hub/firmwares.ex
```diff
@@ -578,8 +533,45 @@ defmodule NervesHub.Firmwares do
...
```
joshk: the `firmware_delta_` feels superfluous, and adding the status also feels a little odd.

How about having the event name be `status_update` or `progress_update`, and adding the status to the payload?

lib/nerves_hub/devices.ex
```diff
@@ -1724,116 +1740,73 @@ defmodule NervesHub.Devices do
...
+    source_firmware_id_query =
+      Firmware
+      |> where(uuid: ^source_uuid)
+      |> select([f], f.id)
+
+    delta =
+      FirmwareDelta
+      |> where([fd], fd.source_id == subquery(source_firmware_id_query))
+      |> where([fd], fd.target_id == ^target_id)
+      |> Repo.one()
```
joshk, line 1806: should we also add a `where` for the status being `complete`?

lib/nerves_hub_web/components/deployment_group_page/settings.ex
```diff
@@ -38,8 +38,9 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Settings do
           <div class="flex p-6 gap-6">
             <div class="w-1/2 flex flex-col gap-6">
               <.input field={@form[:name]} label="Name" placeholder="Production" />
-              <.input field={@form[:delta_updatable]} type="checkbox" label="Delta updates" phx-click="toggle-delta-updates">
+              <.input field={@form[:delta_updatable]} type="checkbox" label="Delta updates">
                 <:rich_hint>
+                  When enabled, deployment group will only send delta updates.
```
joshk:
```suggestion
                  When enabled, the deployment group will only send delta updates.
```

lib/nerves_hub_web/components/deployment_group_page/settings.ex
```diff
@@ -38,8 +38,9 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.Settings do
           <div class="flex p-6 gap-6">
             <div class="w-1/2 flex flex-col gap-6">
               <.input field={@form[:name]} label="Name" placeholder="Production" />
-              <.input field={@form[:delta_updatable]} type="checkbox" label="Delta updates" phx-click="toggle-delta-updates">
```
joshk, line 41: why was the `phx-click` removed?

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -324,11 +324,9 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
                 </div>
               </form>
             </div>
-            <div class="flex pt-2 gap-4 items-center">
+            <div :if={@device.deployment_group} class="flex pt-2 gap-4 items-center">
               <span class="text-sm text-nerves-gray-500">Assigned deployment group:</span>
-              <span :if={is_nil(@device.deployment_group)} class="text-sm text-nerves-gray-500">No assigned deployment group</span>
```
joshk, line 329: Was there a reason for removing this message?

joshk (changes requested):
This looks awesome!

Some feedback, nothing is a required change.

lib/nerves_hub/devices.ex
nshoes: I don't think so. The only time we hit this code path is when we know the delta is complete with the check I added in the UI. This code path is also not resilient to failures, such as passing `nil` to `Firmwares.get_firmware_url/1`, and I don't particularly care to tackle that here. Though it should be handled :)

lib/nerves_hub_web/components/deployment_group_page/settings.ex
nshoes: I removed it because I really dislike that this form can be submitted manually, but for some reason this checkbox updates async... 😬

lib/nerves_hub_web/components/device_page/details_tab.ex
nshoes: Yes! It was duplicating info, IMO. It felt silly to state that there is no deployment group right above the UI to select the deployment group in the deployment group card. It now looks like:

CleanShot 2025-09-25 at 12 31 22@2x

lib/nerves_hub_web/components/deployment_group_page/settings.ex
nshoes: Done!

lib/nerves_hub/firmwares.ex
nshoes: Love it, implemented.

lib/nerves_hub/devices.ex
joshk: Ok, I can live with this. The benefit of adding the `where` is that if we see it exploding (sentry error) then we know something funky happened.

lib/nerves_hub_web/components/deployment_group_page/settings.ex
joshk: Ok, I can get behind that.

lib/nerves_hub_web/components/device_page/details_tab.ex
joshk: Oh, great improvement! thanks for the screenshot!

joshk (approved):
<3

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
nshoes: Agreed and implemented. However, I put the function in `ManagedDeployments` since we're passing a deployment group for it's `id` and `firmware_id`. Please let me know if you hate this.

joshk: Love this! Thank you!

lib/nerves_hub/devices.ex
```diff
@@ -835,6 +834,22 @@ defmodule NervesHub.Devices do
...
+      |> select([f], f.id)
+
+    query =
+      FirmwareDelta
+      |> where([fd], fd.source_id == subquery(source_firmware_id_query))
+      |> where([fd], fd.target_id == ^target_id)
+      |> where([fd], fd.status == :completed)
+
+    Repo.exists?(query)
+  end
```
joshk: Guess we need to scope the top query (based on our learnings from the last deploy)

lib/nerves_hub/devices.ex
```diff
@@ -1724,116 +1740,73 @@ defmodule NervesHub.Devices do
...
 
-  def get_delta_if_ready(_device, _target_firmware), do: {:firmware, :no_device_firmware_metadata}
+  @spec get_delta_url(Device.t(), Firmware.t()) ::
+          {:ok, String.t()}
+          | {:error, :failure}
+  def get_delta_url(%Device{firmware_metadata: %{uuid: source_uuid}}, %Firmware{id: target_id}) do
+    source_firmware_id_query =
+      Firmware
+      |> where(uuid: ^source_uuid)
+      |> select([f], f.id)
```
joshk, line 1800: We should add `product_id` scoping to this query

joshk (changes requested):

lib/nerves_hub/devices.ex
nshoes: Done!

lib/nerves_hub/devices.ex
nshoes: Done!

PR #2265, opened 2025-08-10
Use jemalloc for our memory allocation

joshk:
This adds two new build stages, one for building jemalloc with GCC 14, and one for building a statically linked fwup.

(This fwup change was included in this PR because it made testing the Dockerfile changes easier - I have a Mac.)

Using [jemalloc](https://github.com/facebook/jemalloc) (this link leads to the facebook fork of jemalloc, which is under active development) reduces memory usage significantly by better managing memory fragmentation.

Changing to a statically built fwup, along with reducing the packages we install, helps reduce the final docker image size from _**619MB**_ to _**293MB**_.

I've also tweaked the Erlang alloc settings for the binary and eheap allocs, which came from various recommendations to reduce memory fragmentation.


.cspell.json
```diff
@@ -20,10 +20,13 @@
   ],
   "ignorePaths": [
     "**/*.csv",
+    "docker/*",
+    "Dockerfile",
```
jjcarstens, line 24: I have been trying to be very cautious and restrictive in what gets wholly ignored since the here is no indication of ignored paths. Adding so much here reduces usefulness of the tool, especially to our future selves.

There are several ways in which we can configure to still be useful in checking such as the `docker` dictionary which can be global or just to the Dockerfile. The `docker` directory is also not even committed to the repo, so I would leave it out until necessary. If some artifacts are being made, then maybe it belongs in .gitignore?

Point being I just want to challenge us to be thoughtful with this tool. Don't have to waste time, but some effort into keeping it generally useful 🎉

joshk: Ahhhh sorry, `docker/*` was an artifact of when I was testing out different memory allocators. Sorry for not cleaning that up.

I ignored `Dockerfile` because how apt packages, and things like `jemalloc` were being picked up. Sorry if I used a blunt stick to solve that problem.

PR #2268, opened 2025-08-11
Support TLS Version as an Option within SMTP

jclarkgp:
In order for NervesHub to work against AWS SES, we need to be able to set TLS Version to 1.2.  The default that happens with Swoosh isn't supporting what SES expects.  This change permits the option to set TLS_VERSION or leave it to default within the Swoosh


joshk:
@jclarkgp could you please run a `mix format` and then I can merge this in.

PR #2279, opened 2025-08-13
Relax user role for device event channel

kybishop:
The device event channel is read only and simply forwards events, no need for it to map to a manage role.


kybishop:
@nshoes added what is effectively an integration test. Let me know if this is the kind of thing you had in mind, or something else.

PR #2280, opened 2025-08-13
Fix LiveView warning when uploading device certificates, rehydrate state properly

nshoes:
`Phoenix.LiveView. consume_uploaded_entry/3` expects a return of `{:ok, _return}` or `{:postpone, _return}`. We were returning a socket, causing a warning in the logs.

I also fixed an issue when uploading a device certificate. We were previously using `update_in/2`, but on an unloaded Ecto association. This was copied over from the old UI, but this code now sits in a component, and we don't preload device certificates in the parent LiveView. Fixing this uncovered another bug if you deleted the device's last cert and then uploaded a new one. `Repo.preload/3` won't preload on a loaded empty association, so `force: true` is needed.

### DIFF
--- a/lib/nerves_hub_web/components/device_page/settings_tab.ex
+++ b/lib/nerves_hub_web/components/device_page/settings_tab.ex
@@ -22,7 +22,7 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
   end
 
   def render(assigns) do
-    device = Repo.preload(assigns.device, :device_certificates)
+    device = Repo.preload(assigns.device, :device_certificates, force: true)
 
     changeset = Ecto.Changeset.change(assigns.device)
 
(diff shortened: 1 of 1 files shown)


joshk:
You are a star @nshoes 

Do we need tests for any of this?

joshk:
Totally understand. 

Does your fix mean certificate upload wasn't working at all? or were these just warnings?

I thought we had tests for this flow, so now I'm just wondering if they are actually working.

lib/nerves_hub_web/components/device_page/settings_tab.ex
```diff
@@ -431,25 +431,26 @@ defmodule NervesHubWeb.Components.DevicePage.SettingsTab do
...
-         {:ok, db_cert} <- Devices.create_device_certificate(device, otp_cert) do
-      updated = update_in(device.device_certificates, &[db_cert | &1])
+         {:ok, _db_cert} <- Devices.create_device_certificate(device, otp_cert) do
+      updated = Repo.preload(device, :device_certificates)
 
-      assign(socket, :device, updated)
-      |> put_flash(:info, "Certificate Upload Successful")
+      {:ok,
+       assign(socket, :device, updated)
+       |> put_flash(:info, "Certificate Upload Successful")}
```
joshk:
```suggestion
      socket
      |> assign(:device, updated)
      |> put_flash(:info, "Certificate Upload Successful")
      |> ok()
```

just a suggestion

nshoes: Much nicer to read, I'll implement.

nshoes: Done!

PR #2284, opened 2025-08-14
Improve SSL certificate logging

joshk:
This PR improves the SSL verification failure logging from:

```
2025-08-14T00:46:33Z app[48e4199b764518] iad level=notice msg="TLS :server: In state :certify at ssl_handshake.erl:2201 generated SERVER ALERT: Fatal - Handshake Failure - :unknown_ca" depth=20 report_cb=&:ssl_logger.format/1
```

to

```
2025-08-14T00:45:50Z app[48e4199b764518] iad level=info msg="SSL certificate verification failed" reason=":unknown_ca" event="nerves_hub.ssl.fail" cert_serial="1234" cert_subject="blah"
```

It involves some log filtering of messages from the Erlang SSL application, and then hooking into some telemetry we already send.


lib/nerves_hub/utils/ssl_logging_filter.ex
jjcarstens, line 1: Do we need a whole other module for this? Maybe it could be a function in the `NervesHub.Logger` module instead?

lib/nerves_hub/utils/ssl_logging_filter.ex
```diff
@@ -0,0 +1,19 @@
...
+  The Erlang SSL application will log issues or failures related to verification of certificates.
+
+  This filter is designed to ignore certain SSL handshake errors that are not relevant to the application's operation.
+
+  eg. TLS :server: In state :certify at ssl_handshake.erl:2201 generated SERVER ALERT: Fatal - Handshake Failure - :unknown_ca
+  """
+
+  def filter(log_event, _opts) do
+    case log_event do
+      %{msg: {:report, %{alert: {:alert, _, _, %{file: ~c"ssl_handshake.erl"}, _, :unknown_ca}}}} ->
```
jjcarstens: I don't know about this. You will capture the events that make it to `NervesHub.SSL` via the verify_fun we've implemented. But with this ignore statement, I think we will miss problems and events logging that terminate in the TLS stack _before_ getting to our verify_fun. But on the other hand, `unknown_ca` is the most common and generic, so maybe not a problem to miss out?

jjcarstens:
But also, this is actually pretty exciting and could be way more helpful 😍

lib/nerves_hub/utils/ssl_logging_filter.ex
joshk: I can move it into `Logger`

joshk: Done

lib/nerves_hub/utils/ssl_logging_filter.ex
joshk: After some chatting with Jon in Slack, I have broadened the match by removing `:unknown_ca`, which is only one of the possible log messages we might get from the SSL logger, but also made it more specific by adding `statename: :certify` so we only filter messages during the `:certify` state, allowing other messages from `ssl_handshake.erl` through.

jjcarstens (approved):

PR #2291, opened 2025-08-19
Use `:zip` and `Plug.Upload.random_file/1` where appropriate

joshk:
Switch from system shelling out for `zip` and `unzip` and use `:zip` instead.

Also, favour `Plug.Upload.random_file/1` for safe cleanup of downloaded or generated files.

### DIFF
--- a/lib/nerves_hub/firmwares/update_tool/fwup.ex
+++ b/lib/nerves_hub/firmwares/update_tool/fwup.ex
@@ -23,6 +23,7 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
          {:ok, tool_metadata} <- get_tool_metadata(meta_conf_path) do
       {:ok,
        %{
+         path: meta_conf_path,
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -49,26 +50,27 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
...
+        {:error, reason} ->
+          Logger.warning("Could not create a firmware delta: #{inspect(reason)}",
+            source_url: source_url,
+            target_url: target_url
+          )
 
-        {:error, :delta_not_created}
+          {:error, :delta_not_created}
+      end
+    after
```
nshoes, line 72: TIL about `after`, cool!

nshoes (approved):
Spoke to soon, deltas were broken for me locally because of bad data 😬 They are humming just fine in our QA.

:shipit:

PR #2293, opened 2025-08-19
Introduce Quokka for automatic lint fixes (defs and line lengths only)

kybishop:
This introduces [Quokka](https://github.com/emkguts/quokka), an automatic formatting tool that respects and fixes issues highlighted by credo configurations. We use Quokka across multiple apps at Smartrent and adoption continues to grow. I've personally found it saving me tens of cumulative hours not having to think about various formatting idiosyncrasies.

This PR is just a step 1 including only line length fixups. After this the diffs of each included fixup module will be significantly smaller (and the wins larger).


lawik:
I am fine with it :)

Want @joshk to weigh in as well.

joshk (approved):
Looks good. I see Quokka likes a line length of 98, which is all good by me.

kybishop:
> Looks good. I see Quokka likes a line length of 98, which is all good by me.

Yeah, this is the kind-of-undocumented default for mix format and Quokka respects it, it just does a bit more to de-clutter multi-line defs that don't need to be such. We can always switch it up if we want.

> Before I merge it in I should finish reviewing and merging in some other related PRs, which might mean that we need to run Quokka again to update the PR.

Makes sense, very easy to just run `mix format` on this one vs asking others to do the same on theirs 👍

nshoes:
Rebased and re-formatted. @josh any other PRs you were looking at merging before merging this work?

PR #2294, opened 2025-08-19
Run Oban migrations (bump to v13)

joshk:
Current Oban migrations on NervesCloud are v12, this makes sure we are up to date.


priv/repo/migrations/20250819222435_upgrade_oban_to_v13.exs
```diff
@@ -0,0 +1,7 @@
+defmodule NervesHub.Repo.Migrations.UpgradeObanToV13 do
+  use Ecto.Migration
+
+  def up, do: Oban.Migrations.up(version: 13)
+
+  def down, do: Oban.Migrations.down(version: 13)
```
nshoes, line 6: Should this be 12 or?

joshk: honestly, I was a little confused, I just went off the example in the docs.

Also, I'm not sure SR is on v12 or earlier?

nshoes: Would we be on different versions?? Now I am the confused one 😆

joshk: You likely are, as its all about when you set up your installation.

nshoes (approved):

PR #2295, opened 2025-08-19
Implement logging of transfer savings from deltas with Postgres

nshoes:
Based on the ClickHouse implementation in #2189, but adapted for PostgreSQL.


lib/nerves_hub/devices.ex
```diff
@@ -974,6 +978,11 @@ defmodule NervesHub.Devices do
         deployment_device_updated(device)
       end
 
+    _ =
+      if previous_metadata do
+        UpdateStats.log_update(device, previous_metadata)
+      end
```
joshk: If you change/add a `log_update/2` function which pattern matches `previous_metadata` to be `nil`, then this line can just become `_ = UpdateStats.log_update(device, previous_metadata)`

lib/nerves_hub/devices/update_stats.ex
```diff
@@ -0,0 +1,228 @@
...
+            num_updates: non_neg_integer()
+          }
+  def stats_by_device(%Device{} = device) do
+    Repo.one(
+      from(s in UpdateStat,
+        where: s.device_id == ^device.id,
+        select: %{
+          total_update_bytes: sum(s.update_bytes),
+          total_saved_bytes: sum(s.saved_bytes),
+          num_updates: fragment("count(*)")
```
joshk: I recommend `total_updates` instead of `num_`

lib/nerves_hub/devices/update_stat.ex
```diff
@@ -0,0 +1,54 @@
...
+      params
+      |> Map.put(:device_id, device.id)
+      |> Map.put(:product_id, device.product_id)
+
+    %__MODULE__{}
+    |> cast(params, @required ++ @optional)
+    |> cast_assoc(:device)
+    |> cast_assoc(:product)
+    |> cast_assoc(:deployment)
+    |> validate_required(@required)
```
joshk: I believe the `cast_assoc` and the casting of id's is double work and we just need to pick one way.

I'm also a fan of using function args over everything being in the params, but I think that is a broader topic which we can address later

lib/nerves_hub/devices/update_stats.ex
```diff
@@ -0,0 +1,228 @@
...
+        # where: s.target_firmware_uuid == ^deployment_group.firmware.uuid,
+        group_by: [s.target_firmware_uuid],
+        select: %{
+          total_update_bytes: sum(s.update_bytes),
+          total_saved_bytes: sum(s.saved_bytes),
+          num_updates: fragment("count(*)"),
+          target_firmware_uuid: s.target_firmware_uuid
+        }
+      )
+    )
```
joshk: We seem to be mixing Ecto styles, could we look at using the pipeline version for consistency?

Or at the very least, move the `Repo.x` functions to after the ecto compositions

joshk (changes requested):
Honestly, looks great to me, lets ship ASAP and we can improve from there

lib/nerves_hub/devices.ex
nshoes: Note: This is an old comment and is no longer relevant. Though it turns out I implemented exactly what you asked for here 😂

lib/nerves_hub/devices/update_stats.ex
nshoes: Done!

lib/nerves_hub/devices/update_stat.ex
nshoes: Done!

lib/nerves_hub/devices/update_stats.ex
nshoes: Done!

config/dev.exs
```diff
@@ -95,7 +95,7 @@ config :nerves_hub, NervesHub.ObanRepo,
 
 if System.get_env("ANALYTICS_ENABLED", "true") == "true" do
   config :nerves_hub, NervesHub.AnalyticsRepo,
-    url: System.get_env("CLICKHOUSE_URL", "http://default:@localhost:8123/default")
+    url: System.get_env("CLICKHOUSE_URL", "http://default:@localhost:8123/default_dev")
```
nshoes, line 90: This, along with the change to the test config, are changes from Lars' original work, but I'm inclined to keep them.

test/nerves_hub_web/live/new_ui/deployment_groups/show_test.exs
nshoes, line 1: Tests originally here are moved to `test/nerves_hub_web/live/new_ui/deployment_groups/settings_test.exs`.

lib/nerves_hub/devices/update_stats.ex
```diff
@@ -0,0 +1,218 @@
...
+  import Ecto.Query
+
+  require Logger
+
+  @spec stats_by_device(Device.t()) ::
+          %{
+            total_update_bytes: non_neg_integer(),
+            total_saved_bytes: integer(),
+            total_updates: non_neg_integer()
+          }
```
joshk, line 27: Not a deal breaker, just a little easier for reading, we could move this spec return typing (and the following functions) to be their own types

lib/nerves_hub/devices/update_stats.ex
```diff
@@ -0,0 +1,218 @@
...
+              "deployment:#{device.deployment_id}:internal",
+              "stat:logged",
+              %{}
+            )
+          end
+
+        :ok
+
+      {:error, %Ecto.Changeset{} = changeset} = error ->
+        Logger.warning(
```
joshk, line 162: is it worth creating a sentry error for this?

joshk (approved):
Love it!

Two small comments, you can do as you want with them

lib/nerves_hub/devices/update_stats.ex
nshoes: Yes! Good call.

lib/nerves_hub/devices/update_stats.ex
nshoes: I don't feel strongly either way here; I'm going to leave as is.

PR #2296, opened 2025-08-20
Remove default subscriptions setup when a device connects

joshk:
All devices, when connecting to a topic ("device", "console", "extensions"), are lumped into a PubSub topic using the channel name (the name used in join/3).

Since these names are generic, what ends up happening is that every device is subscribed to a shared topic name that has fastlaning enabled, meaning messages can be sent directly to devices.

I believe this could be a slight security issue, as pubsub messages can be sent to all connected devices with just a simple:

```elixir
Phoenix.Channel.Server.broadcast(NervesHub.PubSub, "device", "any_event_name", %{just: "a payload"})
```

Additionally, this topic isn't needed or used, so we can unsubscribe from it.

### DIFF
--- a/lib/nerves_hub_web/channels/console_channel.ex
+++ b/lib/nerves_hub_web/channels/console_channel.ex
@@ -62,6 +62,11 @@ defmodule NervesHubWeb.ConsoleChannel do
   def handle_info({:after_join, payload}, socket) do
     socket = assign(socket, :version, payload["console_version"])
 
+    # all devices are lumped into a `console` topic (the name used in join/3)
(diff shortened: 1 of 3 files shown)


lawik:
This seems like a good move to me.

Ideally we'd switch nerves_hub_link to use `"device:" <> identifier` but I guess that'll be a longer more intricate transition if so.

I'll let Nate and Jon take a look to make sure there isn't something I'm missing.

joshk:
@jjcarstens @nshoes I've tested this on NC and everything works as expected

PR #2297, opened 2025-08-20
Smart Device PubSub fastlaning

joshk:
This builds upon https://github.com/nerves-hub/nerves_hub_web/pull/2296

Behind the scenes, after the Device has connected, and after unsubscribing from `device`, I setup a new subscription of `device:[device.id]`, along with fastlaning to the underlying transport pid.

I've also created a new module of `NervesHub.DeviceEvents` to encapsulate broadcasts specific to connected devices. I've also used this module to encapsulate additional logic around the event.

Three events can now be fastlaned: **identify**, **reboot**, and **update** (scheduled and manual)

And two events are 'intercepted' for processing by the channel **updated** and **deployment_updated**.

I've also slipped in a change where `DeviceChannel` no longer directly logs, but instead uses telemetry, which also makes the logging format consistent.


lib/nerves_hub/devices/inflight_update.ex
```diff
@@ -31,4 +31,11 @@ defmodule NervesHub.Devices.InflightUpdate do
       name: :inflight_updates_device_id_deployment_id_index
     )
   end
+
+  def update_status(inflight_update, status) do
```
nshoes:
```suggestion
  def update_status_changeset(inflight_update, status) do
```

lib/nerves_hub/devices.ex
```diff
@@ -824,8 +825,7 @@ defmodule NervesHub.Devices do
       |> Ecto.Changeset.put_change(:deployment_id, deployment_group.id)
       |> Repo.update!()
 
-    _ =
-      broadcast(device, "devices/deployment-updated", %{deployment_id: deployment_group.id})
+    DeviceEvents.deployment_assigned(device)
```
nshoes, line 816: This is so much nicer 😭

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -87,98 +98,16 @@ defmodule NervesHubWeb.DeviceChannel do
...
 
-  # manually pushed
-  def handle_info(%Broadcast{event: "devices/update-manual", payload: payload}, socket) do
-    :telemetry.execute([:nerves_hub, :devices, :update, :manual], %{count: 1})
-    push(socket, "update", payload)
-    {:noreply, socket}
-  end
-
+  # we can ignore this message
   def handle_info(%Broadcast{event: "deployments/update"}, socket) do
```
nshoes, line 102: Why are we subscribed to the devices' deployment topic? I believe we no longer respond to any messages from it in this channel; it should be safe to remove (`update_deployment_group_subscription/2`).

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -225,9 +152,37 @@ defmodule NervesHubWeb.DeviceChannel do
...
+
+    socket =
+      socket
+      |> update_device(device)
+      |> maybe_send_archive(audit_log: true)
+
+    {:noreply, socket}
+  end
+
+  # def handle_out("deployment_updated", %{deployment_id: deployment_id}, socket) do
```
nshoes:
```suggestion
```

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -225,9 +152,37 @@ defmodule NervesHubWeb.DeviceChannel do
...
+  def handle_out("deployment_updated", payload, socket) do
+    device = %{socket.assigns.device | deployment_id: payload.deployment_id}
+
+    socket =
+      socket
+      |> update_device(device)
+      |> maybe_send_archive(audit_log: true)
```
nshoes, line 183: Good addition 💯

lib/nerves_hub/devices/inflight_update.ex
joshk: love it! I'll get this updated!

lib/nerves_hub/devices.ex
joshk: Thank you 😊

lib/nerves_hub_web/channels/device_channel.ex
joshk: We need that subscription for when a deployments archive is updated. Rethinking this would be awesome, but I think its best to wait for another PR.

lib/nerves_hub_web/channels/device_channel.ex
joshk: 😊

nshoes (approved):
I tested rebotting, manual, and scheduled updates in our QA, and everything works! 🚢 🚢🚢

joshk:
💥

PR #2308, opened 2025-08-26
Only update the target firmware file with deltas that are smaller

joshk:
Instead of creating a new firmware file from scratch, this only updates/replaces files which have changed.

This is an extension of #2307


joshk:
I have deployed this to NervesCloud for testing, and it has helped reduce my 200KB delta down to 150KB

PR #2314, opened 2025-09-01
Track a device's firmware validation status, and if an auto-revert was detected

joshk:
This is the Hub side of https://github.com/nerves-hub/nerves_hub_link/pull/331

By tracking a devices firmware validation status, and if an auto revert is detected, we can now show this in the UI.

Later on we can look at excluding devices from receiving firmware updates until they they validated, and potentially block updates for X period of time if a revert is detected, which can prevent devices being stuck in update-revert loops.

Here is what I have added to the UI:

Screenshot 2025-09-01 at 1 51 33 PM

Screenshot 2025-09-01 at 1 51 49 PM

Screenshot 2025-09-01 at 1 51 02 PM


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -29,10 +29,16 @@ defmodule NervesHubWeb.DeviceChannel do
   alias NervesHub.Repo
   alias Phoenix.Socket.Broadcast
 
+  require Logger
+
   intercept(["updated", "deployment_updated"])
 
   @decorate with_span("Channels.DeviceChannel.join")
   def join("device", params, %{assigns: %{device: device}} = socket) do
+    Logger.metadata(device_id: device.id, device_identifier: device.identifier)
```
nshoes, line 38: This seems so obvious in hindsight... 😆

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -58,6 +64,22 @@ defmodule NervesHubWeb.DeviceChannel do
     end
   end
 
+  defp maybe_sanitize_device_api_version(%{"device_api_version" => version} = params) do
+    case Version.parse(version) do
```
nshoes, line 68: Did you run into this in the wild? I'm all for the defensiveness, but we control both link + hub and think it's fair to assume that we'll get a valid version here.

nshoes (approved):
Nice work! Tests are great as well.

🚢 🚢 🚢 🚢 🚢

lib/nerves_hub_web/channels/device_channel.ex
joshk: I have to thank @lawik for this

lib/nerves_hub_web/channels/device_channel.ex
joshk: this came up when I wrote a test with an invalid device api version.

I generally agree with you, and yet decided to be defensive as the library is open source, so we should account for forks or custom libs.

nshoes: Cool, agreed, thanks.

PR #2318, opened 2025-09-02
Quokka block fixups

kybishop:
The next step in the Quokka train 💪


.formatter.exs
```diff
@@ -6,6 +6,8 @@
   plugins: [Phoenix.LiveView.HTMLFormatter, Quokka],
   quokka: [
     only: [
+      :blocks,
+      :comment_directives,
```
kybishop, line 10: comment directives don't actually do anything unless you manually add one, e.g. `# quokka:sort`

lib/nerves_hub_web/live/devices/settings.ex
```diff
@@ -72,11 +72,13 @@ defmodule NervesHubWeb.Live.Devices.Settings do
...
+    db_cert = Enum.find(certs, &(&1.serial == serial))
+
+    case Devices.delete_device_certificate(db_cert) do
+      {:ok, _db_cert} ->
+        updated_certs = Enum.reject(certs, &(&1.serial == serial))
+        {:noreply, assign(socket, device: %{device | device_certificates: updated_certs})}
+
       _ ->
         {:noreply, put_flash(socket, :error, "Failed to delete certificate #{serial}")}
     end
```
kybishop, line 84: Sometimes I'm still amazed at how cool Quokka's automatic fixups are

nshoes (approved):
🚂 🚂 🚂 🚂

lib/nerves_hub_web/live/devices/settings.ex
joshk: this is an interesting style enforcement. I generally agree that `with` likely isn't the best, but I do like the flow.

PR #2319, opened 2025-09-02
Quokka config and deprecation fixups

kybishop:
Configs are now alphabetized automatically and deprecations removed (in this case just using the native `to_timeout` instead of the erlang `:timer`.

### DIFF
--- a/lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
+++ b/lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
@@ -61,7 +61,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
       PubSub.subscribe(NervesHub.PubSub, "orchestrator:deployment:#{deployment_group.id}")
 
     # trigger every two minutes, plus a jitter between 1 and 10 seconds, as a back up
-    interval = :timer.seconds(120 + :rand.uniform(20))
+    interval = to_timeout(second: 120 + :rand.uniform(20))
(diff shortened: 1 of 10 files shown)


lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
```diff
@@ -61,7 +61,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
       PubSub.subscribe(NervesHub.PubSub, "orchestrator:deployment:#{deployment_group.id}")
 
     # trigger every two minutes, plus a jitter between 1 and 10 seconds, as a back up
-    interval = :timer.seconds(120 + :rand.uniform(20))
+    interval = to_timeout(second: 120 + :rand.uniform(20))
```
joshk, line 64: hmmm, why `to_timeout` vs `:timer`?

kybishop: Just to make sure we're on the same page these are all just automatic changes from quokka, not anything I'm personally implementing.

If you're asking me as a maintainer of quokka though, I would say because `to_timeout` is a more extensible version that happens to be Elixir native while `:timer` is an Erlang library. Personally I don't have a strong opinion either way.

joshk: Oh, totally understand. Sorry, I should have said `Quakka, why ...`

I personally think `:timer.seconds` reads better, and I think its being used under the hood, but I'm fine with these changes.

PR #2320, opened 2025-09-02
Quokka pipe fixups

kybishop:


.formatter.exs
```diff
@@ -10,7 +10,9 @@
       :comment_directives,
       :configs,
       :defs,
-      :deprecations
-    ]
+      :deprecations,
+      :pipes
+    ],
+    exclude: [piped_functions: [:subquery]]
```
kybishop, line 16: There's one instance of a subquery used in a pipe that Quokka was moving to the bottom of said pipe.

At Smartrent this is the only diversion from the default rules that we utilize, but I don't have strong enough feelings to fight for it, so happy to defer to others on keeping it or not.

joshk (approved):

PR #2258, opened 2025-08-07
Resolve Credo issues and enforce it in CI

taras-tyshko:
## 🎯 Objective

Improve code quality and enforce coding standards by making Credo checks mandatory in CI, while resolving all existing code quality issues.

### Enforced Credo in CI
- Update `--mute-exit-status` from `.github/workflows/ci.yml` on `--min-priority low` (full comment here https://github.com/nerves-hub/nerves_hub_web/pull/2258#discussion_r2262908419)
- Credo will now fail the build when code quality issues are detected
- All existing issues have been resolved

https://github.com/nerves-hub/nerves_hub_web/pull/2258#discussion_r2262908419

### Software Design Issues (21 files)
- **Nested modules aliasing**: Added proper `alias` statements for nested modules like `X509.Certificate.Extension`, `Phoenix.Channel.Server`, `NervesHub.Telemetry.Customizations`
- **Examples**: `lib/nerves_hub/certificate.ex`, `lib/nerves_hub_web/plugs/configure_uploads.ex`, `lib/nerves_hub/extensions/health.ex`

### Readability Issues (6 files)  
- **Alias ordering**: Fixed alphabetical ordering of alias statements
- **Explicit try blocks**: Converted to implicit try blocks in `lib/nerves_hub/firmwares/update_tool/fwup.ex`
- **Zero-arity functions**: Added parentheses to functions like `def spec()` in `lib/nerves_hub_web/api_spec.ex`
- **Large numbers**: Formatted with underscores (e.g., `578088` → `578_088`)
- **String literals**: Replaced multi-quote strings with sigils in tests

### Refactoring Opportunities (3 files)
- **Cyclomatic complexity**: Refactored `friendly_blocked_until/1` in `lib/nerves_hub_web/components/device_update_status.ex` using pattern matching
- **Complex macros**: Split `updated_live_view/0` and `hooked_component/1` into smaller, focused functions
- **Enum optimizations**: Replaced `Enum.map |> Enum.join` with `Enum.map_join` in tests

### Configuration Updates
- **Disabled checks**: Made `TODO` comments and `@moduledoc` tags optional in `.credo.exs`
- **Why**: `TODO` comments are acceptable in our codebase for tracking technical debt, and `@moduledoc` requirements would be too disruptive for existing modules

## ✅ Results

- **0 Credo warnings** across 385 source files
- Improved readability and maintainability of complex macros  
- Ensured code quality enforcement for future PRs via CI

PR #2323, opened 2025-09-03
Quokka autosort

kybishop:
This is the second to last Quokka fixup. We'll want to get #2258 before the final PR for `module_directives`.


kybishop:
Closing in favor of a broken up version to make this PR a little less daunting

PR #2324, opened 2025-09-03
Quokka autosort defstruct and schema

kybishop:


lib/nerves_hub/devices/device.ex
```diff
@@ -37,58 +37,56 @@ defmodule NervesHub.Devices.Device do
...
+    # Also remove index from NervesHub.Repo.Migrations.AddConnectionStatusIndexToDevices.
+    # field(:connection_status, Ecto.Enum,
+    #   values: [:connected, :disconnected, :not_seen],
+    #   default: :not_seen
+    # )
+    # field(:connection_established_at, :utc_datetime)
+    # field(:connection_disconnected_at, :utc_datetime)
+    # field(:connection_last_seen_at, :utc_datetime)
+    # field(:connection_metadata, :map, default: %{})
+    # field(:connection_types, {:array, Ecto.Enum}, values: [:cellular, :ethernet, :wifi])
```
kybishop: Quokka really doesn't like these being at the bottom of the schema and tries to push the comments out of the schema block. Perhaps better for these to be at the top anyways to remind folks it needs to be cleaned up 🤷 😅

joshk: @kybishop, would you mind creating a separate PR that removes these columns? I think its time to clean them up.

lib/nerves_hub/accounts/invite.ex
```diff
@@ -11,13 +11,13 @@ defmodule NervesHub.Accounts.Invite do
...
-
-    field(:email, :string)
-    field(:token, Ecto.UUID)
     field(:accepted, :boolean)
+    field(:email, :string)
     field(:role, Ecto.Enum, values: Ecto.Enum.values(OrgUser, :role))
+    field(:token, Ecto.UUID)
+
+    belongs_to(:invited_by, User)
+    belongs_to(:org, Org)
```
joshk: oh sooo these changes leave me a little torn, as I do like to group related information together. is this something we can opt out of?

lib/nerves_hub/accounts/org_user.ex
```diff
@@ -9,12 +9,12 @@ defmodule NervesHub.Accounts.OrgUser do
...
   schema "org_users" do
+    field(:deleted_at, :utc_datetime)
+    field(:role, Ecto.Enum, values: [:admin, :manage, :view])
+
     belongs_to(:org, Org, where: [deleted_at: nil])
     belongs_to(:user, User, where: [deleted_at: nil])
 
-    field(:role, Ecto.Enum, values: [:admin, :manage, :view])
-    field(:deleted_at, :utc_datetime)
-
```
joshk: for me, this is a good example of a change that doesn't feel beneficial. I know that ordered info can be easier to find, on the flip side it is degrouping information and putting less important relationship at the top.

lib/nerves_hub/devices/device.ex
joshk: Funnily enough, this is the one example of moving stuff to the top that I like 😆

joshk (changes requested):
I like the concept and half of the changes, but I'm not sure I'm behind all the reordering of ecto schema definitions, ungrouping fields etc.

lib/nerves_hub/accounts/invite.ex
nshoes: @kybishop is this possible?

kybishop: for sure, autosorting is completely optional. Personally I love it for not having to figure out where to put things but it's not something we need to incorporate for the benefits of Quokka. There is also the comment directive `# quokka:skip-sort`, which can be used to skip sorting a particular example.

For now I've edited this PR to just be sorting of defstruct so it's a smaller piece and we can perhaps revisit ecto schemas later with usage of `# quokka:skip-sort` where the current order is preferred.

kybishop: also apologies for the delay, I got sucked in to some other work but I'm back on Nerves things for a bit :)

nshoes (approved):

PR #2326, opened 2025-09-04
A small optimization to `Devices.update_attempted/2`

joshk:
The `DeviceChannel` doesn't use the `update_attempts` information, so instead of reloading the device and updating the field via a changeset, I've taken advantage of Postgres's ability to append to the array via and SQL `update` query.


joshk:
@nshoes THE SUSPENSE!

PR #2336, opened 2025-09-09
Move device connection pubsub updates to the `Connections` context

joshk:
While the Tracker gets a rewrite, this improves a bit of code I came across where we announce a device online twice, once in the `DeviceSocket` and once in `DeviceChannel`.

I still send two, but now I send `connecting` and `online`.

I've also encapsulated all connection related pubsub messages to `Tracker`, and the `Connections` context is the primary user of Tracker.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -7,6 +7,7 @@ defmodule NervesHub.Devices.Connections do
   alias NervesHub.Devices.Device
   alias NervesHub.Devices.DeviceConnection
   alias NervesHub.Repo
+  alias NervesHub.Tracker
 
   @doc """
   Get all connections for a device.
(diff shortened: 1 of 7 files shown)


lib/nerves_hub/devices/connections.ex
```diff
@@ -128,8 +134,12 @@ defmodule NervesHub.Devices.Connections do
       timeout: 60_000
     )
     |> case do
-      {1, _} -> :ok
-      _ -> :error
+      {1, _} ->
+        Tracker.offline(device)
```
nshoes, line 138: Since we only pass a device to `offline/1` we can remove the definition that takes the identifier as the arg in Tracker.

joshk: Great spotting!

PR #2344, opened 2025-09-17
Remove the default manual update submit button

joshk:
It's essentially duplicated for some unknown reason

### DIFF
--- a/lib/nerves_hub_web/components/device_page/details_tab.ex
+++ b/lib/nerves_hub_web/components/device_page/details_tab.ex
@@ -367,7 +367,7 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
           </div>
 
           <div :if={Enum.any?(@firmwares)} class="flex p-4 gap-4 items-center border-t border-zinc-700">
-            <form id="push-update-form" phx-change="select-firmware-version" phx-submit="push-update" class="flex gap-2 items-center w-full">
+            <form phx-change="select-firmware-version" class="flex gap-2 items-center w-full">
               <div class="grow grid grid-cols-1">
                 <label for="firmware" class="hidden">Firmware</label>
                 <select
@@ -393,7 +393,6 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
                 Send delta update
               </.button>
               <.button
-                :if={@delta_available?}
                 type="button"
                 disabled={disconnected?(@device_connection)}
                 aria-label="Send firmware update"
@@ -402,15 +401,6 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
                 phx-click="push-update"
               >
                 Send full update
-              </.button>
-              <.button
-                :if={not @delta_available?}
-                type="submit"
-                disabled={disconnected?(@device_connection)}
-                aria-label="Send firmware update"
-                data-confirm="Are you sure you want to send this firmware to the device?"
-              >
-                Send update
               </.button>
             </form>
           </div>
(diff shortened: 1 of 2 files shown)


lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -367,7 +367,7 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
           </div>
 
           <div :if={Enum.any?(@firmwares)} class="flex p-4 gap-4 items-center border-t border-zinc-700">
-            <form id="push-update-form" phx-change="select-firmware-version" phx-submit="push-update" class="flex gap-2 items-center w-full">
+            <form phx-change="select-firmware-version" class="flex gap-2 items-center w-full">
```
lawik: I believe the ID prevented the form from being unusuable during updates when a lot of progress report messages made the UI redraw. LiveView/morphdom is smarter about not updating elements with IDs unnecessarily.

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -404,15 +403,6 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
               >
                 Send full update
               </.button>
-              <.button
```
lawik, line 407: There were two buttons as a fallback in case a device has trouble with delta updates for whatever reason. Or for test purposes.

So if there is a delta available we'd show two buttons, one for delta and one for full update.

It also allows testing and comparing the speed of updates between delta and full.

lawik (changes requested):
I don't think this was an unnecessary button. Happy to discuss if we want to change how this is presented. Maybe with a dropdown for the Full update as that'd be less common when a delta is available.

Also the title of the PR confused me a bit. After reading it I understand it wasn't the word "update" as a generic term. I thought this would be a settings form or something which didn't need a manual "Save" button. But it was literally about firmware updates :D

lib/nerves_hub_web/components/device_page/details_tab.ex
joshk: hmmmm, thats interesting, I'll have to test that out and see whats going on as this sounds like a slight more complex thing going on, like extra data being reloaded.

I'll add the id back for now.

An issue I did address is on line 379 where we weren't telling LV what firmware version had previously been selected. Thus, if any new firmware versions came through, this was being overwritten.

lib/nerves_hub_web/components/device_page/details_tab.ex
joshk: We still have that, because the extra button, which was doing a standard form submit, was using the same `handle_info` as the one that I've kept, which is using `push-update` via `phx-click`.

The delta button is shown if the firmware supports it, but the full update option is always shown.

joshk:
> I don't think this was an unnecessary button. Happy to discuss if we want to change how this is presented. Maybe with a dropdown for the Full update as that'd be less common when a delta is available.
> 
> Also the title of the PR confused me a bit. After reading it I understand it wasn't the word "update" as a generic term. I thought this would be a settings form or something which didn't need a manual "Save" button. But it was literally about firmware updates :D

I was likely too brief in my PR description.

There were two buttons that both did the same thing, a full update. This PR removed one of them, opting to keep the `phx-click` version and removing the `phx-submit` on the form.

Everything works as expected, but now we have just two buttons.

lawik:
Ah, then I misread the diff :)

lawik (approved):

PR #2345, opened 2025-09-17
Enable delta updates for new deployments

joshk:


lib/nerves_hub/firmwares.ex
```diff
@@ -144,6 +144,10 @@ defmodule NervesHub.Firmwares do
   @spec get_firmware(Org.t(), integer()) ::
           {:ok, Firmware.t()}
           | {:error, :not_found}
+  def get_firmware(_, nil) do
+    {:error, :not_found}
+  end
```
nshoes, line 149: Why is this needed? Can we handle this case at the caller instead of here?

joshk: During my testing I discovered an error we have have in the 'create deployment' workflow for a while. If you click 'save' but you haven't selected a firmware, an Ecto query builder error is thrown because you can't use `where([f], f.id == ^id)` when `id` is `nil`.

This felt like a reasonable fix for the error as `:not_found` felt within the bounds of the firmware id being `nil`.

PR #2347, opened 2025-09-18
Add `Connected` and `Last Seen` to the Devices General Info box

joshk:
When the device is disconnected:

Screenshot 2025-09-18 at 4 20 59 PM

And when it is connected:

Screenshot 2025-09-18 at 4 22 14 PM


lawik (approved):
Nice!

PR #2349, opened 2025-09-18
Switch from `phx-change` to `phx-click`

joshk:
This should help stop LV thinking the checkbox has changed and firing an event.


lawik (approved):
Wild..

PR #2354, opened 2025-09-19
Org settings: Firmware proxy URL

joshk:
This is a customer request that I felt had merit.

A NervesCloud customer requires custom firmware download URLs due to some of the complexities associated with whitelisting URLs with their customers.

This PR addresses this feature request by adding a `settings` field (and embedded schema) to the `Org` schema, adding a `firmware_proxy_url` field, and using this URL when creating the update payloads we send to devices.


lawik (approved):
Looks good. I was a bit tripped up by some code shifting over into DeviceEvents for a second but it makes sense :)

PR #2361, opened 2025-09-24
Support log event exclusions

jjcarstens:
In some setups, you may not want the full verbosity of events in the logs. This adds a mechanism to include a `LOGGER_EXCLUSIONS` env var with a list of event keys that will be filtered out before telemetry attachment so that they will not be logged as well


joshk:
Jon and I had a good chat about this and we agreed on a slightly different approach where logging can be detached at application start up, thus allowing us to keep the info logging as the default.

nshoes:
You're right, I think I confused `:telemetry.attach/4` as a mechanism for the telemetry events itself, but it only has to do with attaching the handler function that we use for logging. I'll adjust.

PR #2363, opened 2025-09-25
Use `System.monotonic_time/1` for quick interval calculations

joshk:
Just a small change from `DateTime.utc_now` as we don't need 100% accurate dates and times.


joshk:
> Do we have any tests for this?

for heartbeats? yes

not for the use of `monotonic_time/1` as that is an implementation detail

fhunleth (approved):
I really like this. Even though I know this runs on a server with a really well managed clock, I'm such of fan of the simplicity of monotonic time.

PR #2368, opened 2025-09-28
Encapsulate logic related to resetting update attempts into a changeset function

joshk:
Just a tiny little change which caught my eye


nshoes (approved):
Get that Ecto out of my contexts!! 🥇

PR #2374, opened 2025-10-06
Switch all Product nested UIs to the updated design

joshk:
This PR removes the 'New UI' switch, making the new UI available, and default, to all users.

This temporarily removes the Pinned devices UI on the Orgs list page. 

I'd recommend adding that back in another PR which also implements the new design for the Org pages.


mix.exs
nshoes, line 1: I ran into a similar issue recently and solved it by [enabling the code reloader only in dev](https://github.com/nerves-hub/nerves_hub_web/pull/2262/files#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308R196). Would you mind if we approached it that way instead of checking env vars?

nshoes (approved):
Tested locally, LGTM!

mix.exs
joshk: ahhhhh, much smarter! yes, that sounds better!

PR #2377, opened 2025-10-10
Fix registration problems where people's names are called invalid

lawik:
It is now used as a display name.


joshk:
Is this format validation too restrictive on the characters?

joshk:
Totally clicks, thanks for the confirmation!

joshk (approved):

nshoes:
@lawik Ha! Thanks for the explanation :)

PR #2381, opened 2025-10-17
Remove controller and live view param whitelisting

joshk:
Ecto Changesets are the right way to do whitelisting, so this isn't needed.

(Built upon #2380)


nshoes (approved):
LGTM! I hate whitelisting outside of changesets.

PR #2382, opened 2025-10-17
A better Device tab UI experience when there is latency

joshk:
When a tab is clicked, select it straight away, reduce the opacity of the current tab content by 50%, and when the tab has finished loading, fade the new content in.

https://github.com/user-attachments/assets/482fc102-ee25-478b-97ba-c343c41b2146

This is using the latency simulator set to 300ms


nshoes (approved):
Oh yeah, that is NICE 🥇 🥇 🥇

lawik:
Nice!

PR #2391, opened 2025-11-20
Tools for inspecting timing and performance during production

lawik:
I think we'd benefit from building up a small toolset for just introspecting the running system in various ways and getting relevant information. This stuff in particular were things I used while making sense of the slow performance of the device index page.

If you have other ideas of how to include tooling like this, happy to hear it :)

### DIFF
--- a/lib/nerves_hub/debug.ex
+++ b/lib/nerves_hub/debug.ex
@@ -1,0 +1,126 @@
+defmodule NervesHub.Debug do
+  @moduledoc """
+  Tools for debugging performance in production.
+  """
+
+  def mark(label) do
+    t = Process.get(:t)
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/debug.ex
```diff
@@ -0,0 +1,128 @@
...
+
+          Enum.each(measurements, fn {k, v} ->
+            if v > threshold do
+              ms = System.convert_time_unit(v, :native, :millisecond)
+              IO.inspect("#{inspect(event)} #{k}: #{ms} milliseconds")
+            end
+          end)
+        end
+
+        # IO.inspect(measurements)
```
nshoes:
```suggestion
```

PR #2397, opened 2025-11-24
Fix deployment group tags being merged when editing settings

### LINKED ISSUE #2393: Viewing Deployment Group settings merges all tags together
When viewing the `settings` tab of a deployment group, all tags are merged together in the form. This is problematic for groups with 2+ tags as saving any other setting will rewrite the tags with this incorrect value. If not caught, this may produce unexpected results in grouping and filtering

Image

Image

kybishop:
Previously they were merged into a single string with no comma separation. Fixes #2393 

https://github.com/user-attachments/assets/2caee870-bfdb-40d2-b08e-0998509e8b51


nshoes (approved):
Sweeeet! :shipit:

Issue #2395, opened 2025-11-21
Connecting Code uses console channel which can cause memory overload


jjcarstens:
We had an incident where nodes memory started filling out of control and crashing with OOM. We suspected a bad deploy, rolled back to a version we had been using for quite a while, and it began to occur again.

What we found was that that problem was actually our connecting code we had put in deployment groups targeting a few hundred thousand devices. CC uses the [console channel and sends one character at a time](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/channels/console_channel.ex#L82-L100), so our PubSub mailboxes were filling with millions of messages like this:

```elixir
{:forward_to_local, "user:console:12345",
    %Phoenix.Socket.Broadcast{
      topic: "user:console:12345",
      event: "up",
      payload: %{"data" => "n"}
    }, Phoenix.Channel.Server}
```

This is because when you push to the console IO, it also echos back what it receives, and we attempt to broadcast that to any user console tabs open. This meant we were making `N characters * N nodes * N device connection`. Our nodes sit at ~60% capacity for device connections. In our case, we would see a slow rise in memory because a fair number of device connections would flip quicker than the PubSub mailboxes would GC their binary heap, leading to continual growth. When one node would, a cascading effect would happen - All device connections would attempt to be spread across the remain nodes, but connections would fail at a faster rate causing more single character message to every node until the PubSub mailboxes were too full to process. Run queues could frequently spike 4-8k.

I suggest shifting to use the support scripts path for sending executable code to the device rather than the console path since that is intended for live user input. We can also incorporate tracking into the success/failure of connecting code as well in the future

PR #2399, opened 2025-11-24
Use scripts plumbing for connecting code

nshoes:
Solves #2395. 

This uses `scripts/run` to push connecting code to the device rather than broadcasting it to every device node.

I added telemetry events for when the connecting code succeeds or fails, as well as logging when it fails. The contract for "failing" a script run is in NHL, which I don't love, but can be fixed at a later time.

### DIFF
--- a/lib/nerves_hub_web/channels/device_channel.ex
+++ b/lib/nerves_hub_web/channels/device_channel.ex
@@ -82,10 +82,10 @@ defmodule NervesHubWeb.DeviceChannel do
 
       [false, false] ->
         connecting_code = Enum.join(connecting_codes, "\n")
+        text = ~s/#{connecting_code}\n\r/
         topic = "device:console:#{device.id}"
 
-        socket.endpoint.broadcast_from!(self(), topic, "dn", %{"data" => connecting_code})
-        socket.endpoint.broadcast_from!(self(), topic, "dn", %{"data" => "\r"})
+        socket.endpoint.broadcast_from!(self(), topic, "dn", %{"data" => text})
 
       _ ->
         :ok


lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -172,8 +188,23 @@ defmodule NervesHubWeb.DeviceChannel do
     {:noreply, socket}
   end
 
+  def handle_in("scripts/run", %{"ref" => "connecting_code", "return" => "nil", "output" => output}, socket)
+      when byte_size(output) > 0 do
+    :telemetry.execute([:nerves_hub, :devices, :connecting_code_failure], %{output: output})
```
jjcarstens:
```suggestion
    :telemetry.execute([:nerves_hub, :devices, :connecting_code_failure], %{output: output, identifier: socket.assigns.device.identifier})
```
Can we put the identifier in here so we can also track by device if needed?

lib/nerves_hub/logger.ex
```diff
@@ -94,6 +95,13 @@ defmodule NervesHub.Logger do
     )
   end
 
+  def log_event([:nerves_hub, :devices, :connecting_code_failure], _, metadata, _) do
+    Logger.info("Connecting code failure",
+      event: "nerves_hub.devices.connecting_code_failure",
+      output: metadata[:output]
```
jjcarstens:
```suggestion
      identifier: metadata[:identifier],
      output: metadata[:output]
```
And also include it here 😉

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -172,8 +188,23 @@ defmodule NervesHubWeb.DeviceChannel do
     {:noreply, socket}
   end
 
+  def handle_in("scripts/run", %{"ref" => "connecting_code", "return" => "nil", "output" => output}, socket)
```
jjcarstens: I don't think this is quite right. `return` could be a caught exception which is what the `nil` accounts for. But there are [other conditions that `result == error`](https://github.com/nerves-hub/nerves_hub_link/blob/main/lib/nerves_hub_link/socket.ex#L535-L551) that we should probably consider for a while until fixed up

```suggestion
  def handle_in("scripts/run", %{"ref" => "connecting_code", "result" => result, "return" => return, "output" => output}, socket) when result == "error" or return == "nil"
```

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -63,7 +63,23 @@ defmodule NervesHubWeb.DeviceChannel do
...
+      [
+        get_in(device, [Access.key(:deployment_group), Access.key(:connecting_code)]),
+        device.connecting_code
+      ]
+      |> Enum.join("\n")
+
+    if is_binary(connecting_code) and byte_size(connecting_code) > 0 do
+      # connecting code first incase it attempts to change things before the other messages
+      push(socket, "scripts/run", %{"text" => connecting_code, "ref" => "connecting_code"})
+    end
```
jjcarstens, line 92: We probably need to do the [same thing checking for api version when running a script](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub_web/channels/device_channel.ex#L83-L103) since we may need a fallback to the [old way using the console channel](https://github.com/nerves-hub/nerves_hub_web/blob/main/lib/nerves_hub/scripts/runner.ex#L65-L73) (though we should adapt that to see all characters at once instead of one at a time). At SmartRent, we have ~1300 devices still on < 2.1.0 of the device api

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -172,8 +188,23 @@ defmodule NervesHubWeb.DeviceChannel do
...
+    :telemetry.execute([:nerves_hub, :devices, :connecting_code_success], %{count: 1})
+
+    {:noreply, socket}
+  end
+
   def handle_in("scripts/run", params, socket) do
-    if pid = socket.assigns.script_refs[params["ref"]] do
+    refs = Map.get(socket.assigns, :script_refs, %{})
+
+    if pid = refs[params["ref"]] do
```
jjcarstens: NBD, but just a passing thought - If you wanted to avoid assignment, you could leverage `get_in` here to fetch the ref if it exists
```suggestion
    if pid = get_in(socket.assigns, [:script_refs, params["ref"]) do
```

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Done!

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Ah, I'm not on `main` locally for Link and missed those changes. Thanks for the callout.

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Cool, I like this much more than the double bracket access!

lib/nerves_hub_web/channels/device_channel.ex
nshoes: Done!

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -63,7 +64,31 @@ defmodule NervesHubWeb.DeviceChannel do
...
+      connecting_code = Enum.join(connecting_codes, "\n")
+      # connecting code first incase it attempts to change things before the other messages
+      push(socket, "scripts/run", %{"text" => connecting_code, "ref" => "connecting_code"})
+    else
+      connecting_code = Enum.join(connecting_codes, "\n")
+      text = ~s/#{connecting_code}\n# [NERVESHUB:END]/
+      topic = "device:console:#{device.id}"
+
+      Endpoint.broadcast_from!(self(), topic, "dn", %{"data" => text})
+      Endpoint.broadcast_from!(self(), topic, "dn", %{"data" => "\r"})
```
jjcarstens: I don't think we needed this `NERVESHUB:END` marker since we're not planning to do anything with the output. That was previously used to try and group script output for return to the caller way back when

```suggestion
      text = ~s/#{connecting_code}\n\r/
      topic = "device:console:#{device.id}"

      Endpoint.broadcast_from!(self(), topic, "dn", %{"data" => text})
```

PR #2401, opened 2025-11-25
Update dep confuse to v0.2.1

lawik:


lawik:
Fixes a parsing issue with escaped quotes that Nate hit.

PR #2402, opened 2025-12-02
Add devices online/offline to org and product views

lawik:
This is primarily to help people doing development or with few devices but a couple of orgs set up. It helps for figuring out where your active devices actually are. I've had this issue a hilarious number of times.

The loading is deferred so it won't stop someone from navigating if it becomes slow. There are only two queries made so it shouldn't be particularly impactful in performance and the information isn't live-updated.

It might also indicate when something is really wrong or weird I suppose. But mostly it helps indicate which orgs and products are actually in use. Especially since the development seed produces a bunch of them.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -59,7 +59,7 @@ defmodule NervesHub.Devices.Connections do
         }
   def get_connection_status_by_products(product_ids) when is_list(product_ids) do
     q =
-      from(d in DeviceConnection)
+      DeviceConnection
       |> join(:inner, [d], p in assoc(d, :product))
       |> select([d, p], [p.id, count(d.id)])
       |> where([_, p], p.id in ^product_ids)


lawik:
Images of this impressive UI.

Screenshot 2025-12-02 at 22 42 06
Screenshot 2025-12-02 at 22 41 54

lib/nerves_hub/devices/connections.ex
```diff
@@ -20,6 +20,76 @@ defmodule NervesHub.Devices.Connections do
...
 
+  @doc """
+  Get device connection information for Orgs.
+  """
+  @spec get_connection_status_by_orgs(org_ids :: [non_neg_integer()]) :: %{
+          non_neg_integer() => %{online: non_neg_integer(), offline: non_neg_integer()}
+        }
+  def get_connection_status_by_orgs(org_ids) when is_list(org_ids) do
+    q =
+      from(d in DeviceConnection)
```
joshk:
```suggestion
      DeviceConnection
```

lib/nerves_hub_web/live/orgs/index.ex
```diff
@@ -18,22 +20,35 @@ defmodule NervesHubWeb.Live.Orgs.Index do
...
+      socket
+      |> assign(:page_title, "Organizations")
+      |> assign(:show_all_pinned?, false)
+      |> assign(:device_info, %{})
+      |> assign(:pinned_devices, Devices.get_pinned_devices(user.id))
+      |> assign(:device_statuses, statuses)
+      |> assign(:device_limit, @pinned_devices_limit)
+      |> subscribe()
+
+    if connected?(socket), do: send(self(), :load_extras)
```
joshk, line 33: if you create a `maybe_load_extras` function we can just add it to the set of pipe calls

lib/nerves_hub/devices/connections.ex
```diff
@@ -20,6 +20,76 @@ defmodule NervesHub.Devices.Connections do
...
+
+  @doc """
+  Get device connection information for Products.
+  """
+  @spec get_connection_status_by_products(product_ids :: [non_neg_integer()]) :: %{
+          non_neg_integer() => %{online: non_neg_integer(), offline: non_neg_integer()}
+        }
+  def get_connection_status_by_products(product_ids) when is_list(product_ids) do
+    q =
+      from(d in DeviceConnection)
```
joshk:
```suggestion
      DeviceConnection
```

lib/nerves_hub_web/live/orgs/index.ex
lawik: I find that mildly less obvious. But I see your point :)

PR #2404, opened 2025-12-05
Track release history

kybishop:
Screenshot 2025-12-05 at 1 17 44 PM

### DIFF
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -200,11 +200,12 @@ defmodule NervesHub.ManagedDeployments do
   @spec update_deployment_group(DeploymentGroup.t(), map, User.t()) ::
           {:ok, DeploymentGroup.t()} | {:error, Changeset.t()}
   def update_deployment_group(deployment_group, params, user) do
+    deployment_group = Repo.preload(deployment_group, :firmware)
+
     result =
       Repo.transact(fn ->
         changeset =
           deployment_group
-          |> Repo.preload(:firmware)
           |> DeploymentGroup.update_changeset(params)
 
         create_deployment_release? =
@@ -212,7 +213,6 @@ defmodule NervesHub.ManagedDeployments do
             Map.has_key?(changeset.changes, :archive_id)
 
         with {:ok, deployment_group} <- Repo.update(changeset),
-             deployment_group = Repo.preload(deployment_group, [:firmware], force: true),
              :ok <- create_audit_logs!(deployment_group, changeset),
              {:ok, _deployment_group} <-
                if(create_deployment_release?,


priv/repo/migrations/20251206184125_add_cascade_delete_to_deployment_releases.exs
```diff
@@ -0,0 +1,23 @@
+defmodule NervesHub.Repo.Migrations.AddCascadeDeleteToDeploymentReleases do
```
kybishop, line 1: Foreign key validation does not need to be separated out to a second migration because there are no deployment_release records until after this PR.

lib/nerves_hub/managed_deployments.ex
```diff
@@ -193,27 +195,31 @@ defmodule NervesHub.ManagedDeployments do
...
-            {deployment_group, changeset}
-
-          {:error, changeset} ->
-            Repo.rollback(changeset)
+        create_deployment_release? =
+          Map.has_key?(changeset.changes, :firmware_id) or
+            Map.has_key?(changeset.changes, :archive_id)
+
+        with {:ok, deployment_group} <- Repo.update(changeset),
+             deployment_group = Repo.preload(deployment_group, [:firmware], force: true),
```
nshoes: Is this force needed? 🤔

test/nerves_hub/managed_deployments_test.exs
```diff
@@ -134,12 +137,42 @@ defmodule NervesHub.ManagedDeploymentsTest do
...
+      {:ok, deployment_group} = ManagedDeployments.create_deployment_group(params, product, user)
+
+      releases = ManagedDeployments.list_deployment_releases(deployment_group)
+
+      assert length(releases) == 1
+      [release] = releases
+
+      assert release.deployment_group_id == deployment_group.id
+      assert release.firmware_id == firmware.id
+      assert release.archive_id == nil
```
nshoes:
```suggestion
      refute release.archive_id
```

nshoes (approved):
This LGTM. I'll publish and run through a few tests in our QA.

lib/nerves_hub_web/components/deployment_group_page/release_history.ex
```diff
@@ -9,9 +9,61 @@ defmodule NervesHubWeb.Components.DeploymentGroupPage.ReleaseHistory do
...
+                  <th class="text-left px-4 py-3 text-sm font-medium text-zinc-400">Archive</th>
+                  <th class="text-left px-4 py-3 text-sm font-medium text-zinc-400">Released By</th>
+                </tr>
+              </thead>
+              <tbody>
+                <tr :for={release <- @releases} class="border-b border-zinc-800 hover:bg-zinc-800/50">
+                  <td class="px-4 py-3 text-sm text-zinc-300">
+                    <div class="flex flex-col">
+                      <span>{Calendar.strftime(release.inserted_at, "%B %d, %Y")}</span>
+                      <span class="text-xs text-zinc-500">{Calendar.strftime(release.inserted_at, "%I:%M %p")}</span>
```
nshoes:
```suggestion
                      <span class="text-xs text-zinc-500">{Calendar.strftime(release.inserted_at, "%I:%M %p")} UTC</span>
```

lib/nerves_hub/managed_deployments.ex
kybishop: it was not, and was actually inefficient in the first place since we preloaded earlier but never saved the result. Good catch, fixed.

PR #2410, opened 2025-12-09
Fix expiring inflight updates on device connect

nshoes:
The assumption on connect is that if a device joins with `currently_downloading_uuid`, we assume the device is updating and don't clear inflight updates. However, a device will always join with this param, so the pattern match would never allow inflight updates to be cleared on connect.

### DIFF
--- a/lib/nerves_hub/device_link.ex
+++ b/lib/nerves_hub/device_link.ex
@@ -130,7 +130,9 @@ defmodule NervesHub.DeviceLink do
     :ok
   end
 
-  defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => uuid}) when not is_nil(uuid), do: :ok
+  defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => uuid})
+       when is_binary(uuid) and byte_size(uuid) > 0,
+       do: :ok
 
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/device_link.ex
```diff
@@ -130,7 +130,7 @@ defmodule NervesHub.DeviceLink do
     :ok
   end
 
-  defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => _uuid}), do: :ok
+  defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => uuid}) when not is_nil(uuid), do: :ok
```
jjcarstens:
```suggestion
  defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => uuid}) when is_binary(uuid) and byte_size(uuid) == 36, do: :ok
```
You could also maybe flip this so it is the expected success case? (Could also be `byte_size > 0`)

```elixir
iex(1)> byte_size "df1b7b8f-66f8-5f1a-09a6-1c84d0b69126"
36
```

nshoes: O! I like this. I used `byte_size > 0`.

Issue #2409, opened 2025-12-09
Offline device count for product is wrong


nshoes:
I have seven devices, but it's showing 12 offline  🤔 

<img width="513" height="114" alt="Image" src="https://github.com/user-attachments/assets/51223206-7c79-4f76-9ac2-4a0d41ccc3de" />

<img width="839" height="378" alt="Image" src="https://github.com/user-attachments/assets/ef1fdfc0-6abb-4a80-a6f3-c13055a7b198" />

I think we're not accounting for stale connections? Or maybe only a distinct is needed.

PR #2412, opened 2025-12-10
Only take into account a device's latest connection when processing online and offline counts for products and orgs

nshoes:
Solves #2409

Looking at the entire `device_connections` table when collecting these counts produced incorrect results. We should only be taking a device's `latest_connection` into account, ignoring the old or stale records that will be eventually cleaned up by workers.

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -27,22 +27,12 @@ defmodule NervesHub.Devices.Connections do
           non_neg_integer() => %{online: non_neg_integer(), offline: non_neg_integer()}
         }
   def get_connection_status_by_orgs(org_ids) when is_list(org_ids) do
-    q =
-      DeviceConnection
-      |> join(:inner, [d], p in assoc(d, :product))
-      |> select([d, p], [p.org_id, count(d.id)])
-      |> where([_, p], p.org_id in ^org_ids)
-      |> group_by([_, p], p.org_id)
-
-    online =
-      q
-      |> where([d], d.status == :connected)
-      |> Repo.all()
-
-    offline =
-      q
-      |> where([d], d.status != :connected)
-      |> Repo.all()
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/devices/connections.ex
```diff
@@ -82,6 +62,19 @@ defmodule NervesHub.Devices.Connections do
...
+    |> join(:inner, [dc], p in assoc(dc, :product))
+    |> join(:inner, [dc], dev in assoc(dc, :device), on: dev.latest_connection_id == dc.id)
+  end
+
+  defp connection_status_counts(query) do
+    %{
+      online: where(query, [dc], dc.status == :connected) |> Repo.all(),
+      offline: where(query, [dc], dc.status != :connected) |> Repo.all()
+    }
+  end
```
kybishop, line 76: I totally get the desire to DRY these up, but I feel hesitant about them since they introduce mystery elements into the query:
- For the base query, we now have these inner joins which we may or may not be interacting with either/both on.
- For the connection status we're depending on the first `from` being the device connection table, which I'll give is generally true, but we're separating the things we actually want in the end from the code that is calling it

I won't stop the PR, but I feel at least somewhat strongly that queries should be pretty expressive right where they stand, with little left to mystery that needs to be pieced together from various functions wherever possible.

nshoes: I wrote a paragraph but it doesn't matter 😆 These two queries are tied together, and changes should be made to both. It's the same data from slightly different angles. There are no plans to extend or change this, and if there is, it needs to be re-worked anyways.

I don't necessarily disagree with you, but the needs you layed out represent something more akin to a query module with a base query, and this is not the time nor the place. Though I'd like to get there soon for NervesHub 😄

nshoes: Guess I still ended up writing a paragraph 🤣

PR #2413, opened 2025-12-11
Introduce a priority queue

kybishop:
This PR introduces priority queues to deployment groups.

The problem:

1. Often times you'll have a giant batch of hardware sitting in a warehouse that is already preloaded with some old firmware version.
2. Installers then go into the field and set up these devices, but they need to update right away since they could have been sitting in the warehouse for ages, with some known to be buggy version, or a version that might not even work with current APIs the hardware might need to reach out to.
3. Installers then have to wait around while the device updates, often a very expensive wait at that.

Currently the update could be clogged up behind the standard update pool, perhaps behind a bunch of devices taking up all the slots due to slow connections. This problem only compounds if installers are installing multiple devices. What we really want is for those devices to be updated independent of the in-use devices, and without having to go through the process of making an entire duplicated deployment group for them.

Solution:

Set a threshold version for firmware, under which any device instead enters a priority queue that has its own size configurations. This allows newly installed devices to immediately get updates without having to increase the size of the normal queue, which would inadvertently affect devices that are already out on the field.

Screenshot 2025-12-10 at 11 18 21 PM

### DIFF
--- a/lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
+++ b/lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
@@ -142,7 +142,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
     :telemetry.execute([:nerves_hub, :deployments, :trigger_update], %{count: 1})
 
     # Process priority queue first, if enabled
-    skipped_updates = maybe_do_priority_update(deployment_group)
+    skipped_priority_updates = maybe_do_priority_update(deployment_group)
 
     # Process normal queue
     slots = available_slots(deployment_group)
@@ -151,7 +151,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
       available = Devices.available_for_update(deployment_group, slots)
       updated_count = schedule_devices!(available, deployment_group, false)
 
-      if length(available) != updated_count or skipped_updates > 0 do
+      if length(available) != updated_count or skipped_priority_updates > 0 do
         # rerun the deployment check since some devices were skipped
         send(self(), :trigger)
       end
--- a/test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
+++ b/test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
@@ -5,6 +5,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorTest do
 
   alias NervesHub.Devices
   alias NervesHub.Devices.Connections
+  alias NervesHub.Devices.InflightUpdate
   alias NervesHub.Firmwares
   alias NervesHub.Firmwares.FirmwareDelta
   alias NervesHub.Firmwares.UpdateTool.Fwup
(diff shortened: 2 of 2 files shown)


priv/repo/migrations/20251208171533_add_priority_queue_to_deployment_groups.exs
```diff
@@ -0,0 +1,15 @@
...
+    alter table(:deployments) do
+      add :priority_queue_enabled, :boolean, default: false, null: false
+      add :priority_queue_concurrent_updates, :integer, default: 5, null: false
+      add :priority_queue_firmware_version_threshold, :string
+    end
+
+    alter table(:inflight_updates) do
+      add :priority_queue, :boolean, default: false, null: false
+    end
+  end
```
kybishop, line 14: According to fly.io, this is safe in PG >= 11 (which itself is 7 years old)

https://fly.io/phoenix-files/migration-recipes/#adding-a-column-with-a-default-value

lib/nerves_hub/managed_deployments/deployment_group.ex
```diff
@@ -208,6 +217,26 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
     |> validate_required([:status])
   end
 
+  defp validate_priority_queue_version_threshold(changeset) do
+    threshold = get_field(changeset, :priority_queue_firmware_version_threshold)
+
+    if threshold && threshold != "" do
+      case Version.parse(threshold) do
```
nshoes, line 233: Since the column is being used in the query as `<= version`, we need to make sure a user can't submit a string with the operator prefix. We can't catch this with only `Version.parse/1`. Maybe there's another `Version` function we could use or string split and grab the second element?

lib/nerves_hub/devices.ex
```diff
@@ -740,6 +740,58 @@ defmodule NervesHub.Devices do
...
+  Get devices eligible for priority queue updates.
+
+  Similar to `available_for_update/2` but filters devices whose firmware version
+  is less than or equal to the priority_queue_firmware_version_threshold.
+  """
+  @spec available_for_priority_update(DeploymentGroup.t(), non_neg_integer()) :: [Device.t()]
+  def available_for_priority_update(deployment_group, count) do
+    threshold = deployment_group.priority_queue_firmware_version_threshold
+
+    if threshold && threshold != "" do
```
nshoes: Can we handle this in a guard?

lib/nerves_hub/devices.ex
```diff
@@ -740,6 +740,58 @@ defmodule NervesHub.Devices do
     |> Repo.all()
   end
 
+  @doc """
+  Get devices eligible for priority queue updates.
+
+  Similar to `available_for_update/2` but filters devices whose firmware version
+  is less than or equal to the priority_queue_firmware_version_threshold.
+  """
+  @spec available_for_priority_update(DeploymentGroup.t(), non_neg_integer()) :: [Device.t()]
```
nshoes: I'd argue this is a good time for a base query. I can see us extending this query  in the future, and with this feature specifically, we're only adding a single clause.

lib/nerves_hub/devices.ex
```diff
@@ -740,6 +740,58 @@ defmodule NervesHub.Devices do
...
+      |> where([d], not is_nil(d.firmware_metadata))
+      |> where([d, firmware: f], fragment("(? #>> '{\"uuid\"}') != ?", d.firmware_metadata, f.uuid))
+      |> where([inflight_update: ifu], is_nil(ifu))
+      |> where([d], is_nil(d.updates_blocked_until) or d.updates_blocked_until < ^now)
+      |> where([d], fragment("semver_match(? #>> '{\"version\"}', ?)", d.firmware_metadata, ^"<= #{threshold}"))
+      |> then(fn query ->
+        case deployment_group.queue_management do
+          :FIFO ->
+            order_by(query, [d, latest_connection: lc],
+              desc: d.priority_updates,
```
nshoes: Could we remove the usage of this column in this PR? It's not used and the priority queue was the original direction of priority updates. It's junk.

lib/nerves_hub/devices.ex
```diff
@@ -740,6 +740,58 @@ defmodule NervesHub.Devices do
...
+      |> join(:left, [d], ifu in InflightUpdate, on: d.id == ifu.device_id, as: :inflight_update)
+      |> where(deployment_id: ^deployment_group.id)
+      |> where(updates_enabled: true)
+      |> where([d], d.firmware_validation_status in [:validated, :unknown])
+      |> where([latest_connection: lc], lc.status == :connected)
+      |> where([d], not is_nil(d.firmware_metadata))
+      |> where([d, firmware: f], fragment("(? #>> '{\"uuid\"}') != ?", d.firmware_metadata, f.uuid))
+      |> where([inflight_update: ifu], is_nil(ifu))
+      |> where([d], is_nil(d.updates_blocked_until) or d.updates_blocked_until < ^now)
+      |> where([d], fragment("semver_match(? #>> '{\"version\"}', ?)", d.firmware_metadata, ^"<= #{threshold}"))
```
nshoes: I like the usage of `semver_match` here. It's not used anymore, and the SQL was pulled straight from StackOverflow. I'm confident that it was ditched because of feature parity between that function and the `Version` module. Since we validate with `Version`, we'd have to ensure that edge cases are handled with the PG function.

However, in this body of work, we are explicitly comparing a version to a static operator, which is a lot less surface area to cover feature-wise; and it already works 😄 

I'd feel comfortable moving forward with this if we unit tested `semver_match`. I'd like to depend on this function more moving forward because of the overhead of pulling records into memory and checking with `Version`, which we do now. This would start us on that path.

lib/nerves_hub/managed_deployments/deployment_group.ex
kybishop: Do you have an example of a version string you're worried about? It seems like it catches bad versions pretty well:

```
iex(4)> [Version.parse("1.2.3"), Version.parse("<= 1.2.3"), Version.parse("1.2.3.4.5")]
[{:ok, %Version{major: 1, minor: 2, patch: 3}}, :error, :error]
```

lib/nerves_hub/devices.ex
kybishop: Done

lib/nerves_hub/devices.ex
kybishop: FWIW I'm anti base queries if they start hiding joins and things like that, but I'm for having a shared function here that contains the entire query. Done.

lib/nerves_hub/devices.ex
kybishop: Yeah definitely much better to do the work on the db. Will add unit tests

lib/nerves_hub/managed_deployments/deployment_group.ex
nshoes: Oh! I misunderstoof the implementation. That's exactly what I wanted 🙃

lib/nerves_hub/devices.ex
kybishop: Tests added. `semver_match` now goes pretty hard and I believe is fully semver 2.0 compliant (including pre-release support)

lib/nerves_hub/devices.ex
kybishop: Done. The column itself will be removed in a follow-up migration.

lib/nerves_hub/devices.ex
nshoes: Oh man, that makes me so excited. Looks good. Love the tests.

test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
```diff
@@ -591,4 +591,616 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorTest do
...
+
+      device = Devices.update_deployment_group(device, deployment_group)
+
+      {:ok, conn} = Connections.device_connecting(device, product.id)
+      :ok = Connections.device_connected(device, conn.id)
+
+      assert Devices.available_for_priority_update(deployment_group, 10) == []
+    end
+
+    test "priority queue empty when threshold is empty string", %{
```
nshoes: Do we need to handle this case? Why don't we ensure that the column is either `nil` or validated by `Version`?

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
```diff
@@ -140,19 +140,56 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
...
   defp do_trigger_update(deployment_group) do
     :telemetry.execute([:nerves_hub, :deployments, :trigger_update], %{count: 1})
+
+    # Process priority queue first, if enabled
+    priority_updated =
+      if deployment_group.priority_queue_enabled do
+        do_trigger_priority_update(deployment_group)
+      else
+        0
+      end
```
nshoes: Tiny nit: can we move this to a guard?

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
```diff
@@ -140,19 +140,56 @@ defmodule NervesHub.ManagedDeployments.Distributed.Orchestrator do
...
+    # Process normal queue
     slots = available_slots(deployment_group)
 
     if slots > 0 do
       available = Devices.available_for_update(deployment_group, slots)
-      updated_count = schedule_devices!(available, deployment_group)
+      updated_count = schedule_devices!(available, deployment_group, false)
 
-      if length(available) != updated_count do
+      if length(available) != updated_count or priority_updated > 0 do
```
nshoes: I'm assuming the thought was "get back to the priority queue as fast as possible" here, but wanted to check. I can't think of any other reason to trigger manually when we know priority devices were updated.

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
kybishop: done

test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
kybishop: good call, fixed

lib/nerves_hub/managed_deployments/distributed/orchestrator.ex
kybishop: this was a mistake on my part, fixed it to be like the other check, which assures we keep processing if we didn't use all available slots

nshoes: Thanks for fixing! 

Last thing: could we get tests for:
1.) Updates are sent when there are available priority slots
2.) Updates are not sent when there aren't priority slots
3.) Updates are re-queued when there are available slots

There's probably already a test that covers the first one.

kybishop: done! https://github.com/nerves-hub/nerves_hub_web/pull/2413/commits/a756ef406fd2336c93ded30b092980d341fd2b49

PR #2414, opened 2025-12-15
Short-circuit firmware delta generation for things that will always fail

lawik:
A firmware without support for deltas will never succeed, so lets not retry it 5 times.

This will lead to processing being much faster in the orchestrator as well for those cases.


nshoes (approved):
Sweet! Thanks.

PR #2416, opened 2025-12-15
Device status was not properly capturing fwup errors

lawik:
At some point the "FWUP error : ${inspect(reason)}" device status from NervesHubLink stopped matching what we do on the serve side.

This restores that functionality.

It can be tested by checking out the original code and running the new test. The one that matches what NervesHubLink does fails.


lawik:
@nshoes I bet we could have this notify the orchestrator. Do you know top of mind  what we could broadcast?

lawik:
Yeah. The lack of some kind of message format is a bit of a pain. We'd need to support the existing formats for a decent bit of time even if we add a new more enforceable thing but it would be good to have a module or two that covers that the messages.

@nshoes On notifying the orchestrator. I think clearing the inflight is entirely right. And maybe we gain some consistency in performance if the orchestrator always works on its (currently 3 second?) ticks. But we'd gain some speed (0-3 seconds) if it was notified that inflight was cleared. I think I may be over-indexed on trying to make the orchestrator be more lively because I've had some bad run-ins recently. Which prompted this investigation for one thing :D 

We can hold off on the notifying.

lawik:
Yeah, I figured we had some broadcasts for the orchestrator. So this would make sense to me. But either way.

Well I had one goof that was invalid data in NervesCloud from a client that had somehow made invalid tags or tags that were once valid but then couldn't be unpacked into a struct and that made my orchestrators not start and that made updates work really, really poorly for obvious reasons.

Then I had the fwup error not being caught causing 500 updates to get stuck for 1 hour. 

I did just (today) have a device with an inflight that I think stuck around for a long time or something. Not sure. I think it ended up queued behind itself or something. It wasn't updating and it let me hit the Skip the queue, which I did. It was the only device in the product :D 

This good to merge for you?

nshoes:
The only thing left to decide is whether a FWUP error from the device counts as an update attempt. I think it should, but am open to hear why we shouldn't if you feel strongly.

lawik:
I mean we should definitely stop trying if not immediately then after a few attempts. Typically they'll need manual remediation.

I've had fwup errors when the device hasn't received a new signing key yet. That's recoverable but requires the device to reconnect. Many fwup errors would be due to some missing state. But presumably it could be due to some intermittent issue as well so probably fine to do  few retries and then flag the device.

nshoes:
Good, those all sound like reasons to count a real attempt. To implement, we can pass `socket.assigns.update_started?` to `DeviceLink` when the channel receives `"status_update"`, then `Devices.update_attempted/1` if the update hasn't started.

PR #2418, opened 2025-12-16
Improve device details layout for slimmer display

lawik:
This prevents long Version names from wrapping outside of the box.

It prevents long information from overlapping or covering the device identifier or compressing the online/offline indicator dot.


lawik:
Before:
Screenshot_20251216_090609

After:
Screenshot_20251216_090130

nshoes (approved):

PR #2421, opened 2025-12-23
Product api keys

lawik:
Adding Product API keys to start helping most users move away from User API keys for automation.

Currently need to test if there are a bunch of API end points that will fall over looking for a `user` assign. But overall I think this is sound. If you see things that can be tightened up, feel free. I'm not precious and some of the tests were certainly vibed and then reviewed which we'll see how well I did I suppose.


lawik:
Is draft because I need to go through the API controllers and ensure they work with product keys.

PR #2426, opened 2026-01-14
Make deployment group devices counter clickable

kybishop:
This allows you to click the devices counter from the deployment group's show page, which will bring you to the Device list page with the deployment group selected.

With the mouse hovered over said button:
Screenshot 2026-01-14 at 11 14 30 AM


nshoes (changes requested):
Solid! Could we get a test? :)

PR #2427, opened 2026-01-14
Quokka module directives

kybishop:
This brings the last bit of Quokka goodness to Nerves. It auto-sorts module directives and automatically aliases modules.

### DIFF
--- a/lib/nerves_hub_web.ex
+++ b/lib/nerves_hub_web.ex
@@ -393,6 +393,7 @@ defmodule NervesHubWeb do
     quote do
       use PhoenixHTMLHelpers
       use Gettext, backend: NervesHubWeb.Gettext
+
       import NervesHubWeb.ErrorHelpers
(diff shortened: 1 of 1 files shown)


lib/nerves_hub/extensions/geo.ex
```diff
@@ -51,7 +52,7 @@ defmodule NervesHub.Extensions.Geo do
     Connections.merge_update_metadata(socket.assigns.reference_id, %{location: location})
 
     _ =
-      Phoenix.Channel.Server.broadcast(
+      Server.broadcast(
```
nshoes: Can you adjust all of these instances to `alias Phoenix.Channel.Server as ChannelServer`? I am generally against `as` usage, but `Server` is so generic I think it's more painful 😆

kybishop: done!

nshoes (approved):

PR #2435, opened 2026-01-22
When matching deployment groups with a device, prioritize matching tags

nshoes:
When attemptig to find matching deployment groups for a device, we do not take into consideration how many tags match. This PR addresses the issue by checking tag count when sorting. 

I originally went down the path of leveraging our Postgres function `semver_match` to query and sort, but there's a bit more work to do there before we can.

### DIFF
--- a/test/nerves_hub/managed_deployments_test.exs
+++ b/test/nerves_hub/managed_deployments_test.exs
@@ -737,7 +737,7 @@ defmodule NervesHub.ManagedDeploymentsTest do
 
     test "older deployment groups are prioritized if deployment groups have the same firmware and there are no matching tags",
          state do
-      %{org: org, product: product, firmware: firmware} = state
+      %{org: org, product: product, firmware: firmware, deployment_group: %{id: oldest_deployment_group_id}} = state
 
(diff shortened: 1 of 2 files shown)


test/nerves_hub/managed_deployments_test.exs
```diff
@@ -710,6 +710,81 @@ defmodule NervesHub.ManagedDeploymentsTest do
                  false
                ])
     end
+
+    test "matchings tags are prioritized if deployment groups have the same firmware", state do
```
lawik, line 714: The improvement looks great. I'm confused by this test name :D 

It seems like the deployment groups don't have the same firmware? And the test name indicates they should? I'm guessing this test case is doing a different check now than it did originally?

nshoes: They do have the same firmware, I'm passing `firmware` as the first argument to each call of `Fixtures.deployment_group_fixture/2`.

lawik: Ah, the "version" key is the version constraint. That makes sense.

PR #2437, opened 2026-01-23
Storing Device Network Interface 

nshoes:
This PR introduces storing a device's network interface type. The end goal is to target releases within a deployment group by interface type, e.g. "I want to update all ethernet devices to v1.3.4". This will be handled next by Ky and his rollout logic work.

The lifecycle is simple:
- Device reports network interface
  - If NH holds the same value, we skip
  - If it's different, we set it

Actual interface names are translated to a friendly version (`eth0 -> ethernet`), and garbage values will result in `"unknown"`, giving us an indicator that something is wrong. If a device never reports it will stay as `nil`. Through application code, you cannot explicitly set this field to `nil`.

On the Link side of things, we'll report if there's mismatch between what the network interface is at time of device connect, and what it is when we go to download firmware. The whole point of this feature is to be confident on what interface the device uses, but currently we have no data to tell us devices _never_ switch. So, we log if there is a mismatch and move on. We'll be monitoring this in our systems after firmware is updated with the Link changes to see if it's a fully realized issue or not.

I've also added a test helper for asserting against SQL queries in our websocket tests. Take a look at the module doc for `EctoTelemetryHandler` for more info.

### DIFF
--- a/lib/nerves_hub/devices/device.ex
+++ b/lib/nerves_hub/devices/device.ex
@@ -145,6 +145,7 @@ defmodule NervesHub.Devices.Device do
     |> validate_required([:network_interface])
   end
 
+  @spec humanized_network_interface_name(String.t()) :: :wifi | :ethernet | :cellular | :unknown
   def humanized_network_interface_name(interface) do
     cond do
       String.starts_with?(interface, "wlan") -> :wifi


lib/nerves_hub/devices/device.ex
```diff
@@ -87,6 +87,8 @@ defmodule NervesHub.Devices.Device do
     # To be removed in a migration in the next release
     # field(:priority_updates, :boolean, default: false)
 
+    field(:network_interface, :string)
```
kybishop: This could be an ecto enum to guarantee its value is one that we expect:


```suggestion
    field(:network_interface, Ecto.Enum, values: [:wifi, :ethernet, :cellular, :unknown])
```

Also has the benefit of giving us back a known set of atoms to work with :)

kybishop (approved):
Just one thought about making the field an Ecto enum, but otherwise lgtm :tada:

lib/nerves_hub/devices/device.ex
nshoes: This was my original direction, but I found it annoying to convert the incoming value to an atom or converting what we have to a string to check if it has changed. 

I can't decide what's more annoying though, the former or having to manually keep track of the values. Thoughts?

test/nerves_hub_web/channels/websocket_test.exs
```diff
@@ -1327,17 +1440,7 @@ defmodule NervesHubWeb.WebsocketTest do
         })
         |> ManagedDeployments.update_deployment_group(%{is_active: true}, user)
 
-      device =
```
nshoes, line 1330: Diff is weird here because many of these tests are similar in shape.

lib/nerves_hub/devices/device.ex
kybishop: Shouldn't `Changeset.cast` take care of that conversion for you? We use Ecto.Enum a lot internally and don't have to do any string to atom conversions

kybishop: Poking around a bit I think you'd just need to change the output values of `friendly_network_interface_name` to atoms.

Also bikeshed but "friendly" sounds hilarious to me. Thoughts on `humanized` or `human_readable` instead?

nshoes: Yeah I hear you, I'll convert and report back.

> Also bikeshed but "friendly" sounds hilarious to me

Ha! You're right, `humanized` is the more appropriate term here. I'll change.

nshoes: Donezo!

PR #2438, opened 2026-01-30
Add option for streaming console output through the device code API call

lawik:
If streaming is switched on it will subscribe to the results from the console and shove them down the HTTP as chunked responses. This allows a crude but workable console experience on the CLI side. Or just a way to run RingLogger.attach remotely and watch the device.

Required for https://github.com/nerves-hub/nerves_hub_cli/pull/310

Happy to take input on this.

A better implementation would be to have a channel for the **not device** websocket to stream the console and then just send the commands as requests. I think this is an okay step that we could drop later though.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -29,7 +29,8 @@ config :flop, repo: NervesHub.Repo
 config :mime, :types, %{
   "application/pem" => ["pem"],
   "application/crt" => ["crt"],
-  "application/fwup" => ["fw"]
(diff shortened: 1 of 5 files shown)


lawik:
Nothing urgent really. I rabbit-holed a bit on solving it with what was available. But I do want to expand the scope of that web socket so can do that when I find some time :)

This PR can be adapted to that so will keep it around :)

PR #2441, opened 2026-02-05
Rollout options 🎉 🚀 🚢

kybishop:
This PR adds support for specifying rollout options. These constrain releases to the selected network interfaces and/or specified tags:

Screenshot 2026-02-04 at 8 26 56 PM
Screenshot 2026-02-04 at 8 37 17 PM

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -31,16 +31,17 @@
   ],
   "words": [
     "castore",
-    "chartjs",
     "certifi",
+    "chartjs",
     "circuitboard",
     "citext",
     "clickhouse",
     "clipcopy",
     "comeonin",
     "conzole",
-    "extralight",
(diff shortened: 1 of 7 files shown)


lib/nerves_hub/devices/device.ex
```diff
@@ -35,7 +35,8 @@ defmodule NervesHub.Devices.Device do
     :firmware_validation_status,
     :firmware_auto_revert_detected,
     :first_seen_at,
-    :custom_location_coordinates
+    :custom_location_coordinates,
+    :network_interface
```
nshoes: I'm assuming this isn't needed?

kybishop: Indeed, artifact from rebasing onto main

PR #2442, opened 2026-02-05
Clean up after creating firmware deltas

kybishop:
Turns out we were never actually cleaning up after ourselves when creating deltas.

This PR fixes things so that we do said cleanup, and adds real E2E tests to firmware delta generation such that we can have real (or at least improved) confidence things are working.

Switched to Req for fetching the firmware so that we can get its built in test stubbing to make dx a little more enjoyable as well.

### DIFF
--- a/test/nerves_hub/workers/firmware_delta_builder_test.exs
+++ b/test/nerves_hub/workers/firmware_delta_builder_test.exs
@@ -31,9 +31,6 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilderTest do
           [_full, uuid] ->
             # Find firmware by UUID and serve the local file
             case Repo.get_by(Firmwares.Firmware, uuid: uuid) do
-              nil ->
-                Plug.Conn.send_resp(conn, 404, Jason.encode!(%{error: "not found"}))
-
               firmware ->
                 # Get local path from upload_metadata
                 local_path =
@@ -46,9 +43,6 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilderTest do
                 |> Plug.Conn.put_resp_content_type("application/octet-stream")
                 |> Plug.Conn.send_resp(200, body)
             end
-
-          nil ->
-            Plug.Conn.send_resp(conn, 404, Jason.encode!(%{error: "invalid path"}))
         end
       end)
 
@@ -131,9 +125,7 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilderTest do
         }
       }
 
-      # Worker returns {:error, ...} on failure
-      result = FirmwareDeltaBuilder.perform(job)
-      assert match?({:error, _}, result) or match?(:error, result)
+      assert {:error, :test_failure} = FirmwareDeltaBuilder.perform(job)
 
       delta = Repo.reload(delta)
       assert delta.status == :failed
@@ -170,8 +162,7 @@ defmodule NervesHub.Workers.FirmwareDeltaBuilderTest do
       }
 
       # Worker returns {:error, ...} on failure
-      result = FirmwareDeltaBuilder.perform(job)
-      assert match?({:error, _}, result) or match?(:error, result)
+      assert {:error, :test_failure} = FirmwareDeltaBuilder.perform(job)
 
       delta = Repo.reload(delta)
       assert delta.status == :processing, "Should remain processing for retry"


lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -70,8 +70,17 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
...
+        Logger.warning(
+          "Firmware delta creation failed with exception: #{inspect(e)}, stacktrace: #{inspect(__STACKTRACE__)}",
+          source_url: source_url,
+          target_url: target_url
+        )
+
+        {:error, :download_or_delta_failed}
     after
-      File.rmdir(work_dir)
+      File.rm_rf(work_dir)
```
kybishop, line 83: rmdir only applies to empty directories 😬

nshoes: Ah! Much better. When I took a pass at deltas there wasn an error with logic (only in tests) that resulted in `work_dir` being cwd... it was very confusing why I couldn't rerun tests and why suddenly git wasn't working and OMG I just blasted away the entire project locally 😅🫡

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -355,13 +364,14 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
...
   defp dl!(url, filepath) do
-    {:ok, :saved_to_file} =
-      :httpc.request(
-        :get,
-        {url |> to_charlist(), []},
-        [],
-        stream: filepath |> to_charlist()
-      )
+    # Download and write to file
+    response = Req.get!(url, Application.get_env(:nerves_hub, :firmware_download_options, []))
```
nshoes, line 368: Will this raise on a non-200 response?

test/nerves_hub/workers/firmware_delta_builder_test.exs
```diff
@@ -1,61 +1,180 @@
...
+        attempt: 5,
+        args: %{
+          "source_id" => source_firmware.id,
+          "target_id" => target_firmware.id
+        }
+      }
+
+      # Worker returns {:error, ...} on failure
+      result = FirmwareDeltaBuilder.perform(job)
+      assert match?({:error, _}, result) or match?(:error, result)
```
nshoes: Why can this be either?

test/nerves_hub/workers/firmware_delta_builder_test.exs
```diff
@@ -1,61 +1,180 @@
...
+      # Set up Req.Test to intercept HTTP requests
+      # (plug is configured in config/test.exs)
+      Req.Test.stub(NervesHub, fn conn ->
+        # Extract UUID from the request path
+        # Expected format: http://localhost:1234/firmware/{org_id}/{uuid}.fw
+        case Regex.run(~r|/firmware/\d+/([a-f0-9-]+)\.fw$|, conn.request_path) do
+          [_full, uuid] ->
+            # Find firmware by UUID and serve the local file
+            case Repo.get_by(Firmwares.Firmware, uuid: uuid) do
+              nil ->
```
nshoes: Why are we handling this and the other `nil` case on line 50?

lib/nerves_hub/firmwares/update_tool/fwup.ex
kybishop: It will not, only on network failures

test/nerves_hub/workers/firmware_delta_builder_test.exs
kybishop: it cannot, that was me not looking closely enough at some AI aid. Fixed.

test/nerves_hub/workers/firmware_delta_builder_test.exs
kybishop: good catch, removed.

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -355,13 +364,14 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
...
+
+    if response.status != 200 do
+      raise "HTTP download failed with status #{response.status}"
+    end
+
+    File.write!(filepath, response.body)
```
joshk: @nshoes @kybishop one downside to this change is that the the download will be fetched into memory instead of streamed to a file. This will impact memory bloat.

PR #2445, opened 2026-02-06
Add connection events and console channel to Device Event socket

lawik:
Anyone listening to the sidechannel socket for a device now gets connectivity messages.

Added a channel that can be joined to receive console messages.

This is mainly in service of improving nerves_hub_cli. Allowing it to both watch and interact with the console.


joshk:
I think I've added most of this functionality in a few recently merged PRs.

How about we close this for now and open a few issues detailing any missing bits, which I can then take a look at.

PR #2446, opened 2026-02-06
Remove unused files in delta generation

lawik:
A `complete` or other task can contain file resources never used in the upgrade. This change will detect these unnecessary files and remove them from the archive. This should be completely equivalent as previously but potentially save a fair bit if you have some files jingling around in your firmware.

It also means that if you do things carefully you can make very lean firmware by dropping files that aren't needed for a particular upgrade but might be needed again later but then you better be gating your versions very well. So not sure I'd recommend that.


test/nerves_hub/firmwares/update_tool_test.exs
```diff
@@ -652,6 +652,51 @@ defmodule NervesHub.Firmwares.UpdateToolTest do
...
+      File.write!(data_path_1, data_1)
+
+      data_path_2 = Path.join(dir, "data-2")
+      data_2 = for _ <- 1..100, into: data_1, do: Ecto.UUID.generate()
+      File.write!(data_path_2, data_2)
+
+      fw_a = build_fw!(Path.join(dir, "a.fw"), fwup_conf_path, data_path_1)
+      fw_b = build_fw!(Path.join(dir, "b.fw"), fwup_conf_path, data_path_2)
+      %{size: source_size} = File.stat!(fw_a)
+      %{size: target_size} = File.stat!(fw_b)
```
kybishop: Do we do this for a particular reason? Seems like these vars are unused.

lawik: Ah, usually we check the sizes but that's covered by another test. We can drop those :)

PR #2453, opened 2026-02-15
Deployment Release improvements

joshk:
First off, I'm sorry for the big PR.

This PR started as a focus on UI improvements for Deployment Releases, but grew into also fixing some delta race conditions, complexity with how delta statues were managed, and tests that didn't exists.

The core improvements include:
- The UI for creating a release is now in a dialog
- Added badges based on the deployment group status to indicate if deltas are being generated or have errored
- Removed the need to listen to Firmware PubSub events in the Orchestrator and instead only listen for Deployment Group events. These status updates are tied into Firmware delta processing which, upon finishing, tells all Deployment Groups using the target firmware to recalculate their status.
- And finally, I've added a way to regenerate deltas if deltas have been deleted for a deployment group

_Please be aware that I've had to update FWUP to a pinned commit as this fixes an issue with downloading zlib when building a static library of FWUP._

https://github.com/user-attachments/assets/02c33e27-71f4-4925-a7aa-d5c21e450972

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -311,7 +311,7 @@ defmodule NervesHub.Firmwares do
 
     Repo.transact(fn ->
       with {:ok, firmware} <- Repo.delete(delta),
-           :ok = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(delta.target_id),
+           {:ok, _} = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(delta.target_id),
            {:ok, _} <- Oban.insert(delete_delta_job) do
         {:ok, firmware}
       end
@@ -504,7 +504,7 @@ defmodule NervesHub.Firmwares do
     with {:ok, firmware_delta} <- Repo.update(changeset),
          :ok <- firmware_upload_config().upload_file(delta_file_metadata.filepath, upload_metadata),
          {:ok, _firmware_delta} <- notify_firmware_delta_target({:ok, firmware_delta}) do
-      :ok = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(firmware_delta.target_id)
+      {:ok, _} = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(firmware_delta.target_id)
 
       Logger.info("Created firmware delta successfully.")
 
@@ -585,9 +585,10 @@ defmodule NervesHub.Firmwares do
     |> notify_firmware_delta_target()
     |> case do
       {:ok, firmware_delta} ->
-        if recalculate_deployment_statuses do
-          :ok = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(target_id)
-        end
+        _ =
+          if recalculate_deployment_statuses do
+            {:ok, _} = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(target_id)
+          end
 
         {:ok, firmware_delta}
 
@@ -605,7 +606,7 @@ defmodule NervesHub.Firmwares do
       |> Repo.update()
       |> notify_firmware_delta_target()
 
-    :ok = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(firmware_delta.target_id)
+    {:ok, _} = ManagedDeployments.recalculate_deployment_group_status_by_firmware_id(firmware_delta.target_id)
 
     {:ok, firmware_delta}
   end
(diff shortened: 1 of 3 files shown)


assets/ui-rework/app.js
```diff
@@ -129,3 +129,10 @@ window.addEventListener("ca:new:jitp", () => {
     document.getElementById("jitp_form").classList.add("hidden")
   }
 })
+
+window.addEventListener("phx:close-modal", ({ detail }) => {
+  let modal = document.getElementById(detail.id)
+  if (modal) {
+    liveSocket.execJS(modal, modal.getAttribute("data-cancel"))
```
nshoes, line 136: What is this doing?

lib/nerves_hub/firmwares.ex
```diff
@@ -395,16 +397,21 @@ defmodule NervesHub.Firmwares do
...
+      else
+        where(query, [fd], fd.status in ^status)
+      end
+    end)
+    |> Repo.one()
     |> case do
-      [] -> {:error, :not_found}
-      [firmware_delta] -> {:ok, firmware_delta}
+      nil -> {:error, :not_found}
+      firmware_delta -> {:ok, firmware_delta}
```
nshoes, line 446: FWIW I checked our prod and there are no deltas with the same target + source id.

lib/nerves_hub/firmwares.ex
```diff
@@ -519,51 +531,79 @@ defmodule NervesHub.Firmwares do
...
           {:ok, FirmwareDelta.t()} | {:error, Ecto.Changeset.t()}
-  def start_firmware_delta(source_id, target_id) do
-    FirmwareDelta.start_changeset(source_id, target_id)
+  def start_firmware_delta(source_id, target_id, recalculate_deployment_statuses \\ true) do
+    _ =
+      get_firmware_delta_by_source_and_target(source_id, target_id, [:failed, :timed_out])
+      |> case do
+        {:ok, firmware_delta} -> {:ok, _} = Repo.delete(firmware_delta)
+        {:error, _} -> :ok
+      end
```
nshoes, line 580: This seems overly defensive, did you run into a case where this happens?

test/nerves_hub_web/live/deployment_groups/show/activity_tab_test.exs
```diff
@@ -0,0 +1,23 @@
+defmodule NervesHubWeb.Live.NewUI.DeploymentGroups.Show.ActivityTabTest do
+  use NervesHubWeb.ConnCase.Browser, async: false
+
+  # alias NervesHub.AuditLogs
+  # alias NervesHub.ManagedDeployments
+  # alias NervesHub.ManagedDeployments.DeploymentGroup
+  # alias NervesHub.Repo
```
nshoes:
```suggestion
```

test/nerves_hub_web/live/deployment_groups/new_test.exs
```diff
@@ -76,4 +89,62 @@ defmodule NervesHubWeb.Live.DeploymentGroups.NewTest do
...
+    |> submit()
+    |> assert_path("/org/#{org.name}/#{product.name}/deployment_groups/Canaries")
+
+    deployment_group = Repo.one!(from(d in DeploymentGroup, where: d.name == "Canaries"))
+    assert deployment_group.delta_updatable
+  end
+
+  test "disable delta updates when creating a deployment group", %{conn: conn, org: org, product: product} do
+    conn
+    |> assert_has("input[name='deployment_group[delta_updatable]']", checked: true)
```
nshoes, line 109: I thought this was unchecking the default, which is a checked box enabling deltas, but that doesn't make sense. What is this doing? This reads as "I'm asserting it's checked".

assets/ui-rework/app.js
joshk: This allows us to execute different LiveView JS actions when a modal is closed, by embedding them in the `data-cancel` attribute.

lib/nerves_hub/firmwares.ex
joshk: Yep, there can't be as there is a constraint.

lib/nerves_hub/firmwares.ex
joshk: It was easier for me to delete the firmware record than to update the current one as I would need to update the `inserted_at` field which is used for calculating time outs.

test/nerves_hub_web/live/deployment_groups/new_test.exs
joshk: this is confirming that the deltas are enabled before disabling them. deltas are enabled by default for new deployment groups.

lib/nerves_hub/firmwares.ex
nshoes: Got it, that makes sense

test/nerves_hub_web/live/deployment_groups/new_test.exs
nshoes: I must have forgotten my glasses; I missed the `uncheck/2` call on line 111.

PR #2457, opened 2026-02-16
Add product banner image

lawik:
It can be nice to be able to tell products apart at a glance and also feel like NervesHub shows what you do. Not just a sleek UI. So this adds a product banner image. Currently used on the front page. Need to do some thinking about integration in the overall product view.

Screenshot_20260216_155958

And upload/replace/remove is in settings:

Screenshot_20260216_160014

I used Claude a fair bit in this. I have looked at the code. But do let me know if I let anything slip past.


joshk:
I really love this idea, and think it would look great when we have a 'summary'/'insights' page. We could also use it as a background image on the org/product selector.

As for the implementation, I saw a few things I would like to tweak. I need to finish off my PR first, so I'd recommend not merging this in just yet but I will get to this soon.

lawik:
Yeah, bunch of places it could be used to give you a sense of what you are looking at I think. This is a start and it doesn't have to go in right away :)

lawik:
And then...

Screenshot_20260218_083323

lawik:
Note to self: Add some aesthetically pleasing default varieties for people that don't have a product photo yet.

lawik:
Updated screenshots:

Screenshot_20260521_150421

Screenshot_20260521_150439
Screenshot_20260521_152722
/>
Screenshot_20260521_150520
Screenshot_20260521_152704

lawik:
Adding some pre-packaged banners for your selecting pleasure.

lawik:
Pre-baked banners offered as well as the custom support. All CC-0 images.

Screenshot_20260522_120111

lawik:
I think this is in pretty good shape. The form above shows up in onboarding, new product and product settings.

PR #2462, opened 2026-02-19
Firmware connected to Releases can't be deleted

joshk:
Adding `no_assoc_constraint` stops this from blowing up, allowing us to show a nice message instead.

There is a question to be answered (outside of this PR) if we should loosen this restriction, but for now showing a message is better than errors that are going to Sentry and not being shown to users.


lawik (approved):
Love it!

nshoes (approved):

PR #2463, opened 2026-02-19
Update the various Firmware lists and drop downs when new Firmware is available

joshk:
This uses the magic of PubSub to live update firmware lists.

I've also added a new flash type of 'notice' for messaging which isn't a success or failure, but just a little 'notice' of sorts.

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -451,9 +451,7 @@ defmodule NervesHub.Firmwares do
           :ok
           | {:error, Ecto.Changeset.t() | :no_delta_support_in_firmware}
   def generate_firmware_delta(firmware_delta, source_firmware, target_firmware) do
-    Logger.info(
-      "Creating firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}."
-    )
+    Logger.info("Creating firmware delta between #{source_firmware.uuid} and #{target_firmware.uuid}.")
 
     {:ok, source_url} = firmware_upload_config().download_file(source_firmware)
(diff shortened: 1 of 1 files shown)


test/nerves_hub/managed_deployments/deployment_group_test.exs
```diff
@@ -1,5 +1,5 @@
 defmodule NervesHub.ManagedDeployments.DeploymentGroupTest do
-  use NervesHub.DataCase, async: true
+  use NervesHub.DataCase, async: false
```
lawik: Does this fix the flaky test we've seen or something? Seems separate from the PR overall.

joshk: I ran into some flaky tests related to Oban usage, but with Oban no longer separated, this might be able to be changed back. That might be a different PR though.

lawik: Fair enough :)

lib/nerves_hub_web/live/deployment_groups/show.ex
```diff
@@ -38,7 +38,12 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Show do
     updating_count = Devices.updating_count(deployment_group)
     releases = ManagedDeployments.list_deployment_releases(deployment_group)
 
-    :ok = socket.endpoint.subscribe("deployment:#{deployment_group.id}:internal")
+    Logger.metadata(user_id: user.id, product_id: product.id, deployment_group_id: deployment_group.id)
+
+    if connected?(socket) do
```
nshoes, line 43: This is to ensure that we don't subscribe until the second mount is called, correct?

PR #2469, opened 2026-02-21
bump 1.19.5 and OTP 28

pxp9:
The project already compiles to 1.19 and OTP 28 

- bumping the tool versions
- mix format
- fix tests
- ignore some dialyzer warns 
Previous PR was blocked bc false positives in dialyzer , but this PR takes the approach of ignoring them and when the false positives are fixed , we can remove them by put it them in the dialyzer ignore file
- Update Dockerfile [image](https://hub.docker.com/layers/hexpm/elixir/1.19.5-erlang-28.3.1-ubuntu-noble-20260210.1/images/sha256-a61e0c725c61755b6a07d40eb3ac02be9066502286300e9e80ed8a40ab813a1c)

this PR in favor of this one

https://github.com/nerves-hub/nerves_hub_web/pull/2375

so we should close that one

also previous PR is stale (4 moths ago)

### DIFF
--- a/lib/nerves_hub/accounts/remove_account.ex
+++ b/lib/nerves_hub/accounts/remove_account.ex
@@ -22,8 +22,6 @@ defmodule NervesHub.Accounts.RemoveAccount do
   alias NervesHub.Repo
   alias NervesHub.Scripts.Script
 
-  @dialyzer {:nowarn_function, remove_account: 1}
-
   def remove_account(user_id) do
     Multi.new()
     |> Multi.run(:user_id, fn _, _ -> {:ok, user_id} end)
--- a/.dialyzer_ignore.exs
+++ b/.dialyzer_ignore.exs
@@ -5,5 +5,7 @@
   # Ecto.Multi.t() opaque type false positives — MapSet.new() returns a concrete
   # %MapSet{map: %{}} that Dialyzer can't unify with the opaque MapSet.internal(_) type,
   # causing spurious call_without_opaque warnings on Ecto.Multi pipe chains.
-  {"lib/nerves_hub/accounts.ex", :call_without_opaque}
+  {"lib/nerves_hub/accounts.ex", :call_without_opaque},
+  {"lib/nerves_hub/accounts/remove_account.ex", :call_without_opaque},
+  {"lib/nerves_hub/devices.ex", :call_without_opaque}
 ]
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -41,9 +41,6 @@ defmodule NervesHub.Devices do
 
   require Logger
 
-  @dialyzer {:nowarn_function,
-             delete_device: 1, update_attempted: 2, move: 3, update_device_with_audit: 4, save_device_health: 1}
-
   def get_device(device_id) when is_integer(device_id) do
     Repo.get(Device, device_id)
   end


test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
```diff
@@ -171,12 +169,9 @@ defmodule NervesHubWeb.Live.NewUi.Devices.HealthTabTest do
...
 
-    organized_metrics =
-      ~s([{"y":#{value},"x":"#{now}"}])
-      |> html_escape()
-      |> safe_to_string()
-
-    assert html =~ ~s(data-metrics="#{organized_metrics}")
+    assert html =~ ~s(data-metrics=")
+    assert html =~ ~s(&quot;y&quot;:#{value})
+    assert html =~ ~s(&quot;x&quot;:&quot;#{now}&quot;)
```
joshk: Did the change in Elixir and Erlang versions cause this to fail?

test/nerves_hub_web/live/devices/health_test.exs
```diff
@@ -170,12 +168,9 @@ defmodule NervesHubWeb.Devices.HealthTest do
...
 
-    organized_metrics =
-      ~s([{"y":#{value},"x":"#{now}"}])
-      |> html_escape()
-      |> safe_to_string()
-
-    assert html =~ ~s(data-metrics="#{organized_metrics}")
+    assert html =~ ~s(data-metrics=")
+    assert html =~ ~s(&quot;y&quot;:#{value})
+    assert html =~ ~s(&quot;x&quot;:&quot;#{now}&quot;)
```
joshk: Same question as my other other comment.

test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
pxp9: the order of iteration changes I did a small proof with Nix packages

```bash
$ nix-shell -p beam27Packages.erlang beam27Packages.elixir_1_18 --run "elixir --eval \"IO.puts(System.version()); IO.puts(:erlang.system_info(:otp_release)); m = %{y: 1, x: 2}; IO.inspect(:maps.to_list(m))\"" 2>/dev/null
1.18.4
27
[y: 1, x: 2]
```

```bash
nix-shell -p beam28Packages.erlang beam28Packages.elixir_1_19 --run "elixir --eval \"IO.puts(System.version()); IO.puts(:erlang.system_info(:otp_release)); m = %{y: 1, x: 2}; IO.inspect(:maps.to_list(m))\"" 2>/dev/null
1.19.5
28
[x: 2, y: 1]
```

so thats why i change the test to only change the order of the result

test/nerves_hub_web/live/devices/health_test.exs
pxp9: same for this

test/nerves_hub_web/live/new_ui/devices/health_tab_test.exs
pxp9: this proof does the following , runs the following bash script in a shell with the corresponding Erlang and Elixir

```bash
elixir --eval "IO.puts(System.version()); IO.puts(:erlang.system_info(:otp_release)); 
m = %{y: 1, x: 2}; 
IO.inspect(:maps.to_list(m))"
```

lib/nerves_hub/accounts/remove_account.ex
```diff
@@ -22,6 +22,8 @@ defmodule NervesHub.Accounts.RemoveAccount do
   alias NervesHub.Repo
   alias NervesHub.Scripts.Script
 
+  @dialyzer {:nowarn_function, remove_account: 1}
```
joshk: Are we able to move these to the dialyzer ignore file as well?

lib/nerves_hub/managed_deployments/distributed/orchestrator_registration.ex
```diff
@@ -92,6 +92,10 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorRegistration do
     :ok
   end
 
+  defp report_errors({:error, errors, :rollback}) do
```
joshk, line 95: Thanks for catching this!

lib/nerves_hub/accounts/remove_account.ex
pxp9: yup , sorry , i forgot these ones

pxp9: [Fix last dialyzer warns](https://github.com/nerves-hub/nerves_hub_web/pull/2469/commits/7e9af5fec9574b3b3d5d7175f7ef0606006f4e86)

lib/nerves_hub/managed_deployments/distributed/orchestrator_registration.ex
pxp9: is it correct behavior like this ? just Send it to Sentry or we should do something else like Add that is from a Rollback ?

pxp9:
> @joshk this was your note from the previous PR btw:
> 
> > Dialyzer issues are a blocker.
> > This is being looked into by Elixir core.

Indeed , this is why i took the approach of ignoring it , bc from Core they did not fix it since October , probably they are fixing other stuff , but this allow us to continue to a new version and not get stucked.

lib/nerves_hub/managed_deployments/distributed/orchestrator_registration.ex
joshk: we don't enable `:rollback`, so this should never happen. Adding this is fine as it is.

PR #2475, opened 2026-02-22
Add default org and product when user registers

lawik:
This is not necessarily the end game of user onboarding but it should get users started a lot faster.

It pre-seeds an org, a product and gives the product a shared secret.

They also get redirected to the devices list and that now has an empty state containing the pre-seeded credentials ready to go into a config.

This should remove a lot of friction from the process of getting started.

Open to config or disabling this if it seems disruptive when onboarding people to existing orgs.


lawik:
Dangit, forgot the screenshot. Will fix.

lib/nerves_hub/accounts.ex
```diff
@@ -79,6 +79,28 @@ defmodule NervesHub.Accounts do
...
+  Returns `{org_name, product_name}`.
+  """
+  @spec generate_onboarding_names(String.t()) :: {String.t(), String.t()}
+  def generate_onboarding_names(user_name) do
+    slug = slugify_name(user_name)
+    {"#{slug}-team", "#{slug}ifier"}
+  end
+
+  defp slugify_name(name) do
+    case Slug.slugify(name) do
```
nshoes, line 98: Do we really need an external library for this? Can't we just lowercase username and move on?

Related, I was really surprised to see that Phoenix doesn't have a builtin util for slugs.

joshk: I think using a lib is the easiest as we don't have to create our own dictionary.

lib/nerves_hub_web/live/devices/index.html.heex
```diff
@@ -111,80 +111,131 @@
...
+            <p class="text-sm text-zinc-400 mb-6">Get started by adding <code class="text-zinc-300">nerves_hub_link</code> to your Nerves project.</p>
+            <% active_secret = Enum.find(@product.shared_secret_auths, &is_nil(&1.deactivated_at)) %>
+            <%= if active_secret do %>
+              <div class="bg-zinc-900 border border-zinc-700 rounded-lg p-4">
+                <h3 class="text-sm font-semibold text-zinc-300 mb-3">1. Add the dependency to your <code>mix.exs</code></h3>
+                <pre class="bg-zinc-950 rounded p-3 text-sm text-green-400 font-mono mb-4 overflow-x-auto"><code>&#123;:nerves_hub_link, "~&gt; 2.0"&#125;</code></pre>
+
+                <h3 class="text-sm font-semibold text-zinc-300 mb-3">2. Add this to your <code>config/config.exs</code></h3>
+                <pre class="bg-zinc-950 rounded p-3 text-sm text-green-400 font-mono overflow-x-auto"><code>config :nerves_hub_link,
+  host: "{URI.parse(NervesHubWeb.Endpoint.url()).host}",
```
joshk: we might need to configure this slightly as the endpoint for us will be `devices.nervescloud.com` but this will show `manage.nervescloud.com`. I think we might need to take this from the application config.

lib/nerves_hub/accounts.ex
lawik: I understand your hesitation.

What's the slug for Örjan Rådgärd and how do you get to it? The goal of the slug is to be as basic ascii as possible? There are ways to do it. I'd rather not relearn all the sharp corners.

My expectation is that we will use this more in the future as well as things acquire names and maybe we don't want them all to be underscored and dashed to be nice identifiers when we display them?

Anyway, fully get the  question.

lib/nerves_hub_web/live/devices/index.html.heex
lawik: Good catch

lib/nerves_hub/accounts.ex
nshoes: Heard and agreed!

PR #2490, opened 2026-02-24
Move to using a `current_release` association on `DeploymentGroup`

joshk:
This is a first step towards removing `firmware_id` and `archive_id` from `DeploymentGroup`, and instead use `Firmware` and `Archive` from the most current release.

I've also given each `DeploymentRelease` a number field, which is scoped to the `DeploymentGroup`, as well as a backfill migration.


lib/nerves_hub/managed_deployments.ex
```diff
@@ -259,6 +274,18 @@ defmodule NervesHub.ManagedDeployments do
     end
   end
 
+  def load_current_release(deployment_group, force \\ false) do
```
nshoes: Could we change the `force` option here and in `load_current_release/2` to a keyword list? The readibility is lacking;  seeing booleans passed in and no option name making the reader have to go to the definition to understand is frustrating.

lib/nerves_hub_web/plugs/api/device.ex
```diff
@@ -3,7 +3,7 @@ defmodule NervesHubWeb.API.Plugs.Device do
 
   alias NervesHub.Devices
 
-  @preloads [:product, :latest_connection, :firmware]
```
nshoes, line 6: Why can't/shouldn't we do this anymore? I think I'm missing something. I think of a device's firmware as independent from the current release -> firmware relationship.

test/nerves_hub_web/live/deployment_groups/show/releases_tab_test.exs
```diff
@@ -20,19 +17,6 @@ defmodule NervesHubWeb.Live.NewUI.DeploymentGroups.Show.ReleasesTabTest do
     %{context | conn: conn}
   end
 
-  test "shows a nice message when there are no releases", %{
```
nshoes, line 23: Why remove?? The view hasn't been touched in this PR.

lib/nerves_hub/managed_deployments.ex
joshk: agree, will get onto that

test/nerves_hub_web/live/deployment_groups/show/releases_tab_test.exs
joshk: Creating a deployment creates a release, so this situation is impossible to reach.

lib/nerves_hub_web/plugs/api/device.ex
joshk: This is a bit deceptive as `:firmware` actually preloaded the firmware of the deployment group associated with the Device (https://github.com/nerves-hub/nerves_hub_web/pull/2490/changes#diff-03454c312d5c891b1ae7d916a1eb171cdf023ca00f9aa5155ea5633f7c5cbe81L283)

nshoes: Oop! That'll do it. I assumed it looked at it's metadata.

test/nerves_hub_web/live/deployment_groups/show/releases_tab_test.exs
nshoes: Ah! Missed that.

lib/nerves_hub/managed_deployments.ex
joshk: Fixed!

PR #2493, opened 2026-02-25
Move Account Tokens to `/account`, and to the new UI

joshk:
I think this is the final page.

Next up, ripping out all the old JS and CSS from the old UI...


lib/nerves_hub_web/live/account_tokens.ex
jjcarstens, line 1: Is this file still needed? Looks like it has been kept around empty

test/nerves_hub_web/live/account_test.exs
```diff
@@ -123,4 +123,70 @@ defmodule NervesHubWeb.Live.AccountTest do
...
     end
   end
+
+  describe "generate new access token" do
+    test "accepts a token note", %{conn: conn, user: user} do
+      conn
+      |> visit("/account")
+      |> assert_has("div", text: "No access tokens have been created")
+      # PhoenixTest doesn't work with JS commands in the phx-click
+      # |> click_button("Generate Access Token")
```
jjcarstens, line 133: Is this still needed since the button is clicked below?

lib/nerves_hub_web/live/account_tokens.ex
joshk: Oops, thanks for pointing that out

test/nerves_hub_web/live/account_test.exs
joshk: Well, that commented out button shows the modal popup, which PhoenixTest can't do.

jjcarstens: O I get it. I thought "generate" and "create" Access token buttons were the same. But there is multi steps

joshk: Its a little confusing.

PR #2494, opened 2026-02-25
Remove JS and CSS used by the old UI

joshk:
Less JS, less NPM libs, and an updated Node.js

(based upon #2493)


joshk:
I have another commit incoming to fix the 400, 404, and 500 pages

PR #2497, opened 2026-02-25
Consolidate and remove duplicated UI tests

joshk:
When the new UI was made the default, and only, UI to use, some 'new ui' tests still existed and needed moving or removing.


test/nerves_hub_web/live/devices/show/settings_tab_test.exs
```diff
@@ -1,5 +1,5 @@
-defmodule NervesHubWeb.Live.Devices.SettingsTest do
-  use NervesHubWeb.ConnCase.Browser, async: true
+defmodule NervesHubWeb.NewUi.Devices.SettingsTabTest do
```
lawik: We want the namespace NewUi? I am not quite following :)

joshk: Arg, nope, missed that

PR #2498, opened 2026-02-26
Allow for the unique Device identifier to be scoped to Orgs

### LINKED ISSUE #2495: Gotta allow the same device identifier in different products/orgs
This is such a common issue when people do a test with a device and later want to set it up properly. They have to track it down and add it and the error isn't obvious so they just think it doesn't work.

I also hit it all the time due to demo:ing the same device and forgetting to remove it from the previous account.

joshk:
This change allows for Device identifiers to be reused across Orgs.

Device identifiers are unique across the entire platform, and this will remain the default.

To allow for device identifiers to be reused, the env var `PLATFORM_UNIQUE_DEVICE_IDENTIFIERS` must be set to any other value than `true`,  `TRUE`, or `1` (default is `true`).

Additionally, you will need to run the following SQL in your DB 
  `DROP INDEX devices_identifier_index;` 
  and then `CREATE INDEX devices_identifier_org_id_index ON devices(identifier, org_id);`

There are two issues which require discussion:

- [x] the API routes `/api/devices/:identifier`
- [x] the `DeviceEventsStreamChannel` websocket channel

**The API routes `/api/devices/:identifier`**

The 'short' API routes are only enabled when PLATFORM_UNIQUE_DEVICE_IDENTIFIERS is enabled. I've also hidden them from the Swagger UI when PLATFORM_UNIQUE_DEVICE_IDENTIFIERS is false.

**The `DeviceEventsStreamChannel` websocket channel**

I've kept support for `device:[device_identifier]` if `PLATFORM_UNIQUE_DEVICE_IDENTIFIERS` is `true`, otherwise users will need to use `org:[org_name]:device:[device_identifier]`


jjcarstens:
I haven't gotten very far, but before it gets too deep, we should not be considering using an enumerating database ID in any public facing situation. That has been a common security anti-pattern as it greatly increases the likelihood of guess based attacks and interceptions. We are actively moving away from them in many situations.

We also should be careful in error responses. While saying "this api request has multiple returns for the identifier" is very helpful, it may provide too much information from a security stance. 

One could argue that the "internal" topics should not apply, but they are still discoverable to an extent through browser and web tooling quite easily, so I would be hesitant to include enumerating IDs there too.

joshk:
@nshoes @jjcarstens I've addressed the two issues/points I called out in my PR description, and have updated the description to include how it now works.

Essentially, no change for self-hosted installs, only NervesCloud (and NervesCloud users) are affected.

lawik (approved):

lib/nerves_hub_web/channels/device_events_stream_channel.ex
```diff
@@ -3,12 +3,14 @@ defmodule NervesHubWeb.DeviceEventsStreamChannel do
   Phoenix Channel for external services to subscribe to device updates.
   Currently only supports firmware update progress.
 
-  External services can join device-specific channels using the topic pattern "device:\#{device_identifier}"
+  External services can join device-specific channels using the topic pattern "device:\#{device_id}"
```
nshoes: We still support `device:\#{device_identifier}`. Can you add that `device:\#{device_id}` is supported as well through the`:platform_unique_device_identifiers` setting?

nshoes (approved):
Sweet, this looks good to me now. Thanks for the putting in the work after the feedback.

lib/nerves_hub_web/channels/device_events_stream_channel.ex
joshk: Ah, yes, oops, I meant to change this, one sec.

PR #2500, opened 2026-03-01
Reuse the tmp firmware file created by LiveView.Upload

joshk:
This is a small optimization which slightly reduces file system operations.


joshk:
> LGTM!
> 
> Could we get some documentation around the two GenServer calls? Do we have any tests covering this codepath?

Added some code comments to explain this a bit more. 

I don't think specific tests are needed as `Plug.Upload` is well tested.

Also, I just realized that I combined two PRs into one by accident, one for the upload file reuse, and one for a better error message. Oops.

PR #2501, opened 2026-03-02
Update pacakage.lock from recent asset changes

nshoes:


nshoes:
@joshk nope! This is it.

PR #2502, opened 2026-03-02
Use more tmp dirs in tests, and use Breifly for delta generation

joshk:
Briefly allows for temp directory generation and cleanup. This is a better approach than how we were using Plug.Upload

All tests are using ExUnits tmp directory generation.

And I think the flaky delta tests are gone 🤞


test/nerves_hub_web/live/deployment_groups/tabs/releases_tab_test.exs
lawik, line 1: Was this intentionally removed? :)

joshk: yep

https://github.com/nerves-hub/nerves_hub_web/pull/2502/changes/d7b9fcf97ab2281cf29a3d340d99fdc208581d30

PR #2506, opened 2026-03-03
Add `join_ref`s to Channel messages missing a `join_ref`

joshk:
Because of https://github.com/phoenixframework/phoenix/commit/c73bbfcfeafce3d42cb270a7ecb41a4ca39ae393, and that we don't control when devices update their Slipstream so that they work with this Phoenix change (https://github.com/CuatroElixir/slipstream/pull/84), this commit fills in the `join_ref` when its missing.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -30,6 +30,7 @@
     "test"
   ],
   "words": [
+    "bandaid",
     "castore",
     "certifi",
     "chartjs",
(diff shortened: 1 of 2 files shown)


lawik (approved):
This seems relatively cursed. Good job.

Maybe add a comment or three about why we do this. Affected versions and so on. So we have some idea of if we can ever remove it :D

PR #2510, opened 2026-03-03
Update all Node deps, which also resolves a security warning with semver

joshk:


nshoes:
@joshk how deep did you investigate UI-things working?

PR #2513, opened 2026-03-04
Generate a delta when adding a device

kybishop:
Currently we don't generate a delta when a device is newly added to a deployment group. This PR fixes that.

### DIFF
--- a/test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
+++ b/test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
@@ -3,6 +3,7 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorTest do
   use Mimic
   use AssertEventually, timeout: 500, interval: 50
 
+  alias NervesHub.Accounts.Scope
   alias NervesHub.Devices
   alias NervesHub.Devices.Connections
   alias NervesHub.Devices.InflightUpdate
@@ -320,6 +321,63 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorTest do
     :sys.get_state(pid)
   end
 
+  test "the orchestrator 'trigger's when a group (bulk) of devices are added to a deployment group", %{
+    deployment_group: deployment_group,
+    org_key: org_key,
+    product: product,
+    device: device1,
+    device2: device2,
+    user: user,
+    tmp_dir: tmp_dir
+  } do
+    # An ugly set of expectations
+    # `Devices.available_for_update` should be called:
+    # - once upon Orchestrator startup
+    # - and once when a bulk number of devices are added to the deployment group
+    # - and no more times after that
+    Devices
+    |> expect(:available_for_update, 1, fn _deployment_group, _slots ->
+      []
+    end)
+    |> expect(:available_for_update, 1, fn _deployment_group, _slots ->
+      []
+    end)
+    |> reject(:available_for_update, 2)
+
+    firmware = Fixtures.firmware_fixture(org_key, product, %{dir: tmp_dir})
+
+    {:ok, deployment_group} =
+      ManagedDeployments.update_deployment_group(
+        deployment_group,
+        %{
+          concurrent_updates: 2,
+          firmware_id: firmware.id
+        },
+        user
+      )
+
+    deployment_group_topic = "orchestrator:deployment:#{deployment_group.id}"
+    Phoenix.PubSub.subscribe(NervesHub.PubSub, deployment_group_topic)
+
+    {:ok, pid} =
+      start_supervised(%{
+        id: "Orchestrator##{deployment_group.id}",
+        start: {Orchestrator, :start_link, [deployment_group, false]},
+        restart: :temporary
+      })
+
+    allow(Devices, self(), pid)
+
+    Devices.move_many_to_deployment_group(Scope.for_user(user), [device1.id, device2.id], deployment_group)
+
+    assert_receive %Broadcast{topic: ^deployment_group_topic, event: "bulk-devices-added"}, 500
+
+    Mimic.reject(&Devices.available_for_update/2)
+
+    # allows for db connections to finish and close
+    :sys.get_state(pid)
+  end
+
   test "the orchestrator is 'triggered' when a device is reenabled to accept updates", %{
     user: user,
     deployment_group: deployment_group,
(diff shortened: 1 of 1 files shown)


priv/repo/migrations/20260306153000_add_firmware_delta_query_indexes.exs
```diff
@@ -0,0 +1,24 @@
+defmodule NervesHub.Repo.Migrations.AddFirmwareDeltaQueryIndexes do
+  use Ecto.Migration
+  @disable_ddl_transaction true
+
+  def change do
+    # Critical: Index for firmware UUID extraction in delta availability query
+    # This enables efficient lookup of source firmware by UUID from device metadata
+    create_if_not_exists index(
```
joshk: is this index needed now that we have reworked the query?

priv/repo/migrations/20260306153000_add_firmware_delta_query_indexes.exs
```diff
@@ -0,0 +1,24 @@
...
+      "devices",
+      ["(firmware_metadata #>> '{uuid}')"],
+      name: :devices_firmware_uuid_idx,
+      concurrently: true
+    )
+
+    # Important: Composite index including status for firmware_delta filtering
+    # Helps with queries that check if deltas exist and are completed
+    create_if_not_exists index(
```
joshk: we already have an index for source id and target id, which is unique, so this is likely redundant

lib/nerves_hub/devices.ex
```diff
@@ -973,19 +1011,28 @@ defmodule NervesHub.Devices do
...
-      |> Ecto.Changeset.change()
-      |> Ecto.Changeset.put_change(:deployment_id, deployment_group.id)
-      |> Repo.update!()
+    deployment_group = ManagedDeployments.load_current_release(deployment_group, force: true)
 
-    DeviceEvents.deployment_assigned(device)
-
-    # let the orchestrator know that a device has been added to the deployment group
-    DeploymentOrchestratorEvents.device_added(device)
+    Repo.transact(fn ->
```
joshk: we don't need to use a transaction when there is only one sql statement in there

lib/nerves_hub/devices.ex
```diff
@@ -790,42 +790,80 @@ defmodule NervesHub.Devices do
...
-    |> where(deployment_id: ^deployment_group.id)
-    |> where(updates_enabled: true)
-    |> where([d], d.firmware_validation_status in [:validated, :unknown])
-    |> where([latest_connection: lc], lc.status == :connected)
-    |> where([d], not is_nil(d.firmware_metadata))
-    |> where(
-      [d],
-      fragment("(? #>> '{\"uuid\"}') != ?", d.firmware_metadata, ^deployment_group.current_release.firmware.uuid)
+    |> ManagedDeployments.join_current_release()
+    |> join(
```
joshk: can you please break out these joins into their own functions, like what has been done with `maybe_filter_by_network_interfaces`

joshk (changes requested):
Thanks for making the changes.

I still think we shouldn't be using the orchestrator for this, based on the race condition I explain earlier. 

We would use `update_deployment_group` to trigger any delta jobs, which would then be picked up by the query used by the orchestrator to find devices to update.

priv/repo/migrations/20260306153000_add_firmware_delta_query_indexes.exs
kybishop: indeed, removed.

priv/repo/migrations/20260306153000_add_firmware_delta_query_indexes.exs
kybishop: it is not. Removed.

lib/nerves_hub/devices.ex
kybishop: fixed

lib/nerves_hub/devices.ex
kybishop: I made this change to avoid holding anything up, but I would be remiss not to mention that I feel pretty strongly against hiding joins in separate functions.

My thinking being that one should avoid hiding side effects that later calls depend on. I think it has a short term benefit of making the query more readable, but a long term negative in that it hides things about the query from those editing it. More than anything, I feel like it hurts the locality of behavior to hide away the join since you have to go hunt for it if you need it from some later conditional in the query, and if you want to change the join itself you're now hunting for anywhere that uses it or the function it is contained in (plus has the chance of others calling our new helper function, making the spaghetti/coupling even tighter.

lib/nerves_hub/devices.ex
kybishop: actually brought back now that we're generating the deltas from the update function again 😅

lib/nerves_hub/devices.ex
```diff
@@ -1382,16 +1446,37 @@ defmodule NervesHub.Devices do
...
+            [d],
+            d.firmware_metadata["architecture"] ==
+              ^deployment_group.current_release.firmware.architecture
+          )
+          |> Repo.update_all([set: [deployment_id: deployment_id]], timeout: to_timeout(minute: 2))
+
+        # Queue delta generation for any new device firmware combinations immediately
+        # after the device updates within the same transaction
+        if deployment_group.delta_updatable do
+          _ = ManagedDeployments.trigger_delta_generation_for_deployment_group(deployment_group)
```
joshk:
```suggestion
        _ = if deployment_group.delta_updatable do
          ManagedDeployments.trigger_delta_generation_for_deployment_group(deployment_group)
```

test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
```diff
@@ -327,63 +326,6 @@ defmodule NervesHub.ManagedDeployments.Distributed.OrchestratorTest do
     :sys.get_state(pid)
   end
 
-  test "the orchestrator 'trigger's when a group (bulk) of devices are added to a deployment group", %{
```
joshk, line 330: why were these tests removed?

kybishop: Ah, I noticed they were in a really funky spot since they're really just testing a side effect of `Devices.update_deployment_group` (that it triggers `DeploymentOrchestratorEvents.device_added(device)` and likewise for the bulk actions). The orchestrator is a receiver of those events, but has nothing to do with how they're triggered. Since `Devices.update_deployment_group` is called from many top-level entry points, I moved the tests to be unit tests instead of integration style on controller, UI, etc.

The move came about because I was initially adding the delta expectations to copies of these tests in this file before being like "wait a minute... the tests shouldn't live here".

kybishop: One of the moves: https://github.com/nerves-hub/nerves_hub_web/pull/2513/changes#diff-bd272c7a2e0914c069300eecc49b5fd05bcf9a67bd82b08fef40281bc49d46fdR729

lib/nerves_hub/devices.ex
joshk: Thanks for sharing this.

I think in this case it isn't about hiding but about having a more readable function.

test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
joshk: I'm a little torn. I don't see this as a side effect but as a core behaviour of the orchestrator. By moving the tests away from the orchestrator we are no longer testing that the orchestrator responds to this event.

lib/nerves_hub/devices.ex
```diff
@@ -976,16 +1017,40 @@ defmodule NervesHub.Devices do
...
+        {:ok, updated_device}
+      end)
 
+    # Always notify the device about its deployment assignment
     DeviceEvents.deployment_assigned(device)
 
-    # let the orchestrator know that a device has been added to the deployment group
-    DeploymentOrchestratorEvents.device_added(device)
+    # Only notify the orchestrator if the deployment actually changed
+    if deployment_changed do
```
joshk: I wonder if we should be checking if the deployment group has changed much earlier, eg. a function signature that checks for this, and exit straight away.

lib/nerves_hub/devices.ex
```diff
@@ -1382,16 +1447,38 @@ defmodule NervesHub.Devices do
...
+              ^deployment_group.current_release.firmware.architecture
+          )
+          |> Repo.update_all([set: [deployment_id: deployment_id]], timeout: to_timeout(minute: 2))
+
+        # Queue delta generation for any new device firmware combinations immediately
+        # after the device updates within the same transaction
+        _ =
+          if deployment_group.delta_updatable do
+            ManagedDeployments.trigger_delta_generation_for_deployment_group(deployment_group)
+          end
```
joshk: should we move the `if` into a `trigger_delta_generation_for_deployment_group` function guard?

lib/nerves_hub/devices.ex
```diff
@@ -1654,6 +1741,9 @@ defmodule NervesHub.Devices do
   end
 
   def told_to_update(device_id, deployment_group, opts) do
+    # Reload deployment_group from database to ensure we have the latest current_release_id
+    deployment_group = Repo.get!(ManagedDeployments.DeploymentGroup, deployment_group.id)
```
joshk: did you encounter issues when this wasn't up to date?

if we are doing a 'get!' here, maybe its better for us to combine this with the load and preload below.

test/nerves_hub/managed_deployments/distributed/orchestrator_test.exs
kybishop: Ah, I see, the `expect`s on `available_for_update` were being used as a proxy for the orchestrator being called as a result of the various inputs.

I'm going to noodle on this, testing internal implementation of the orchestrator through a proxy feels super wonky. If I can find a way to more concretely test behavior/outcomes instead of implementation I'll go with that, but if not I'll bring back the expects.

PR #2515, opened 2026-03-06
Move to using Phoenix Scopes for better auth organization

joshk:
This is the beginning of moving all context functions which fetch, create, update, or delete to requiring the use of `Scope` in their function args.

I've also cleaned up our plug usage as some were no longer needed, and I'm focused on making API and UI use as much of the same authorization fundamentals since they now have `Scope` use in common.

Its a big PR, but there was no way around that.

There is more work to be done around context functions, and tidying up ones which aren't needed or could be consolidated.

### DIFF
--- a/lib/nerves_hub_web/plugs/api/require_authenticated_user.ex
+++ b/lib/nerves_hub_web/plugs/api/require_authenticated_user.ex
@@ -4,7 +4,7 @@ defmodule NervesHubWeb.API.Plugs.RequireAuthenticatedUser do
   end
 
   def call(conn, _opts) do
-    if conn.assigns.current_scope && conn.assigns.current_scope.user do
+    if get_in(conn.assigns.current_scope.user) do
       conn
     else
       raise NervesHubWeb.UnauthorizedError


lib/nerves_hub_web/helpers/authorization.ex
```diff
@@ -1,69 +1,78 @@
...
+  alias NervesHub.Accounts.Scope
   alias NervesHub.Accounts.User
 
   def authorized!(org_user, permission) do
     authorized?(org_user, permission) || raise NervesHubWeb.UnauthorizedError
   end
 
-  def authorized?(:"organization:update", %OrgUser{role: ur}), do: role_check(:admin, ur)
-  def authorized?(:"organization:delete", %OrgUser{role: ur}), do: role_check(:admin, ur)
+  def authorized?(:"organization:update", role), do: role_check(:admin, role)
```
lawik, line 10: I think all these should take the name/binding `scope` rather than `role`? They get unpacked to role later right?

joshk: https://github.com/nerves-hub/nerves_hub_web/pull/2515/changes#diff-f5e5f7f36dab4d2d5c78925b32d60b74cd6ae40843432c1dc6315a5f7c0789baR69-R75

that takes an `OrgUser` or role

config/config.exs
```diff
@@ -88,6 +89,43 @@ config :nerves_hub, Oban,
...
+  user: [
+    default: true,
+    module: Scope,
+    assign_key: :current_scope,
+    access_path: [:user, :id],
+    schema_key: :user_id,
+    schema_type: :id,
+    schema_table: :users
+    # test_data_fixture: MyApp.AccountsFixtures,
+    # test_setup_helper: :register_and_log_in_user
```
nshoes, line 102: Cruft? As well as other comments in this file?

lib/nerves_hub/products.ex
```diff
@@ -20,6 +21,34 @@ defmodule NervesHub.Products do
...
+    |> where([_, _, ou], is_nil(ou.deleted_at))
+    |> where([p], p.name == ^product_name)
+    |> where([_, o], o.id == ^current_scope.org.id)
+    |> where([_, _, ou], ou.user_id == ^current_scope.user.id)
+    |> Repo.one!()
+  end
+
+  @spec get_products(Scope.t()) :: [Product.t()]
+  def get_products(%Scope{user: user, org: org}) do
+    from(
```
nshoes, line 39: Why use `from` here?

lib/nerves_hub_web/live/deployment_groups/new.ex
```diff
@@ -139,9 +135,9 @@ defmodule NervesHubWeb.Live.DeploymentGroups.New do
   # end
```
nshoes, line 135: Let's just delete all of this?

lib/nerves_hub_web/plugs/api/require_authenticated_user.ex
```diff
@@ -0,0 +1,13 @@
+defmodule NervesHubWeb.API.Plugs.RequireAuthenticatedUser do
+  def init(opts) do
+    opts
+  end
+
+  def call(conn, _opts) do
+    if conn.assigns.current_scope && conn.assigns.current_scope.user do
```
nshoes: Tiny nit
```suggestion
    if get_in(conn.assigns.current_scope.user) do
```

config/config.exs
joshk: I kept these as a reminder to relook at our fixtures, but I'm happy to remove them as well

lib/nerves_hub/products.ex
joshk: This was a function that I just reused and added Scopes on top of. On my list of things to clean up.

lib/nerves_hub_web/live/deployment_groups/new.ex
joshk: Good idea! I'll do that in another PR though.

PR #2524, opened 2026-03-08
Add support for Lucide icons

joshk:
This gives us a clear and consistent icon library.

https://lucide.dev/

The other strong contender was https://phosphoricons.com/

The Lucide library is imported as a hex lib, and implemented using a similar Tailwind plugin that Phoenix generates for Heroicons.

I've had to implement the different stoke widths using different 'weights' via a suffix at the end of the class name:

```
      let weights = [
        ["--verythin", "0.8"],
        ["--thin", "1"],
        ["--light", "1.25"],
        ["--semilight", "1.5"],
        ["", ""],
        ["--semibold", "2.25"],
        ["--bold", "2.5"],
      ]
```

The default is `2`, but it felt a bit thick for our use.

**Org sidebar before:**
Screenshot 2026-03-09 at 11 11 22 AM

**Org sidebar after:**
Screenshot 2026-03-09 at 11 12 03 AM

**Product sidebar before:**
Screenshot 2026-03-09 at 11 12 38 AM

**Product sidebar after:**
Screenshot 2026-03-09 at 11 13 00 AM

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -24,6 +24,7 @@
     "Dockerfile",
     "config",
     "assets/{node_modules,static}",
+    "assets/tailwind.config.js",
     "mix.exs",
     "priv",
     "rel",


assets/tailwind.config.js
```diff
@@ -101,6 +104,54 @@ export default {
         },
         { values: theme("transitionDelay") }
       )
+    }),
+    plugin(function ({ matchComponents, theme }) {
+      let iconsDir = path.join(__dirname, "../deps/lucide/icons")
+      let values = {}
+      let weights = [
+        ["--verythin", "0.8"],
```
lawik, line 112: does it need to be double dash? I feel like I don't see that in Tailwind in general..

lawik (approved):
One question :)

assets/tailwind.config.js
joshk: it doesn't need to be, but i've seen it used but other icon plugins since stroke width (weight) is usually its own css modifier.

joshk:
I just updated the PR to give an increased stroke width to selected icons.

Screenshot 2026-03-09 at 11 36 46 AM

I liked the added highlight

PR #2550, opened 2026-03-13
Use existing auth tokens for device console WebSocket

elielgordondensity:
## Summary
- Replace Phoenix.Token-based authentication in `UserSocket` with direct API token and session token support
- `UserSocket.connect/2` now accepts `token` (API token, e.g. `nhu_...`) or `session_token` (base64-encoded session token) params
- Remove Phoenix.Token generation from console and local shell tab components — they now use the session token already available in LiveView assigns
- Assign `user_token` (base64 session token) in both plug pipeline and LiveView mount for use by JS hooks

## Motivation
Previously, the device console WebSocket required a Phoenix.Token generated specifically for the connection. This meant API clients (e.g. mobile apps) had no way to connect to the console WebSocket using their existing auth credentials.

Now clients can connect using the same tokens they already have:
- **Web UI**: Session token flows via `@user_token` assign → `data-user-token` → JS sends `{session_token: "..."}` → `UserSocket` decodes and verifies
- **API/mobile clients**: Send `{token: "nhu_..."}` → `UserSocket` verifies via `fetch_user_by_api_token` (backward compatible — same `token` param key as before)

## Changes
- `user_socket.ex` — Two `connect/2` function heads: one for API tokens (`token`), one for session tokens (`session_token`)
- `auth.ex` — Assign `user_token` as base64-encoded session token instead of Phoenix.Token, in both plug and LiveView mount paths
- `console_tab.ex` / `local_shell_tab.ex` — Remove per-render Phoenix.Token generation, rely on `@user_token` from assigns
- `console.js` / `localShell.js` — Send `session_token` param instead of `token`
- `account_auth.ex` — Assign `user_token` to LiveView socket

## Test plan
- [x] `local_shell_tab_test.exs` — console UI renders without KeyError on `user_token`
- [ ] Manual: web console connects and relays I/O via session token
- [ ] Manual: API client connects via `{token: "nhu_..."}` param


joshk (changes requested):
I think we can skip the need for this and instead update the websocket auth to accept api auth token.

joshk:
Hey @elielgordondensity 

I'm sorry that this never got further replies from us.

I've deployed a variation of your PR which adds API token auth support to `/api/socket`, along with slight channel joining modifications using the device identifier. 

This allows us to keep the use of Phoenix.Token for web UI requests, while having a path for API usage where we can also encapsulate restrictions around max consoles and connections a user can have going at once.

I've tested these changes with a CLI experiment, as well as making sure it works with the Device 'local shell' extension.

Is it alright if I close this?

elielgordondensity:
of course! @joshk thanks for the update

PR #2560, opened 2026-03-24
Prevent future missing indexes and add remaining missing fkey indexes

kybishop:
Depends on #2558 and #2559


test/nerves_hub/database/index_test.exs
```diff
@@ -0,0 +1,52 @@
...
+          JOIN pg_catalog.pg_class      c2 ON c2.oid = t.confrelid
+          JOIN pg_catalog.pg_namespace  n2 ON n2.oid = c2.relnamespace
+          JOIN pg_catalog.pg_attribute  a2 ON a2.attrelid = t.confrelid AND a2.attnum = t.confkey[1]
+          WHERE t.contype = 'f'
+          AND NOT EXISTS (
+            SELECT 1
+            FROM pg_catalog.pg_index i
+            WHERE i.indrelid = t.conrelid
+            AND i.indkey[0] = t.conkey[1]
+          )
```
esvinson, line 77:
```suggestion
            AND i.indkey[0] = t.conkey[1]
            AND i.indpred IS NULL
          )
```
This makes it so partial indexes don't count.

priv/repo/migrations/20260324203540_add_missing_foreign_key_indexes.exs
```diff
@@ -0,0 +1,37 @@
...
+    create index(:firmwares, [:org_id], concurrently: true)
+    create index(:firmwares, [:org_key_id], concurrently: true)
+    create index(:inflight_deployment_checks, [:deployment_id], concurrently: true)
+    create index(:inflight_deployment_checks, [:device_id], concurrently: true)
+    create index(:inflight_updates, [:firmware_id], concurrently: true)
+    create index(:invites, [:invited_by_id], concurrently: true)
+    create index(:invites, [:org_id], concurrently: true)
+    create index(:jitp, [:product_id], concurrently: true)
+    create index(:org_keys, [:created_by_id], concurrently: true)
+    create index(:org_users, [:user_id], concurrently: true)
```
esvinson:
```suggestion
    create index(:org_users, [:user_id], concurrently: true)
    create index(:org_users, [:org_id], concurrently: true)
```
With the change below this one also needs to be created.

joshk (commented):
While I've left some feedback in a comment, I thought I'd also leave it as a proper PR review.

Creating an index for every FK will lead to index bloat as these new indexes won't be used by our queries. 

This test will force us to create indexes which we won't need or use. For NervesCloud, we don't want that.

kybishop:
@joshk updated with a whitelist like we discussed 🙂 

Includes a single missing index that I believe to be valid.

nshoes (approved):

priv/repo/migrations/20260519000000_add_devices_org_id_index.exs
```diff
@@ -0,0 +1,10 @@
+defmodule NervesHub.Repo.Migrations.AddDevicesOrgIdIndex do
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def change() do
+    create(index(:devices, [:org_id], concurrently: true))
```
joshk, line 8: Is it worth adding this to the known missing foreign indexes?
is this index needed?

kybishop: See previous comment 😉

https://github.com/nerves-hub/nerves_hub_web/pull/2560#issuecomment-4491079327

joshk: oops, missed that.

joshk (approved):

PR #2577, opened 2026-04-02
Handle race condition with a deleted deployment group

kybishop:
We saw this crop up a few times

### DIFF
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -634,10 +634,20 @@ defmodule NervesHub.ManagedDeployments do
   end
 
   @spec verify_deployment_group_membership(Device.t()) :: Device.t()
-  def verify_deployment_group_membership(
-        %Device{deployment_id: deployment_id, firmware_metadata: %{version: device_version}} = device
-      )
+  def verify_deployment_group_membership(%Device{deployment_id: deployment_id} = device)
       when not is_nil(deployment_id) do
+    with {:ok, deployment_group} <- get_or_clear_deployment_group(device) do
+      audit_device_deployment_group(device, deployment_group)
+    end
+  end
+
+  def verify_deployment_group_membership(device), do: device
+
+  # This is meant to safeguard against race conditions where the deployment group is deleted
+  # Simply returns the device if the deployment group was cleared since there isn't any more checking to do
+  defp get_or_clear_deployment_group(device) do
+    deployment_id = device.deployment_id
+
     deployment_group =
       DeploymentGroup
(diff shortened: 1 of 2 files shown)


lib/nerves_hub/managed_deployments.ex
```diff
@@ -634,23 +634,43 @@ defmodule NervesHub.ManagedDeployments do
...
+    with {:ok, deployment_group} <- get_or_clear_deployment_group(device) do
+      audit_device_deployment_group(device, deployment_group)
+    end
+  end
+
+  def verify_deployment_group_membership(device), do: device
+
+  # This is meant to safeguard against race conditions where the deployment group is deleted
+  # Simply returns the device if the deployment group was cleared since there isn't any more checking to do
+  defp get_or_clear_deployment_group(device) do
```
joshk, line 648: this function is expected to return a deployment, but on line 663 it will return the updated device. 

It might be best to return `{:error, :deployment_cleared, device}` for consistency, although this would mean adding an `else` to the `with` so that the `device` can be returned.

lib/nerves_hub/managed_deployments.ex
```diff
@@ -634,23 +634,43 @@ defmodule NervesHub.ManagedDeployments do
...
       |> from(as: :deployment_group)
       |> where([d], d.id == ^deployment_id)
       |> join_current_release(true)
       |> Repo.one()
 
+    if deployment_group do
+      {:ok, deployment_group}
+    else
+      device
+      |> Ecto.Changeset.change(%{deployment_id: nil})
```
joshk, line 662: I think it would be nice to move this to `Device.clear_deployment()`

joshk: we can then apply this to line 703

lib/nerves_hub/managed_deployments.ex
```diff
@@ -634,23 +634,43 @@ defmodule NervesHub.ManagedDeployments do
...
+      |> Repo.update!()
+    end
+  end
+
+  defp audit_device_deployment_group(device, deployment_group) do
```
joshk, line 667: I don't think this should be broken out into a separate function, since this is where the verification happens. The auditing is just a side effect of the deployment.

joshk (changes requested):
I'm surprised you have seen these errors, as this is a very, very specific race condition, requiring a deployment to be deleted just as a device has been loaded after its connection to NH.

kybishop:
I think I'm just going to close this out for now as we haven't seen this crop up again. I'll come back to it if we see the bug repeat itself

PR #2580, opened 2026-04-03
SR - Remove compression

kybishop:
🚧 👷 🚧


lawik:
I assume this was to address something in SmartRent's usage fairly immediately. Feels like this could be a config option for the longer term.

Curious about the reason :)

joshk:
This is a compile time setting, and there is no runtime alternative.

lawik:
> there is no runtime alternative

how fun :D

nshoes:
Superceded by https://github.com/nerves-hub/nerves_hub_web/pull/2865

PR #2598, opened 2026-04-09
Pass associated user when seeding deployment group, format `priv`

nshoes:
I unfortunately nuked my local database and ran into an error when seeding. Deployment group and release creation needs the associated org/product user on creation.

I also added `priv` to the formatter since `seeds.exs` was excluded, and then formatted all the things. I don't have a problem with this, and think it's right. Open to others thoughts.

I split the work into two commits so changes are easy to review.


joshk (approved):
cool cool cool

PR #2606, opened 2026-04-16
Exclude analytics repo from migrations if disabled

thiagoesteves:
During the installation of NervesHub in our Cloud environment, we found that when the analytics repository is not in use, the analytics migration was still being run. This PR skips that migration when analytics is not enabled, following the same pattern used in [lib/nerves_hub/application.ex:L132](https://github.com/nerves-hub/nerves_hub_web/blob/0216931beadcceeeba859d75461456c3768a95e2/lib/nerves_hub/application.ex#L132).


joshk:
Ahhhhh, nice, thank you.

PR #2610, opened 2026-04-17
Add download endpoint to firmware api

OleHenFo:
Having this would be very nice for our automation tools.
Would be used to fetch the latest firmware from a deploy group when flashing new cards for instance.

PR #2613, opened 2026-04-18
Add role validation to api firmware download

OleHenFo:
Follow up to PR #2610 since role validation was missing


joshk (approved):
You are a star!

PR #2615, opened 2026-04-20
Add a minimum delta version

kybishop:
After all the work to fwup to improve delta performance, it's clear we cannot update some of our devices until they get a fully updated version of fwup. As such, this PR introduces a gate on deltas so you can specify a minimum version. This will allow us to constrain deltas to a firmware version that we know includes a fwup with all the perf improvements we need.

Screenshot 2026-04-20 at 1 26 56 PM

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -983,6 +983,20 @@ defmodule NervesHub.Devices do
     Repo.exists?(query)
   end
 
+  defp meets_delta_minimum_version?(%Device{} = device, %DeploymentGroup{delta_minimum_version: min_version})
+       when not is_nil(min_version) do
+    case {Version.parse(device.firmware_metadata.version), Version.parse(min_version)} do
+      {{:ok, device_vsn}, {:ok, min_vsn}} ->
+        Version.compare(device_vsn, min_vsn) != :lt
+
+      _ ->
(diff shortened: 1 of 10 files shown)


kybishop:
Turning this into a draft in the hope that we can have this computed automatically behind the scenes.

lawik:
@kybishop Confuse can figure out some feature-support: https://github.com/underjord/confuse/blob/main/lib/confuse/fwup.ex

In this case it isn't quite that things aren't supported but rather pathologically slow right? So not sure we can just detect that considering the current deltas would work for lots of cases. Is there something detectable in the new firmware that clearly indicates that it wants at least fwup version X.Y.Z+, maybe some config about those cache/block/window dealios I saw you writing about.

If the new firmware has that, don't make deltas for unless fwup version is at least this high..

kybishop:
@lawik yeah that was the plan: If a device has an encrypted partition, and is below a certain fwup version, _don't_ use a delta for it, because it could be a multi hour operation. Also, PR for you to accomplish this 😉 https://github.com/underjord/confuse/pull/7

Awaiting some feedback from @joshk on this since it would affect all folks that have an encrypted partition, even if their CPU is fast enough for the delta to update in a reasonable time.

kybishop:
closing this since we're going with a different approach where one simply uses deployment group routing to get to a minimum firmware version

PR #2618, opened 2026-04-24
Use source window of device's cache (when available)

kybishop:


lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -183,7 +183,11 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
             source_work_dir,
             target_work_dir,
             output_work_dir,
-            all_delta_files
+            all_delta_files,
+            # Note: We intentionally use the target firmware's block cache size in case you need to update
+            # this value on the fly, perhaps due to running out of memory during delta application on the device.
+            # This allows you to rebuild the target firmware with a new block cache size if you need to.
+            Map.get(tool_metadata, :block_cache_size_mb)
```
joshk: I think it would be best to either pass the full `tool_metadata` to the function, and then move the note to the function, or move the note and `Map.get/2` to above the function call. (I would vote for option 1)

lib/nerves_hub/firmwares/update_tool/fwup.ex
```diff
@@ -256,7 +261,12 @@ defmodule NervesHub.Firmwares.UpdateTool.Fwup do
 
       case File.stat(source_filepath) do
         {:ok, %{size: f_source_size}} ->
-          args = ["-A", "-S", "-f", "-s", source_filepath, target_filepath, output_path]
+          b_args =
```
joshk: could we rename the variable to something descriptive like `block_cache_args`

joshk (changes requested):
Did we want to merge this before @lawik can release a new Confuse version?

lib/nerves_hub/firmwares/update_tool/fwup.ex
nshoes: I vote option one as well

lib/nerves_hub/firmwares/update_tool/fwup.ex
kybishop: For xdelta3 this corresponds to the source window size. Renamed to match and added a brief comment for context

lib/nerves_hub/firmwares/update_tool/fwup.ex
kybishop: done

joshk (approved):
LGTM

joshk:
@lawik any thoughts on the `confuse` version?

lawik:
Yeah, can make a release

lawik:
v0.4.0 is out

PR #2620, opened 2026-04-27
When a device is disconnecting, consider no DeviceConnection records updated successful

nshoes:


lib/nerves_hub/devices/connections.ex
```diff
@@ -72,7 +72,7 @@ defmodule NervesHub.Devices.Connections do
         :ok
 
       _ ->
-        :error
+        :ok
```
lawik, line 75: But it shouldn't ensure that it is marked online in tracker state?

nshoes: I'm unsure on how strict we want to be here. If a device is having issues connecting, and this path is hit where no device connection is updated, I don't think we should.

kybishop (approved):

PR #2621, opened 2026-04-27
Log unhappy-path result when updating metadata on device connect

nshoes:
We saw this in our last deploy:

```
MatchError: no match of right hand side value:

    :error

    lib/nerves_hub/device_link.ex:47: Elixir.NervesHub.DeviceLink.NervesHub.DeviceLink.update_connection_metadata/2
    lib/nerves_hub/device_link.ex:22: Elixir.NervesHub.DeviceLink.NervesHub.DeviceLink.join/3
    lib/nerves_hub_web/channels/device_channel.ex:41: Elixir.NervesHubWeb.DeviceChannel.NervesHubWeb.DeviceChannel.join/3
    lib/phoenix/channel/server.ex:401: Elixir.Phoenix.Channel.Server.Phoenix.Channel.Server.channel_join/4
    lib/phoenix/channel/server.ex:315: Elixir.Phoenix.Channel.Server.Phoenix.Channel.Server.handle_info/2
    gen_server.erl:2434: gen_server.:gen_server.try_handle_info/3
    gen_server.erl:2420: gen_server.:gen_server.handle_msg/3
    proc_lib.erl:333: proc_lib.:proc_lib.init_p_do_apply/3
```

The source cause may be a timing issue, and _could_ be solved in https://github.com/nerves-hub/nerves_hub_web/pull/2620. To further validate, we're allowing the error through to log the result. If we don't see this log then we know these two errors are related.


kybishop (approved):

lib/nerves_hub/extensions/geo.ex
```diff
@@ -49,7 +49,7 @@ defmodule NervesHub.Extensions.Geo do
 
   @impl NervesHub.Extensions
   def handle_in("location:update", location, socket) do
-    Connections.merge_update_metadata(socket.assigns.reference_id, %{location: location})
+    _ = Connections.merge_update_metadata(socket.assigns.reference_id, %{location: location})
```
joshk:
```suggestion
    :ok = Connections.merge_update_metadata(socket.assigns.reference_id, %{location: location})
```

joshk: I think its best to fail on this not updating, as it might help shine a light on a larger issue

lib/nerves_hub_web/channels/device_channel.ex
```diff
@@ -194,9 +194,10 @@ defmodule NervesHubWeb.DeviceChannel do
 
   @decorate with_span("Channels.DeviceChannel.handle_in:connection_types")
   def handle_in("connection_types", %{"values" => types}, socket) do
-    DeviceLink.update_connection_metadata(socket.assigns.reference_id, %{
-      "connection_types" => types
-    })
+    _ =
```
joshk:
```suggestion
    :ok =
```

joshk: (same as above)

joshk (changes requested):
Love the direction, thanks for this!

joshk (approved):
❤️

PR #2623, opened 2026-04-27
Fix a UI bug related to the deployment group filter being reset

joshk:
If a deployment group was selected in the devices list filtering, and then another select drop-down filter was used, the UI reset the deployment group filter even though it was still present.

This fixes the UI bug.


lib/nerves_hub_web/components/filter_sidebar.ex
```diff
@@ -56,7 +56,7 @@ defmodule NervesHubWeb.Components.FilterSidebar do
                     <input class="sidebar-text-input" type="text" name={filter.attr} id={"input_#{filter.attr}"} value={@current_filters[filter.attr]} phx-debounce="500" />
                   <% :select -> %>
                     <select class="sidebar-select" name={filter.attr} id={"input_#{filter.attr}"}>
-                      <option :for={{label, value} <- filter.values} value={value} selected={to_string(@current_filters[filter.attr]) == value}>
+                      <option :for={{label, value} <- filter.values} value={value} selected={to_string(@current_filters[filter.attr]) == to_string(value)}>
```
nshoes, line 59: Ah, that'll do it 😆

nshoes (approved):

PR #2625, opened 2026-04-29
Support running bulk actions on all listed devices, including matching devices when using filters

joshk:
I've added a way to 'select all' devices, including when using filters.

And then I've added support for bulk actions to accept an Ecto query.

And I've used LiveView `start_async` for background processing.

More thought and love needs to be applied to the various 'edge' cases when selecting all devices, then maybe paginating, or removing a device. But I'm going to leave that for another PR.

https://github.com/user-attachments/assets/98f05878-5b76-4a72-8a94-a8b4f2cd5b1b

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -609,6 +609,17 @@ defmodule NervesHub.Firmwares do
         {:ok, %FirmwareDelta{}} ->
           {:ok, :delta_already_exists}
 
+        {:delta_insert,
+         {:error,
+          [
+            {:unique_firmware_delta,
+             {"has already been taken", [constraint: :unique, constraint_name: "source_id_target_id_unique_index"]}}
+          ]}} ->
+          # a race condition exists where multiple devices connect, are added to a deployment group, and
+          # try to create a delta, only for one to 'start' but all others fail.
+          # This should be regarded as an `:ok` response.
+          {:ok, :delta_already_exists}
+
         {:delta_insert, {:error, changeset}} ->
           Logger.warning("Failed to insert firmware delta for #{source_id} -> #{target_id}")
 
--- a/lib/nerves_hub_web/live/deployment_groups/show.ex
+++ b/lib/nerves_hub_web/live/deployment_groups/show.ex
@@ -3,6 +3,7 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Show do
 
   alias NervesHub.AuditLogs.DeploymentGroupTemplates
   alias NervesHub.Devices
+  alias NervesHub.Devices.BulkActions
   alias NervesHub.Helpers.Logging
   alias NervesHub.ManagedDeployments
   alias NervesHubWeb.Components.DeploymentGroupPage.Activity, as: ActivityTab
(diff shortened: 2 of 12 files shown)


lib/nerves_hub/firmwares.ex
```diff
@@ -609,6 +609,17 @@ defmodule NervesHub.Firmwares do
...
+         {:error,
+          [
+            {:unique_firmware_delta,
+             {"has already been taken", [constraint: :unique, constraint_name: "source_id_target_id_unique_index"]}}
+          ]}} ->
+          # a race condition exists where multiple devices connect, are added to a deployment group, and
+          # try to create a delta, only for one to 'start' but all others fail.
+          # This should be regarded as an `:ok` response.
+          {:ok, :delta_already_exists}
+
```
nshoes, line 622: Nice catch

lib/nerves_hub_web/live/deployment_groups/show.ex
```diff
@@ -92,8 +93,8 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Show do
     move_devices = fn ->
       devices = ManagedDeployments.matched_device_ids(deployment_group, in_deployment: false)
 
-      Devices.move_many_to_deployment_group(scope, devices, deployment_group)
-      |> then(fn {:ok, %{updated: updated_count, ignored: ignored_count}} ->
+      BulkActions.move_many_to_deployment_group(devices, deployment_group, scope.user)
```
nshoes, line 96: Awesome

lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -303,33 +326,37 @@ defmodule NervesHubWeb.Live.Devices.Index do
...
+          {:noreply, put_flash(socket, :error, "You haven't selected any devices")}
+
+        {:successful, false} ->
+          {:noreply, put_flash(socket, :error, "No devices were successfully tagged")}
+
+        {:has_errors, true, result} ->
+          socket
+          |> put_flash(
+            :info,
+            "#{Enum.count(result[:ok])} devices were successfully tagged and #{Enum.count(result[:error])} devices had errors."
```
nshoes, line 354: I have similar work where I'm deciding how to handle errors in a batch action like this. I don't want to ship something that doesn't surface errors, but I'm unsure of the right way of handling and displaying them.

lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -734,6 +792,137 @@ defmodule NervesHubWeb.Live.Devices.Index do
     |> noreply()
   end
 
+  def handle_async(:enable_updates_for_devices, {:ok, results}, socket) do
```
nshoes, line 795: Nice to handle these with LV primitives, good call

nshoes (commented):
LGTM! Could we get some videos of this in action?

lib/nerves_hub/firmwares.ex
joshk: Oops, that was supposed to be in a different PR, it got caught up in my commit.

lib/nerves_hub_web/live/devices/index.ex
joshk: Might be good to discuss this on a Zoom or in an issue. I don't think surfacing errors is very helpful, especially if you have hundreds of errors. Instead I think errors should be escalated to Sentry as they are unexpected, and likely something the user cant fix.

These errors can also be a memory hog as the Ecto Changeset struct can be big, and if you have 50k errors, well, that isn't useful for the user, but we could send the errors to Sentry during the stream loop.

nshoes: Cool, agreed. Sending them to Sentry is reasonable. This PR or a follow-up?

nshoes (approved):

lib/nerves_hub_web/live/devices/index.ex
joshk: A different PR. There is more cleanup and tweaks that I would like to make, but its not high on the priority list.

PR #2631, opened 2026-04-30
Tweak the Device export to use HTTP streaming (chunks)

joshk:
The core change is generalising `devices_csv/1` to instead act as a reducer so we can pass in a custom function which allows us to stream the response to the user.

I've also tweaked the function to select less information from the database, fetch more rows, and use a 90 second timeout for the transaction.

Oh, and I added a test for the HTTP part of this.

Side note: This is way more memory efficient as we aren't creating a 500k row CSV and then sending it to the client, instead we are sending rows as they become available.


nshoes (approved):
This is so sick, great stuff

PR #2661, opened 2026-05-13
Ignore messages to Extensions if they aren't attached

joshk:
I've been hunting down this issue for ages, and I finally found it!

Its possible for Extensions to be forwarded `handle_info` messages when they aren't attached.

This ignores messages, and finally fixing the Sentry error that keeps popping up under a race condition.


nshoes (approved):
Great find!

PR #2662, opened 2026-05-13
Simplify the Oban queues into three

joshk:
This implements an Oban best practice of less queues is better.

Less queues, grouped based on function (and essentially priority).


nshoes (approved):
Much better

PR #2665, opened 2026-05-13
Track device firmware history

joshk:
This is the start of firmware history tracking.

This will soon be expanded to work with Deployment Workflows.

Screenshot 2026-05-13 at 11 29 07 PM


nshoes (approved):
LGTM!

PR #2680, opened 2026-05-15
Start a health check endpoint when running a Device node

joshk:
This is useful for containerized deployments, like k8s or fly.io, where the health check can't use the full hostname, causing SSL to fail.


nshoes (approved):
Love it!

PR #2688, opened 2026-05-21
Improve the PubSub topic used for internal device notifications

joshk:
- `internal:device:[device_id]` allows for pattern matching based on topic, eg. `"internal:device:" <> device_id`
- remove an unused `Tracker.confirm_online/1`, and any `handle_info`s which listened for `connection:status`
- switch to using `device.id` instead of `device.identifier` in the topics


lawik (approved):
Seems reasonable. Eyes glazed over a few times so not sure how thorough the review is. This must have been exciting to do :D

nshoes (approved):

PR #2689, opened 2026-05-21
Enhance InflightUpdate to include more information

joshk:
This builds upon #2688 

- Inflight update information is created for manual updates
- Stage and progress is persisted to the DB (progress is only persisted every 10 seconds, max)
- The persisted information allows for the Devices list and show page to show an updates progress straight away
- The persisted information also works as a heartbeat, which allows us to remove the need for 'inflight update expiration' and instead rely on the heartbeat (if there are no updates for 15 mins then the update is expired)

And I slipped in a little UI update to the show page progress bar.

Screenshot 2026-05-21 at 10 02 07 PM


priv/repo/migrations/20260519233623_change_inflight_update_constraint.exs
```diff
@@ -0,0 +1,17 @@
+defmodule NervesHub.Repo.Migrations.ChangeInflightUpdateConstraint do
+  use Ecto.Migration
+
```
nshoes, line 6:
```suggestion
@disable_ddl_transaction true
@disable_migration_lock true

```

nshoes (approved):
LGTM 😎

priv/repo/migrations/20260519233623_change_inflight_update_constraint.exs
joshk: I don't think this is needed as the InflightUpdate table is very very small, but I'll go with this none the less.

PR #2690, opened 2026-05-21
Accept firmware/archive UUIDs on deployment group updates

### LINKED ISSUE #2654: Question: supported API/CLI path for CI to create a Managed Deployment release from a firmware UUID?
Hi NervesHub team,

I’m trying to understand the intended way to automate the “create a new release on a deployment group” step for Managed Deployments / Deployment Groups.

## What we are trying to build

We have a GitHub Actions firmware pipeline for a Nerves device fleet:

1. Build firmware on merge to `main`.
2. Sign and publish the firmware to Hosted NervesHub.
3. Later, after explicit release approval, promote that already-published firmware UUID to a staging deployment group.
4. Let NervesHub offer the update to the device.
5. Let the device-side app/operator decide when to apply it, so updates are not applied at an unsafe time.

So the goal is **not** to force-install firmware from CI. We only want CI to create/select the deployment group release. The device/operator still controls when the offered update is applied.

## What works today

If I manually use the Hosted NervesHub UI and create a new release on the deployment group with the most recent firmware, the deployment group advances and the device receives the offer as expected.

Once device-side firmware updates are enabled, this flow works correctly.

## What I expected to work

I expected one of these CLI/API flows to create the same deployment-group release:

```sh
nh firmware publish path/to/firmware.fw --product my_product --deploy my_deployment
```
or, after publish:
```sh
nh deployment update my_deployment firmware <firmware_uuid> --product my_product
```
## What I observed
The firmware publishes successfully, and the CLI/API path appears to accept the deployment update, but the deployment group’s current release does not move to the requested firmware UUID. In our CI guard we poll nh deployment list afterward, and the deployment group still reports the previous firmware UUID.

Creating the release manually in the Hosted NervesHub UI does move it.

## Source reading
Looking at current upstream code, I may be misunderstanding the intended API shape:

- `nerves_hub_cli` sends deployment updates using the public firmware UUID under the firmware key.
- `nerves_hub_web` appears to create a new deployment release only when the update payload contains firmware_id or archive_id.
- The firmware JSON response exposes UUID/version/platform/etc., but not the internal firmware id.

So from the public CLI/API surface, I do not see a stable way to say: “create a new DeploymentRelease for this deployment group using this already-published firmware UUID.”

## Question
What is the intended supported way for a CI pipeline to create a new Managed Deployment / Deployment Group release from an already-published firmware UUID?

Is one of these expected?

1. `nh firmware publish --deploy <deployment>` should create a new DeploymentRelease.
2. `nh deployment update <deployment> firmware <uuid>` should create a new DeploymentRelease.
3. There is another public API/CLI endpoint we should use.
4. This is currently UI-only and needs a new API/CLI feature.

If this is a missing feature, I’m happy to help test or adjust our pipeline around the intended future API shape.

lawik:
The release-creation clause of the API deployment-group update was gated on the internal `firmware_id` / `archive_id` keys. The public CLI only knows the UUID and sends `firmware: <uuid>`, which fell through to the metadata-update clause; that path translated the UUID to `firmware_id` and called `update_deployment_group/3`, but the deployment-group changeset doesn't cast `:firmware_id`, so the field was silently dropped and no new `DeploymentRelease` was created.

Widen the release-creation clause's guard to also match the UUID forms and resolve them via `Firmwares.get_firmware_by_product_and_uuid/2` and `Archives.get/2`. The metadata-update clause now only handles non-firmware/archive changes.

Closes #2654

### DIFF
--- a/lib/nerves_hub_web/controllers/api/deployment_group_controller.ex
+++ b/lib/nerves_hub_web/controllers/api/deployment_group_controller.ex
@@ -67,9 +67,10 @@ defmodule NervesHubWeb.API.DeploymentGroupController do
         "name" => name,
         "deployment" => deployment_group_params
       })
-      when is_map_key(deployment_group_params, "firmware_id") or is_map_key(deployment_group_params, "archive_id") do
-    firmware = Firmwares.get_by_id(product, deployment_group_params["firmware_id"])
-    archive = Archives.get_by_id(product, deployment_group_params["archive_id"])
+      when is_map_key(deployment_group_params, "firmware_id") or is_map_key(deployment_group_params, "archive_id") or
+             is_map_key(deployment_group_params, "firmware") or is_map_key(deployment_group_params, "archive") do
+    firmware = resolve_firmware(product, deployment_group_params)
+    archive = resolve_archive(product, deployment_group_params)
 
     with {:ok, deployment_group} <-
            ManagedDeployments.get_deployment_group_by_name(product, name),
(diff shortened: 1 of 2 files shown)


lawik:
I had Claude confirm the issue using my weirdo end-to-end test thing for nerveshub: https://github.com/lawik/test_nerves_hub/commit/c029b55908d132d7762fd2099c0df1aae24dc149

joshk (approved):

PR #2691, opened 2026-05-21
Make support script output panels uniquely identifiable per script

### LINKED ISSUE #2607: Previous support script overrides output of subsequent scripts
**Describe the bug**
When running multiple support scripts without closing the previous one, the earlier script will override or appear in place of later ones. Results seem to be in a FIFO order rather than being what was recently executed. 
For example, if you:

- Run Script 1
- Run Script 2
- Close Script 1
- Run Script 3

The output would correspond to Script 2 instead of Script 3

**To Reproduce**
1. Run one support script
2. Run another support script without closing the prior one

**Expected behavior**
The displayed output should correspond to the most recently executed script. Alternatively, prior scripts should be automatically closed before running the next. 

**Screenshots**
Image

lawik:
The output panel rendered the same DOM ids ("support-script" / "support-script-output") for every script with output. The xterm.js hook called `document.getElementById("support-script-output")`, which always returned the first match, so subsequent scripts' terminals rendered the earlier script's output.

Suffix both ids with `script.id` and have the JS hook look up its matching content div via a `data-output-id` attribute.

Fixes #2607.

### DIFF
--- a/assets/js/hooks/supportScriptOutput.js
+++ b/assets/js/hooks/supportScriptOutput.js
@@ -36,7 +36,8 @@ const defaultTermOptions = {
 
 export default {
   mounted() {
-    let content = document.getElementById("support-script-output").textContent
+    let outputId = this.el.dataset.outputId || "support-script-output"
+    let content = document.getElementById(outputId).textContent
     let contentRows = content.split("\n")
 
     if (contentRows.length <= 20) {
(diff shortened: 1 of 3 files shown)


lawik:
The experience is still not particularly great. But it is not completely bonkers now.

nshoes (approved):

PR #2698, opened 2026-05-24
MFA support

pxp9:
This PR adds an extra layer of security to NervesHub while developing it and while using it.

Adds MFA support for extra layer of security in the login flow


joshk:
Thanks for the PR. Can you please add some screenshots and screencasts of this in action?

pxp9:
> Thanks for the PR. Can you please add some screenshots and screencasts of this in action?

Here is the video , unfortunately is too big to fit on github (only accepts 10 MB files)

so i uploaded to Google drive and put it public read there 

https://drive.google.com/file/d/1I43r0FKZZSwTA00OGEQy8vSEPDB9-aYO/view?usp=sharing

lib/nerves_hub_web/live/account_templates/edit_account_template.html.heex
```diff
@@ -102,6 +102,104 @@
...
+        <div :if={@user.mfa_enabled_at && is_nil(@mfa_recovery_codes)} class="text-base-300 mb-4 text-sm">
+          MFA is enabled for your account.
+        </div>
+
+        <div :if={is_nil(@user.mfa_enabled_at) && is_nil(@mfa_setup)} class="space-y-4">
+          <p class="text-base-300 text-sm">
+            Add an authenticator app to require a time-based code the next time you sign in.
+          </p>
+
+          <.form :let={f} id="mfa-start-form" for={@mfa_password_form} phx-submit="start-mfa-setup">
```
joshk, line 117: I don't think we need to require users confirm their password to setup MFA

This is what it looks like on hex.pm

Image

lib/nerves_hub_web/live/account_templates/edit_account_template.html.heex
```diff
@@ -102,6 +102,104 @@
...
+            Save these recovery codes now. They will not be shown again.
+          </div>
+
+          <div class="grid grid-cols-2 gap-2">
+            <code :for={code <- @mfa_recovery_codes} class="bg-base-800 text-base-100 rounded px-2 py-1 text-sm">{code}</code>
+          </div>
+        </div>
+
+        <div :if={@user.mfa_enabled_at && is_nil(@mfa_recovery_codes)} class="mt-4">
+          <.form :let={f} id="mfa-disable-form" for={@mfa_password_form} phx-submit="disable-mfa">
```
joshk, line 182: Same for removing MFA, we should show a modal to ask for confirmation, but I don't think we need to ask for the users password.

This is what it looks like on hex.pm

Image

lib/nerves_hub/accounts.ex
```diff
@@ -145,6 +149,189 @@ defmodule NervesHub.Accounts do
...
+
+  def confirm_mfa_setup(%User{} = user, secret, code) when is_binary(secret) and is_binary(code) do
+    if used_at = accepted_totp_time(secret, code) do
+      recovery_codes = generate_recovery_codes()
+      now = DateTime.utc_now() |> DateTime.truncate(:second)
+
+      Multi.new()
+      |> Multi.update(
+        :user,
+        Changeset.change(user, %{
```
joshk, line 179: these changesets should be added to `User`, eg. `User.confirm_mfa_changeset/2` and have the functions for generating recovery codes and encrypting the secret in `User`

lib/nerves_hub/accounts.ex
```diff
@@ -145,6 +149,189 @@ defmodule NervesHub.Accounts do
...
+        session_tokens = repo.all(UserToken.by_user_and_contexts_query(user, ["session"]))
+
+        Enum.each(session_tokens, fn session_token ->
+          Endpoint.broadcast("users_sessions:#{Base.url_encode64(session_token.token)}", "disconnect", %{})
+        end)
+
+        {deleted_count, _} = repo.delete_all(UserToken.by_user_and_contexts_query(user, ["session"]))
+
+        {:ok, deleted_count}
+      end)
```
joshk, line 196: During testing I was confused why confirming MFA setup caused me to be logged out. I then looked a bit deeper into the code and found this. We don't need to log a user out when they setup MFA, its a pretty disrupting and confusing UI flow.

lib/nerves_hub_web/controllers/session_controller.ex
```diff
@@ -37,6 +43,36 @@ defmodule NervesHubWeb.SessionController do
     end
   end
 
+  def mfa_new(conn, _params) do
+    if pending_mfa_user_id(conn) do
+      render(conn, :mfa, form: Phoenix.Component.to_form(%{}, as: "mfa"))
```
joshk, line 48: I don't think `Phoenix.Component` is needed

assets/js/hooks/mfaCodeInput.js
```diff
@@ -0,0 +1,66 @@
...
+
+      const nextEmpty = digitInputs.find((digitInput) => digitInput.value === "")
+      ;(nextEmpty || digitInputs[digitInputs.length - 1]).focus()
+
+      submitWhenComplete(form, hiddenInput)
+    })
+  })
+}
+
+const MfaCodeInput = {
```
joshk, line 60: this LiveHook is a little non standard when compared to our other hooks. Is there are a reason why each function is not part of the `MfaCodeInput` object?

joshk (changes requested):
This is a great start.

Regarding the UI for setting up MFA, can you please implement this as a modal, similar to how hex.pm does it:

Image

joshk:
Also, I just noticed that the tests are a little lacking for the MFA setup process.

joshk:
Hi @pxp9 

I'm going to close this PR for the time being as it's gone stale.

If you are interested in relooking at adding 2FA, please sing out.

Issue #2482, opened 2026-02-23
Color choice for the delta creation pulse


lawik:
This is a minor thing. The status pill thing near the title while generating deltas is currently a burnt yellow-ish color. Since it isn't a problem or warning I'd love for it to be blue or purple. It isn't all systems go yet (that'd be green). But it also isn't a problem.

My immediate reaction when it shows up is that something is slightly wrong, not just that processing is progressing.

joshk:
I believe this has been resolved by Lars.

PR #2699, opened 2026-05-27
Use notice color for the delta-preparing pill (#2482)

lawik:
The "Preparing: Generating Deltas" pill used `bg-warning` (amber), which reads as a problem at a glance even though delta generation is just progress. Switch to `bg-notice` (indigo) so it reads as informational.


lawik:
Screenshot_20260527_153854

kybishop (approved):

PR #2700, opened 2026-05-27
Don't crash device join when firmware_metadata is nil

### LINKED ISSUE #2430: Device without firmware metadata uuid can cause device channel to crash
During some client work we had an update run that for one reason or another didn't set a firmware UUID in the uboot env.

This is of course not ideal for update status and bookkeeping but it is very important that the device is still able to connect even when data is missing.

The error:

```
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]ts=2026-01-15T18:58:34.004 level=error msg="an exception was raised:
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]    ** (KeyError) key :uuid not found in: nil
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]If you are using the dot syntax, such as map.field, make sure the left-hand side of the dot is a map
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (nerves_hub 2.0.0+b4472d07) lib/nerves_hub/devices.ex:1042: NervesHub.Devices.firmware_update_successful/2
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (nerves_hub 2.0.0+b4472d07) lib/nerves_hub/device_link.ex:18: NervesHub.DeviceLink.join/3
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (nerves_hub 2.0.0+b4472d07) lib/nerves_hub_web/channels/device_channel.ex:40: NervesHubWeb.DeviceChannel.join/3
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (phoenix 1.8.1) lib/phoenix/channel/server.ex:394: Phoenix.Channel.Server.channel_join/4
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (phoenix 1.8.1) lib/phoenix/channel/server.ex:308: Phoenix.Channel.Server.handle_info/2
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (stdlib 6.2.2) gen_server.erl:2345: :gen_server.try_handle_info/3
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (stdlib 6.2.2) gen_server.erl:2433: :gen_server.handle_msg/6
2026-01-15T18:58:34Z app[e82031b77593d8] sjc [info]        (stdlib 6.2.2) proc_lib.erl:329: :proc_lib.init_p_do_apply/3"
```

Currently travelling so I'm logging it first and will see if I can get a fix in later.

lawik:
A device whose `firmware_metadata` is nil — e.g. a fresh row whose uboot env never reported a `nerves_fw_uuid` — would crash the device channel on join. `update_firmware_metadata/4` called `Map.from_struct(nil)` and, in older code paths, `firmware_update_successful/2` dereferenced `device.firmware_metadata.uuid`.

Skip the DeviceFirmwares report when there's no firmware metadata to attach to, and short-circuit `firmware_update_successful/2` for the same case — without metadata there's no real update to record.

Closes #2430.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -724,7 +724,9 @@ defmodule NervesHub.Devices do
           nil ->
             # Device joined without any known firmware metadata (e.g. its
             # uboot env had no `nerves_fw_uuid`). Nothing to report against,
-            # so just persist the device-level values.
+            # so just persist the device-level values. This is unexpected,
+            # so raise a product notification to surface it.
+            _ = ProductNotifications.create_missing_firmware_metadata_notification!(device)
             attrs
 
           current_metadata ->
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -719,18 +719,27 @@ defmodule NervesHub.Devices do
         firmware_auto_revert_detected: auto_revert_detected?
       }
 
-      firmware_metadata = Map.from_struct(device.firmware_metadata)
-
       attrs =
-        DeviceFirmwares.add_or_update_reported_firmware(
-          device,
-          firmware_metadata,
-          validation_status,
-          auto_revert_detected?
-        )
-        |> case do
-          :ok -> attrs
-          {:ok, df} -> Map.put(attrs, :current_device_firmware_id, df.id)
+        case device.firmware_metadata do
+          nil ->
+            # Device joined without any known firmware metadata (e.g. its
+            # uboot env had no `nerves_fw_uuid`). Nothing to report against,
+            # so just persist the device-level values.
+            attrs
+
+          current_metadata ->
+            firmware_metadata = Map.from_struct(current_metadata)
+
+            DeviceFirmwares.add_or_update_reported_firmware(
+              device,
+              firmware_metadata,
+              validation_status,
+              auto_revert_detected?
+            )
+            |> case do
+              :ok -> attrs
+              {:ok, df} -> Map.put(attrs, :current_device_firmware_id, df.id)
+            end
         end
 
       update_device(device, attrs)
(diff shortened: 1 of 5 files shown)


nshoes (changes requested):
LGTM, could we add tests?

lawik:
I figured it would be awkward, but on second thought it should make sense as a regression test at least.

lawik:
@nshoes tests added

joshk:
From my understanding a device connecting to NH would never report empty firmware metadata. So this has me wonder, should we enforce and require firmware metadata (with a nicer error code path) or allow this even though it shouldn't happen?

lawik:
This is based on observed behavior on a device. During development but we'd see the same from a mildly corrupted KV or simply one where through some mistake the uuid was never set.

I'd much prefer to be resilient to that and able to deliver a new firmware so the device probably recovers than reject the connection with an error.

joshk:
On second review of the code path, I think I understand this better.

The function you have updated (`update_firmware_metadata/4`) is already expecting `nil` from the firmware params sent by the device, and its that the `firmware_metadata` of the Device in the system is nil. So what threw me was that your comment here https://github.com/nerves-hub/nerves_hub_web/pull/2700/changes#diff-03454c312d5c891b1ae7d916a1eb171cdf023ca00f9aa5155ea5633f7c5cbe81R722 is not correct as it reads as though no metadata is sent by the device when its connecting.

Its valid and fine for devices not to have any `firmware_metadata`, eg. a device is created via the UI or API, but we should then expect metadata to be sent when it connects, and it shouldn't go down this code path as this path is for when there are no firmware metadata changes.

Your test is also a little brittle as you are testing updating firmware metadata with no new firmware metadata, which isn't a use case we see or should accept (that is up for debate). 

Do we see this issue happening in the wild? should we allow devices to connect without firmware metadata?

joshk:
@lawik sorry, I posted my second message without seeing your reply (the UI didn't update)

I'm generally on board with being less strict during the connection process. This sort of issue would point to a pretty major problem with Nerves or the installed system, and this would still hide it (granted, it's kinda hidden right now too).

This would be a good use of Product Notifications, and maybe a warning banner when viewing the device if firmware metadata is nil but the device has been 'provisioned'.

lawik:
Will be rebasing.

joshk (approved):
Issues mostly addressed

PR #2708, opened 2026-05-30
PNPM 11 instead of NPM

pxp9:
Replaces NPM usage with PNPM 11 usage because PNPM is secure by default to supply chain attacks which recently happened in the JS ecosystem.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -177,6 +177,8 @@ defmodule NervesHub.MixProject do
   # See the documentation for `Mix` for more info on aliases.
   defp aliases() do
     [
+      "assets.install": ["cmd pnpm install --dir assets"],
+      "assets.build": ["esbuild default", "tailwind default"],
       "assets.deploy": ["tailwind default --minify", "esbuild default --minify", "phx.digest"],
       "assets.setup": ["assets.install", "assets.build"],
       "ecto.setup": [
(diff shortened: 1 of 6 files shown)


pxp9:
it seems there is a flacky test breaking the PR CI.

joshk (commented):
I'll need to review this later in the week as I want to test it locally and confirm a few questions I have around pnpm setup.

mix.exs
```diff
@@ -177,6 +177,8 @@ defmodule NervesHub.MixProject do
   # See the documentation for `Mix` for more info on aliases.
   defp aliases() do
     [
+      "assets.install": ["cmd pnpm install --dir assets"],
+      "assets.build": ["esbuild default", "tailwind default"],
```
joshk, line 181: why was `assets.build` added?

joshk (changes requested):
Thanks for the PR

Why was the default `package.lock` not removed?

I also noticed that the Dockerfile hasn't been updated to install and use pnpm.

mix.exs
joshk: I think these are the correct changes:

```
"assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing", "cmd pnpm install --dir assets"],
"assets.build": ["compile", "tailwind default", "esbuild default"],
"assets.deploy": ["tailwind default --minify", "esbuild default --minify", "phx.digest"],
```

PR #2709, opened 2026-06-02
Full integration of all firmware update statuses

joshk:
- functions related to updating or cleaning `InflightUpdate` have been extracted to a new `FirmwareUpdates` context
- `ignored`, `rescheduled`, and `failed` firmware update statuses have been implemented to audit log and block updates, depending if the update was manually requested or scheduled by a deployment
- a fairly complete test suite of the various statuses expected
- modify the `inflight_updates` table to allow for null `firmware_uuid`s and `firmware_id`s so that a devices which are currently updating while a NH deployment don't cause a ton of false-ish errors

next up, saving the 'failed', 'rescheduled', and 'ignored' results as `DeviceFirmware` schemas


nshoes (approved):
This LGTM, the tests are very appreciated!

PR #2716, opened 2026-06-04
Use Elixir 1.20.0 and Erlang 29.0.1

joshk:


kybishop (approved):
Nice! Love it when these don't require any changes :)

PR #2720, opened 2026-06-06
Add a UI based log in flow when using the CLI

joshk:
This adds the supporting API for a UI based CLI login flow

https://github.com/user-attachments/assets/a07f32a4-546d-42b1-be84-8631a513c6c9

Short lived CLI tokens are stored in Mnesia and cleaned out after 5 mins.

The API endpoint used by the CLI to check on the auth status is rate limited based on IP, 20 requests every 60 seconds. PubSub is used to sync the rate limiting across the cluster.


lawik:
Stuff looks great to me.

Because this is login code I asked for some extra review via Claude. The review found something I completely missed thinking about:

The CLI login approval endpoint is a GET that performs a state-changing, security-critical action with no consent step and no CSRF protection. get("/auth/cli/:token", SessionController, :cli) (router.ex:214) is gated only by :require_authenticated_user (cookie-based). SessionController.cli/2 (session_controller.ex:73) calls Accounts.verify_cli_session_token(user, token) (accounts.ex:1087), which, for a :waiting session, immediately mints a full API token via create_user_api_token(user, "") and binds it to the session record (status: :ready, user_id: user.id). The rendered cli.html.heex is a static "You're all set" page — there is no confirmation button, no POST, no CSRF token, and no user_code for the user to compare. Phoenix's protect_from_forgery plug does not validate GET requests, so the approval is fully forgeable. The session cookie uses the default SameSite=Lax, which still sends the cookie on top-level cross-site navigations (a clicked link or redirect).
- Exploit Scenario:
  a. Attacker calls the unauthenticated POST /api/auth/cli_session (user_controller.ex:68, security: []) and receives {token, url} where url = https://<host>/auth/cli/<token> and token is a UUID the attacker now knows.
  b. Attacker lures a logged-in NervesHub user to navigate to that approval URL (link in email/Slack, attacker-controlled page redirect, window.open). Under SameSite=Lax the victim's session cookie is sent on the top-level GET.
  c. The GET silently mints the victim's API token and marks the session :ready — no interaction beyond loading the page.
  d. Attacker polls the unauthenticated GET /api/auth/cli_session/:token (check_cli_session, user_controller.ex:87, security: []), which returns the full user_token.

Result: complete account takeover with no credential phishing. The rate-limit Attack plug on the poll endpoint does not gate disclosure.
- Recommendation:
  - Require explicit, intentional consent: render a confirmation page on GET and mint/bind the token only on a POST carrying the Phoenix CSRF token, so protect_from_forgery applies.
  - Display what is being approved (requesting client/IP/user-agent) and a short verification code that the CLI also shows, so the user confirms the codes match (as with the OAuth 2.0 Device Authorization Grant user_code).
  - Consider a separate CLI-held device secret (returned only in the original POST response, never in the browser URL) required when polling, so an approved session cannot be drained by a third party.

joshk:
@lawik thanks for the feedback, or should I be thanking Claude? :D

I've added a confirmation code, for visual verification, and also require the CLI session be confirmed.

I've also updated the screencast in the PR description.

lawik:
Awesome!

lawik (approved):

PR #2735, opened 2026-06-11
A full pass addressing the missing pieces of the OpenAPI specs

joshk:
(built on top of #2733, which Jon will be reviewing soon)

This completes the missing pieces of the OpenAPI spec, as well as fixing a bunch of issues and inconsistencies.


lib/nerves_hub_web/controllers/api/ca_certificate_controller.ex
```diff
@@ -5,11 +5,16 @@ defmodule NervesHubWeb.API.CACertificateController do
   alias NervesHub.Certificate
   alias NervesHub.Devices
   alias NervesHub.Devices.CACertificate.CSR
+
```
joshk:
```suggestion
```

joshk: A small formatting warning addressed

lawik (approved):
Looks like specs 🤷

PR #2740, opened 2026-06-14
Remove Memento and switch to ETS + PubSub

joshk:
Memento required manual, and difficult, clustering configuration.

This switches to ETS + PubSub + a little bit of syncing logic.


nshoes (commented):
What specifically was difficult about Momento configuration? I don't mind baking something simple on our own, but I don't understand the push to switch.

joshk:
> What specifically was difficult about Momento configuration? I don't mind baking something simple on our own, but I don't understand the push to switch.

the configuration was not documented, the one GH issue about cluster config ended with the person saying they couldn't get it working.

Since I couldn't understand it, I went for something i could understand, and didn't require an extra library.

nshoes (approved):
LGTM 🥇

PR #2762, opened 2026-06-18
A few more tweaks noticed after deploying

joshk:
oh, and a new user collapsible sidebar

https://github.com/user-attachments/assets/4033d146-10ea-431d-aa5c-4f04690cdbe8


assets/js/hooks/sidebarToggle.js
lawik, line 1: Could it have been a LiveView.JS command or does the fact that it is in the layout prevent that?

lawik (approved):

assets/js/hooks/sidebarToggle.js
joshk: LV JS command requires going to the server and back, this is all client side.

Plus this code is mostly for storing the state in local storage.

PR #2771, opened 2026-06-20
Add tag matching operator to Deployment Group matching conditions

### LINKED ISSUE #2650: UI/UX - Deployments use of tags is "OR" but it would be really useful to have "AND"
Device tags are the primary way to assign devices to deployments.

Tags are also the only way to categorize devices. 

For example, tags can be used to categorize,
- by pipeline: production, staging, qa, sandbox, etc
- by deployment use case: canary, beta_testing, deploy_last
- by region: europe, north_america, asia, anzac, etc
- by device type: type1, type2, etc.

Unfortunately, given the "OR" nature used for assigning tagged devices to deployments, it can get very confusing to make sure a device auto-assigns to the correct deployment group. For example, if there is a deployment group for `production` and a deployment group for `production europe`, a device tagged `production europe` matches both. 

Pre- NervesHub 2.0 and Deployment Groups, we frequently ended up with ping pong deployments due to the OR nature of the old deployments and tags. Groups assignment helps but does not eliminate the issue.

It would be nicer user experience, and dare I say "more intuitive", if there was maybe a check box or some other affordance to change the matching mode from "OR" to "AND". That way one could define deployment groups as,

- `production europe` = all production devices in Europe
- `production type1` = all type1 production devices
- and still allow the default `production` = all devices tagged with only with `production`

joshk:
resolves #2650

### DIFF
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -723,9 +723,30 @@ defmodule NervesHub.ManagedDeployments do
     |> where(
       [d],
       fragment(
-        "?->'tags' <@ to_jsonb(?::text[]) or (jsonb_array_length(?->'tags') = 0 and ?::text[] is null)",
+        """
+        (
+          COALESCE(?->>'tag_operator', 'and') = 'and'
+          AND (
+            ?->'tags' <@ to_jsonb(?::text[])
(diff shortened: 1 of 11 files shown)


lawik (approved):

nshoes:
Love this!

PR #2777, opened 2026-06-21
Move network interface tracking to the DeviceConnection schema

joshk:
This gives us a history of interface usage for a device

### DIFF
--- a/priv/repo/migrations/20260620235408_update_latest_device_connections_table.exs
+++ b/priv/repo/migrations/20260620235408_update_latest_device_connections_table.exs
@@ -9,8 +9,7 @@ defmodule NervesHub.Repo.Migrations.UpdateLatestDeviceConnectionsTable do
       add(:lib, :string)
       add(:lib_version, :string)
 
-      add(:interface, :string)
-      add(:available_interfaces, {:array, :string})
+      add(:network_interface, :string)
     end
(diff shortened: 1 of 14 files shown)


priv/repo/migrations/20260620235408_update_latest_device_connections_table.exs
```diff
@@ -9,8 +9,7 @@ defmodule NervesHub.Repo.Migrations.UpdateLatestDeviceConnectionsTable do
       add(:lib, :string)
       add(:lib_version, :string)
 
-      add(:interface, :string)
-      add(:available_interfaces, {:array, :string})
```
lawik, line 13: no more available interfaces? Seems like that is probably helpful for figuring out why a device might be on the "wrong" interface, ie, signal that something expected is missing.

lawik (approved):
One Q, but looks good

priv/repo/migrations/20260620235408_update_latest_device_connections_table.exs
joshk: True, but we don't have that data being sent at the moment, I added it in a different PR but haven't hooked it up. I thought its best to leave that for another PR since it need changes to DeviceChannel, and NHL

PR #2778, opened 2026-06-22
Drop dead connection columns from devices table

### LINKED ISSUE #1819: Remove `device.connection_metadata` to `device_connection.metadata`
We are moving away from storing connection state on a `device`. We only use `device.connection_metadata` for location-related features; that work should be moved to save/update/read/query off of `device.latest_connection.connection_metadata` instead.

Original conversation in https://github.com/nerves-hub/nerves_hub_web/pull/1807.

mvanhorn:
## Summary

Adds a single Ecto migration that drops the dead `connection_metadata` and `connection_types` columns from the `devices` table, along with the now-unused `connection_type` PostgreSQL enum.

## Why this matters

Per #1819, connection state has moved off the `device` record. Connection metadata now lives on `device_connection.metadata` (DeviceConnection), and the device schema fields for these two columns are already commented out in `lib/nerves_hub/devices/device.ex`. No live code reads or writes the `devices` table columns: the remaining `connection_types` and `update_connection_metadata` references in `device_channel.ex` and `device_link.ex` operate on the new DeviceConnection path, and `device_filtering.ex` reads `latest_connection.metadata["connection_types"]`, not the device column. As joshk noted on the issue, all that remains is to delete the columns from the DB.

## Changes

- New migration `priv/repo/migrations/20260622000000_remove_connection_columns_from_devices.exs`:
  - Removes `connection_metadata` and `connection_types` from the `devices` table.
  - Drops the `connection_type` enum that only existed to back the `connection_types` array column.
- Uses reversible `remove/3` calls and an `execute/2` with both up and down SQL, mirroring the original creating migrations (20240707230955 and 20230319193211), so `mix ecto.rollback` restores both columns and the enum.

No schema or runtime code change is needed: the schema fields are already commented out and there are no live readers or writers of these columns.

## Testing

- `mix ecto.migrate` applies cleanly; both columns and the enum are removed.
- `mix ecto.rollback` restores both columns and recreates the `connection_type` enum.
- Existing test suite is unaffected since no code path depends on the dropped columns.

Fixes #1819


joshk:
Thanks for the PR. If you look at the Device schema you will notice a few other connection related columns which are also scheduled to be removed, can you add these to the migration and remove the related comments from the schema.

mvanhorn:
Done in 1f1edeb - added connection_status, connection_established_at, connection_disconnected_at, and connection_last_seen_at to the migration (matching their original `:string` / `:utc_datetime` types) and removed the deprecated-field comment block from the Device schema. Removing connection_status also drops its index from AddConnectionStatusIndexToDevices automatically. I confirmed the app reads connection state from the latest_connection (DeviceConnection) association rather than these columns, so nothing else references them.

nshoes:
@joshk this looks good to me, any more thoughts?

joshk (approved):
LGTM

mvanhorn:
Cheers @joshk, thanks for merging the dead-column cleanup on the devices table.

joshk:
Thanks for the PR @mvanhorn

PR #2779, opened 2026-06-22
Fix concurrency issue with CLI Session cache

lawik:
Claude flagged a concurrency issue in the new cache mechanism that was covered by mnesia transaction support previously.

This seems to make sense.


nshoes (approved):
This makes sense to me

joshk (approved):

PR #2785, opened 2026-06-23
Show devices with flapping connections on the Insights page

joshk:
image

Flapping connections are when devices make 10 or more connections per hour.


nshoes (approved):
LGTM! Let the flappers be known

PR #2791, opened 2026-06-24
Write firmware uploads directly to the temp dir

mcrumm:
This change uses the [`Phoenix.LiveView.UploadWriter`](https://phoenix-live-view.hexdocs.pm/Phoenix.LiveView.UploadWriter.html) behaviour to avoid double-writing to temp when uploading firmware.

Note that there is an opportunity to better handle errors from the writer process, but I wanted to get feedback on change before taking it any further. Thanks!


joshk:
Hey @mcrumm 

This is a smart approach, I feel silly for not thinking of this. The double write always annoyed me.

Since `FirmwareUploadWriter` doesn't have anything specific to firmware in its code, it might be better to call this `BrieflyUploadWriter`.

Please do continue with this magic.

joshk:
@nshoes this is ready for review. I think its good to merge, but since I've worked on it I think you should review.

mcrumm:
@joshk Fwiw those are roughly the same changes I was going to make re: the error handling. Sorry for the delay!

joshk:
@mcrumm nothing to apologise for, thanks for creating the PR to start the discussion.

joshk (approved):

PR #2792, opened 2026-06-25
A query syntax for advanced device filtering

joshk:
This allows for advanced device list filtering, including brackets, with a nice UI and suggestions.

https://github.com/user-attachments/assets/07568311-5aa9-4e69-9f89-84b3edfd3242

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -32,6 +32,7 @@
   ],
   "words": [
     "animatedelay",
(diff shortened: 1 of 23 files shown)


lawik (approved):
Eyes kind of glazed over in terms of reviewing. But love the feature and the sensitive bits looked like they were properly parameterized :)

PR #2794, opened 2026-06-25
Rework how Support Scripts are selected and run from the Device page

joshk:
This squashes a UI bug, and deals with long support script lists better 
(combo select and autocomplete when more than 10 scripts)

https://github.com/user-attachments/assets/6200364b-c9ec-4d27-afb0-f40cd448fc23

https://github.com/user-attachments/assets/bb5f359a-5f86-4ac9-b818-e59ca0444f54


nshoes:
Could we get a screen cap of this in action?

nshoes (approved):

PR #2807, opened 2026-07-05
Support login redirect to related NervesHub URLs

joshk:
The enables related NervesHub apps to be connected to the running platform, hooking into auth and redirecting correctly.


nshoes (approved):
LGTM 💯

PR #2822, opened 2026-07-15
feat: add deployment group notes

### LINKED ISSUE #2821: feat: add deployment group notes
### Problem: 

Often times a deployment group might get created for a purpose (which might be short lived) but then we don't have a way to capture this, as a result, as devices move around you might end up with (empty) deployment groups where its name might not suffice to know important related context of its purposes

### Proposed solution: 

Add an optional text `notes` field on deployment groups to capture the group creation purpose just so users can use it to help to remember, which can be helpful when deciding to delete a group or to remember why it exists in the first place

viniarck:
Closes #2821 

### Local tests

- Made sure that the introduced optional `<input>` that sets deployments notes is minimalistic by default:

Screenshot 2026-07-15 at 2 36 52 PM

- Explored locally making sure that a deployments notes with 1000 characters would still render nicely:

Screenshot 2026-07-15 at 2 37 25 PM

- Explored creation/update, and removal (by setting to empty):

Screenshot 2026-07-15 at 2 49 24 PM

### DIFF
--- a/lib/nerves_hub/managed_deployments/deployment_group.ex
+++ b/lib/nerves_hub/managed_deployments/deployment_group.ex
@@ -62,6 +62,7 @@ defmodule NervesHub.ManagedDeployments.DeploymentGroup do
     field(:healthy, :boolean, default: true)
     field(:penalty_timeout_minutes, :integer, default: 1440)
     field(:connecting_code, :string)
+    field(:notes, :string)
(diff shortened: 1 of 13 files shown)


nshoes (approved):
LGTM! CI reporting mix dependecies failed, but isn't blocking.

PR #2829, opened 2026-07-24
Add fallback for device search filters to do broad search

lawik:
Video shows the new functionality. We've had this feedback from a few different places now and it hits a major workflow with an inconvenient regression so I feel we should fall through to searching.

https://github.com/user-attachments/assets/67c8d3b8-4d6d-4358-853a-daf6b2c3af87


joshk:
we already have an 'identifier' category, why not reuse that?

lawik:
> we already have an 'identifier' category, why not reuse that?

Because I want to search more fields :)

nshoes (approved):
Works for me 😎

PR #2830, opened 2026-07-24
Improve keyboard behavior in the advanced device search

lawik:
Enter and Tab now complete suggestions the way you'd expect from a command palette, instead of requiring arrow-key navigation first:

- Tab accepts the highlighted suggestion, or the first one
- Enter does the same while mid-word (a non-empty prefix), so typing "hea" + Enter gives "health" rather than running a failed search
- Enter on a structurally incomplete query (mid-term, dangling and/or, unclosed paren) advances to the next slot and shows its suggestions - e.g. out of a finished "like" operator and into the value - instead of submitting a query the server is guaranteed to reject
- Enter on a trailing unterminated string closes the quote and submits
- A finished query still submits with Enter even while the "and"/"or" hints are showing; Escape-dismissed suggestions are never accepted


.github/workflows/ci.yml
```diff
@@ -134,6 +134,10 @@ jobs:
       - name: Run Credo
         run: mix credo --min-priority low
 
+      # Dependency-free: runs on Node's built-in test runner, no npm install.
+      - name: Run JS unit tests
+        run: npm test --prefix assets
```
nshoes, line 139: TIL! This is cool

nshoes (approved):
Tested locally, LGTM

PR #2847, opened 2026-08-04
Decorate device penalty box logging with useful info

nshoes:
This will allow us to create useful graphs and alerts with observaiblity tools until we get penalty box introspecition from the Deployment Workflows feature. I didn't add a test because `NervesHub.DevicesTest` has multiple test scenarios that put a device in a penalty box. Preload work was done inline because context of callers differs greatly.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -1211,7 +1211,17 @@ defmodule NervesHub.Devices do
     :ok = DeviceTemplates.audit_firmware_upgrade_blocked(deployment_group, device)
     _ = FirmwareUpdates.clear_inflight_update(device)
 
-    Logger.info("Device #{device.identifier} put in penalty box until #{blocked_until}")
+    device =
+      Repo.preload(device, [:org, :product, :current_device_firmware, deployment_group: [current_release: :firmware]])
+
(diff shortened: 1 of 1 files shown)


kybishop (approved):

lib/nerves_hub/devices.ex
```diff
@@ -1211,7 +1211,17 @@ defmodule NervesHub.Devices do
     :ok = DeviceTemplates.audit_firmware_upgrade_blocked(deployment_group, device)
     _ = FirmwareUpdates.clear_inflight_update(device)
 
-    Logger.info("Device #{device.identifier} put in penalty box until #{blocked_until}")
+    device =
+      Repo.preload(device, [:org, :product, :current_device_firmware, deployment_group: [current_release: :firmware]])
+
+    Logger.info("Device #{device.identifier} put in penalty box until #{blocked_until}", %{
```
rraub, line 1217: This is missing context to answer **why** it is being put in the penalty box, but its still a step in the right direction. Thank you for adding this!

nshoes: Oh! Good point, how about this: https://github.com/nerves-hub/nerves_hub_web/pull/2849

PR #2849, opened 2026-08-04
Update penalty box logging

nshoes:
Adds the reason for putting a device in the penalty box.


nshoes:
@rraub we could add what the actual rate and threshold are on the deployment group, but I'm not sure if that has any value when we can just reference the deployment group if needed.

rraub:
> @rraub we could add what the actual rate and threshold are on the deployment group, but I'm not sure if that has any value when we can just reference the deployment group if needed.

I am mostly interested in the mode, which this PR adds! The specific rate and threshold values are interesting but not super useful in the log. This lets us clearly see if this was a fast failure (rate) or a slow failure (threshold). Thank you for adding this!

kybishop (approved):

PR #2850, opened 2026-08-06
Add deployment group lock to prevent device reassignment

nshoes:
This feature will help alleviate the opaque behavior that is device deployment group membership on connect. Sometimes, you don't want any devices in or out automatically. This PR does that, as well as:

- Don't consider locked deployment groups when looking for an appropriate group for a device; and a test
- Refactor `ManagedDeployments.verify_deployment_group_membership`, it's needed it for awhile (my bad)
- Claude wrote a function doc for `ManagedDeployments.matching_deployment_groups/2`. I'm getting tired of re-reading that SQL to remember what it does every 3-4 months 😆


kybishop (approved):
lgtm

PR #2851, opened 2026-08-07
Use the `:group` library for targeted device -> UI PubSub

joshk:
## Summary

Moves several device- and web-scoped pub/sub flows off cluster-wide `Phoenix.PubSub` broadcasts onto the [`:group`](https://hex.pm/packages/group) library, so messages are delivered only to the nodes that actually care. `Phoenix.PubSub` fans every message out to every node and filters locally; `Group.dispatch/3` delivers only to nodes with a joined member for the key, and named clusters keep web-only membership state off the (more numerous) device nodes.

Throughout, events are dispatched as `%Phoenix.Socket.Broadcast{}` structs identical to what PubSub delivered, so receiver `handle_info(%Broadcast{...})` clauses are unchanged — only the subscribe/broadcast call sites move.

## Changes

**Device → UI events (`internal:device:<id>`)** — default cluster.
Connection status, heartbeats, firmware-update progress, location, and log lines now dispatch to the LiveViews/channels watching a device via a small `NervesHub.Devices.PubSub` wrapper (`subscribe`/`unsubscribe`/`broadcast`). Quiet devices generate no cross-node traffic. Uses the default cluster because senders run on device nodes and watchers on web nodes.

**Rate limiting + CLI session cache** — named `"web"` cluster.
`RateLimitPubSub` (PlugAttack IP-throttle sync for `check_cli_session`) and `CLISessionCache` only matter to web nodes, so they move to a named `"web"` cluster that device nodes don't join. `Group.dispatch` has no self-exclusion, so each stamps the origin node and skips its own echo (`origin == node()`). A reintroduced `GroupClusterConnection` connects non-device nodes to `"web"`, and both processes are gated off device nodes. `CLISessionCache` startup warm-up now pulls from actual `"web"` group members (not `Node.list()`, which could hit a device node), deferred and retried to cover async membership propagation.

**Deployment orchestrator events (`orchestrator:deployment:<id>`)** — default cluster.
Exactly one `Orchestrator` per deployment (held by ProcessHub) consumed this topic, yet every device-node event fanned out cluster-wide. It now `Group.join`s the key and `DeploymentOrchestratorEvents` dispatches to it, so events reach only the orchestrator's node. Default cluster because senders are device nodes; `join` (not `register`) is used so ProcessHub keeps owning the singleton guarantee while the test suite can still co-observe events. The multi-subscriber `deployment:<id>` topic stays on `Phoenix.PubSub`.

**Console + local shell terminal I/O** — default cluster.
The device console and local-shell byte streams (keystrokes, output, file transfer) plus their liveness/connect control-plane move behind a `NervesHub.Consoles.PubSub` wrapper. Terminal I/O is now delivered only to the nodes holding the paired channels rather than every device node, and `Tracker.console_active?` / the shell tab's `shell_active?` become `Group.members(...) != []` instead of a 500ms broadcast-and-wait (also removing the console/shell tab load stall). Default cluster because the device-side channels run on device nodes and the user channels/LiveViews on web nodes.

**Device extensions events (`device:<id>:extensions`)** — default cluster.
The per-device extensions channel (health/geo config in, health reports out) moves behind a `NervesHub.Extensions.PubSub` wrapper. The topic is **split into two keys by direction** — `device:extensions/<id>` (web→device: `health:check`/`attach`/`detach`, consumed by the device-side `ExtensionsChannel`) and `device:extensions:reports/<id>` (device→web: `health_check_report`, consumed by the Show LiveView's HealthTab/DetailsTab hooks). The split preserves the old `broadcast_from!` self-exclusion structurally: the channel originates the report but is not a member of the reports key, so it never receives (and pushes back to the device) its own report. Product-wide `product:<id>:extensions` stays on `Phoenix.PubSub` (dense fan-out to every device's extensions channel — no targeted-dispatch win, and moving it would trade a rare broadcast for continuous per-connect membership churn) but is routed through the same `Extensions.PubSub` wrapper for consistency.

**Sparse UI fan-out topics** — default cluster.
The product/firmware UI-facing topics move to targeted dispatch; their only consumers are LiveViews open for a product (all on web nodes), so PubSub was fanning every event to all device nodes to be discarded. `product:<id>` (`firmware/created`, `firmware/deleted`) gets a new `NervesHub.Products.PubSub` whose `broadcast_from/3` is **self-excluding** — the publisher is itself a subscriber (`create_firmware`/`delete_firmware` run inside the Firmware LiveView), which already refreshes/flashes directly, so it must not also receive the broadcast; since `Group.dispatch` has no self-exclusion we send to `Group.members` minus `self()`. `firmware:<id>` (delta build status) and `product_notifications:<id>` (`created`/`dismissed`) convert their existing context wrapper functions to `Group.join`/`dispatch` (plain — the publisher isn't a subscriber, or the subscriber relies on self-receipt). `product:<id>:devices` is orphaned (a devices-index subscription with no publisher) and left untouched.

## Cluster model

- **Default cluster** for cross-role flows (device-node sender → web-node consumer): device→UI, orchestrator events, console/local-shell I/O, device extensions events, and the sparse product/firmware UI fan-outs.
- **Named `"web"` cluster** for web-only state (rate limiting, CLI session cache), keeping that membership off device nodes.

## Testing

`mix format` / `mix credo --min-priority low` / `mix dialyzer` clean; full suite green apart from unrelated pre-existing failures (two `UpdateToolTest` cases that fail identically on `main` due to local `mtools`, and a ClickHouse-backed `connection_history_test` eventual-consistency flake that passes on re-run/in isolation). Adds targeted tests for the device→UI wrapper, rate-limit sync, the CLI cache warm-up self-exclusion, a `NervesHub.Consoles.PubSub` suite (registry auto-leave, connect handshake, byte relay, file-data transfer), and a `NervesHub.Extensions.PubSub` suite asserting the direction split (report not echoed to the device key; report subscriber does not receive web→device commands) and the product-wide path (delivery to subscribers, caller self-exclusion preserved). Adds a `NervesHub.Products.PubSub` suite asserting the self-excluding `broadcast_from` (delivers to other subscribers, never to the caller). Preserves the orchestrator behaviour tests.


nshoes (approved):
Tested thoroughly in our QA, did not see any issues 🚀

nshoes:
Retesed on our QA after the rebase, LGTM. I'm going to take a human-pass on the changes just to make sure Claude didn't add anything funky; I already asked but I don't trust it 😆

nshoes (approved):

PR #2852, opened 2026-08-07
Split dependency submission and review into separate CI jobs

joshk:
## Problem

The `report_mix_deps` job runs `erlef/mix-dependency-submission`, which submits the dependency graph via GitHub's Dependency Submission API and therefore needs `contents: write`. The job ran on **every** event, including `pull_request`.

For pull requests **from forks**, GitHub grants the workflow a read-only `GITHUB_TOKEN` regardless of the job's `permissions:` block (a deliberate security measure for untrusted forks). So the submit step 403s and the job fails for any contributor outside the org. Org members don't hit it because their PRs are branches on the repo, where the token can be write.

## Fix

Split the single job into two least-privilege jobs:

- **`submit_deps`** — runs on non-PR events (`push` to `main`/tags, `workflow_dispatch`), `contents: write`, submits the dependency graph. The graph only needs updating from the default branch, and fork PRs can't submit it anyway.
- **`dependency_review`** — runs on `pull_request` only, `contents: read`, reviews the PR's dependency changes. It's read-only, so it works on fork PRs.

## Behaviour

- **Fork PR:** `submit_deps` is skipped; `dependency_review` runs read-only → green (unblocks it as a required check).
- **push to `main` / tag / manual dispatch:** `submit_deps` runs with the write token it needs, keeping the graph current.

Note: `submit_deps` is gated `!= 'pull_request'` (rather than strictly `== 'push'`) so it still runs on tag pushes and manual dispatch — the original non-PR triggers — and only PRs are excluded. Easy to tighten to push-only if preferred.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,8 @@ jobs:
       - uses: actions/dependency-review-action@v4
 
   compile-and-test:
+    name: "Compile & Test"
+
     runs-on: ubuntu-latest
     timeout-minutes: 14
 
@@ -193,6 +195,8 @@ jobs:
         run: mix test --include mtools
 
   build-and-publish:
+    name: "Build & Publish Docker image"
+
     runs-on: ubuntu-latest
 
     needs: compile-and-test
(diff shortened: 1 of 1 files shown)


nshoes:
Ah! Nice, ty, and ty Claude

PR #2863, opened 2026-08-09
Add per-product setting to require unique firmware versions

joshk:
The final PR in the firmware version-handling stack. **Base is #2858** (current stack tip), so this sits on top of the whole chain.

## What & why
Adds a `require_unique_firmware_version` boolean setting on products. When on, uploading firmware whose version is already used **for the same platform + architecture** in the product is rejected with a changeset error on `:version`.

**Defaults (per Josh's ask — off for existing, on for new):**
- Migration column default `false` → existing products unaffected.
- Schema field default `true` → products created from now on require unique versions. `create_product/1` picks this up; test fixtures opt out via `product_params/0` so they can keep creating multiple firmwares sharing a version (the real "on for new" default is asserted directly in `products_test`).

**Uniqueness scope: per (product, platform, architecture, version).** Multi-platform products legitimately reuse a version across platforms (1.0.0 for rpi4 *and* rpi0), so only a same-target re-upload is blocked. (Flagged this in the channel; building on that lean — easy to tighten to product-wide if you'd prefer.)

**Why app-level, not a DB unique index:** the rule is conditional on a setting that lives on the `products` table, which a partial unique index on `firmwares` can't reference. Enforced in `create_firmware/3` inside the existing transaction.

## Files
- `products` migration + `Product` schema field/changeset; `Products.update_product/2`.
- `Firmwares.create_firmware/3`: `validate_version_uniqueness/1` step + `firmware_version_taken?/1`.
- Product → Settings LiveView + template: a "Require unique firmware versions" toggle (a real product setting, separate from device extensions).
- `product_params/0` fixture defaults the setting off; one delta test (`devices_test`) that creates products directly + uploads same-version firmware opts out explicitly.

## Testing
- New tests: create-time default (`true`), opt-out at creation, `update_product/2` toggle, upload enforcement (reject same target / allow across platforms / allow when off), and the settings-toggle render state.
- Full suite: **1417/1419**, the only 2 failures being the pre-existing `UpdateToolTest` fwup `same_fat_files?` environment failures (reproduce on the base branch). `mix format` clean on changed files; compiles `--warnings-as-errors`. (`mix credo --strict` flags only the pre-existing missing `@moduledoc` on `NervesHub.Firmwares` / `NervesHub.Products`.)

## Review focus
- **Security/robustness:** new upload-time validation over firmware metadata; the check runs inside the create transaction. There's a small TOCTOU window (two concurrent uploads of the same version could both pass) — acceptable for firmware uploads, and noted here rather than hidden.
- **Concurrency default note:** confirm the uniqueness scope (per platform+arch vs product-wide) matches intent.
- The migration-default-`false` / schema-default-`true` split is what delivers "off for existing, on for new" — worth a sanity check.

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -332,6 +332,7 @@ defmodule NervesHub.Firmwares do
     Repo.transact(
       fn ->
         with {:ok, params} <- build_firmware_params(org, filepath),
+             :ok <- validate_version_uniqueness(params),
              {:ok, firmware} <- insert_firmware(params),
              :ok <- upload_file_2.(filepath, firmware.upload_metadata) do
           {:ok, firmware}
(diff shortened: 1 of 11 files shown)


joshk:
Superseded by WALL-E's PR7 for the same feature — we both picked this up within ~12s of each other, and WALL-E's is the implementation under review (and matches Josh's preference to put the check in the Firmware changeset rather than the context). Closing this duplicate so the stack keeps a single top.

PR #2864, opened 2026-08-09
Add per-product setting requiring unique firmware versions

### LINKED ISSUE #2697: Feature: Add setting for preventing duplicate firmware version numbers
Based on #2403

joshk:
## What

Adds a per-Product setting, `require_unique_firmware_version`. When enabled,
uploading firmware whose version already exists for the product **(same platform
and architecture)** is rejected. Off for existing products, on for new ones.

Last PR in the firmware-version-handling stack.

## Behaviour

- **Setting default:** the DB column defaults to `false` (existing products keep
  today's behaviour), while the schema field defaults to `true`, so products
  created through `Product.changeset/2` (i.e. new products) get it on.
- **Scope:** uniqueness is per `(product, platform, architecture)`. The same
  version built for different targets (e.g. `1.2.0` for `rpi3` and `rpi4`) is
  still allowed; re-uploading `1.2.0` for the same target is rejected. Firmware
  version comes from the signed fwup metadata.
- **Enforcement:** a `validate_unique_version/2` step in
  `Firmware.create_changeset/2` adds a `:version` error
  (`"has already been taken for this product"`) when the product requires
  uniqueness (flag threaded through params from the product setting) and a
  matching firmware exists. Keeping it in the schema changeset matches the repo
  convention (`DeploymentGroup`/`DeviceCertificate` changesets already query
  `Repo`). The API and LiveView upload surface it the same way they already
  surface the "uuid taken" error; a generic changeset clause was added to the
  upload LiveView so the message shows instead of the generic "Unknown error".
- **UI:** a "Require unique firmware versions" toggle on the product settings
  page (mirrors the existing extension toggles; gated on `product:update`).

## Notes

- **Application-level check, not a unique index.** The requirement is a
  per-product toggle, so a plain/partial unique index can't express it
  (it can't reference the product's setting). This leaves a small TOCTOU window
  for two *truly-concurrent* uploads of the same version, which the check does
  not close — documented in a code comment. Concurrent same-version uploads for
  one product are rare in practice.
- **Test fixtures** default the setting to `false` (`product_params/0`) so the
  many tests that create several firmwares per product are unaffected;
  enforcement is covered by explicit tests that set it `true`.

## Tests

- `firmwares_test.exs`: rejects a duplicate version when required; allows the
  same version for a different platform; allows duplicates when not required.
- `products_test.exs`: new products default the setting to `true`;
  `update_product/2` toggles it.
- `settings_test.exs`: the toggle renders on the product settings page.
- Full `mix check` substantive steps + test suite green (same pre-existing
  fwup/spellcheck non-issues noted in the earlier stack PRs).

### DIFF
--- a/lib/nerves_hub/firmwares.ex
+++ b/lib/nerves_hub/firmwares.ex
@@ -857,7 +857,9 @@ defmodule NervesHub.Firmwares do
   defp resolve_product(params) do
     case Products.get_product_by_org_id_and_name(params.org_id, params.product_name) do
       {:ok, product} ->
-        Map.put(params, :product_id, product.id)
+        params
+        |> Map.put(:product_id, product.id)
+        |> Map.put(:require_unique_firmware_version, product.require_unique_firmware_version)
 
       _ ->
         params
--- a/lib/nerves_hub/firmwares/firmware.ex
+++ b/lib/nerves_hub/firmwares/firmware.ex
@@ -2,12 +2,14 @@ defmodule NervesHub.Firmwares.Firmware do
   use Ecto.Schema
 
   import Ecto.Changeset
+  import Ecto.Query
 
   alias __MODULE__
   alias NervesHub.Accounts.Org
   alias NervesHub.Accounts.OrgKey
   alias NervesHub.ManagedDeployments.DeploymentRelease
   alias NervesHub.Products.Product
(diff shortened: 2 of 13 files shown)


nshoes (approved):
We've talked about this one for awhile, exciting!

PR #2872, opened 2026-08-11
Add CMD-K command palette for logged-in users

joshk:
## What

Adds a keyboard-driven command palette (**Cmd/Ctrl+K**) available on every authenticated page for quick search and navigation.

- **Search** devices by `identifier`, deployment groups by `name`, and firmware by `uuid`.
- **Navigate** via static commands (Devices, Firmware, Deployment Groups, Settings, org/product pages) filtered by the query.

## Scope-aware search

Results are scoped to where the user currently is, derived from `@current_scope`:

| Where the user is          | Entity search covers                    |
|----------------------------|-----------------------------------------|
| Inside a product           | that product only                       |
| Inside an org (no product) | all products in that org                |
| Main dashboard (no org)    | every product the user belongs to       |

In every case results are constrained to products the user is a member of, so the palette never surfaces entities from orgs the user can't otherwise see.

## How it's built

- **`NervesHub.CommandPalette`** — a context with a single `search/3` entry point returning grouped results. It reuses the existing device `identifier` trigram index (fast substring, even cross-org) and uses an index-friendly `uuid` prefix match for firmware.
- **`NervesHubWeb.Components.CommandPalette`** — a LiveComponent rendering the overlay. Open/close state is **server-owned** so re-renders (fired on every keystroke) don't clobber it. Entity results load via **`assign_async`** so the DB lookups never block the LiveView; the static nav commands don't hit the DB and render immediately.
- **`assets/js/hooks/commandPalette.js`** — a hook owning the global Cmd/Ctrl+K shortcut and in-palette arrow/enter/escape keyboard navigation.
- Mounted once in both authenticated shells (`sidebar` and `no_sidebar` layouts).

## Tests

- Context tests for each scope band (product / org / dashboard), cross-org isolation, firmware uuid prefix, and soft-delete exclusion.
- LiveComponent render tests (correct deep links, scope-gated commands).
- PhoenixTest UI tests driving search → scoped results → correct destinations.
- JS unit tests for the hook's keyboard-decision helpers (`node --test`).

Also scopes two bare `"form"` selectors (deployment group + certificate authority tests) now that the palette adds a search form to the layout.

Full suite: **1411 passed**, JS: **13 passed**.

https://github.com/user-attachments/assets/e1fd49a8-6f25-4814-b727-5508f7984c68


nshoes (approved):
Oh yeah that's nice 🤤

PR #2873, opened 2026-08-11
Context cleanup — Phase 1: dead code, spec fixes, auth hardening

joshk:
First in a **stacked series** of PRs cleaning up the `NervesHub` context layer. This phase is deliberately mechanical and low-risk (dead-code removal + spec/naming fixes + two small correctness/encapsulation fixes). Later phases (deduplication, encapsulation, module splits) will branch on top of this one.

### Dead code removed (grep-verified: zero call sites in `lib/` and `test/`)
| Module | Removed |
|---|---|
| `Accounts` | `get_org_with_org_keys/1`, `get_org_by_name_and_user!/2`, `user_in_org?/2` |
| `ManagedDeployments` | `deltas_processing?/1`, `preload_firmware_and_archive/1`, `preload_with_firmware_and_archive/1` |
| `Firmwares` | `get_firmware_by_product_and_uuid!/2`, `unsubscribe_firmware_delta_target/1` |
| `Devices` | `preload_product/1`, `get_minimal_device_location_by_product/1` (+ now-unused `DeviceConnection` alias) |
| `Products` | `get_product_by_org_id_and_name!/2` |
| `AuditLogs` | single-arg `logs_for_feed/1` (only the `/2` clause is used) |

### Misleading `@spec` / naming fixes
- `Scripts.filter/2` and `Archives.filter/2` declared `{[Product.t()], …}` but return Scripts/Archives.
- `Firmwares.get_firmware_delta_by_source_and_target` spec was missing its 3rd (`status`) arg.
- `ManagedDeployments.get_device_count` spec was `term() | nil` for a count → `non_neg_integer()`.
- `Devices.clean_up_soft_deleted_devices`: local `two_weeks_ago` actually held a `-12 day` value → renamed `twelve_days_ago` (no behaviour change; if 14 days was intended, change the value in a follow-up).

### Correctness
- **`Accounts.get_user_by_email_and_password/2` skipped `Repo.exclude_deleted`**, unlike the sibling auth paths (`authenticate`, `get_user_by_email`), allowing a soft-deleted user to authenticate. Now excludes deleted users.

### Encapsulation
- Added the missing **`Uploads.delete/1`** facade (the `@callback` existed with no public wrapper) and routed `Workers.DeleteArchive` through it instead of reaching into `Application` config for the backend.

### Deliberately deferred
[...]

PR #2874, opened 2026-08-11
Context cleanup — Phase 2: deduplication (Repo.fetch + extension settings)

joshk:
Second in the **stacked** context-cleanup series. **Base is the Phase 1 branch** (`cleanup/phase-1-dead-code-and-fixes`, #2873), not `main` — review/merge Phase 1 first. Net **−125 lines**.

### `Repo.fetch/2`
Adds a small helper and removes the pervasive not-found boilerplate:
```elixir
def fetch(queryable, error \\ :not_found) do
  case one(queryable) do
    nil -> {:error, error}
    result -> {:ok, result}
  end
end
```
Replaced the repeated `|> Repo.one() |> case do nil -> {:error, :not_found}; x -> {:ok, x} end` with `Repo.fetch/1` across **28 getters** in `accounts`, `devices`, `products`, `firmwares`, `managed_deployments`, `scripts` (incl. 3 prefix-form `case Repo.one(...)` variants). Behaviour is identical.

Left untouched: sites with a different error atom (`:org_not_found`, `:invite_not_found`) or a transformed success branch (e.g. `{:ok, Org.with_org_keys(org)}`).

### Extension settings
Collapsed the near-identical `enable_extension_setting`/`disable_extension_setting` pairs in `Products` and `Devices` into a private `set_extension_setting/3` (parameterized on `enabled?` + the `"attach"`/`"detach"` event).

### Deliberately NOT deduplicated (with reasons)
- **`platforms/1` + `architectures/1`** (Devices): parameterizing the JSON key breaks `SELECT DISTINCT … ORDER BY` — Postgres binds the select and order-by as *different* params (`$1` vs `$3`) and rejects it with `ORDER BY expressions must appear in select list`. (Caught by the test suite; kept as separate literal-fragment queries.)
- **The `*_count` families** (Devices): their `WHERE` clauses and join targets differ enough (`Device` vs `InflightUpdate`, `latest_connection` vs `latest_health`) that a shared counter would reduce readability.

### Verification
- `mix format` ✓, credo (no new issues) ✓, dialyzer (0 unskipped) ✓
- **Full test suite: 1396 passing, 0 failures.**

### DIFF
--- a/lib/nerves_hub/accounts.ex
+++ b/lib/nerves_hub/accounts.ex
@@ -236,11 +236,7 @@ defmodule NervesHub.Accounts do
 
   def get_org_user(org, user_id) do
     get_org_user_query(org, user_id)
-    |> Repo.one()
-    |> case do
-      nil -> {:error, :not_found}
-      org_user -> {:ok, org_user}
-    end
+    |> Repo.fetch()
   end
 
   defp get_org_user_query(org, user_id) do
@@ -376,11 +372,7 @@ defmodule NervesHub.Accounts do
 
     query
     |> Repo.exclude_deleted()
-    |> Repo.one()
-    |> case do
-      nil -> {:error, :not_found}
-      user -> {:ok, user}
-    end
+    |> Repo.fetch()
   end
 
   def get_user!(user_id) do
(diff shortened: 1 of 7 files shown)


nshoes (approved):
I can get behind this

PR #2887, opened 2026-08-12
CI: parallelize checks and cache Docker layers

joshk:
## What

Speeds up the GitHub Actions CI run by splitting the serial `compile-and-test` job into parallel jobs, partitioning the test suite, and caching Docker layers.

### Job split

The single job ran everything sequentially — install fwup/xdelta3/mtools, compile, format, deps check, spelling, credo, JS tests, DB setup, dialyzer, then the suite — so a fast check waited behind all the slow ones. Now:

- **`lint`** — `compile --warnings-as-errors`, tailwind install, format, `deps.unlock --unused`, spelling, Credo. No DB or fwup/mtools, so it skips the slow `apt-get`.
- **`dialyzer`** — its own job, off the test critical path.
- **`test`** — a `[1, 2, 3]` matrix running `mix test --partitions 3 --include mtools`. Each partition is an isolated runner with its own Postgres/ClickHouse, so the fixed test DB name is fine. `fail-fast: false`.
- **`js-test`** — Node-only, no longer queued behind the Elixir compile.

Shared BEAM setup + `deps`/`_build` caches are extracted into a composite action at `.github/actions/setup` (same cache keys as before).

### Docker

Added `docker/setup-buildx-action` + `cache-from/cache-to: type=gha` so PR builds reuse layers instead of building from scratch. `needs` now points at the four new jobs, preserving the previous gating.

## Notes

- Warnings-as-errors still lives only in `lint`; siblings restore the previous run's `_build` at start (mid-run cache saves aren't visible across jobs), so lint recompiles changed files and catches warnings as before.
- Partition count is 3 — `mix test --partitions` divides by file, so if one partition lags on the first runs, the matrix can be adjusted.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,48 @@
+name: "Setup Elixir and dependencies"
+description: "Sets up the BEAM, restores the deps/build caches, and fetches dependencies"
+
+outputs:
+  elixir-version:
+    description: "Resolved Elixir version"
+    value: ${{ steps.beam.outputs.elixir-version }}
+  otp-version:
+    description: "Resolved OTP version"
+    value: ${{ steps.beam.outputs.otp-version }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Elixir
+      id: beam
+      uses: erlef/setup-beam@v1
+      with:
+        version-file: .tool-versions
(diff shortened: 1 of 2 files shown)


joshk:
This essentially halves our CI wall clock time

nshoes (approved):

PR #2893, opened 2026-08-13
Extract a transport-independent DeviceLink contract

joshk:
Extracts a contract for everything the platform decides about a connected device, so that the decisions stop being tangled up with the transport that happens to deliver them.

Today a device's behaviour is spread across `DeviceChannel`, `DeviceSocket`, `ExtensionsChannel` and the contexts they reach into, and all of it is written in terms of `Phoenix.Socket`. Extensions take a socket, read assigns off it, and push through it. That makes the websocket the only way a device can talk to us — not because anything about the logic requires it, but because the logic is written in websocket vocabulary.

This is the groundwork for changing how devices connect at all. If we want to offer MQTT, or anything else, the platform's answer to "this device just said X, what happens now?" has to be expressible without a `Phoenix.Socket` in the middle. After this branch it is.

## The shape

`NervesHub.DeviceLink` is the contract. It takes plain data and returns plain data:

```elixir
{session, effects} = DeviceLink.device_message(session, "fwup_progress", params)
```

`Session` carries what the platform needs to know about a connection. Effects say what should happen to it — `{:push, event, payload}`, `{:subscribe, topic}`, `{:send_after, key, msg, delay}`, `{:scrollback_append, chunk}`. Nothing in the contract knows what a socket is.

`DeviceChannel` drops from 335 lines to 84 and becomes an interpreter: it receives effects and performs them. Extensions were rewritten the same way — they take state and return effects instead of manipulating a socket. Authentication, the connection lifecycle, extension dispatch, and certificate verification all moved behind the contract.

A dispatcher seam (`DeviceLink.Dispatcher`) sits between the contract and its implementation, with `Local` calling in-process. Local dispatch is the default and is what every existing node uses, so nothing about how the platform runs today changes.

## Risk

This is a refactor of the path every device is on, which is worth being blunt about. The mitigations:

- Local dispatch is unconditional unless configured otherwise, and is a direct function call — the same work, in the same process, as before.
- The device-facing wire protocol is untouched. No device sees a difference.
- `DeviceChannel` and `ExtensionsChannel` tests were kept and now exercise the effects path.
- New tests cover dispatch, serialization of everything crossing the contract, certificate verification, and script reference lifecycle.

One behaviour change worth calling out: `DeviceSocket.connect/3` now refuses with `503` and an `nh-connection-error-reason` header when authentication cannot be reached, instead of raising into a `500`. A device in that position should retry, not re-check its credentials, and `401` told it the wrong thing. Local dispatch never raises there, so this is only reachable under a dispatch mode no current node uses.

## Also in here

- `health_check_report` moved from `device:<id>:extensions` to `internal:device:<id>`, which is where the LiveViews consuming it already listen. It had been relying on `broadcast_from!(self())` to keep the sender from receiving its own message — a property of who is broadcasting rather than of the topic, and it meant a device-facing topic was carrying platform-internal traffic.
- Script references now release their timers on both completion *and* timeout. Previously a timed-out script left an entry behind for the life of the connection, which is a slow leak on long-lived connections.
- Local shell scrollback moved onto the connection, with a size guard.


nshoes (approved):
LGTM 🚀

PR #2894, opened 2026-08-14
Show support script syntax errors while editing

### LINKED ISSUE #2890: Show syntax errors when writing support scripts in the UI
When writing support scripts, I am unsure if the script has errors until I actually try running it. 

To combat this, I tend to write the code on an IDE with a LSP/linter first, and then paste it in. It would be nice if I can get feedback in the app without having to leave it and open an IDE.

This would make support scripts faster to write, and reduce the risk of a bad script reaching the device.

floze-the-genius:
## Summary

- validate support script text as Elixir during the existing form validation
- include parser line and column in field feedback without executing the script
- reuse a static parser atom to avoid growing the VM atom table from form input
- cover feedback on the new support script form

Closes #2890

## Testing

- `git diff --check`
- targeted changeset/parser compile smoke with Elixir 1.20.2 / OTP 29

Full `mix test test/nerves_hub_web/live/support_scripts_test.exs` was not run locally because the dependency graph is not installed in this sparse checkout; CI should exercise the LiveView test.


nshoes:
@floze-the-genius thanks for this! Can you please look at and fix errors in CI?

joshk:
This is very cool. Thank you. Can you also add a screenshot to the PR showing what it looks like when in use?

joshk (changes requested):
This is a great start, but I need to make some adjustments based on our recent trials with our rust agent. I'll have some updates shortly.

joshk:
Superseded by #3014

PR #2896, opened 2026-08-15
Record the identities a device holds on other networks

joshk:
Devices increasingly reach the outside world over something other than their NervesHub socket — an iroh endpoint, a NetBird or Tailscale peer, a WireGuard interface. Each of those networks names a device by a long-lived public key, and that key is what you need in order to reach it by any route other than NervesHub. Until now there was nowhere to keep it.

For example, at NervesCloud we have a customer on NetBird, and iroh is close behind, so this is the generic version rather than a table per service.

## What's here

- `Devices.ExternalIdentity` + a `Devices.ExternalIdentities` sub-context, following the pattern of `Devices.Pinning` / `Devices.Health`.
- An opt-in `external_identity` extension. It asks once per connection; there's no interval, because an identity is long-lived.
- A panel on the device details tab.

## Two fields, not one

`service` is the protocol and stays a closed `Ecto.Enum`. `instance` names *which* endpoint of it, so a device running both an iroh console and an iroh application records both without one overwriting the other. Singletons like Tailscale stay on `"default"` and the UI doesn't label them.

Telling endpoints apart by `identifier` instead looks cheaper but breaks on key rotation: the upsert would insert a second row rather than update the first, leaving the dead key behind with nothing marking it stale. A discriminator has to be stable, which rules out the value being tracked.

## What `identifier` means

The value the device cryptographically proves it holds — its iroh endpoint id, or its WireGuard/NetBird/Tailscale public key. Reassignable control-plane handles (a NetBird peer id, an assigned overlay IP) go in `details`. Keeping that line is what stops the column becoming meaningless as integrations are added, and it's what would let NervesHub answer "which of my devices holds this key?" for an external service's auth callout later.

## Uniqueness

- `(device_id, service, instance)` — one identity per endpoint.
- `(service, identifier)` — no two devices may claim the same key. This is what surfaces a cloned SD card: imaging a provisioned device copies its `/data`, so the whole batch boots holding the identity of the device that was imaged. Without it that's N valid rows and a fleet that misbehaves inexplicably.

`instance` is `NOT NULL` with a default deliberately — a nullable column would silently defeat the first index, since `NULL != NULL` in Postgres.

## Trust boundary

Device reports are bounded and never fatal to the connection: an unknown service is discarded, a malformed entry skipped, oversized input refused, and a device may **not** overwrite an identity an operator recorded by hand. That last case is logged rather than applied — a device disagreeing about its own identity means it was reflashed, its state partition was wiped, or something is claiming to be it, and that's the signal worth keeping.

Nothing is stored that is secret. `details` is explicitly non-authoritative and visible to anyone who can view the device.

## Scope

Server side only. The client half is a [companion PR](https://github.com/nerves-hub/nerves_hub_link/pull/460) on `nerves_hub_link`; a management API is deliberately left for later, once the shape has settled in practice.

## Screenshot

This is what the `External Identities` section looks like on the Device show page:

Screenshot 2026-08-15 at 8 55 17 PM


lib/nerves_hub/devices/external_identity.ex
```diff
@@ -0,0 +1,163 @@
...
+  @max_identifier_length 255
+
+  # Likewise a ceiling rather than an expectation: an iroh ticket plus its relay
+  # URLs is well under a kilobyte.
+  @max_details_bytes 4_096
+
+  schema "device_external_identities" do
+    belongs_to(:device, Device)
+
+    field(:service, Ecto.Enum, values: [:iroh, :netbird, :tailscale, :wireguard])
```
lawik, line 84: I don't think we win much by having this be an Enum. Considering how many additional things it could be.

You might want to announce the Zerotier or Yggdrasil identity. Or you want to announce what the thing is when accessed over Zigbee or Z-Wave. Or LoRAWAN, LoRa, RFM69. Or a company has their own special dealio set up.

The list will just keep expanding. If we want to have some special behavior for known ones we can certainly have that to provide better integration but I don't think restricting it wins us a lot.

lib/nerves_hub/devices/external_identity.ex
```diff
@@ -0,0 +1,163 @@
...
+  Devices increasingly reach the outside world over something other than their
+  NervesHub socket — an iroh endpoint, a NetBird or Tailscale peer, a plain
+  WireGuard interface. Those networks each name a device by a long-lived public
+  key, and each keeps *where* the device currently is (endpoint addresses, relay
+  assignment) separate and changeable. This schema records that pairing so an
+  operator can see, and act on, a device's identity elsewhere.
+
+  ## What `identifier` means
+
+  `identifier` is **the value the device cryptographically proves possession
```
lawik: how does the device cryptographically prove this? I feel like Claude is doing improv or I'm missing something.

lawik (dismissed):

lib/nerves_hub/devices/external_identity.ex
joshk: You make an interesting point.

I would suggest we open it up after using this in the wild for a bit first. 

Its easier to open it up then it is to restrict it once data has been added.

lib/nerves_hub/devices/external_identity.ex
joshk: You are right, this should be 'the value the device claims possession of`

lib/nerves_hub/devices/external_identity.ex
```diff
@@ -0,0 +1,163 @@
...
+  of** — its iroh endpoint id, or its WireGuard/NetBird/Tailscale public key.
```
joshk:
```suggestion
  `identifier` is **the value the device claims possession of** — its iroh endpoint id, 
  or its WireGuard/NetBird/Tailscale public key.
```

jjcarstens (approved):

PR #2901, opened 2026-08-18
Version connection history rows at microsecond resolution

joshk:
## The problem

`device_connection_history` is a `ReplacingMergeTree(version)` sorted by `(org_id, product_id, device_id, established_at)`. Every history row for a single connection — connecting, connected, heartbeats, disconnected — shares all four of those values, so `version` is the only thing deciding which row the merged (`FINAL`) view keeps.

`version` was `DateTime.to_unix(now)`: whole seconds. A connect/disconnect cycle usually completes well inside one second, so the rows come out tied, and ClickHouse resolves version ties arbitrarily.

Probe of a real run, 1.5s after the writes, all three rows present:

```
=== RAW (3) ===
  version=1787015348 last_seen=...858927Z disc_at=nil    reason=""
  version=1787015348 last_seen=...861415Z disc_at=nil    reason=""
  version=1787015348 last_seen=...862130Z disc_at=...    reason="Stale connection"
=== FINAL (1) ===
  version=1787015348 disc_at=nil reason=""
```

The merged view reports a closed connection as still open. That matters in production, not just in tests: `Connections.clean_stale_connections_from_analytics/0` reads with `final: 1` looking for rows with a nil `disconnected_at`, so a cleanly closed connection can be picked up and written back as a spurious stale disconnect. The same merged view backs the product insights queries.

It also made `test "device_disconnected/2 records a history row with the disconnect details"` flaky — 3 failures in 6 runs on an unmodified `main`. Waiting longer never helped; the disconnect row had already landed.

## The fix

Version rows at microsecond resolution. A connection's rows are built sequentially in the caller, so versions strictly increase and the most recent write always wins the merge.

No migration needed — the column is already `UInt64`, and a microsecond version (~1.8e15) outranks every existing second-scale one (~1.8e9), which is the correct outcome since those rows are older.

One caveat: during a rolling deploy, an old node writing a seconds-scale version for a connection whose earlier rows came from a new node will lose the merge. That is one deploy window, affecting only connections straddling it, and is unavoidable for any change of version scale.

## Tests

- The disconnect test now asserts the merged view collapses to exactly one row carrying the disconnect — that is the deterministic guarantee, and it fails loudly if the tie-break regresses.
- Corrected a stale comment on the `device_connected` test claiming `FINAL` may keep both rows.
- Added unit tests for strictly increasing versions across successive rows, and for `mark_as_stale_and_disconnected_changeset/1` bumping past the row it carries forward.

### DIFF
--- a/lib/nerves_hub/devices/device_connection_history.ex
+++ b/lib/nerves_hub/devices/device_connection_history.ex
@@ -43,8 +43,18 @@ defmodule NervesHub.Devices.DeviceConnectionHistory do
     |> put_change(:lib, connection.lib)
     |> put_change(:lib_version, connection.lib_version)
     |> put_change(:network_interface, to_string(connection.network_interface))
-    |> put_change(:version, DateTime.utc_now() |> DateTime.to_unix())
+    |> put_change(:version, current_version())
   end
+
+  # The `ReplacingMergeTree` dedupes on (org_id, product_id, device_id,
+  # established_at), which every row for a single connection shares, and keeps
+  # the row with the highest `version`. A connection's rows (connecting,
+  # connected, heartbeats, disconnected) are usually written within the same
+  # second, so a second-resolution version leaves them tied and ClickHouse picks
+  # between them arbitrarily - the merged view could report an already
+  # disconnected connection as still open. Microseconds keep the writes strictly
+  # ordered, so the most recent one always wins.
+  defp current_version(), do: DateTime.to_unix(DateTime.utc_now(), :microsecond)
 
   @doc """
   Builds a new history row from an existing one.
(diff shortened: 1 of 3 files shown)


lib/nerves_hub/devices/device_connection_history.ex
```diff
@@ -43,9 +43,19 @@ defmodule NervesHub.Devices.DeviceConnectionHistory do
...
 
+  # The `ReplacingMergeTree` dedupes on (org_id, product_id, device_id,
+  # established_at), which every row for a single connection shares, and keeps
+  # the row with the highest `version`. A connection's rows (connecting,
+  # connected, heartbeats, disconnected) are usually written within the same
+  # second, so a second-resolution version leaves them tied and ClickHouse picks
+  # between them arbitrarily - the merged view could report an already
+  # disconnected connection as still open. Microseconds keep the writes strictly
+  # ordered, so the most recent one always wins.
+  defp current_version(), do: DateTime.to_unix(DateTime.utc_now(), :microsecond)
```
lawik, line 57: Shouldn't this be some kind of monotonic time?

I'll leave it to you though.

lawik (approved):

lib/nerves_hub/devices/device_connection_history.ex
joshk: Not sure, but this is better than what is was, so I'll take it.

PR #2902, opened 2026-08-18
Fix device events stream never forwarding update progress

joshk:
`DeviceEventsStreamChannel` matched on a broadcast that is never sent: event `"fwup_progress"` with a `:percent` atom key. What `NervesHub.FirmwareUpdates` actually broadcasts is `"firmware_update_progress"` with string keys.

The handler wasn't fixed when some recent changes were merged in.

The existing test hid this by hand-broadcasting the fake message rather than going through the real path. It now calls `FirmwareUpdates.update_inflight_update/4`, so a mismatch between what is broadcast and what is handled fails the test instead of passing under it.

The pushed payload gains `stage` alongside `percent`, which is additive for existing consumers.

### DIFF
--- a/lib/nerves_hub_web/channels/device_events_stream_channel.ex
+++ b/lib/nerves_hub_web/channels/device_events_stream_channel.ex
@@ -28,10 +28,14 @@ defmodule NervesHubWeb.DeviceEventsStreamChannel do
     end
   end
 
+  # `NervesHub.FirmwareUpdates` broadcasts `firmware_update_progress` with string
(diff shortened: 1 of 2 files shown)


jjcarstens (approved):

nshoes:
> The existing test hid this by hand-broadcasting the fake message rather than going through the real path

This is a pretty common occurence with our test suite, and I've noticed Claude has picked up that pattern and uses it consistently when they're writing tests.

PR #2905, opened 2026-08-18
Add ESP-IDF application image support, off by default

joshk:
This PR adds support for ESP-IDF application images (`.bin`), alongside fwup archives. 

Everything it needs is already in the image: `esp_app_desc_t` sits at a fixed offset of `0x20` and carries the project name, version, and a SHA-256 of the source ELF; `esp_image_header_t` carries the chip ID. A firmware author adds nothing to their build beyond a `PROJECT_VER` that reads as SemVer.

## UI changes

**Product → Settings → Firmware**
The two new toggles. The second only appears once the first is on, and neither appears unless the instance has ESP_IDF_FIRMWARE_ENABLED.

image

**Signing keys**
Keys now carry a scheme, so an org can hold both fwup Ed25519 keys and Secure Boot v2 RSA keys.

image

image

**Firmware list**
An ESP-IDF image alongside fwup archives — the tool, the chip as platform/architecture, and the key that verified it.

image

## Enabling it

Off unless `ESP_IDF_FIRMWARE_ENABLED=true`, and then per product:

| Setting | Column | Default |
| --- | --- | --- |
| Accept ESP-IDF application images | `products.allowed_update_tools` | `["fwup"]` |
| Allow unsigned ESP-IDF images | `products.allow_unsigned_esp_idf_firmware` | `false` |

No existing product changes behaviour, and a product cannot list a format the instance has not enabled. Both settings are readable and settable over the product API, which previously exposed only a product's name.

## Signatures

Secure Boot v2 RSA-3072: a 1216-byte block at the start of a 4 KB sector appended to the image, RSA-PSS (MGF1/SHA-256, 32-byte salt) over everything before that sector, with the modulus and signature stored byte-reversed for the ESP32 RSA peripheral.

`org_keys` gains a `scheme`, so an org can hold both fwup Ed25519 keys and Secure Boot v2 RSA keys; each tool considers only keys of its own scheme. The public key embedded in the signature block is ignored — anyone can self-sign. ECDSA blocks share the magic byte but not the layout, and are refused rather than misread.

The layout is verified against an image signed by the real `espsecure.py`, committed under `test/fixtures/esp_idf/`. The synthetic signer in `test/support/esp_idf.ex` is checked against that fixture, so the two cannot share a mistake.

## Not supported yet

Deltas are generated and stored but always sent in full — applying an xdelta3 patch needs a device to read back its inactive OTA slot, which no agent does yet. The client library at [nerves-hub-link-esp32](https://github.com/nerves-hub/nerves-hub-link-esp32) is early.

## Also here: two dead routes

`/org/:org_name/:product_name/firmware/upload` and the archives equivalent routed to their LiveViews with a `:upload` action that `apply_action/3` has no clause for, so both URLs raised a `FunctionClauseError` for anyone who reached them. They predate this branch — I found the firmware one while taking the screenshots above. Both are uploaded from their index pages, so there was nothing for either route to show.

"upload" now falls through to the `:show` route as an unknown identifier: a 404 for firmware, and a 400 for archives, whose lookup casts the value to a UUID first.


config/runtime.exs
```diff
@@ -43,6 +43,10 @@ config :nerves_hub,
   deploy_env: System.get_env("DEPLOY_ENV", to_string(config_env())),
   log_include_mfa: System.get_env("LOG_INCLUDE_MFA", "false") == "true",
   web_title_suffix: System.get_env("WEB_TITLE_SUFFIX", "NervesHub"),
+  # Accept ESP-IDF application images (.bin) in addition to fwup archives.
```
lawik: Strip the comment. I don't think we need it :)

docs/esp_idf_support.md
```diff
@@ -0,0 +1,152 @@
...
+
+Deltas are generated and stored as whole-image `xdelta3` patches, but **are never
+sent to devices** — `device_update_type/2` always returns `:full`. Applying a
+patch requires the device to read back its inactive OTA slot and patch into it,
+which no device agent does yet. Deployment groups with delta updates enabled will
+still send complete images to ESP-IDF devices.
+
+### Device connectivity
+
+There is **no device agent yet**. NervesHub can store and serve ESP-IDF images,
```
lawik: Reference the client library you made.

docs/esp_idf_support.md
```diff
@@ -0,0 +1,152 @@
...
+}
+```
+
+Note that the device does **not** send a UUID. NervesHub derives it from
+`app_elf_sha256` using the same rule it applied when the image was uploaded, so
+an agent cannot get it subtly wrong.
+
+`update_tool` is optional — NervesHub also recognises a device by the keys it
+sends, and a device sending nothing recognisable is read as a Nerves device, as
+it was before this seam existed. Sending it explicitly is preferred.
```
lawik: "as it was before this seam existed". I think Claude needs to be told to go over and just tighten this documentation. People don't need to know what the technical decision making was at the time :)

docs/esp_idf_support.md
```diff
@@ -0,0 +1,152 @@
...
+### Reporting update progress
+
+Send `update_progress` with a `value` percentage and an optional `stage`
+(`downloading`, `updating`):
+
+```json
+{"value": 42, "stage": "downloading"}
+```
+
+`fwup_progress` is the same message under its original name. Every deployed
```
lawik: Also not documentation of this feature. This is documenting the work that was done.

Making review notes on this where it bothers me and you can just ask Claude to check the review for my comments. So hey, Claude, go through all the descriptive text, the comments and documentation and make sure it is describing how the thing works. Not how you decided to implement things as a moment-in-time.

lawik (dismissed):
A few notes :) mostly have Claude do a pass to prune the walls of text and tighten documentation

jjcarstens (approved):

PR #2908, opened 2026-08-19
Verify ESP-IDF Secure Boot v2 RSA signatures

joshk:
An ESP-IDF image carrying a Secure Boot v2 signature block now has to verify against a key the organization registered, rather than being accepted as though it were unsigned.

- Every key until now was an fwup Ed25519 public key and the format check assumed it. `org_keys.scheme` is backfilled to `ed25519`, with `secure_boot_v2_rsa` for a PEM RSA-3072 public key.
- `key` also widens from `varchar(255)` to `text` — an Ed25519 key is 44 characters, an RSA-3072 PEM is around 625. A test caught that, not review.
- Each tool filters by scheme rather than trying every key: fwup would otherwise hand a multi-line PEM to `fwup --verify` as an argument.

## What is verified

- CRC32 over the first 1196 bytes of the block
- The block's SHA-256 against the image preceding the signature sector
- RSA-PSS (MGF1/SHA-256, 32 byte salt) over that digest

**The public key embedded in the block is deliberately ignored.** Verifying against it would prove only that *somebody* signed the image, and anyone can self-sign. Candidates are the org's registered keys, exactly as the fwup tool tries each Ed25519 key.

The modulus and signature are stored **byte-reversed** in the block for the ESP32's RSA peripheral, and are reversed back before use. An ECDSA block (version `0x03`) has a different layout entirely — a hash-type byte where RSA has padding, key and signature elsewhere — and is refused rather than misread as RSA.

## Fixtures come from `espsecure.py`

`test/fixtures/esp_idf/` holds an image signed by ESP-IDF's own tooling, not by this repository:

```bash
espsecure.py generate_signing_key --version 2 --scheme rsa3072 signing_key.pem
espsecure.py sign_data --version 2 --keyfile signing_key.pem --output signed_rsa3072.bin unsigned.bin
```

That byte-reversal is why. A hand-rolled signer written next to the verifier would share any mistake about endianness and pass while production failed. The previous synthetic `Support.EspIdf.signed_image/1` is removed for exactly that reason — the coverage is stronger for it, not weaker.


joshk:
Combined into #2909 — signature verification you cannot register a key for is not shippable on its own, so reviewing them apart was more work for less signal. Both commits are on that PR unchanged.

PR #2909, opened 2026-08-19
Verify ESP-IDF Secure Boot v2 signatures, and require them

joshk:
Verifies ESP-IDF Secure Boot v2 RSA-3072 signatures, requires them, and makes it possible to register the keys that do the verifying.

Originally split into two PRs; combined because signature verification you cannot register a key for is not shippable on its own.

## Verification

- CRC32 over the first 1196 bytes of the signature block
- The block's SHA-256 against the image preceding the signature sector
- RSA-PSS (MGF1/SHA-256, 32 byte salt) over that digest

**The public key embedded in the block is deliberately ignored.** Verifying against it would prove only that *somebody* signed the image, and anyone can self-sign. Candidates are the org's registered keys, exactly as the fwup tool tries each Ed25519 key.

The modulus and signature are stored **byte-reversed** in the block for the ESP32's RSA peripheral, and are reversed back before use. An ECDSA block (version `0x03`) has a different layout entirely and is refused rather than misread as RSA.

## Fixtures come from `espsecure.py`

`test/fixtures/esp_idf/` holds an image signed by ESP-IDF's own tooling, not by this repository. That byte-reversal is why: a hand-rolled signer written next to the verifier would share any mistake about endianness and pass while production failed.

`Support.EspIdf.signed_image/1` does sign in-repo, so other suites can vary product and version — but it is only safe because `EspIdfSignatureTest` checks the real espsecure image through the same verifier. That fixture is the ground truth.

## Organization keys gain a scheme

`org_keys.scheme` is backfilled to `ed25519`, with `secure_boot_v2_rsa` for a PEM RSA-3072 public key. `key` also widens from `varchar(255)` to `text` — an Ed25519 key is 44 characters, an RSA-3072 PEM is around 625. A test caught that, not review.

Each tool filters by scheme rather than trying every key: fwup would otherwise hand a multi-line PEM to `fwup --verify` as an argument.

## Registering a key was impossible

Verification would have been unreachable in practice, because neither surface could register a Secure Boot v2 key:

**UI** — the form gains a scheme picker, and the key field becomes a textarea (a PEM is multi-line and could not be pasted into a text input). The list shows the scheme and, for a PEM, only its boundary lines rather than 600 bytes of base64.

**API** — `KeyController.create/2` whitelisted only `name` and `key`, so a posted `scheme` was **silently dropped**, defaulted to `ed25519`, and an RSA PEM came back as *"not a valid Ed25519 public key"*. Now accepted, returned, and documented with its enum; omitting it still means `ed25519`, so existing clients are unaffected.

## Firmware API

`tool` is exposed on firmware responses — without it an API consumer cannot tell an fwup archive from an ESP-IDF image.

The `create` operation documented its 422 as a changeset error, but failures raised *before* the changeset (unrecognised format, unsigned image, non-SemVer version) return `{"errors": {"detail": ...}}`. A client generated from that spec would mis-parse them. Now a `oneOf` of both shapes.

## ⚠️ Migrations to collapse before merge

Requiring signatures means no firmware can be unsigned, so `firmwares.org_key_id` goes back to `NOT NULL` and the CHECK scoping unsigned storage to ESP-IDF is dropped. `20260818000000` (in #2905), `20260819000000` and `20260819000001` therefore cancel out and should be squashed into a single "add scheme to org_keys" migration. Kept apart here so the reasoning is visible in review.

**This is likely to change**: per-product settings for ESP-IDF uploads and for allowing unsigned images are under discussion, and the unsigned case would need the nullable column back. Worth settling before squashing.

## The adoption trade-off

Most ESP-IDF builds are unsigned, so requiring signatures means an ESP-IDF shop must set up Secure Boot v2 signing before it can use NervesHub — the same friction fwup users accept, but new for this audience. Deliberate, in favour of consistency with the platform's existing posture.

## Verification

1534/1537 (the three are pre-existing `mtools` and ClickHouse-flake failures that also fail on `main`), dialyzer clean, Credo at parity, format and spellcheck clean.

Stacked on #2905.


joshk:
Folded into #2905. Keeping signature verification separate meant it rewrote #2905's unsigned-firmware story and its migration, so the two land together instead.

PR #2910, opened 2026-08-19
Let a product choose which firmware formats it accepts

joshk:
Sixth and last of the ESP-IDF stack. Builds on #2909.

ESP-IDF support was gated only by `ESP_IDF_FIRMWARE_ENABLED`, which is instance-wide: turning it on opted in every product on the instance, including ones that only ever ship fwup and for which a `.bin` upload can only be a mistake.

Two settings on the product, under **Product → Settings → Firmware**:

| Setting | Column | Default |
| --- | --- | --- |
| Accept ESP-IDF application images | `products.allowed_update_tools` | `["fwup"]` |
| Allow unsigned ESP-IDF images | `products.allow_unsigned_esp_idf_firmware` | `false` |

An existing product keeps refusing anything but fwup until someone opts it in, and the platform flag still bounds the whole thing — a product cannot enable a format the instance has not. The settings are hidden entirely when the flag is off.

### On the unsigned setting

It excuses a *missing* Secure Boot v2 signature block and nothing more. An image that carries a signature is still verified against the org's registered keys, so the setting cannot be used to accept an image signed by a key nobody registered. It is there for board bring-up, before Secure Boot has been provisioned.

### On the database constraint

The rule is now "signed unless the product says otherwise, and only ESP-IDF may say otherwise", which cannot be expressed as a constraint on `firmwares`. So the CHECK covers what it can — `org_key_id IS NOT NULL OR tool IS NOT DISTINCT FROM 'esp-idf'` — and the application covers the rest. fwup still cannot be left unsigned at the database level, and there is a test for each half of that.

(`IS NOT DISTINCT FROM` rather than `=` because `tool` is nullable, and a CHECK passes on NULL.)

### Tested

`test/nerves_hub/firmwares/product_update_tool_settings_test.exs` covers the default, the rejection, the opt-in, the unknown-tool changeset error, the unsigned allowance, and that the allowance extends neither to a bad signature nor to fwup. `settings_test.exs` covers the toggles, including that the second only appears once the first is on.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
+.expert
 .elixir-tools
 .elixir-ls
 .env.*
--- a/docs/esp_idf_support.md
+++ b/docs/esp_idf_support.md
@@ -38,6 +38,27 @@ not the setting is on. See [Firmware signing](#firmware-signing).
 Turning acceptance back off leaves the unsigned setting untouched, so a product
 that is switched off and on again comes back configured as it was.
 
+### Over the API
+
+Both settings are on the product resource, and settable with `PUT` or `PATCH`
+(org role `manage`):
+
+```bash
+curl -X PATCH \
+  -H "Authorization: token $NH_TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{"allowed_update_tools": ["fwup", "esp-idf"]}' \
+  https://<host>/api/orgs/<org>/products/<product>
+```
+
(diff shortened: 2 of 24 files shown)


joshk:
Folded into #2905, along with #2909. The per-product gate rewrote the signing rule those PRs had just established, and the `firmwares` table was being altered three times to land where the first migration already had it.

PR #2924, opened 2026-08-20
Redirect misconfigured devices to the device websocket host

joshk:
## What

Some devices are configured with the management host rather than the device host — `manage.nervescloud.com` instead of `devices.nervescloud.com` — so they connect to the management endpoint. When `redirect_to_devices_websocket_url` is enabled, those connections get a `301` to `devices_websocket_url`, and a notification is raised against the product so the misconfiguration is visible to whoever owns the fleet.

Off by default, behind `REDIRECT_TO_DEVICES_WEBSOCKET_URL`. An unset `devices_websocket_url` means no redirect regardless of the flag — the two are separate environment variables, and redirecting to an empty host would strand every device.

## Why it isn't a plug

A plug on `Endpoint` can't do this. Phoenix registers `plug :socket_dispatch` inside `use Phoenix.Endpoint`, so it runs before anything declared in the endpoint body — confirmed with a temporary probe plug, which set a header on `GET /` but never ran for `/device-socket`.

So the redirect lives in `DeviceSocket.connect/3`. Returning `{:error, {:redirect, host}}` routes to the socket's existing `error_handler` with the conn as it was *before* the upgrade — `WebSockAdapter.upgrade` only runs on the `{:ok, arg}` branch — so the device receives an ordinary HTTP redirect on the pre-upgrade request.

## Ordering: authentication, then redirect

The notification has to name the product and the device, and neither is known until the device identifies itself. So `do_connect/3` runs first and `maybe_redirect/1` decides afterwards. Two consequences:

- a wrong-host device pays for a cert lookup or an HMAC verification before being turned away — the same work as a normal connection, and only when the flag is enabled
- shared-secret auth can create a device, so a device connecting to the management host has its `Device` row created there before being redirected. It would be created on the device host anyway

Devices that fail authentication get the usual `401` rather than the redirect.

## Scoping

Only the management endpoint redirects, matched on `socket.endpoint`. `DeviceEndpoint` is the destination, so redirecting there would loop.

The location preserves the request path and query, so the device lands on the same socket with the transport version it asked for. `devices_websocket_url` is a bare host today, so the URL builder defaults the scheme to `wss://` while tolerating a full URL.

## The notification

`event_key` is per device identifier, so each misconfigured device gets its own row and a device reconnecting bumps `occurrence_count` on that row rather than adding another. It follows `create_duplicate_device_identifier_notification!/3`, the existing precedent for a notification raised from a rejected device connection, and needs no UI work — it surfaces on the product notifications page and the nav badge.

`insert_and_notify!/1` is unchanged, so every occurrence broadcasts and an open notifications page keeps counting up.

## Logging

The redirect emits `[:nerves_hub, :devices, :wrong_websocket_host]`, handled in `NervesHub.Logger` alongside the other device events:

```
Device connected to the wrong host, redirecting
  event=nerves_hub.devices.wrong_websocket_host
  device_id=… device_identifier=… product_id=… host=devices.nervescloud.com
```

Being in that list means it honours `LOGGER_EXCLUSIONS`, so a noisy misconfigured fleet can be silenced with `nerves_hub.devices.wrong_websocket_host` without a deploy.

## Verified against a running server

| condition | result |
|---|---|
| flag on, device socket upgrade | `301` → `wss://devices.nervescloud.com/device-socket/websocket?vsn=2.0.0` |
| flag on, `GET /` | `302` — normal web traffic untouched |
| flag off (default) | `401` with the usual `nh-connection-error-reason` — unchanged |

## Worth knowing

If a client version turns out not to follow the `301`, those devices will reconnect in a loop and every attempt becomes a notification upsert plus a PubSub broadcast. The narrow fix would be a rate limit on this call site rather than a change to `insert_and_notify!`.


nshoes (approved):
This is cool, I love these internal notifications. LGTM

PR #2925, opened 2026-08-20
Record device messages and show them on a Data History tab

joshk:
This records every message that crosses a device's websocket and shows it on a new **Data History** tab on the device page.

This is only available for platforms with a configured Clickhouse database.

## Screenshots

image

image

## What is recorded

Direction (`received` from the device, `sent` to it), the channel it crossed, the event name, and when. What is kept of the body depends on the channel:

| Channel | Body |
| --- | --- |
| `device`, `extensions` | the payload, redacted and capped at 8KB |
| `console` | **size only, never contents** |

Console traffic is raw terminal I/O — its contents are whatever someone typed at a prompt, routinely credentials — and its volume is unlike the others, since a single command is dozens of messages. Recording bodies there would dominate the table while being the part least safe to keep.

## Where recording happens

At the connection boundary, in the channels themselves, so a message is recorded because it *actually crossed the wire* rather than because something intended it to.

The exception is the platform's **fastlaned** sends — `identify`, `reboot`, `update`, `archive` and the public key messages — which Phoenix pushes straight from the broadcast to the transport without the channel process ever seeing them. Those are recorded where they are broadcast, which for them is the same thing.

`updated` and `deployment_updated` are deliberately *not* recorded at the broadcast site: `DeviceChannel` intercepts them and turns them into other messages, so the device never receives them, and recording them would claim a send that did not happen. There's a test asserting this.

## Redaction

Redaction runs *before* encoding, so it reaches values nested inside update and archive payloads rather than just the top level.

- **Presigned firmware/archive URLs** keep their scheme, host and path but lose the query string, which is where the signature lives. Knowing a device was pointed at a particular firmware object is the useful part; the credential that lets anyone fetch it is not.
  ```
  https://firmware.example.com/org/1/fw.fw?X-Amz-Signature=deadbeef…
  → https://firmware.example.com/org/1/fw.fw?[redacted]
  ```
- `token`, `secret`, `password`, `private_key`, `auth`, `authorization` are replaced outright, at any depth.

Both key sets are configurable under `config :nerves_hub, NervesHub.Devices.DeviceMessages`.

## Batching the writes

Rebased onto #2932, which landed `NervesHub.Analytics.Buffer` while this was open. That does everything the bespoke buffer in the first version of this PR did and more — the flush runs in a task so the buffer keeps draining its mailbox, only one flush is in flight at a time, and drops are reported on `[:nerves_hub, :analytics, :buffer, :dropped]` — so the bespoke one is **deleted** and `DeviceMessage` is registered as a third buffered schema alongside `DeviceConnectionHistory` and `LogLine`.

That makes this the highest-volume of the three: every message on every channel, including per-keystroke console traffic. Every write to `AnalyticsRepo` in the codebase now goes through the single `insert_all` in that buffer; there are no unbuffered analytics write paths left.

The one thing dropped in the swap was per-device partitioning. One process per schema is the right call here — per message the buffer does a list prepend and a counter bump, with redaction and JSON encoding already done in the calling channel process, and the flush-in-a-task means a slow ClickHouse doesn't back the mailbox up.

Two things worth a decision rather than a quiet tweak:

- `ANALYTICS_BUFFER_MAX_SIZE` (50k rows) is shared by all three buffers. A connection or log row is a fixed handful of small columns; a device message row carries a payload capped at 8KB, so in the pathological case this buffer holds far more memory than the other two. Typical payloads are tens of bytes and the drop-half valve still bounds it, so I have left the shared knob alone.
- Log lines are rate limited to 5/sec/device; device messages are not, so a chatty device contributes as fast as it can talk. The buffer degrades gracefully rather than growing, but the asymmetry is deliberate to note.

## Redacting a struct

Rebasing surfaced a real bug in the first version of this PR, now fixed with a regression test.

`redact/1` was guarded with `not is_struct(payload)`. The `update` message carries an `%UpdatePayload{}` — and a struct is a map, so redaction skipped it in silence and stored the presigned `firmware_url` **verbatim**. Every test passed because they all used plain maps.

Structs are now normalised through their own JSON encoder before being walked. That fixes the leak, and has a second benefit: the stored row is limited to the fields the device actually receives, rather than the loaded `deployment_group` that hangs off the struct and is excluded by its `@derive`.

Removing the struct clause makes the new test fail on the leaked signature, so it holds the fix in place.

## The tab

Streams live over PubSub, and **can be paused**.  Direction and channel filters live in the URL, so a particular view of a device's traffic can be linked to. An unrecognised filter shows everything rather than nothing, so a stale URL doesn't look like a device with no traffic.

The tab sits at the far right of the device nav as a bug icon, with its label sliding out on hover or when selected.

## Storage and retention

`device_messages`, MergeTree, partitioned by day, ordered `(product_id, device_id, timestamp)`, **3 day TTL** — the same retention `device_log_lines` uses. Recording is on whenever `analytics_enabled` is set — no new product or device extension toggle.


lawik:
Outdated :/

lawik (approved):

PR #2926, opened 2026-08-20
Record the address devices connect from

joshk:
Devices connect to NervesHub over two endpoints, and neither one could tell us where a device actually was.

The device endpoint terminates TLS in the app, because devices authenticate with client certificates. A balancer in front of it can only pass the connection through, so it can't add a forwarding header: it never sees inside the stream. The socket we accept is the balancer's own, opened over the platform's private network.

The web endpoint is ordinary HTTPS terminated ahead of us, so its socket peer is whatever terminated it.

Either way the device's address was lost, and nothing recorded it.

## Device endpoint: the PROXY protocol

`NervesHub.ProxyProtocol` reads a PROXY protocol v2 header off the socket before TLS starts, and `DeviceSSLTransport` grows a mode that listens in the clear, consumes the header, then upgrades by hand and answers `peername/1` with the announced client. Everything downstream (Bandit's `peer_data`, Phoenix's `connect_info`) then sees the device rather than the balancer, with no further changes.

Off by default. `DEVICE_PROXY_PROTOCOL=v2` turns it on, paired with a handler on the balancer; on Fly.io:

```toml
[[services.ports]]
  port = 443
  handlers = ["proxy_proto"]
  proxy_proto_options = { version = "v2" }
```

Only version 2 is supported. Version 1 is a variable-length text line, so reading it means either consuming a byte at a time or overshooting into the TLS ClientHello. An overshoot is unrecoverable, because those bytes can't be handed back to `:ssl.handshake/3`. Version 2 declares its own length, so every read is exact.

## Web endpoint: a forwarded header

`NervesHubWeb.Helpers.ClientIP` reads the address from a request header, taking the rightmost value. Each proxy appends what it saw to the right of whatever was already there, so the last entry is the one written by the hop closest to us, and a device sending its own `x-forwarded-for` only manages to prepend a lie. `Plug.RewriteOn` takes the leftmost value, which is why this isn't built on it.

`x-forwarded-for` is the default: a web endpoint almost always has something in front of it, and reading the peer instead would quietly record one balancer address for an entire fleet. A deployment exposed directly should set `WEB_FORWARDED_IP_HEADER=none`.

The device endpoint trusts no header at all, because TLS terminates there and anything that arrives could only have been written by the device itself.

The header must start with `x-`, enforced where the config is read rather than failing quietly.

## Storage

The address is carried through the socket to the connection record: a new `ip_address` column on `latest_device_connections` and on the ClickHouse `device_connection_history` table, and a row on the device's details tab.

image

image


joshk:
I've tested this on our NervesCloud deployment and it works as expected.

lawik:
Lars' bot here: I went through the diff together with ThousandIsland 1.5.0 / Bandit 1.12.4 internals and Fly's docs. The PROXY protocol implementation itself looks mechanically sound; my concerns are below, roughly in order of severity.

### 1. `x-forwarded-for` rightmost-entry on Fly records Fly's IP, not the device's

`ClientIP.resolve/2` takes the rightmost entry of `x-forwarded-for`. Fly's request-header docs say:

> On the Fly.io platform, the last address (rightmost) in this list will be a shared or dedicated IP address assigned to your app.

and recommend `Fly-Client-IP` when nothing sits in front of Fly. So on Fly the web endpoint will record the app's own Fly IP for every device connecting there — the "one balancer address for an entire fleet" outcome this PR is trying to avoid — and the `raise` text in `config/runtime.exs` actively steers people toward it ("keep `x-forwarded-for` rather than reaching for `fly-client-ip`").

Possible fixes: a configurable number of trailing trusted hops to skip, or — since Phoenix only hands sockets `x-*` headers — an Endpoint plug that copies `fly-client-ip` into an `x-` header before the socket upgrade. Either way it'd be good to verify against a real Fly request before this lands, and to flip the advice in the error message.

### 2. Rollout ordering can take the whole mTLS fleet offline

The moduledoc says both sides have to move together, but on Fly `fly secrets set DEVICE_PROXY_PROTOCOL=v2` without `--stage` restarts every machine *before* the `proxy_proto` handler exists. Every device's ClientHello is then consumed as a PROXY header, fails the signature check, and is hung up on, until the fly.toml deploy lands. Rollback has the mirror-image hazard.

Suggest documenting that the env var belongs in fly.toml `[env]` in the same commit as the `handlers` change, so each machine picks up both atomically.

### Minor

- `:persistent_term.get/1` in `ssl_options/1` raises `ArgumentError` if the key is missing. Unreachable today, but `get/2` with a default → `{:error, :closed}` is cheaper than a crashed connection process if the listener plumbing ever changes.
- A client that sends a valid header and then stalls is held indefinitely by `:ssl.handshake/2` (no timeout). Same as today, just noting that the 5 s header timeout doesn't cover it.

Edit: removed cruft

joshk:
You're right on #1, and the docs quote settles it. I'd built the rightmost rule on a community thread rather than the request-headers page, and probed a live request to check:

```
xff="182.48.134.168, 66.241.125.59"            fly-client-ip="182.48.134.168"
xff="1.2.3.4, 182.48.134.168, 66.241.125.59"   fly-client-ip="182.48.134.168"   # client sent 1.2.3.4
```

Fly keeps what the client sent and appends two entries: the address it observed, then the app's anycast address. So the device is second from the right, and rightmost really was recording one address for the whole fleet.

Fixed with the trailing-hop count you suggested: `forwarded_ip_trailing_hops`, default 0, `WEB_FORWARDED_IP_TRAILING_HOPS=1` on Fly. Counting from the right is what keeps it safe: the second line above shows a forged entry only pushes the list leftwards, so no number of prepended values moves where the count lands. There's a test for exactly that. The `raise` text now points at the hop count instead of steering people wrong.

On the plug idea: it can't work. `plug :socket_dispatch` is registered inside `use Phoenix.Endpoint`, so it's the first plug in the pipeline and a websocket upgrade never reaches anything declared in the endpoint body. Reaching `fly-client-ip` would mean wrapping the endpoint rather than adding to it, which the hop count avoids.

#2: agreed, and the moduledoc now shows `DEVICE_PROXY_PROTOCOL` in fly.toml `[env]` next to the `handlers` block, with a note that `fly secrets set` without `--stage` restarts machines before the handler exists.

Minor 1: taken, `:persistent_term.get/2` with a default answering `{:error, :closed}`.

Minor 2: taken as well, `:ssl.handshake/3` with a 30s timeout. Generous on purpose, since a device doing mutual TLS over bad cellular is the case that mustn't be cut off. Worth noting the swap wasn't free: `:ssl.handshake` returns `{:ok, socket}` where Thousand Island's `upgrade/2` matches the 3-tuple and survives on a passthrough clause, so my first attempt killed every connection until the transport test caught it.

Thanks for the careful read.

lawik (approved):

PR #2929, opened 2026-08-20
Use deployment group IDs in URL slug instead of encoded names

Nezteb:
For a variety of support reasons, it's nice to be able to get the numeric IDs of deployment groups.

Example:
CleanShot 2026-08-20 at 15 20 42@2x

This isn't a hard requirement neccesarily, but it'd be a lovely short-term value add for us. 🙏 

I also added a fallback so that existing URLs that use the encoded name should still resolve correctly.

### DIFF
--- a/lib/nerves_hub/command_palette.ex
+++ b/lib/nerves_hub/command_palette.ex
@@ -99,6 +99,7 @@ defmodule NervesHub.CommandPalette do
     |> order_by([dg], asc: dg.name)
     |> limit(^limit)
     |> select([dg, product: p, org: o], %{
+      id: dg.id,
       name: dg.name,
       org_name: o.name,
       product_name: p.name
--- a/lib/nerves_hub/managed_deployments.ex
+++ b/lib/nerves_hub/managed_deployments.ex
@@ -108,21 +108,53 @@ defmodule NervesHub.ManagedDeployments do
   def get_by_product_and_name!(product, name, with_device_count \\ false)
 
   def get_by_product_and_name!(product, name, true) do
-    subquery =
-      Device
-      |> select([d], %{device_count: count()})
-      |> Repo.exclude_deleted()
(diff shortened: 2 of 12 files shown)


joshk:
Thanks for the PR. I’ll have a look over this a little later today.

joshk:
Hey @Nezteb, before I review your PR I had a quick question. Is the primary reason for the URL change so that you can find the ID of the Deployment Group? or is it because you want to use the ID for navigating to it in the browser?

Nezteb:
@joshk We added some on-connect code to several deployment groups and then wanted to bulk-run a script that would reconnect each device in those deployment groups so the on-connect code would apply. Initially we were planning on:
1. Fetch `/orgs/:org/products/:product/devices?pagination[page]=N` and walk every page to collect all NH serial numbers (in memory).
2. Filter those serial numbers based on the _name_ of the deployment group they are in.
3. For each remaining serial, do a `POST` to NH's `"/devices/#{serial}/reconnect"` endpoint.

However, this ended up being a decent amount of devices and instead of abusing the NH API we decided to just cycle the ECS tasks for our NH deployment so that the hubs would all reconnect on their own anyway.

In an ideal world, NH would have a way of either:
1. Getting all device serials within a given DG.
2. Running a script on every device within a given DG.

Item 1 is probably simpler (and could supersede this PR). Item 2 would be more ideal in the long run but also be more work to implement bulk scripting safely.

nshoes:
I really don't want to make this change. We always come up against security concerns when this change is requested and I really don't see myself or others backing down from that position. We should close this and build the right tools instead.

> Getting all device serials within a given DG.

This is extremely easy, we could easily add a CSV export of device identifiers for a deployment group.

> Running a script on every device within a given DG.

This is still on the todo list. I have some considerations around running scripts in batches, as you guessed, but this will be coming soon™️ 

Thoughts?

PR #2931, opened 2026-08-21
Drop the DeviceSocket join_ref bandaid

joshk:
Bumps Phoenix to 1.8.12 and removes the `join_ref` workaround from `DeviceSocket`.

## Background

Older Slipstream releases only send a `join_ref` on `phx_join`, leaving it `nil` on every message after that ([CuatroElixir/slipstream#84](https://github.com/CuatroElixir/slipstream/issues/84)). Phoenix tightened its staleness check in [c73bbfc](https://github.com/phoenixframework/phoenix/commit/c73bbfc) and started discarding those messages. Since we can't force deployed devices onto a newer Slipstream, `DeviceSocket` decoded each frame, patched the `join_ref` back in, and re-encoded it before handing it to the channel.

[phoenixframework/phoenix#6800](https://github.com/phoenixframework/phoenix/pull/6800) relaxed the check. The guard used to only tolerate a nil ref on the stored side:

```elixir
# 1.8.11 - matches when the server's stored ref is nil (v1 protocol)
%{^pid => {^topic, existing_join_ref}} when existing_join_ref in [join_ref, nil] ->
```

Slipstream joins *with* a `join_ref`, so the stored one is non-nil, and subsequent messages fell through to the "stale, ignore" clause. 1.8.12 tolerates nil on the incoming side, which is exactly our case:

```elixir
# 1.8.12
%{^pid => {^topic, existing_join_ref}}
when is_nil(join_ref) or join_ref === existing_join_ref ->
```

Stale non-nil refs are still dropped. The workaround never affected those anyway. It only ever filled in refs that were nil.

## Changes

- `mix.lock`: phoenix 1.8.11 → 1.8.12, and bandit 1.12.4 → 1.12.5 pulled in transitively. No `mix.exs` change needed, `~> 1.8.4` already allowed it.
- `DeviceSocket`: dropped `maybe_fix_join_ref/2`. The `handle_in/2` override stays because `heartbeat/1` still needs it.
- Added a test covering a legacy client, plus the two client serializers behind it.

## On the test

The suite pins Slipstream 1.2.2, the fixed client, so nothing in it was exercising the path the workaround existed for. Removing the code without a legacy client on the wire would have been unguarded.

`SocketClient.LegacyJoinRefSerializer` and its msgpack counterpart send a `join_ref` on `phx_join` and `nil` on everything after, the way old Slipstream does. The new test joins over one of them, pushes `connection_types`, and asserts the device's connection metadata gets updated.

I pinned back to 1.8.11 to check the test actually fails without the fix:

- **1.8.11**: fails on both wire formats, metadata is `nil`, message silently dropped
- **1.8.12**: passes on both

## A bug that went with it

The removed code had a latent bug. It decoded with `socket.serializer` but re-encoded as JSON unconditionally, so a msgpack device sending a nil `join_ref` would have had JSON handed to the msgpack decoder. Probably unreachable in practice, since msgpack devices are on newer Slipstream, but it's gone now either way.


lawik (approved):

nshoes:
Finally 🙌

PR #2933, opened 2026-08-21
Make connection history versions strictly increasing

joshk:
`device_connection_history` is a ReplacingMergeTree that dedupes on
`(org_id, product_id, device_id, established_at)` - a key every row for a single
connection shares - and keeps the row with the highest `version`. So the
connecting, connected, heartbeat and disconnected rows for one connection are
competing on `version` alone, and if the wrong one wins the merged view reports
a connection that closed as still open.

`version` was a wall clock read, `DateTime.to_unix(DateTime.utc_now(), :microsecond)`,
which leaves two gaps:

- two rows can land in the same microsecond, and ClickHouse then picks between
  them arbitrarily. In a tight loop the old generator returned 24,436 unique
  values out of 80,000 reads, so the clock ties far more often than microsecond
  resolution suggests. Real writes for one connection are much further apart
  than that, but the margin is thinner than the comment claimed.
- `DateTime.utc_now/0` reads raw OS time, which steps backwards when the host's
  clock is adjusted. A backwards step between a heartbeat and the disconnect
  that follows it puts the disconnect behind, which is exactly the case
  `version` exists to prevent. No amount of resolution fixes that one.

`current_version/0` now reads `System.system_time/1`, which comes off the BEAM's
corrected clock and doesn't step backwards, and passes it through a node wide
atomic high water mark, so each value is `max(now, previous + 1)`.
`mark_as_stale_and_disconnected_changeset/1` uses the same generator instead of
deriving a version from its own `DateTime.utc_now/0` call.

The counter is created in `@on_load` rather than on first use. That isn't
stylistic: lazy creation is broken under concurrency, because `:persistent_term.put/2`
is slow enough that racing first callers each create and store their own counter.
Measured: 50,456 unique values out of 80,000. Creating it as the module
loads means it exists before any process can call the module.

### DIFF
--- a/lib/nerves_hub/devices/device_connection_history.ex
+++ b/lib/nerves_hub/devices/device_connection_history.ex
@@ -7,6 +7,12 @@ defmodule NervesHub.Devices.DeviceConnectionHistory do
 
   @type t :: %__MODULE__{}
 
+  @version_counter {__MODULE__, :version}
+
+  # Creating the counter as the module loads means it exists before any process
+  # can call the module, so no two writers can race to create one each.
+  @on_load :init_version_counter
+
   @primary_key false
   schema "device_connection_history" do
     field(:ref, Ch, type: "UUID")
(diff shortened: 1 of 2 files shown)


lawik (approved):
This seems good but might be more complex than needed. I don't have the bandwidth to think through it all but Erlang docs say this:

Determine the order of events by saving a tuple containing [monotonic time](https://www.erlang.org/doc/apps/erts/erlang.html#monotonic_time/0) and a [strictly monotonically increasing integer](https://www.erlang.org/doc/apps/erts/erlang.html#unique_integer/1) as follows:

```
Time = erlang:monotonic_time(),
UMI = erlang:unique_integer([monotonic]),
EventTag = {Time, UMI}
```

System.monotonic_time in Elixir is monotonic (increases or stays the same) and unique_integer is `strictly` monotonic (always increasing). This is true per node, which I think would be sufficient.

But feel free to roll with this for now.

PR #2943, opened 2026-08-23
Cut device node memory: zlib contexts and allocator retention

joshk:
Memory work on the device nodes, driven by measurements from a production machine holding 1420 devices at 1219MB RSS.

Where that 1219MB actually goes:

| | | |
|---|---|---|
| zlib permessage-deflate contexts | 376MB | 31% |
| freed but not returned to the OS | 395MB | 32% |
| process heaps | 239MB | 20% |
| loaded code | 79MB | 6% |
| ETS | 32MB | 3% |
| binaries | 19MB | 2% |

The first two are two thirds of the machine and neither is doing any work. This branch goes after both.

## zlib per connection

Both device sockets set `compress: true`, so Bandit builds a deflate and an inflate context per connection and holds them for the life of the device's connection. I measured it through `PerMessageDeflate.negotiate/2` rather than estimating: 270.6KB per device, 376MB across the fleet. A sample of 25 live connections came back compressed 25, plain 0, so every device is paying it.

Most of that is the deflate hash table, `1 <<< (mem_level + 9)` bytes, which is 128KB at the default mem_level of 8. Device frames do not need it. Over a representative mix of heartbeats, update progress and health reports, mem_level 8, 6, 4 and 2 all emit **byte-identical output**; only mem_level 1 gets worse, by 2.3%. At mem_level 4 a connection holds 150.5KB instead of 270.6KB, which is about 167MB back for no change in what goes over the wire.

The setting has to live in Bandit's server-wide `websocket_options`. Phoenix accepts `:deflate_options` in the per-socket `websocket:` list, but Bandit reads it from the server config, so putting it there would have silently done nothing.

If you would rather have the whole 376MB than keep compression, `compress: false` on both sockets does that. I did not take that route because devices on metered cellular are the ones paying for it, and that is a product call rather than a memory one.

## jemalloc

jemalloc has been LD_PRELOADed since it was added but never given a `MALLOC_CONF`, so it ran on defaults: four arenas per CPU, each holding its own freed pages, purging only when allocation traffic drives the decay. That is the 395MB gap between RSS and `:erlang.memory(:total)`.

**This commit is the one to be careful with.** The 395MB is measured; how much of it comes back is not. Roll it out to a single device machine and compare RSS before merging, and drop the commit if it does not move. `background_thread:true` would purge on a timer instead of on allocation traffic and would probably help more, but this image shells out to fwup and detools and jemalloc background threads across fork have a history of deadlocking, so it stays off.

## vm.args

Six of the eight allocator flags set values OTP already uses. Against the running VM:

```
erlang:system_info({allocator, binary_alloc}) -> options
lmbcs=5242880  smbcs=262144  as=aoffcbf  acul=0
```

eheap_alloc reports the same. So `+MBas`/`+MHas aoffcbf`, `+MBacul`/`+MHacul 0` and `+MBsmbcs`/`+MHsmbcs 256` did nothing except make the file look like more tuning had happened than had. The two `lmbcs` flags do change something and stay, but their comment claimed the reverse of their effect, so it is reworded.

## A process per device, to send one message an hour

1425 processes on the node were sitting in `:timer.interval_loop/5`, one per device, holding about 40MB between them. Since OTP 25 `:timer.send_interval/2` spawns a process per interval that lives as long as whoever asked for it, and `Extensions.Health` asks for one on attach so it can poll the device hourly. `Extensions.Geo` does the same.

`Process.send_after/3` uses the VM's timer wheel and spawns nothing: 364 bytes against 3.9KB for a freshly started interval, and the gap widens once an interval process has been running long enough to grow its heap, which is where the 28.4KB p50 on that node came from.

The wheel only does one-shots, so the repeat is re-armed in `Effects` when the message comes back. That happens in the channel before the message is dispatched, rather than by asking the extension to re-arm itself, because `Extensions.Dispatch` swallows an extension that raises and a re-arm missed that way would quietly end the timer for the rest of the connection.

## What I looked at and left alone

Channel processes are not a problem. DeviceChannel sits at a 3.7KB p50, ExtensionsChannel 3.8KB, ConsoleChannel 2.6KB, about 15MB for all 4265 of them. Phoenix's default `hibernate_after` is doing its job.

Console scrollback is not a problem either. Every device joins the console channel, and I expected those 1024-line buffers to be full; they total 3.6MB with a 2.7KB max, because devices join but never print.

`Sentry.Sources` holds 7.6MB on every node for stack-trace context. `enable_source_code_context` is compile-time and the same image runs web and device nodes, so switching it off for devices means a second build. Not worth it for 0.6%.

TLS is the largest slice of process memory at 62.8MB, but its p50 is 9.1KB against a 1089KB max, so it is a small tail of busy connections rather than a per-connection cost. Left alone.

## Testing

`mix check` is clean, and the full suite passes apart from two `UpdateToolTest` FAT/delta tests that fail the same way on an untouched `main` on my machine. `effects_test.exs` covers the repeat, cancellation, replacement and the process count; reverting to `:timer.send_interval/2` fails three of its five tests.

Every number above was measured on Elixir 1.20.3 / OTP 29, which matches `.tool-versions`, but the zlib figures come from a local VM rather than the production release.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -197,6 +197,18 @@ ENV LC_ALL=en_US.UTF-8
 COPY --from=jemalloc /usr/local/lib/libjemalloc.so.2 /usr/local/lib/libjemalloc.so.2
 ENV LD_PRELOAD=/usr/local/lib/libjemalloc.so.2
 
+# jemalloc was preloaded but never configured, so it ran on defaults: one arena
+# per CPU times four, each holding onto its own freed pages, and purging only
+# when allocation traffic happens to drive the decay. On a device node that
+# left ~395MB of a 1.2GB RSS as memory the BEAM had already freed.
+#
+# Fewer arenas concentrates that retention, and the shorter decay returns pages
+# sooner. `background_thread` would purge on a timer instead of on allocation
+# traffic and would help more, but this image shells out to fwup and detools,
+# and jemalloc background threads across fork have a bad history -- so it stays
+# off.
+ENV MALLOC_CONF=narenas:4,dirty_decay_ms:5000,muzzy_decay_ms:5000
+
 # Copy over the statically built fwup
 COPY --from=fwup /usr/local/bin/fwup /usr/local/bin/fwup
 
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -225,6 +225,22 @@ if config_env() == :prod do
         http_options: [
           log_protocol_errors: false
         ],
+        # The sockets set `compress: true`, so Bandit builds a zlib deflate and
+        # inflate context per connection and holds both for as long as the
+        # device stays connected. At the default mem_level of 8 the deflate
+        # hash table alone is 128KB, and measured end to end that is 271KB per
+        # device -- around 375MB of `:erlang.memory(:system)` on a node holding
+        # 1400 devices.
+        #
+        # Device frames are small and repetitive, so the hash table buys
+        # nothing here: over a representative mix of heartbeats, progress and
+        # health reports, mem_level 4 emits byte-identical output to mem_level
+        # 8. It costs 121KB less per connection.
+        #
+        # This has to live in Bandit's server-wide `websocket_options`. Bandit
+        # reads `deflate_options` from there, not from the per-socket
+        # `websocket:` list in the endpoint.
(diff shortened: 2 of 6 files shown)


lawik (approved):
Love this

PR #2946, opened 2026-08-23
Add RAUC bundle support, off by default

joshk:
Adds RAUC as a third update tool, alongside fwup and ESP-IDF. Off unless `RAUC_FIRMWARE_ENABLED` is set, and off per product until enabled in its settings.

RAUC is one of the two update tools that dominate Yocto-based Linux devices, and the one that fits NervesHub best: signing is mandatory in the format, and a device installs by streaming the bundle over HTTP range requests, fetching only the blocks its target slot does not already hold.

## No `rauc` on the server

A `.raucb` is a payload, a CMS signature, and an eight-byte big-endian length. For a verity bundle the manifest *is* the CMS content, so a single `openssl cms -verify` both checks the signature and returns the manifest. Installing `rauc` server-side would pull in glib, dbus and libcurl for something that never installs a bundle. It only ever reads one.

`openssl` is a runtime dependency, which nothing else in NervesHub has. The runtime image installs it; if it goes missing the upload fails with a message saying so rather than raising.

Plain-format bundles are refused with a message saying so. Their signature is detached and the manifest lives inside the SquashFS, so reading four lines of INI would mean unpacking a filesystem. Verity is what RAUC recommends and what streaming requires.

## No deltas, ever

`supports_deltas?/0` returns false. The saving is in what the device declines to fetch, so generating a patch would cost worker time and object storage to produce something nothing would ask for.

## The UUID

RAUC has no notion of one. It is derived from a SHA-256 over the manifest, which is the digest RAUC itself computes and reports as `hash`. Verified against a bundle built by RAUC 1.13: it matches byte for byte.

What is stored is the first 128 bits of that digest, formatted as a UUID, which is the same truncation RAUC applies when it renders `bundle.hash`. Matching RAUC's truncation is the point, but it does halve the collision resistance of the identifier, so the moduledoc says so rather than leaving it to be discovered.

That identity is what makes it usable rather than merely unique. RAUC records the hash against the slot it installs into (`bundle.hash`, persisted in its status file), so a device reads back the same value NervesHub derived at upload without having been told it.

## Requires RAUC 1.9 at both ends

Two unrelated requirements that happen to share a version:

- `[meta.<label>]` sections arrived in 1.9. Older versions do not reject them, they drop them while rewriting the manifest into the signature, so a bundle built from a manifest containing `architecture=` can arrive without it.
- `bundle.hash` was added to slot status in 1.9. A 1.8 device installs and reboots fine, then cannot say which firmware it is running.

Because of the first, a missing meta *section* is reported differently from a missing key. Reporting "add architecture" would send someone to look at a file that already says `architecture=`.

## Keys, and what they are trusted for

RAUC signs with CMS against an X.509 keyring, so `org_keys.scheme` gains `:x509_certificate`. It holds a certificate rather than a bare public key, because CMS verification is chain verification and a public key alone cannot anchor a chain.

Verification passes `-no-CApath` and `-no-CAstore` as well as `-CAfile`. `-CAfile` reads as "trust exactly this certificate" and does not mean it. It suppresses openssl's default CA *file* and leaves the default CA directory and store in play. Without the other two, a bundle whose signer chains to any root the host trusts verifies against an organization that never issued it, which on a stock Ubuntu image is every public CA. There is a regression test that fails without them.

If you sign with a leaf issued by your own CA, register the certificate that anchors the chain. An organization that registered only a leaf whose issuer it never registered was, until this, relying on the host's trust store to fill the gap.

## One thing to know when pulling this branch

`.raucb` had to be registered as a MIME type, because `allow_upload` refuses an
extension it cannot resolve to one. `mime` reads that config at compile time, so
a stale `_build/mime` raises when the firmware page mounts and takes out the
firmware index for every product, not just RAUC uploads. `mix deps.compile mime
--force` fixes it. A fresh build is unaffected.

## Test endpoint ports

`config/test.exs` reads `DEVICE_ENDPOINT_TEST_PORT` and `WEB_ENDPOINT_TEST_PORT`, defaulting to what they were. Unrelated to RAUC. It lets two worktrees run the suite at the same time without fighting over a port.

## Testing

Bundle fixtures are built with openssl rather than `rauc`, so a test can vary the manifest, sign with the wrong key or corrupt the footer. Two bundles built by real RAUC are checked in as well, doing different jobs: a 1.8 one covers the dropped-meta error path, and a 1.13 one covers the happy path and asserts the UUID matches the digest RAUC reports.

The upload path is covered end to end: tool chosen from the bytes rather than the filename, signature checked against a key the organization holds, product gating, product mismatch, missing certificate, wrong signer, and the instance flag stopping uploads rather than only hiding a checkbox. A LiveView test drives a real `.raucb` through the upload, which is the only level at which the file picker's accept list can be wrong.

Several tests exist to pin things that are correct now and have nothing else stopping them drifting: that no product setting can excuse an unverified bundle, that openssl still says "no content" for a detached signature, that each class of bad footer produces its own error, and that a manifest edited inside a valid CMS is refused.

`mix check` is clean. The device-side counterpart lives in `nerves-hub-link-agent`; the join parameters it sends are pinned by a test here, since that contract spans two repositories.

Documentation is in `docs/rauc_support.md`, alongside the ESP-IDF and AtomVM guides.


lib/nerves_hub_web/live/product/settings.ex
nshoes, line 171: The `"rauc"` map from `@optional_tools` doesn't have a `unsigned_field` prop, so this will fail with a `CaseClauseError`.

lib/nerves_hub/firmwares/update_tool/rauc.ex
```diff
@@ -0,0 +1,504 @@
...
+
+        String.starts_with?(line, "[") and String.ends_with?(line, "]") ->
+          name = line |> String.slice(1..-2//1) |> String.trim()
+          {name, Map.put_new(sections, name, %{})}
+
+        section != nil ->
+          case String.split(line, "=", parts: 2) do
+            [key, value] ->
+              entry = {String.trim(key), String.trim(value)}
+              {section, Map.update(sections, section, Map.new([entry]), &Map.new([entry | Map.to_list(&1)]))}
```
nshoes: Claude said:

---
  The intent is "new entry wins" (prepend new, then Map.new). But Map.new on a list with duplicate keys keeps the last occurrence. Since the existing key is appended after the new entry in the list, the old value survives. Result: repeated keys
  retain their first value, not their last.

  The RAUC INI format (GLib key file) is unambiguous that later values override earlier ones. A manifest with duplicate keys will silently use the wrong value. Not likely to occur in practice with well-formed RAUC bundles, but worth fixing for
  correctness.

  Fix:
  Use Map.put to let new entry override
  {section, Map.update(sections, section, Map.new([entry]), &Map.put(&1, elem(entry, 0), elem(entry, 1)))}

---

I think it's worth a fix and a test, but since Claude hedged with "Not likely to occur in practice...", I'll leave it up to you.

lib/nerves_hub_web/live/product/settings.ex
joshk: Good catch, and it was a real crash rather than only a latent one. Fixed in 9b6510b4 with a `%{}` clause that no-ops, plus a test.

I confirmed it by removing the fix and watching the test fail with `CaseClauseError`. The template gates the toggle on `tool[:unsigned_field]` so it was never clickable, but the event is reachable by anything that can send one, so it was a crash from user-controllable input.

lib/nerves_hub/firmwares/update_tool/rauc.ex
joshk: Fixed in 9b6510b4, with two tests.

I would push back on the "not likely to occur in practice" hedge that made this optional. It is unlikely, but the consequence is not cosmetic: `compatible` is what `product` and `platform` are derived from, so a manifest with a duplicate would have NervesHub and `rauc` reading different values out of the same bundle. That is the one thing this design cannot afford, since the whole metadata scheme rests on the two agreeing.

Confirmed the behaviour before changing it:

```
[update]
compatible=first
compatible=second
```

parsed to `%{"compatible" => "first"}`. `Map.new/1` keeps the last occurrence of a duplicate key, and the older entry was appended after the new one, so the older value survived. GLib key file semantics are last-wins.

Now uses `Map.put`, with tests for both a fresh section and a reopened one.

PR #2955, opened 2026-08-25
Document runtime.exs configuration, drop unused settings

joshk:
There was no single place to look up what `config/runtime.exs` reads, so `docs/runtime_configuration.md` now lists every environment variable in it, with its default and what it does.

Entries are grouped by concern (application role, the two endpoints, Postgres, ClickHouse, uploads, email, accounts, retention, extensions, branding, logging, tracing), and each one notes whether it is prod-only and whether it is required. Required means read with `fetch_env!`, so a missing one refuses the boot rather than falling back to something. There is a short "minimum production configuration" block at the top for anyone standing up a self-hosted instance.

Two behaviours the doc now states outright:

- Booleans are compared against the exact string `"true"`, so `DATABASE_SSL=1` silently disables TLS.
- Integers are parsed strictly, so a non-numeric value crashes the boot.

`AGENTS.md` gains a line asking that the doc be kept in step when a variable is added or removed.

## Removed settings

Writing the doc turned up five settings that are parsed at boot but that nothing reads:

| Variable | Config key |
| --- | --- |
| `DASHBOARD_ENABLED` | `:dashboard_enabled` |
| `DEPLOYMENT_CALCULATOR_INTERVAL_SECONDS` | `:deployment_calculator_interval_seconds` |
| `DEVICE_DEPLOYMENT_CHANGE_JITTER_SECONDS` | `:device_deployment_change_jitter_seconds` |
| `DEVICE_CONNECTION_MAX_AGE_DAYS` | `:device_connection_max_age_days` |
| `DEVICE_CONNECTION_DELETE_LIMIT` | `:device_connection_delete_limit` |

I checked each one by config key and by variable name across the repo; every hit was in `runtime.exs` itself. Nothing calls `Application.get_all_env(:nerves_hub)`, so none of them could be picked up indirectly, and none is set in `config.exs`, `dev.exs` or `test.exs`. The similarly named keys that are used (`device_health_delete_limit` and `device_connection_update_limit`) are untouched.


nshoes (approved):
Sweeeet

PR #2962, opened 2026-08-26
Let a device send a second of log lines in one message

joshk:
`logging:send` carried one log line. NervesHub limits how often a device may
send, not how much it may say: five messages a second per device, silently
dropped past that. One line per message turns that into five *lines* a second,
so a device in a crash loop loses almost everything it writes, and what
survives is an arbitrary sample rather than the interesting part.

## The change

The event now also takes a batch:

```elixir
%{"lines" => [%{"level" => "info", "message" => "..."}, ...]}
```

A batch costs the same one token as a single line, so a device can buffer for a
second and send the lot. That is the point of the whole thing: the same rate
budget now buys a second of logs per message instead of one line.

Single lines are still accepted from any device. Not as a transitional
kindness: the AtomVM ESP32 client and everything else on the 0.0.x extension
sends one line per message and always will.

## Which devices may batch

The version on the extensions join decides. 0.0.x is the single-line extension;
0.2.0 says the device may batch.

The gap between them is load-bearing. `~> 0.0.1` does not match 0.2.0, so a
NervesHub that predates this refuses the extension outright rather than
attaching it and dropping every batch it cannot parse. A device that is told
logging was not attached can report that; lines that vanish server-side are the
silence this is meant to end.

That also means an agent declaring 0.2.0 gets no logging at all until this is
deployed, which is the safe direction: visible and diagnosable rather than
quiet.

## The cap

A message may carry 100 lines. Past that they are dropped and the count is
stored as a log line of its own, so the gap shows up in the UI where the lines
would have been. Worst case per device is unchanged in messages and bounded at
500 lines a second into ClickHouse.

## Two things found on the way

`LogLine.format_message/1` raised on a line with no `"message"` key instead of
leaving it to `validate_required/2`. With batching that is worse than untidy:
one malformed line takes out every other line in the same message and lands in
Sentry rather than in a changeset error. It now falls through to validation.

`log_line_payload` described a map with atom keys. Every caller passes string
keys, which is what arrives over the channel and what `cast/3` needs. An
atom-keyed payload is not a stricter version of it, it is one Ecto refuses.
Dialyzer only caught it because the new call site made it provable.

## Tests

`test/nerves_hub/extensions/logging_test.exs` covers both payload shapes, the
version gate in both directions, the cap and its dropped-line notice, and one
malformed line not taking its neighbours with it. The one that matters most:
three batches of 20 store all 60 lines, where 60 single messages would have
lost all but the burst allowance.

`mix check` is clean, and the log lines, extensions, extensions channel, logs
tab and device log controller suites pass (85 tests, run against a scratch
database).

Pairs with nerves-hub/nerves-hub-link-agent#10, which is the agent side. This
one goes first, since an agent declaring 0.2.0 gets no logging until it is
deployed.

### DIFF
--- a/lib/nerves_hub/extensions/logging/batched.ex
+++ b/lib/nerves_hub/extensions/logging/batched.ex
@@ -48,7 +48,15 @@ defmodule NervesHub.Extensions.Logging.Batched do
     {state, []}
   end
 
+  # Nothing to store, and nothing charged for it. The budget is what a device
+  # needs for the second it does have something to say, and spending it on a
+  # message that carried no lines would take rate limiting further than it is
+  # meant to go.
   @impl NervesHub.Extensions
+  def handle_in("send", %{"lines" => []}, state) do
+    {state, []}
+  end
+
   def handle_in("send", %{"lines" => lines}, state) when is_list(lines) do
     if Logging.allow?(state.device_info) do
       {kept, dropped} = Enum.split(lines, @max_lines_per_message)
@@ -56,6 +64,9 @@ defmodule NervesHub.Extensions.Logging.Batched do
       _ = LogLines.async_create_many(state.device_info, kept)
       _ = record_overflow(state.device_info, length(dropped))
 
+      # Both branches of the `if` end on an atom so that nothing complex is
+      # discarded here. Without it dialyzer reports the changeset tuple this
+      # would otherwise return as an unmatched return.
       :noop
     end
 
--- a/lib/nerves_hub/extensions/logging/batched.ex
+++ b/lib/nerves_hub/extensions/logging/batched.ex
@@ -1,0 +1,89 @@
+defmodule NervesHub.Extensions.Logging.Batched do
+  @moduledoc """
+  Storing the log lines a device sends, a second's worth per message.
+
+  Version 0.1.0 of the logging extension. `NervesHub.Extensions.Logging` is
+  0.0.1, where one message carries one line.
+
+      logging:send  %{"lines" => [%{"level" => .., "message" => ..}, ..]}
+
+  ## Why a second's worth
+
+  NervesHub limits how often a device may send, not how much it may say, and a
+  batch costs the same one token as a single line. That is the whole point: a
+  device in a crash loop can report everything it wrote in the last second
+  rather than losing all but the first few lines to the limiter.
+
+  What stops that becoming an unbounded write is `max_lines_per_message/0`.
+  Anything past it is dropped and the count is stored as a log line of its own,
+  which is the same bargain the device's own buffer makes: a gap someone can
+  see beats a gap they cannot.
+  """
+
+  @behaviour NervesHub.Extensions
+
+  alias NervesHub.Devices.LogLines
+  alias NervesHub.Extensions.Logging
+
+  require Logger
+
+  @max_lines_per_message 100
+
+  @doc "How many lines one message may carry."
+  def max_lines_per_message(), do: @max_lines_per_message
+
+  @impl NervesHub.Extensions
+  defdelegate description(), to: Logging
+
+  @impl NervesHub.Extensions
+  defdelegate enabled?(), to: Logging
+
+  @impl NervesHub.Extensions
+  def attach(state) do
+    {state, []}
+  end
+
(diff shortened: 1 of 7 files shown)


lib/nerves_hub/extensions/logging/batched.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  @impl NervesHub.Extensions
+  def handle_in("send", %{"lines" => lines}, state) when is_list(lines) do
+    if Logging.allow?(state.device_info) do
+      {kept, dropped} = Enum.split(lines, @max_lines_per_message)
+
+      _ = LogLines.async_create_many(state.device_info, kept)
+      _ = record_overflow(state.device_info, length(dropped))
+
+      :noop
```
nshoes, line 70: No need for this?

lib/nerves_hub/extensions/logging/batched.ex
```diff
@@ -0,0 +1,89 @@
...
+  @impl NervesHub.Extensions
+  def handle_info(_, state) do
+    {state, []}
+  end
+
+  defp record_overflow(_device_info, 0), do: :ok
+
+  defp record_overflow(device_info, dropped) do
+    LogLines.async_create(device_info, %{
+      "timestamp" => DateTime.utc_now(),
```
nshoes, line 95: This just gets passed through the `LogLine.create_changeset/3` (lib/nerves_hub/devices/log_line.ex:34), is that expected?

lib/nerves_hub/extensions/logging/batched.ex
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  @impl NervesHub.Extensions
+  def detach(state) do
+    {state, []}
+  end
+
+  @impl NervesHub.Extensions
+  def handle_in("send", %{"lines" => lines}, state) when is_list(lines) do
+    if Logging.allow?(state.device_info) do
```
nshoes, line 61: Is it possible for the `lines` arg to be an exmpty list? It would still eat a rate limit hit if it was.

nshoes (approved):

lib/nerves_hub/extensions/logging/batched.ex
joshk: It reads like a stray value but it is load bearing, which means it needed a comment rather than removing. Added one in f5b5e7d.

Without it the `if` returns `{:ok, %LogLine{}} | {:error, %Ecto.Changeset{}} | nil` and nothing matches on it, so dialyzer reports an unmatched return and `mix check` fails. Ending both branches on an atom is what the `case` in the original single line version was doing too, so this keeps that.

The alternative is wrapping the whole `if` in `_ =`, which also passes. I went with the comment since `:noop` is what was here before.

lib/nerves_hub/extensions/logging/batched.ex
joshk: Yes, expected. `maybe_set_timestamp/1`'s first clause is `%{"timestamp" => _} = params`, so it passes straight through to `cast/3`, which is what that clause is for: devices send `meta.time` as microsecond strings and get converted, internal callers hand over a real `DateTime`. `LogLinesTest` "create!/2" already does the same thing and asserts the value round trips.

Going through `async_create/2` rather than writing the row directly is also deliberate for the other half of what it does: the notice gets broadcast on `logs:received` like any other line, so it shows up in an open logs tab live instead of only on reload. A gap that only appears after a refresh is halfway back to being invisible.

lib/nerves_hub/extensions/logging/batched.ex
joshk: Good catch, and yes it was possible: nothing on the wire stops a device sending `%{"lines" => []}`, and it cost a token to store nothing. Fixed in f5b5e7d with a clause above this one that returns before the limiter is asked. The budget is what a device needs for the second it does have something to say.

The agent never sends one (it only flushes when its buffer is non empty) so this is a guard against a client bug, but a client bug shouldn't cost the device the lines it sends next.

There's a test for it: ten empty batches, then a real one, and the real one has to be stored. Ten is the burst allowance, so it fails if empties are charged for.

PR #2965, opened 2026-08-26
Test suite improvements

nshoes:
This PR increases test coverage and makes the test suite faster:

### Before
```
Finished in 198.0 seconds (122.4s async, 75.6s sync)

Result: 2028 passed

Coverage:   80.05%
```

### After
```
Finished in 189.8 seconds (72.4s async, 117.3s sync)

Result: 2580 passed

Coverage:   89.31%
```


joshk (approved):
TLDR LGTM

PR #2967, opened 2026-08-28
Performant device counts in UI

nshoes:
This PR:

1.) Increases the performance of an org's products query with device counts. I profiled against our production database; the worst case, taking 16s, was cut down in half to 8.
2.) This of course is still too slow, so now the counts load async. Each online/offline pair is now it's own query.
3.) When viewing an org, the device counts now load async as well.
4.) We noticed the insights page was also slower, so the four device counts are now async as well.

Async content is represented with a dash. It loads quick local, so look closely 😆 

<details><summary>Orgs and their products list</summary>
CleanShot 2026-08-27 at 18 06 37
</details> 

<details><summary>Org's products</summary>
CleanShot 2026-08-27 at 18 09 09
</details> 

<details><summary>Insights</summary>
CleanShot 2026-08-27 at 18 10 06
</details>


kybishop (approved):
This is dope!

lib/nerves_hub_web/live/org/show.ex
```diff
@@ -5,17 +5,28 @@ defmodule NervesHubWeb.Live.Org.Show do
...
 
     socket
     |> page_title("Products - #{scope.org.name}")
     |> assign(:org, scope.org)
     |> assign(:products, products)
     |> assign(:product_device_info, %{})
     |> sidebar_tab(:products)
+    |> then(fn socket ->
+      Enum.reduce(products, socket, fn product, acc ->
+        assign_async(acc, :"product_counts_#{product.id}", fn ->
```
joshk, line 18: I'd suggest using strings and not atoms for the async keys as they will be cleaned up. This mostly affects NervesCloud where we have many orgs and products, and thus creating atoms which would just stick around taking up some space.

lib/nerves_hub_web/live/orgs/index.ex
```diff
@@ -34,6 +34,14 @@ defmodule NervesHubWeb.Live.Orgs.Index do
       |> assign(:device_limit, @pinned_devices_limit)
       |> maybe_assign_onboarding()
       |> subscribe()
+      |> then(fn socket ->
+        Enum.reduce(orgs, socket, fn org, acc ->
+          assign_async(acc, :"org_counts_#{org.id}", fn ->
+            key = :"org_counts_#{org.id}"
```
joshk, line 40: same here as well

joshk (approved):
Love this. Some minor suggestions. Thank you!

lib/nerves_hub_web/live/org/show.ex
nshoes: Two things:
1.) Atoms are the only thing that `assign_async/3` supports as its second argument
2.) We need _28,000_ `:org_counts_123` atoms to fill 1MB of memory. We could even shorten them to something silly for even more room `:oc_123` needs 40,000 atoms to fill 1MB of memory.

While I completely agree with your intent, I don't think there's another option here.

joshk: Soooo, you are right that strings are allowed, but it looks like lists are https://github.com/phoenixframework/phoenix_live_view/blob/v1.2.11/lib/phoenix_live_view/async.ex#L130-L131

Which means we can use `[:product_counts, product.id]`

joshk: hmmmm, i also see that https://github.com/nerves-hub/nerves_hub_web/pull/2942/changes#diff-35199de47549e03063cb468e3e342604e193d45d0dcaeab2cd0d46e8ec8109c5R441 had us switch to strings. hmmmmm.

nshoes: Good find on both accounts. I'll implement the list + atom arg.

Claude kindly told me the difference, whish is that `start_async` doesn't interact with socket assigns, so it can use anything as a key.

nshoes: Oops, that doesn't work either:
```
** (ArgumentError) assigns in LiveView must be atoms, got: 155752
```
So now we're back to just implementing atoms 😬

PR #2978, opened 2026-08-29
Device-managed firmware updates

joshk:
Lets a device decide *when* it takes firmware. Its deployment group still decides *what*.

Server side; the client half is nerves-hub/nerves_hub_link#473. Protocol and rules in `docs/device_managed_updates.md`.

## The mode

`updates_enabled` becomes a three-value `update_mode`, migrating `true → :automatic` and `false → :off`.

| Mode | Orchestrator pushes | Device may pull | Manual push |
| --- | --- | --- | --- |
| `automatic` | yes, on the deployment's schedule | yes, if eligible | yes |
| `device_managed` | never | yes, if eligible | yes |
| `off` | never | never | yes |

The deployment group names the target firmware in every mode — `device_managed` picks a moment, not a build. Pulling is gated on **eligibility, not mode**: a device that hasn't reached its wave in a staged rollout is refused whatever mode it's in.

A device-initiated request takes a deployment concurrency slot like any other update, so a fleet waking at 03:00 local can't walk past the rollout pacing. None free means `update_rejected` with a delay.

## The grant

`managed_updates_allowed` is whether a device may put *itself* into `device_managed`. It defaults to `false`, so the feature stays inert until enabled per device or in bulk.

Two rules bound a device-initiated change:

- **`off` is the operator's alone** — a device can neither freeze itself out of reach nor unfreeze itself
- **`device_managed` needs the grant**

Neither binds an operator. Revoking the grant does not move a device already in `device_managed`.

## Messages

```text
server -> device   update_mode       {"mode": "automatic", "managed_updates_allowed": false}
device -> server   check_update      {}
server -> device   update_available  {"available": true, "firmware_meta": {..}}
device -> server   request_update    {}
server -> device   update            (the existing message, unchanged)
server -> device   update_rejected   {"reason": "busy", "delay_for": 5}
device -> server   set_update_mode   {"mode": "device_managed"}
```

`request_update` answers with the **existing** `update` message, so this adds no new firmware-application path on the device. `check_update` works in every mode including `off`, and carries no URL — they're signed and time limited, so one is fetched when it's about to be used.

`device_api_version >= 2.4.0` gates what the server sends *unasked*, not what it accepts.

## UI

The device page header's two-state switch is now an icon and caret opening a menu of the three modes. The settings page carries only the grant. Clearing the penalty box is an × rather than a switch.

Screenshot 2026-08-29 at 9 55 12 PM

Screenshot 2026-08-29 at 9 55 07 PM

Screenshot 2026-08-29 at 9 54 58 PM

## Firmware too old to manage itself

A `device_managed` device connecting on `< 2.4.0` would be stranded — not pushed to, unable to ask, and never told its mode. Most likely arrived there by firmware auto-revert. It's now returned to `:automatic` on join, audited with the device as actor.

⚠️ **Roll the firmware out before setting the mode.** A fleet set to `device_managed` before its firmware supports it will be moved to `automatic` as devices connect, and start receiving pushes.

## Compatibility

`devices.updates_enabled` (backfilled, unread) and `DeviceInfo.device_updates_enabled` both stay for one release, so a rolling deploy and a rollback work. The JSON API keeps reporting `updates_enabled`, derived as `update_mode != :off`, and gains `update_mode`. The query language keeps `updates = "enabled"` / `"disabled"` and gains `"automatic"` / `"device-managed"`.


lawik:
> A device_managed device connecting on < 2.4.0 would be stranded — not pushed to, unable to ask, and never told its mode. Most likely arrived there by firmware auto-revert. It's now returned to :automatic on join, audited with the device as actor.

is this checked on the nerves_hub_link version? Seems like the device could report that ability on join as part of the metadata so we know something is taking responsibility for that on the other end. Rather than gate on the version.

joshk:
> is this checked on the nerves_hub_link version? Seems like the device could report that ability on join as part of the metadata so we know something is taking responsibility for that on the other end. Rather than gate on the version.

The NHL version is essentially the contract of features supported. You do make a good point as we grow our libraries and integrations. I think, for this PR at least, using the version is the right gate, but we should consider a v3 API spec which supports SDKs/integrations announcing their 'capabilities'.

lawik:
I disagree :)

Yes the version number indicates capabilities but here we could fairly easily make it foolproof and backwards /forwards compatible by just passing a field like "device_manages_updates: true".

this means there is explicit opt in as part of defining a management policy on the device. The defaults keep working and you can't end ip in a state where a device and the server disagrees on who runs the updates. Ie. a device is set to manage itself when it can't but has a sufficiently new version.

The server won't be confused and the device won't be confused. We can still gate on the version as well if that server as purpose.

We already have other client implementations. We might as well take care of them I reckon.

joshk:
Lars and I chatted offline and we agreed that punting `capabilities` to another PR is best as we want to make sure we get that right.

lib/nerves_hub/device_link.ex
```diff
@@ -83,6 +92,26 @@ defmodule NervesHub.DeviceLink do
...
+  # firmware that can manage itself.
+  defp revert_update_mode_if_unsupported(%{device_update_mode: :device_managed} = device_info, version) do
+    if Version.match?(version, @update_mode_api_version) do
+      device_info
+    else
+      device = Devices.get_device(device_info.device_id)
+
+      case Updates.revert_unsupported_update_mode(device) do
+        {:ok, device} -> %{device_info | device_update_mode: device.update_mode}
+        _error -> device_info
```
nshoes: How about an error log here? If this fails, the update mode doesn't change, and the orchestrator won't push out an update. A log to help debugging would be better than nothing.

lib/nerves_hub/events/device_events.ex
```diff
@@ -100,6 +111,40 @@ defmodule NervesHub.DeviceEvents do
...
+
+      true ->
+        {:ok, _inflight} = schedule_update(device.id, deployment_group, initiated_by: :device)
+        :ok
+    end
+  end
+
+  # Long enough that a refused fleet does not come straight back, short enough
+  # that a device is not left sitting on old firmware once slots free up.
+  defp retry_after_minutes(_deployment_group), do: 5
```
nshoes: This seems like a silly Claude thing 😆

lib/nerves_hub_web/controllers/api/openapi/device_controller_specs.ex
```diff
@@ -54,7 +54,11 @@ defmodule NervesHubWeb.API.OpenAPI.DeviceControllerSpecs do
         platform: %OpenApiSpex.Schema{type: :string, required: false, example: "rpi4"},
         search: %OpenApiSpex.Schema{type: :string, required: false, example: "sn123"},
         tags: %OpenApiSpex.Schema{type: :string, required: false, example: "prod,staging"},
-        updates: %OpenApiSpex.Schema{type: :string, required: false, enum: ["enabled", "disabled", "penalty-box"]}
+        updates: %OpenApiSpex.Schema{
+          type: :string,
+          required: false,
+          enum: ["enabled", "disabled", "automatic", "device-managed", "penalty-box"]
```
nshoes, line 60: Why did you opt for a hypen separators here? Why not match the database-backed value with an underscore?

nshoes (approved):

lib/nerves_hub_web/controllers/api/openapi/device_controller_specs.ex
joshk: I felt that these read better than using underscores.

lib/nerves_hub/events/device_events.ex
joshk: Thanks for pointing this out

lib/nerves_hub/device_link.ex
joshk: AAAAAAND fixed

PR #2979, opened 2026-08-30
Add "Updated by" and "Updated at" for support scripts

### LINKED ISSUE #2891: Improve traceability of support scripts
Support scripts has no way to tell whether it was recently modified and what was modified, making it a little more difficult to debug issues. When a script results in an error, it sometimes leads to an investigation of whether the support script was changed/broken or if a device's firmware changed. But 

It would be nice to have:
- A section for "last edited at: " or something similar. 
- A way to see previous versions of the support script, so rollbacks can be easy.

gabjc:
This PR:
- Adds "Updated at" and "Updated by" columns in the support script list.
- Adds `Last edited by {User} on {date} at {time} UTC` to the edit page.
- Fixes a crash when sorting by "Updated by".
- Falls back to the script's creator when it has never been edited so every script shows a name rather than an empty cell.

Closes https://github.com/nerves-hub/nerves_hub_web/issues/2891
I wanted to figure out a way to enable versioning on support scripts, but that may require a lot more changes.

<details><summary>Creating a new script</summary>
new-script-list
new-script-edit
create-script
</details> 

<details><summary>Editing a support script</summary>
edit-script
</details>


lib/nerves_hub_web/live/support_scripts/index.html.heex
```diff
@@ -46,6 +46,12 @@
             <th phx-click="sort" phx-value-sort="inserted_at" class="cursor-pointer">
               <Sorting.sort_icon text="Added on" field="inserted_at" selected_field={@current_sort} selected_direction={@sort_direction} />
             </th>
+            <th phx-click="sort" phx-value-sort="updated_at" class="cursor-pointer">
+              <Sorting.sort_icon text="Updated at" field="updated_at" selected_field={@current_sort} selected_direction={@sort_direction} />
```
joshk: If we are going to show last updated at, I think we should remove 'Addon on'.

joshk: And can you change the header to "Last updated"

joshk (changes requested):
Thanks for the PR. 

This is generally good, just a few changes.

Also, for future PRs, please don't include a screenshot of the test run, that doesn't add anything useful.

lib/nerves_hub_web/live/support_scripts/index.html.heex
gabjc: Here is what it looks like now:
last-updated-no-added
last-updated

joshk (approved):

joshk:
Thanks for the PR

PR #2982, opened 2026-08-31
Add error reports as an opt-in device extension

joshk:
Adds exception and error reporting from devices, as an opt-in extension. Devices report exceptions, exits and explicit application errors; NervesHub groups them into issues you can resolve or mute, on a new device tab and a product-wide page.

The device half is not in this PR. `docs/error_reports.md` is the contract clients get written against.

## What it looks like

A new Errors tab on the device page lists the issues that device has hit, with the count on *that* device and the latest stacktrace behind each row. A new Errors page per product lists the same issues across the fleet, with an occurrences-over-time chart, the affected devices, and resolve / mute / reopen.

Off by default. Turn it on per product in settings, and a device can still opt out of it individually, like every other extension.

## The storage split

Two stores, joined by a fingerprint.

`error_groups` in PostgreSQL holds one row per distinct error per product. It is mutable (status, counters). It is bounded: one row per bug rather than one per crash. And it outlives its occurrences, which are dropped after thirty days, so "this has happened forty thousand times since March" is still answerable in June.

`device_error_reports` in ClickHouse holds one row per occurrence, at fleet scale, through the existing analytics buffer. Stacktraces, context, device attribution, time series. It is read on drill-down and never on a list page, which is the whole reason the list page lives in Postgres.

## Grouping

The fingerprint is computed on the server, not the device, so the rule lives in one place and can change without waiting for a fleet to update. It ignores line numbers (a line moves whenever the file above it changes), frames past the third, and the per-occurrence noise in a reason: pids, refs, addresses, numbers of two digits or more.

`fingerprint_version` is part of the hash. Bump it and new occurrences group into new issues rather than silently re-grouping history against a rule that did not produce it. Expect to bump it at least once after seeing real fleets; a report that knows better than the stacktrace can also supply its own key, which is what makes "every failure in the payment integration is one issue" possible.

## Why the write path looks like this

A crash loop is the case that shapes it, and it arrives when the platform is busiest.

Reports are batched per message rather than per line, for the same reason logging went to 0.1.0: the limit is on how often a device may send, so a batch of twenty-five costs what one costs, and a single-report version would throw most of a crash loop away and keep an arbitrary sample.

Group upserts are coalesced per node by `ErrorReports.GroupBuffer`. Hundreds of crashes a second all target the *same* row, so upserting per report would put every device node in a lock queue on it. Coalescing also makes the statement legal, since Postgres refuses an `ON CONFLICT DO UPDATE` that touches one row twice. Reopening a resolved issue is decided in that conflict clause, so it costs nothing extra.

## Device vitals ride in `context`

Uptime, reboot count and free memory travel as context entries rather than columns. Those three are what a BEAM device happens to report; an ESP32 with free heap and a signal strength to send instead would otherwise need three more columns and a migration. The UI renders recognised keys with friendly units, and that knowledge lives only in the view.

`firmware_uuid` is the exception and is a column. It answers which release broke something, it is carried on the issue as well as the occurrence, and the platform fills it in from the connection when a device omits it.

## Before this ships

The reopen rule is blunt. Any occurrence after `resolved_at` reopens the issue. On a fleet still part-way through an update that means it reopens immediately, and mute is the honest answer there. `resolved_in_firmware_uuid` is stored from the start so the better rule (reopen only for firmware built after the fix) can land without a migration. Try it against a real fleet before we rely on it.

Occurrence retention is a ClickHouse TTL rather than a runtime setting, deliberately: a variable read at migration time would work on fresh installations and quietly do nothing on existing ones. `docs/error_reports.md` gives the `ALTER TABLE ... MODIFY TTL` that does apply to stored rows.

"Language-agnostic" is a claim until a second client exists. The payload costs nothing to shape this way now and is close to impossible to retrofit, but only writing the Rust or ESP-IDF side will prove it.

## Second commit

Separable, and found while building the first. Every inactive tab's `cleanup/0` deletes the assigns it names, in `@tab_components` order, so a tab that cleans a key another tab *sets* deletes it whenever it sits later in the list, surfacing as a KeyError in an unrelated tab's render. A new tab doing exactly that took out the Data History tab and thirteen unrelated tests.

`TabCleanupTest` reads the assign keys each tab sets out of its AST and fails if any tab's `cleanup/0` names one another tab assigns. It found two more cases that only list order was keeping safe. Data History cleaned `:analytics_enabled` and `:streaming_enabled`, which Logs and Errors both set. Details cleaned `:available_tags`, which is worse, because Settings stores something different under that name; that one is now `:addable_tags`.


lib/nerves_hub/error_reports/error_group.ex
```diff
@@ -0,0 +1,162 @@
...
+  @spec mute_changeset(t(), User.t()) :: Ecto.Changeset.t()
+  def mute_changeset(%__MODULE__{} = group, %User{} = user) do
+    change(group, %{status: :muted, muted_at: DateTime.utc_now(), muted_by_id: user.id})
+  end
+
+  @doc """
+  Returns a group to the queue, from either resolved or muted.
+  """
+  @spec reopen_changeset(t()) :: Ecto.Changeset.t()
+  def reopen_changeset(%__MODULE__{} = group) do
```
nshoes, line 152: Should `regressed_at` be cleared here?

nshoes (approved):

lib/nerves_hub/error_reports/error_group.ex
joshk: reopen is used in the UI, so it shouldn't modify if/when the error group regressed

PR #2988, opened 2026-09-01
Deployment workflows: staged firmware rollouts

joshk:
A deployment group can now carry a workflow: a JSON definition, uploaded on its settings tab, that turns a rollout into ordered stages rather than one pool of devices. A handful of canaries first, then a wider group, then everyone, with someone's approval in between if you want one.

Nothing changes for a deployment group without a definition. It takes the same path it always did.

## Screenshots

Screenshot 2026-09-01 at 9 11 18 PM

Screenshot 2026-09-01 at 9 13 56 PM

Screenshot 2026-09-01 at 9 21 40 PM

## Defining one

```json
{
  "version": 1,
  "steps": [
    {
      "name": "Canary",
      "description": "wifi and ethernet canary devices",
      "matching_conditions": {
        "tags": ["canary"],
        "network_interfaces": ["wifi", "ethernet"],
        "match_limit": 20
      },
      "concurrent_updates": 10,
      "failure_tolerance": { "devices": 1 }
    },
    {
      "name": "Product sign-off",
      "type": "approval_required"
    }
  ]
}
```

Definitions are checked against a JSON schema on upload, and every problem is reported with the path to it: `steps/1/concurrent_updates: Type mismatch. Expected Integer but got String.` A misspelled key is an error rather than a setting that silently does nothing.

## How it runs

Creating a release expands the definition into steps belonging to that release, with a `catch_all` appended. The orchestrator picks its coordinator from whether the current release has steps.

One step is active at a time. It covers a set of devices decided once and recorded: `match_limit: 20` means twenty canaries, and re-matching every few seconds would pick a different twenty as devices connect and disconnect, so a step would never finish. Which devices a step covers ignores whether they are connected. An offline canary is still a canary, and leaving it out would let a stage be declared good without having been tried. Connectivity is a scheduling question, answered separately against the same eligibility rules the existing coordinator uses.

A step hands on when every device it covers is running the release's firmware. The trailing `catch_all` never does: devices keep arriving and devices revert, so it stays running for the life of the release, absorbing whatever turns up.

## When devices fail

A step fails once its `failure_tolerance` is used up: either a count of devices, or a percentage of the ones it covers. A device counts as failed once it is in the penalty box, so it has spent its own retries first. `update_devices` steps default to failing on the first device; a `catch_all` never fails.

A failed step stops the workflow where it is. Getting out of it is retry or skip, from controls on the step in the diagram. Retry lets the failed devices out of the penalty box and starts the step again; skip releases its devices for a later step. The trailing `catch_all` cannot be skipped, because nothing follows it. Skipping it would quietly end the release instead of moving it along.

A stopped workflow is otherwise quiet: devices go on connecting, the group still reads as active, and the only sign is a diagram nobody is watching. So it is announced as telemetry, a log line, an audit entry, and a product notification. Reaching an approval step is announced the same way.

## Review notes

`Devices.Updates` and `FirmwareUpdates` are shared with every deployment. Every query over `inflight_updates` assumed a row existing means the device is updating, which held because reaching an outcome deletes the row. That rule is now written down as `InflightUpdate.active_statuses/0` and asked for explicitly. It changes nothing today, since every row that exists is already active. It does sit on the path that decides whether any device gets firmware at all.

The `catch_all` derives its devices rather than recording them. Every other step records which devices it covers, because its cohort is a choice. The `catch_all` has no choice to record, and on a 400k device deployment recording it anyway would be 400k rows per release to say "everyone else". It asks the other steps' claims instead.

A finished step keeps no hold on its devices. A canary that took the update, saw its stage complete, and then reverted was not updated again for that release: still claimed by the step that had finished with it, so covered by nothing. Claiming still counts every claim, so a later stage does not spend its `match_limit` on devices an earlier one saw to.

## Migrations

Three, and they need to run in order. The originals were written in May and carried May timestamps, which after the rebase sorted ahead of everything main has added since; they are now dated to follow the migrations they build on. If you have run the earlier versions on a local database, it wants rebuilding.

## Left out

`InflightUpdate`, `UpdateStat` and the step membership table overlap, and there is a good case for one `UpdateHistory` covering successes and failures together. It touches the path every device update goes through, so it should be its own change. The `active_statuses` change here is the first step of it and is useful on its own.


lib/nerves_hub/events/device_events.ex
```diff
@@ -137,7 +137,7 @@ defmodule NervesHub.DeviceEvents do
       not match?(%{available?: true}, Updates.check_update(device)) ->
         {:error, :no_update}
 
-      Orchestrator.available_slots(deployment_group) <= 0 ->
+      DefaultCoordinator.available_slots(deployment_group) <= 0 ->
```
nshoes: This doesn't take into account an active workflow's step concurrency. A device could self-schedule itself past a step when it's not supposed to. Same with the catch all `true` on lne 143 here.

lib/nerves_hub/managed_deployments/orchestrator/workflow_coordinator.ex
```diff
@@ -0,0 +1,192 @@
...
+  defp just_started?(%DeploymentWorkflowStep{started_at: started_at}) do
+    NaiveDateTime.diff(NaiveDateTime.utc_now(), started_at, :second) <= 1
+  end
+
+  defp schedule_step_devices(deployment_group, step) do
+    slots = Workflows.available_slots(deployment_group, step)
+
+    if slots > 0 do
+      available = Updates.available_for_workflow_step(deployment_group, step, slots)
+      updated_count = schedule_devices!(available, deployment_group)
```
nshoes, line 151: This doesn't take into account the prioty queue, should it? I don't think it should, as I don't see any useful intgeration with workflows + priority queue.

lib/nerves_hub/managed_deployments/deployment_release.ex
```diff
@@ -71,4 +76,28 @@ defmodule NervesHub.ManagedDeployments.DeploymentRelease do
...
+
+    steps = maybe_add_catch_all(steps)
+
+    put_assoc(changeset, :steps, steps)
+  end
+
+  defp maybe_add_catch_all(steps) do
+    last = List.last(steps)
+
+    if get_field(last, :type) == :catch_all do
```
nshoes, line 102: This will blow up if `steps` is an empty list.

lib/nerves_hub_web/live/deployment_groups/show.ex
```diff
@@ -74,6 +90,45 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Show do
...
+        # Somebody else approved it between the page rendering and the click.
+        socket |> assign(:awaiting_approval, nil) |> noreply()
+
+      step ->
+        step = Workflows.approve_step(step, user)
+
+        AuditLogs.audit!(
+          user,
+          deployment_group,
+          "User #{user.name} approved workflow step #{step.number} (#{step.name}) for deployment group #{deployment_group.name}"
```
nshoes: I believe this should use `step.label` instead of `step.name`?

lib/nerves_hub/managed_deployments/deployment_release.ex
joshk: The schema enforces `minItems: 1` and `required: ["steps"]`, and `update_changeset` runs the validator, which is the only place `workflow_definition` is cast. So while this edge case should never be reached, I'll add a guard.

lib/nerves_hub_web/live/deployment_groups/show.ex
joshk: Good spotting, thanks.

lib/nerves_hub/managed_deployments/orchestrator/workflow_coordinator.ex
joshk: Agreed. A workflow already expresses ordering explicitly. I think there is something we can do around this in future iterations, but I'd punt this for now.

lib/nerves_hub/events/device_events.ex
joshk: Hmmmmmm, soooo, this is the `device requested` path, and I wonder if slots shouldn't be taken into account as this is the same as manually sending an update to a device. 

The catch all just adds the device to the penalty box for a bit of time, so that would would end up checking again in the future, but again, I think device requested updates should bypass concurrency controls.

lib/nerves_hub/managed_deployments/workflows.ex
```diff
@@ -0,0 +1,517 @@
...
+  defp release_claimed_devices(step) do
+    @join_table
+    |> from(as: :step_device)
+    |> where([step_device: sd], sd.deployment_workflow_step_id == ^step.id)
+    |> Repo.delete_all()
+  end
+
+  # Devices in the deployment group that match the step and have not already been
+  # claimed by one of the release's steps.
+  defp unclaimed_matching_devices_query(deployment_group, step) do
```
nshoes, line 461: This doesn't filter out devices in the penalty box from a previous release, worth considering I think.

lib/nerves_hub/events/device_events.ex
```diff
@@ -137,8 +144,8 @@ defmodule NervesHub.DeviceEvents do
       not match?(%{available?: true}, Updates.check_update(device)) ->
         {:error, :no_update}
 
-      Orchestrator.available_slots(deployment_group) <= 0 ->
```
nshoes, line 140: Why did we remove the deployment group's concurrency check?

priv/repo/migrations/20260901120100_create_deployment_workflow_steps.exs
```diff
@@ -0,0 +1,28 @@
...
+      add(:skipped_by_id, references(:users), null: true)
+      add(:approved_by_id, references(:users), null: true)
+      add(:number, :integer, null: false)
+      add(:name, :string, null: true, size: 50)
+      add(:description, :string, null: true, size: 100)
+      add(:status, :string, null: false)
+      add(:type, :string, null: false)
+      add(:matching_conditions, :map, null: false, default: %{})
+      add(:failure_tolerance, :map, null: true)
+      add(:concurrency, :integer, default: 10)
```
nshoes: Probably should add `null: false` here.

nshoes (approved):
LGTM, just some Claude notes.

lib/nerves_hub/events/device_events.ex
joshk: This relates to a resolved comment. This branch of checks is specific to 'device managed updates', and I compared this to manually requested updates, where we don't check available slots but allow devices to jump the queue.

priv/repo/migrations/20260901120100_create_deployment_workflow_steps.exs
joshk: you want to allow concurrency to be `null`?

lib/nerves_hub/managed_deployments/workflows.ex
joshk: Thanks for spotting this. Another issue it helped me see if that we don't filter out devices that are disconnected. Fixing this.

priv/repo/migrations/20260901120100_create_deployment_workflow_steps.exs
joshk: Thanks for noticing. I asked Claude if `null` should mean unlimited concurrency but it instead tested that `nil` caused an error and instead of making it unlimited it decided that `null: false` was the way to go. I think we should allow for unlimited concurrency, but I'll do that later.

PR #2990, opened 2026-09-01
Route sidebar device filters through the advanced query compiler

lawik:
Every sidebar filter with an advanced-query equivalent now compiles to the same predicate the query language uses, so one implementation defines what each filter means: alarm, alarm status, health status, connection (including the not-seen-in-N-days variants via last_seen), connection type, platform, updates, identifier, untagged, deleted visibility, currently-updating and free-text search.

Three filters keep their own SQL because the query language means something different: firmware_version (the firmware column matches by UUID), deployment_id (deployment_group matches by name), and tags (the sidebar is a fuzzy substring match, contains is exact).

One behavior change rides along: "Untagged" now also matches a device whose tags column holds an empty array rather than NULL - the old array_length(?, 1) = 0 branch could never be true since array_length returns NULL for an empty array.

Claude-Session: https://claude.ai/code/session_01LSwBQEoJvM3Hv1MAvHdSWX


joshk:
These two PRs make a lot of sense.

It has me wonder, should we keep the side bar filtering? if the advanced query filtering is underpinning the sidebar, should we maybe ditch the sidebar?

lawik:
It provides a lot of discoverability and convenience. Lots of filters I don't remember what they are called.

I had missed that the filters don't map into advanced querying though which I think it should so you can learn the filters more easily.

As in, you select some filter in the sidebar and it adds to the query and filters rhat way.

joshk:
Thats a bit of an issue because if you use brackets then it won't directly match with the sidebar.

If its about discoverability, we can do that as a model popup to explain the query items supported.

I'll create an issue about this.

lawik:
I think we can let the UI be kind of dumb. Just slap AND and the filter on the end.

I don't want us to remove the easy way to filter things in favor of the one that is the power user move. I've felt the pain of trying to understand exactly what a DataDog or NewRelic wants me to be typing or trying to figure out what I should actually do in Grafana. Having that filter sidepane is the simple path and it won't ever bother a power user, they don't need to see it. But for a novice they get shown "oh, all my filters turn into these queries" and then they can be led down the path of corruption/glory

PR #2992, opened 2026-09-01
Configureable health profiles

lawik:
Health profiles are not used for anything if you do not have the health extension enabled on a product. This entire branch is nothing in that case.

Right now it operates against the health data in Postgres, it tries quite hard to be mindful about the database and tracks some state for the math in memory as a per-node GenServer per product instead. This will be even less of a concern when health metrics move to clickhouse. Like previously it updates the information on the device so other views don't have to re-derive the state.

The feature lets you configure a profile for your device based on a health metric:

Warning level if `CPU% > 50` for `15m`.
Alert level if `CPU% > 90` for `15m`.

This dovetails nicely with Josh's work on deployment workflows which then can have custom criteria about what constitutes a healthy deployment. For the evaluation we use median, rather than average because it protects developers from a random sensor reading going from 1-200 and suddenly sending 17000 and throwing the device into alert.

Another neat thing is that the graphs can be colored according to these levels and what gets pulled to the front page dashboard becomes better.

Devices are completely unaware of this currently. I might bump the health extension protocol or break it apart, to allow passing back information so the device can know it is considered unhealthy.

Most of the code is about managing the state of things and most of that boils away later with clickhouse and becomes a single query. It isn't too hairy now, just a lot to read.

Manage profile:
Screenshot_20260901_162139

Add new metric and add platform-specific health profile:
Screenshot_20260901_162157

Usage of profile levels:
Screenshot_20260901_162229

Screenshot_20260901_162212

### DIFF
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -411,6 +411,34 @@ defmodule NervesHub.Devices.Connections do
     :ok
   end
 
+  @doc """
+  How many times the device disconnected during each of two trailing
+  windows, as `{count_in_first, count_in_second}` — one query for both.
+
+  Backs the "disconnects" built-in health profile metric (a warning and an
+  alert window per judgement); callers are expected to check that analytics
+  is enabled first.
+  """
+  @spec disconnection_counts(pos_integer(), pos_integer(), pos_integer(), {pos_integer(), pos_integer()}) ::
+          {non_neg_integer(), non_neg_integer()}
+  def disconnection_counts(org_id, product_id, device_id, {first_seconds, second_seconds}) do
+    now = DateTime.utc_now()
+    first_cutoff = DateTime.shift(now, second: -first_seconds)
+    second_cutoff = DateTime.shift(now, second: -second_seconds)
+    widest = if DateTime.before?(first_cutoff, second_cutoff), do: first_cutoff, else: second_cutoff
+
+    DeviceConnectionHistory
+    |> where([dc], dc.org_id == ^org_id and dc.product_id == ^product_id and dc.device_id == ^device_id)
+    |> where([dc], not is_nil(dc.disconnected_at))
+    |> where([dc], dc.disconnected_at >= ^widest)
+    |> select(
+      [dc],
+      {fragment("countIf(disconnected_at >= ?)", ^first_cutoff),
+       fragment("countIf(disconnected_at >= ?)", ^second_cutoff)}
+    )
+    |> AnalyticsRepo.one(settings: [final: 1])
+  end
+
   def flapping_connections(%Product{} = product) do
     DeviceConnectionHistory
     |> where([dc], dc.org_id == ^product.org_id and dc.product_id == ^product.id)
--- a/lib/nerves_hub/devices/connections.ex
+++ b/lib/nerves_hub/devices/connections.ex
@@ -6,7 +6,7 @@ defmodule NervesHub.Devices.Connections do
 
   alias NervesHub.Analytics.Buffer
   alias NervesHub.AnalyticsRepo
-  alias NervesHub.Devices.Device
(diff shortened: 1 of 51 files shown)


nshoes (approved):
Very cool! Appreciate the human-written PR description :)

lib/nerves_hub/devices/connections.ex
```diff
@@ -411,6 +411,34 @@ defmodule NervesHub.Devices.Connections do
...
+  How many times the device disconnected during each of two trailing
+  windows, as `{count_in_first, count_in_second}` — one query for both.
+
+  Backs the "disconnects" built-in health profile metric (a warning and an
+  alert window per judgement); callers are expected to check that analytics
+  is enabled first.
+  """
+  @spec disconnection_counts(pos_integer(), pos_integer(), pos_integer(), {pos_integer(), pos_integer()}) ::
+          {non_neg_integer(), non_neg_integer()}
+  def disconnection_counts(org_id, product_id, device_id, {first_seconds, second_seconds}) do
```
joshk: Just a small note regarding the PR description. The `DeviceConnectionHistory` schema uses Clickhouse and not Postgres.

lib/nerves_hub/devices/connections.ex
```diff
@@ -411,6 +411,34 @@ defmodule NervesHub.Devices.Connections do
     :ok
   end
 
+  @doc """
+  How many times the device disconnected during each of two trailing
+  windows, as `{count_in_first, count_in_second}` — one query for both.
```
joshk: My brain can't grok this.

lib/nerves_hub/devices/metrics.ex
```diff
@@ -64,6 +64,22 @@ defmodule NervesHub.Devices.Metrics do
...
+  """
+  def observed_ranges(product_id) do
+    DeviceMetric
+    |> join(:inner, [dm], d in assoc(dm, :device))
+    |> where([_, d], d.product_id == ^product_id)
+    |> group_by([dm], dm.key)
+    |> select([dm], {dm.key, {min(dm.value), max(dm.value)}})
+    |> Repo.all()
+    |> Map.new()
+  end
```
joshk: have you run this query against NervesCloud prod to check its performance?

lib/nerves_hub_web/channels/device_socket.ex
```diff
@@ -230,6 +234,15 @@ defmodule NervesHubWeb.DeviceSocket do
...
     _ = DeviceLink.disconnect(device_info.connection_ref)
 
+    # The health evaluator only tracks connected devices, and the extension's
+    # own detach only runs when the device *announces* it — an abrupt
+    # disconnect (power loss, network drop, crashed channel) never does. The
+    # transport's terminate runs for all of those, so this is where the
+    # evaluator is told. A local Registry lookup plus a cast: safe here. If
+    # the device already reconnected to this node, the forget merely costs
+    # one re-warm-up on its next report.
+    :ok = HealthEvaluator.forget(device_info.product_id, device_info.device_id)
```
joshk: this should probably be moved into `DeviceLink.disconnect/1`

lib/nerves_hub_web/live/product/settings.html.heex
```diff
@@ -61,6 +61,21 @@
...
+          </.button>
+        </div>
+        <div class="flex flex-col gap-3 p-6">
+          <p class="w-2/3 text-sm">
+            Health profiles set the thresholds that decide when a device counts as unhealthy — per product, or per platform when
+            hardware differences call for different values.
+          </p>
+        </div>
+      </div>
+
```
joshk: If we are to keep with our UI patterns, we should change Product Settings to be a tabbed UI with Health Profiles being a tab.

lawik:
Reworking to match the change of having health metrics in clickhouse.

PR #2995, opened 2026-09-01
Import devices into deployment group by CSV

nshoes:
Adds a button when viewing a deployment group, "Import from CSV". Takes a CSV with an `identifer` column. Uses the already existing `move_many_to_deployment_group` path. Has a few helpful validations so you don't accidentally import a device from the wrong product and/or devices with platform and architecture that don't match the deployment group. A little performance work was done so thousands of devices could be imported without causing a ruckus. 

<details><summary>Importing in action</summary>
CleanShot 2026-09-02 at 12 26 16
</details>


joshk (approved):
LGTM

PR #2996, opened 2026-09-02
Move device metrics to ClickHouse, and add a metrics extension

joshk:
Device metrics move to ClickHouse, and get an extension of their own.

## Why

Metrics were the last fleet-scale write path landing only in PostgreSQL: a row per key per report, on a table nothing truncates cheaply. Log lines, connection history, device messages and error reports are all in ClickHouse already, so this is the existing shape rather than a new one: history in ClickHouse via `Analytics.Buffer`, and whatever has to be queried alongside a device denormalised into PostgreSQL.

The second half is the payload. The health extension carries metrics, metadata and alarms in one report, so a device reporting every minute sends its firmware metadata every minute. That blob changes when the device is updated and at no other time. Across a fleet, most of what health costs is the part that never changes.

## Reviewing

Six commits, meant to be read in order:

1. `3dff3c45`: ClickHouse table, and `Metrics.record/3` as the single write path
2. `119dc3d6`: `device_latest_metrics` in PostgreSQL
3. `32ca4dc8`: charts read ClickHouse, PostgreSQL writes stop
4. `5979ca11`: the `metrics` extension
5. `7f4aeeb3`: drop the retired PostgreSQL table
6. `5d2eb2bd`: runtime configuration docs
7. `81dbb110`: record the firmware each reading was taken under

Rebased onto #2989 and #2990. Those routed the sidebar's metrics filter through the advanced query compiler, which is a better shape than the duplicate implementation this branch originally rewrote, so that rewrite is gone and the compiler is the single definition. The two tests #2989 added now seed through `Metrics.record/3`.

## Storage

`device_metrics` in ClickHouse is one row per key per reading, sorted `(product_id, device_id, key, timestamp)`, with a 30 day TTL. Metric names are device-defined, so a column per metric was never available.

Each reading also records the firmware the device was running when it took it, which is what lets a metric be read per release rather than only per device: whether memory use went up with 1.4.0, whether a temperature regression followed a deploy. It comes from the connection, so devices need no client change. It is not in the sort key, because the read this table exists to serve is one device's range of one metric over time and putting the firmware ahead of `device_id` would scatter that range.

`device_latest_metrics` in PostgreSQL is one row per device, replaced by each report. Both metric filters used a correlated subquery to find each device's newest reading at query time; that "latest" is now a stored fact, so the sidebar filter and the `metric:` query language cost what any other column costs. The devices list compares a metric value alongside tags, firmware and connection state in one query, and ClickHouse cannot be joined into it. That is the same reason `latest_health_id` and `device_connections` exist. Its own table rather than a column on `devices`, because a watched device reports once a minute and that would churn the hottest row in the system.

Nothing is backfilled. Every device fills its own row in on its next report.

## The `metrics` extension

Version 0.1.0, batched, timestamped by the device, paced by the platform the way health is. Contract in [docs/metrics.md](docs/metrics.md), since a Rust and an ESP-IDF client will both want to implement it.

Batching is what lets a device sample every ten seconds and report every ten minutes, and lets one that lost its connection keep what it measured while it was gone. Neither works if a message can only carry the instant it was sent, which is the mistake `Logging.Batched` exists to correct. Readings more than a day off the platform's clock are dropped; there is no telling how late such a reading actually is, and a chart stretching to 1970 is worse than a gap.

Rate limited per device, one token per message whatever it carries. Charging per reading would push a client towards reporting more often, which is the opposite of the point. Its own bucket rather than a share of logging's or error reports', because a device in trouble produces all three at once.

Health goes on accepting metrics at 0.0.1. Every device in the field speaks it, and both paths end at `Metrics.record/3`.

## What a report may carry

Metric names land in a `LowCardinality` ClickHouse column, JSONB keys, and the advanced-query autosuggest list, so one confused client can widen all three permanently. A report is trimmed rather than trusted: names over 64 bytes dropped, at most 20 names kept (`DEVICE_METRICS_MAX_KEYS_PER_REPORT`, sorted so a device over the limit loses the same readings every time), non-numeric values dropped. The rest of the report is stored, and the product's operators get a notification.

The notification dedupes on the device, not the offending name. Keyed by the name, a client generating unbounded names would get a row per name, which is the growth the cap exists to prevent.

## Without ClickHouse

Metric *history* needs it. Nothing else does: the numbers on the device page, the metrics filter and `metric:` advanced queries all read the PostgreSQL latest set. The health tab's charts say why they are empty, and the `metrics` extension is not offered at all, so a device is never told about an extension whose reports go nowhere.

## A behaviour change to flag

`HEALTH_CHECK_DAYS_TO_RETAIN` no longer governs metrics. A ClickHouse TTL is fixed when the table is created and cannot follow a runtime setting. It still governs `device_health`. Anyone who set it to keep metrics for a fortnight now gets thirty days. `docs/runtime_configuration.md` covers this and the three new variables.

## The table drop, and its lock

The old PostgreSQL `device_metrics` table goes in commit 5, along with `DeviceMetricLegacy`, the `has_many` on `Device` and the truncation worker's call.

One thing to know before deploying it. Dropping the table also drops its foreign key into `devices`, and Postgres takes an ACCESS EXCLUSIVE lock on the *referenced* table to do that. `devices` is the busiest table here, so a drop that has to queue behind a long-running read would then have every other query queue behind the drop. The migration sets `lock_timeout = '5s'` so that becomes a failed migration rather than a stall. Migrations run on boot, so hitting it means deploying again at a quieter moment; the work itself is unlinking files and takes no time once the lock is held.

There is also a narrow rolling-deploy window: an old node still carrying the `has_many` will fail to delete a device once the new node has run the migration, because it tries to `DELETE FROM device_metrics` first. It clears as soon as the deploy finishes.

`down/0` recreates the table, so the migration reverses cleanly. The data does not come back.

## What is deliberately left out

Health 0.1.0. Sending metadata only when its hash changes is the larger payload win and needs a matching client change, so it is its own change once `metrics` has a client.


joshk:
@lawik @nshoes before completing and merging in the Health Profiles PRs, I suggest we merge this in first and get it deployed. Happy to talk more about this on a call.

lawik:
Yeah. This seems good to me.

lawik (approved):

PR #3006, opened 2026-09-03
Split device health into status, alarms and connection metadata

joshk:
`device_health` was a report history: one row per report, the whole report payload in `data`, and `devices.latest_health_id` pointing at the newest row.

Two things prompted this. Nothing in the application ever read a historical row. Every query goes through the `latest_health` association, and the only other consumer was the truncation worker deleting rows nothing had looked at. Of the payload, only alarms and metadata were ever read; since device metrics moved to ClickHouse the rest of it duplicates what is already there.

So each part goes where it belongs.

Alarms get their own table, one row per device and raised alarm. A report carries the device's whole current alarm set rather than events, so `Alarms.sync/3` derives raises and resolves by diffing against what is stored, in one transaction. The upsert replaces only `description`. Leaving `raised_at` alone is what keeps it meaning when the episode started rather than when the alarm was last mentioned. Names lose their `Elixir.` prefix on the way in, which retires the trimming three read sites were doing.

Transitions go to ClickHouse as `device_alarm_history`, buffered like every other analytics write. Edges rather than episodes: they fall out of the diff, they are append-only, and an episode's duration is a window function over the pair. Only transitions are written, so the volume stays small enough to keep for 90 days.

Metadata merges onto the device connection, which already had both the field and `merge_update_metadata/2`. It describes the device as this connection found it, a reconnect resets it, and `attach/1` asks for a report ahead of the timer so it repopulates straight away.

`device_health` keeps the verdict alone: status and reasons, one row per device, replaced in place. Status is derived from the profile thresholds and the stored readings, so a past status can be recomputed from ClickHouse rather than needing a history table of its own. That also retires `latest_health_id`, the truncation worker and its cron entry.

### Two behaviour changes

Nothing is backfilled, so alarms, metadata and status are all empty until each device's next report, up to one idle reporting interval (an hour by default).

The `alarm contains` filter used to run `jsonb_each_text` over the whole health payload, which matched metadata and metric values as well as alarms. Against the new table it matches an alarm's name or description, which is what the filter was for.

### Open to opinions on

- 90 days for the alarm history TTL, against 30 for metrics and error reports. Transitions are cheap and I wanted "has this alarm ever fired on this fleet" to stay answerable, but it is one line to change.
- A device that disconnects while alarming keeps its alarms, so product-wide counts include offline devices. That matches the reasoning for health status, and narrowing on connection status at the call site is available if it is wrong.
- `current_alarms_count/1` had no callers. I kept and tested it rather than dropping it, since it is a natural product-level read.

The `device_health` migration is deliberately irreversible: the payloads and the history it drops are not recoverable, so a `down` that rebuilt the old shape would hand back an empty table pretending to be the old one.


lawik (approved):
Seems good

PR #3015, opened 2026-09-07
Configure device websocket compression from the environment

joshk:
The device sockets set `compress: true`, and that is compiled in. There was no way to turn per-message deflate off, or to change what it costs, without building a new image. What it costs is not small: Bandit holds a zlib deflate and inflate context per connection for as long as the device stays connected, 271KB per device measured end to end, around 375MB of `:erlang.memory(:system)` on a node holding 1400 devices.

Bandit already has the switch for it. It negotiates per-message deflate when the socket asks for it *and* the listener allows it:

```elixir
# bandit/lib/bandit/websocket/handshake.ex
if Keyword.get(opts, :compress) && Keyword.get(websocket_opts, :compress, true) do
```

`opts[:compress]` is Phoenix's compile-time per-socket flag. `websocket_opts` is the listener's own `websocket_options`, which lives in the endpoint's `http:`/`https:` config and is therefore already read at boot from `config/runtime.exs`. The sockets keep asking; the listener now decides.

Three variables, read once and applied to both listeners, because a device reaching the web endpoint upgrades to a websocket there just as it does on the device endpoint:

- `DEVICE_WEBSOCKET_COMPRESSION`, default `true`, whether the extension is negotiated at all.
- `DEVICE_WEBSOCKET_DEFLATE_LEVEL`, default `6`, which is the level zlib's own `:default` resolves to.
- `DEVICE_WEBSOCKET_DEFLATE_MEM_LEVEL`, default `4`, the value that was already hardcoded there.

The defaults reproduce today's behaviour, so an existing deployment sees no change until someone sets one of them. Being runtime config, they take a restart rather than a rebuild.

Both levels are checked at boot. zlib does not reject an out of range value until a device upgrades to a websocket, so a typo would otherwise fail every device connection instead of failing the boot, which is a much worse way to find out.

Nothing changes on the device side when compression is turned off. A client that offers `permessage-deflate` to a listener that does not answer with it sends uncompressed frames, so there is no flag day and no minimum agent version.

The comment explaining the mem_level choice moves up to where the values are now computed, and `docs/runtime_configuration.md` gets the three entries.


nshoes (approved):
Amazing!

PR #3017, opened 2026-09-07
Send emails from a background job

### LINKED ISSUE #2078: Move emails to Oban
This gives us error reports and moves something which could be slow to the background.

joshk:
Emails were sent inline. Every `UserNotifier.deliver_*` call ran `Swoosh.Mailer.deliver/1` in the calling process, so in production the whole SMTP conversation (connect, TLS handshake, AUTH, DATA) happened on the web request or LiveView event that asked for the email.

The `deliver_all_tell_*` fan-outs made it worse: they loop over an org's admins with a separate blocking send each, so inviting someone into a ten-admin org is eleven sequential round-trips before the page comes back. Dev uses the Local adapter and test uses the Test adapter, so the latency only shows up in production.

Emails are now queued as `NervesHub.Workers.SendEmail` jobs on their own `email` queue. The job carries the recipient, a template key and that template's assigns rather than a rendered message, so the rendering happens in the job too; the fan-outs build one job per admin and enqueue them with a single `insert_all`.

Two things fall out of the move:

- Delivery failures were discarded by `_ =` at most call sites, and hard-matched at two (`accounts.ex` raised a `MatchError` after the user row was already inserted during Google OAuth signup). A failed send now fails the job and retries with backoff, up to five attempts.
- `deliver_*` returns whatever `Oban.insert/1` and `Oban.insert_all/1` return, so the existing `{:ok, _}` call sites still match. They mean "queued" now, not "delivered", and the three `Accounts` docstrings that promised an email struct say so.

The second commit fixes a separate bug I hit while moving this code. `deliver_org_user_added/3` took `(org, user, invited_by)`, but its only call site passed `(org, invited_by, org_user.user)`, so the "You have been added to <org>" email went to the admin doing the inviting, and the body named the new member as the person who added them. The parameters are reordered to match `deliver_all_tell_org_user_added/3` next door, so both calls in the `send_invite` handler now read the same. The swap survived because the test only asserted the subject, which is identical either way.

Closes #2078 

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -114,6 +114,7 @@ config :nerves_hub, Oban,
   ],
   queues: [
     default: 1,
+    email: 5,
     firmware: 5,
     delete_file: 3,
     cleanup: 2,
--- a/lib/nerves_hub/accounts.ex
+++ b/lib/nerves_hub/accounts.ex
@@ -158,12 +158,12 @@ defmodule NervesHub.Accounts do
   end
 
   @doc ~S"""
-  Delivers an email confirming that a users password has been updated.
+  Queues an email confirming that a users password has been updated.
 
   ## Examples
 
       iex> deliver_user_password_updated(user, &url(~p"/reset-password/#{&1}"))
-      {:ok, %{to: ..., body: ...}}
+      {:ok, %Oban.Job{}}
 
   """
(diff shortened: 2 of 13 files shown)


joshk:
Closes https://github.com/nerves-hub/nerves_hub_web/issues/2078

nshoes (approved):
Ah, this is obvious in hindsight, whoops

Issue #2408, opened 2025-12-09
Track, display and filter on firmware validation status


lawik:
Definitely in the detail view but also very nice to filter on potentially if you have issues with devices not validating or similar.

PR #3018, opened 2026-09-07
Add firmware validation status to the devices filter and advanced query

joshk:
Devices report whether their running firmware was validated, and the device page shows it, but there was no way to find the devices that never validated without opening them one at a time.

## Sidebar filter

A "Firmware Validation" select on the device list: All / Validated / Not validated / Unknown. It goes through the same filter map as every other sidebar filter, so `filters[firmware_validation_status]` works on the device list API too. The OpenAPI spec documents it.

The filter is guarded on the three valid values, so an old bookmark carrying a junk value is ignored instead of raising.

## Advanced query

A `firmware_validation_status` column supporting `=` and `!=`:

```
firmware_validation_status = "not_validated"
firmware_validation_status != "validated" and connection = "connected"
```

Autosuggest picks the column up from the schema whitelist, and the live view hands its values to the editor in the schema JSON.

## The nullable column

`devices.firmware_validation_status` is nullable, so the compiler folds NULL into `unknown` on both sides of the comparison. Without that, `!= "validated"` would silently drop any NULL row. It's the same treatment `health_status` already gets.

## Not included

The device page also shows "Revert detected", which is the separate `firmware_auto_revert_detected` boolean. That could be its own column and filter later.

### DIFF
--- a/lib/nerves_hub/devices/advanced_query/compiler.ex
+++ b/lib/nerves_hub/devices/advanced_query/compiler.ex
@@ -168,6 +168,25 @@ defmodule NervesHub.Devices.AdvancedQuery.Compiler do
 
   defp comparison_dynamic("architecture", "!=", value), do: dynamic([d], d.firmware_metadata["architecture"] != ^value)
 
+  # Whether the device reported its running firmware as validated. The column is
+  # nullable, and a device that has never reported reads as "unknown" - so the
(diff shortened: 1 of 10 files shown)


joshk:
Closes https://github.com/nerves-hub/nerves_hub_web/issues/2408

lawik (approved):

Issue #1387, opened 2024-07-11
Add 404 tests for all pages


joshk:
This confirms we are aren't showing 500 error pages when a db record can't be found

PR #3019, opened 2026-09-07
Return 404 for missing records in the web UI

joshk:
Started as a question about whether we test that the 404 page shows up when a record can't be found. The API turned out to be well covered; the web UI was not, and probing it turned up three routes that returned 500 instead.

## Missing download targets returned 500

`DownloadController.archive/2`, `DownloadController.firmware/2` and `DeploymentGroupController.export_audit_logs/2` each matched hard on `{:ok, x} =` against a lookup that returns `{:error, :not_found}`. A bad UUID in a download URL raised a MatchError and served the 500 page. All three had happy-path-only tests, which is how it survived.

They now raise `NervesHubWeb.NotFoundError`, the same way the rest of the app reaches the 404 page.

## Devices ignored the product in their own URL

`Devices.get_by_identifier_query/3` filtered by org but never by product, even on routes carrying a `:product_name` segment. So `/org/acme/ProductB/devices/<id>` returned 200 and rendered a device belonging to ProductA, with ProductB in the page context.

Products are not a permission boundary today, so this is an inconsistency rather than a cross-tenant leak. But firmware and deployment groups are product-scoped and 404 correctly, which left devices as the odd one out.

The scope only carries a product when the route has one to give it, so the filter is conditional and the top-level `/api/devices/:identifier` routes work as before. The two console channels build a scope from the user alone and are unaffected.

One thing reviewers should see: `DeviceControllerTest` was itself depending on the bug. Its setup shadowed the conn case's product with a freshly created one while the device still belonged to another, so its URL paired a device with a product it was not in. It passed only because the lookup ignored the segment. I removed the setup so the test names the device's own product.

## On the new test file

`test/nerves_hub_web/live/not_found_test.exs` asserts the response status rather than the exception type. A 404 gets here two ways (a bang lookup raising `Ecto.NoResultsError`, or `NotFoundError`), and neither is visible from the page that depends on it. Asserting the status means moving a lookup between the two won't break the test, but losing the 404 will.

Web UI coverage before this was three tests across 49 LiveView files, and only one of them asserted a 404 response at all.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -271,25 +271,38 @@ defmodule NervesHub.Devices do
     |> Repo.fetch()
   end
 
-  defp get_by_identifier_query(%Scope{org: org}, identifier, preload_assoc) when not is_nil(org) do
+  defp get_by_identifier_query(%Scope{org: org} = scope, identifier, preload_assoc) when not is_nil(org) do
     Device
     |> join(:left, [d], o in assoc(d, :org), as: :org)
     |> where(identifier: ^identifier)
     |> where(org_id: ^org.id)
+    |> scope_to_product(scope)
     |> preload([org: o], org: o)
     |> join_and_preload_deployment_group_and_current_release()
     |> join_and_preload(preload_assoc)
   end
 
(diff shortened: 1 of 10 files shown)


joshk:
Closes https://github.com/nerves-hub/nerves_hub_web/issues/1387

lawik (approved):

Issue #684, opened 2020-11-03
firmware filter for devices should populate from devices


jjcarstens:
The `firmware` filter of the device index page is not that useful. It only lists firmware versions of which you still have a firmware record for. But firmwares get cleaned up and sometimes you have devices connecting which are on an old firmware that doesn't exist.

This makes it so you can't filter devices on reported firmware versions. You also can't filter devices that are "unknown" firmware (have yet to connect). In the screenshot below, you can see I have multiple firmwares that aren't listed in the firmware filter dropdown

<img width="492" alt="image" src="https://user-images.githubusercontent.com/11321326/98026735-d2122000-1dc8-11eb-9066-5e00c8adb855.png">

It would be nice if we could populate the filter option based on all versions reported from devices on the product.

Mrjaco12:
I agree that updating this in some way would be a good idea. I just want to add context here about how we arrived at the way it works now. Since the live view only tracks devices that are being displayed, generating a list of firmware ids from the list of devices that the liveview knows about is not sufficient. The alternative is to fetch firmware versions for each device when the liveview mounts, which we decided was kinda expensive and it wouldn't necessarily have all the correct data if devices were added or removed while the user was on the page. It was also slightly more complicated to factor in filtering for the "unknown" state. So given all that we elected to use the easiest approach.

To summarize, there are a few considerations about tradeoffs that need to be decided before this feature can be fully implemented but I think this issue is bringing things in the right direction.

jjcarstens:
Personally, I think its a worthwhile tradeoff to do a single DB query to select a single field from applicable device firmwares for the versions on mount. Or, it can be done when opening the "show filter" button as well.

PR #3020, opened 2026-09-07
Filter devices by the firmware versions they are running

joshk:
The Firmware dropdown in the devices list sidebar was built from every firmware uploaded to the product. On my dev database the SmartKiosk product has 35 uploads covering 34 distinct versions, but its 10 devices are running 7 of them, so 27 of the options matched no devices at all.

This builds the list from `devices.firmware_metadata` instead, so it only offers versions devices are actually running.

Since a version you expected to see can now be missing, the filter label has a hint tooltip explaining why:

> Only versions currently running on devices in this product are listed. Uploaded firmware which no device is running won't appear.

### Notes

- `Devices.firmware_versions/1` sits alongside `platforms/1` and `architectures/1`, which already derive their dropdowns from the devices in the same way. Like them, it doesn't exclude soft-deleted devices. Otherwise the sidebar's "Only deleted devices" option would have nothing to filter by.
- The list moved out of `mount/3` and into the async `assign_filter_data` load with the rest of the dropdown data, so it no longer blocks the first render.
- `Firmwares.get_firmware_versions_by_product/1` had no other callers and is gone.
- `FilterSidebar`'s `:filter` slot takes an optional `hint`. Filters without one render exactly as before.

I couldn't hover the tooltip on a running devices page, so I rendered the same markup against the compiled CSS to confirm it fits inside the sidebar without clipping. The positioning itself comes from the `ToolTip` hook and `data-placement="left"` already used on the Insights page.

### DIFF
--- a/lib/nerves_hub/devices.ex
+++ b/lib/nerves_hub/devices.ex
@@ -761,6 +761,29 @@ defmodule NervesHub.Devices do
   end
 
   @doc """
+  Get distinct firmware versions currently reported by the product's devices,
+  newest first.
+
+  This is the set of versions actually in use, which is a subset of the
+  firmware uploaded to the product.
+  """
+  def firmware_versions(product_id) do
+    Device
+    |> where([d], d.product_id == ^product_id)
+    |> where([d], not is_nil(fragment("?->>'version'", d.firmware_metadata)))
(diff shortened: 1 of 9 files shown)


joshk:
Closes https://github.com/nerves-hub/nerves_hub_web/issues/684

lawik (approved):

PR #3024, opened 2026-09-08
Track which CA signed each device certificate

### LINKED ISSUE #659: Show devices using a specific CA
It would be cool if we had someway to see what devices are using a CA (either on the device show page, or the certificates page of the product settings)

Thoughts?

joshk:
A device certificate names its signer by key id: `device_certificates.aki` holds the CA's `ski`. There's no foreign key, and nothing in the UI followed that link. Opening a device told you it had a certificate but not who signed it, and the Certificate Authorities list gave no way to tell a signer with a fleet behind it from one nothing uses.

Closes https://github.com/nerves-hub/nerves_hub_web/issues/659

## Device settings

Each certificate now shows `Signer CA:` alongside its serial and validity, linking to the CA.

The lookup is scoped to the device's org, so a certificate signed by another org's CA reads "Unknown" rather than exposing that CA. Certificates whose signer was never registered read "Unknown" too. That includes the NervesHubCA-signed certificates the AKI validation lets through from other orgs.

## A show page for a CA

There wasn't one. The only per-CA page was the edit form, whose save is gated on `certificate_authority:update`, so it's the wrong place to send a viewer who just wants to look. `/org/:org/settings/certificates/:serial` shows the CA's details and a Devices card: the total, then a row per product linking to that product's device list filtered by this CA.

The device list is scoped to one product and a CA is scoped to the org, so per-product rows are the only shape a working link can take.

The list page loses its per-row Edit and Delete buttons and the `Check expiration?` / `JITP Enabled?` columns; the serial links to the show page, and Delete lives there now. It gains a Devices count.

## Filtering

A `Signer CA` select in the device list sidebar, built from the CAs that signed a certificate held by a device in this product. That's the same treatment the firmware version filter got in #3020, and it carries the same kind of hint tooltip explaining why a CA you expected might be missing. A "No known CA" option covers shared secret devices and unregistered signers.

It goes through the same filter map as every other sidebar filter, so `filters[signer_ca]` works on the device list API. There's a matching `signer_ca` advanced query column supporting `=` and `!=`:

```
signer_ca = "613033017807900175"
signer_ca != ":not_set" and connection = "connected"
```

Autosuggest shows the CA's description and sends the serial, the way `firmware` shows a version and sends a UUID. The OpenAPI spec documents both.

## The index

`device_certificates` had no index on `aki`, so every count and every filtered device list would sequential-scan it. This adds `(aki, device_id)` concurrently. The device id is in there so the counts can be answered from the index alone.

## Two things in `CAHelpers` that stopped working

Both are casualties of #2494, which deleted the old SCSS. The components survived; the CSS they depended on didn't.

`check_expiration_tooltip/1` styles nothing: `tooltip-info` and `tooltip-text` have no rules behind them any more, and `display: none` was the line that made it a tooltip. Without it the text is always on, which is the paragraph sitting in the middle of the new and edit CA forms. It's rebuilt on the `ToolTip` hook, the same shape `HealthStatus` uses, so it degrades to hidden rather than to visible. The show page calls it instead of carrying its own copy.

`certificate_status/1` had the same problem with more at stake: `.certificate-status-expired` (red) and `.certificate-status-expiring-soon` (amber) went with the SCSS, so an expired CA rendered in the same colour as a current one. It now returns `text-alert-content` / `text-warning-content` / `text-base-400` rather than class names Tailwind's `@source` scanning can't see.

Writing the first test for `certificate_status/1` turned up a third problem. "Expiring Soon" was unreachable:

```elixir
DateTime.after?(DateTime.shift(DateTime.utc_now(), month: -3), assigns.not_after)
```

A negative shift puts the cutoff three months in the *past*, so the clause only matched certificates that expired over three months ago, which the `Expired` clause above it had already taken. The shift is now positive, and the status has test coverage for all three branches.

## Noticed but not fixed

The edit page looks a CA up by serial alone, with no org check. Serials are globally unique, so a member of one org can open and save another org's CA. The show page added here uses `get_ca_certificate_by_org_and_serial/2`. Fixing edit is a separate change.

### DIFF
--- a/lib/nerves_hub/devices/advanced_query/compiler.ex
+++ b/lib/nerves_hub/devices/advanced_query/compiler.ex
@@ -111,6 +111,27 @@ defmodule NervesHub.Devices.AdvancedQuery.Compiler do
     )
   end
 
+  defp signer_ca_exists() do
+    dynamic(
+      [d],
+      fragment(
+        "EXISTS (SELECT 1 FROM device_certificates dc JOIN ca_certificates ca ON ca.ski = dc.aki WHERE dc.device_id = ?)",
+        d.id
+      )
+    )
+  end
+
+  defp signer_ca_exists(serial) do
+    dynamic(
+      [d],
+      fragment(
+        "EXISTS (SELECT 1 FROM device_certificates dc JOIN ca_certificates ca ON ca.ski = dc.aki WHERE dc.device_id = ? AND ca.serial = ?)",
+        d.id,
+        ^serial
+      )
+    )
+  end
+
   # `like`/`not like` use SQL ILIKE (case-insensitive); the value is the user's
   # pattern, so they supply `%`/`_` wildcards themselves.
   defp comparison_dynamic("identifier", "like", value), do: dynamic([d], ilike(d.identifier, ^value))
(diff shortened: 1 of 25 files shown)


nshoes (approved):
Sweet, love this

PR #3027, opened 2026-09-08
Require users to accept organization invitations

### LINKED ISSUE #460: If user is registered, adding user to organization does not send an invitation.
If an organization wants to invite a user and the user is already registered with NervesHub, but not part of that organization, the UI says it will send an invitation. It doesn't and the user is added immediately to the organization.

Essentially, the user is not given a choice of joining or not joining this organization.

**To Reproduce**
1. Register a user - via the CLI (as of 1 May 2019) UI possibly in near future
2. Create or open an organization and log in to it.
3. Click on 'Invite User'
4. Enter email of the user from step 1.
5. Click on 'Send Invite'
6. User is immediately added to organization and no email sent.

**Desired behavior**
If a registered user is invited to join an organization, that user should have the choice of accepting or not accepting that invitation. The organization does not need to know if the user is registered.

### LINKED ISSUE #1337: Invites to join an Org require the user to accept the invite

joshk:
Inviting someone to an organization used to add them straight away if they already had an account. Only people without an account ever saw an invitation. Now everyone gets one, and nobody joins an org until they accept.

Closes https://github.com/nerves-hub/nerves_hub_web/issues/460
Closes https://github.com/nerves-hub/nerves_hub_web/issues/1337

## The invite page

`/invite/:token` moved out of the `redirect_if_user_is_authenticated` scope, because a signed-in user now has a reason to land there. It shows one of four things depending on who is holding the session:

| Visitor | Sees |
| --- | --- |
| Signed in as the invited address | Join / Decline |
| Signed in as somebody else | Which account the invite is for, and a way to sign out |
| Signed out, has an account | Sign in to accept / Decline |
| Signed out, no account | Registration form / Decline |

Signing in from the invite page (including through an OAuth provider) returns you to it, so accepting is the next click.

Declining sets `declined_at`, which drops the invite out of the org's outstanding list and out of `get_valid_invite/1`. It needs only the token, so somebody without an account can decline without first registering.

The page no longer prints the invited address anywhere. An invite link is a bearer token that can be forwarded or pasted into a shared channel, and it does not need to tell whoever holds it who was invited.

## Managing outstanding invites

The invites table used to print each invite URL in full, ungated, to every member of the org. A `view` member could read a pending `admin` invite and register through it. The column is gone, replaced by Copy invite link, Resend invite and Rescind buttons that are disabled for anyone who is not an org admin. The token is left out of `data-copy-value` and `phx-value-invite_token` for those users, so it never reaches their DOM.

Resending rotates the token and restarts the 48-hour window, then sends the email again. Rotation kills any leaked copy of the previous link, and an invite is usually resent because it expired or went astray. Expiry now keys off `updated_at` rather than `inserted_at`; for an invite that has never been resent the two are the same.

The table also gained Role and Invited by columns.

## API

`POST /api/orgs/:org_name/users/invite` and the deprecated `/users` both always invite now. **This changes the response for a user who already has an account: 204 instead of 201 with the org_user body.** Inviting somebody who is already a member returns 422.

## redirect_if_user_is_authenticated

The plug tested `conn.assigns[:user]`. Nothing in the pipeline sets that; `fetch_current_user` assigns `:current_scope`. It has never redirected anybody.

Now that it reads the scope, `/confirm/:token` and `/password-reset/:token` moved to an unguarded scope. Both are driven by a mailed token and have to work regardless of who holds the session. An invite-registered user has a session while still unconfirmed, and would otherwise never be able to confirm.

## Migration

Adds `declined_at` to `invites`.


lib/nerves_hub/accounts.ex
```diff
@@ -679,9 +674,55 @@ defmodule NervesHub.Accounts do
...
+      Changeset.add_error(changeset, :email, "has already been invited to this organization")
+    else
+      changeset
+    end
+  end
+
+  defp pending_invite_exists?(org, email) do
+    Invite
+    |> where([i], i.org_id == ^org.id)
+    |> where([i], i.email == ^email)
```
nshoes, line 711: We should probably consider case here?

lib/nerves_hub/accounts.ex
```diff
@@ -706,16 +746,45 @@ defmodule NervesHub.Accounts do
...
+    end
+  end
+
   def delete_invite(org, token) do
     query =
       Invite
       |> where([i], i.org_id == ^org.id)
       |> where([i], i.token == ^token)
-      |> where([i], i.accepted == false)
+      |> unaccepted()
```
nshoes: This is filtering out declined invites. When we call this in the LiveView, a `:not_found` from this query displays the `"Invite couldn't be rescinded as the invite has been accepted."` message, even though the invite was declined.

nshoes (approved):
Small nits, otherwise LGTM

lib/nerves_hub/accounts.ex
joshk: Great spotting! Fixed!

lib/nerves_hub/accounts.ex
joshk: You legend! Fixed with some PG wizardry.

PR #3029, opened 2026-09-08
Retire firmware instead of destroying it

### LINKED ISSUE #2787: Device firmware history information prevents deleting firmware
If you want to tidy up firmware but it was used at some point by something, even if nothing uses it anymore it will totally prevent deleting.

joshk:
Deleting firmware usually failed, and failed badly: a raw foreign key error.

Three tables reference `firmwares` with `ON DELETE NO ACTION`, and the one that
bites is `device_firmwares`: it holds a row for every device that has ever
reported a firmware version. Any firmware that actually shipped was undeletable.

Closes https://github.com/nerves-hub/nerves_hub_web/issues/2787

## What changed

Firmware is now soft deleted. The row stays, marked with `deleted_at` and
`deleted_by_id`, and only the file behind it goes, via the existing
`DeleteFirmware` Oban job. Device firmware history and past deployment releases
keep resolving.

Deltas referencing the firmware are deleted outright, since they are derived data
and useless without both endpoints. Their files are queued for removal too, and
the deployment groups whose status was computed from them are recalculated.

### What blocks a delete

Two things, both of which would leave a device reaching for a file that is gone:

- the firmware backs a deployment group's current release
- the firmware has updates in flight

The show page disables the Delete button and names which of the two it is, with
the deployment groups listed. Being a *delta source* only warns, since losing
delta generation costs bandwidth rather than breaking an update. That warning
goes in the confirmation instead.

The check runs again inside the delete transaction. Once the row survives
deletion the foreign keys no longer refuse anything, so that re-check is all
that stands between a click and firmware a deployment group started rolling out
a moment ago.

### Deleted firmware in the UI

- The firmware list has a **Show deleted** toggle, held in the URL so the view
  is shareable and survives a reload. Deleted rows are dimmed and badged.
- A deployment group's release history badges releases whose firmware is gone.
  The link still works; without the badge the history reads as though the
  firmware is still there.
- The firmware page itself renders a deleted state naming who deleted it and
  when, and drops Download and Delete.


nshoes (changes requested):
Some notes outside of the files that were changed (everything else LGTM):

- The `FirmwareDeltaBuilder` uses a raw `Repo.get` for retreiving firmware. If either the source or target have been deleted, that job should should cancel itself.

- The same firmware can't be re-uploaded because because of the `firmwares_product_id_uuid_index`. I vote for an option to undelete firmware, but do not care if you go another direction.

joshk:
> Some notes outside of the files that were changed (everything else LGTM):
> 
> * The `FirmwareDeltaBuilder` uses a raw `Repo.get` for retreiving firmware. If either the source or target have been deleted, that job should should cancel itself.
> * The same firmware can't be re-uploaded because because of the `firmwares_product_id_uuid_index`. I vote for an option to undelete firmware, but do not care if you go another direction.

Hey @nshoes 

Great spotting, you are a ⭐ 

Both issues are now fixed.

On re-upload, I went with the partial index rather than undelete. A Restore button only works if the file outlives the delete, and right now the file deletion is enqueued immediately, so restoring would hand back firmware you can't download or deploy. Making that real means a grace window plus a sweep worker, mirroring what devices already do with restore_device/1 and the 12-day cleanup. I think this is worth considering, but I punted this to another PR.

nshoes (approved):

PR #3032, opened 2026-09-09
Add and remove tags in bulk from the device list

### LINKED ISSUE #1741: Ability to add tags alongside/instead of setting tags
Originally discussed in #1727 

We want to add tags to existing devices. Currently there's only setting tags which requires typing all the existing tags.

**A use case**
Imagine having 10 or 100 devices with a set of tags (there's a unique tag per device), and you want to add a common tag, in this case, you need to add the tag to the devices one by one otherwise you risk resetting all of them.

It would be great if there's a nice way to delete a tag too :)

joshk:
Closes #1741

Bulk tagging can now add and remove tags, not just replace them.

The tag box in the device list sidebar has an operation select:

1. "Set tags (replaces existing tags)": what the box did before
2. "Add tags": each device keeps its own tags and gains the new ones
3. "Remove tags": each device keeps everything except the named tags

All three work for a hand-picked selection and for "Select all matching".

Screenshot 2026-09-09 at 2 42 11 PM

Screenshot 2026-09-09 at 2 42 47 PM

Screenshot 2026-09-09 at 2 42 55 PM

### What add and remove do per device

`Devices.add_tags/3` and `Devices.remove_tags/3` derive each device's new tag list from its own current tags:

- Tag already on the device: no-op, not an error
- Device gains or loses nothing: skipped, no write, no audit entry
- Audit names the tags that changed: `User X added the tags field-trial to device device-21`

The single-tag `add_tag/3` and `remove_tag/3` used by the device details page now delegate to these.

### Bug fixed along the way

The tag input's `value` read `@current_filters[:tag]`; the submit button's enabled state read `@device_tags`. The operation select gave the form a second reason to re-render, which cleared the typed text while leaving the button enabled. With "Set" chosen, pressing it would have stripped the tags from every selected device. Both now read `@device_tags`.

Side effect: the tag-filter pre-fill is gone. It put text in the box but left the button disabled until you typed, so it was already half-broken.


lib/nerves_hub_web/live/devices/index.ex
```diff
@@ -314,42 +317,45 @@ defmodule NervesHubWeb.Live.Devices.Index do
...
-      start_bulk_async(socket, :tag_devices, [socket.assigns.current_scope.user, tags])
+      start_bulk_async(socket, bulk_action, [socket.assigns.current_scope.user, tags])
     else
       %{selected_devices: selected_devices, current_scope: scope} = socket.assigns
+      {applied, attempted} = tag_action_wording(bulk_action)
 
       with {:devices, devices} when is_list(devices) and devices != [] <-
              {:devices, Devices.get_devices_by_id(scope, selected_devices)},
-           result = BulkActions.tag_devices(devices, scope.user, tags),
+           result = apply(BulkActions, bulk_action, [devices, scope.user, tags]),
```
nshoes, line 340: Small nit, but the compiler can't pick this call up because of the runtime value `bulk_action`. I'd rather pattern match on a function arg and explicitly call the funciton.

lib/nerves_hub/devices/bulk_actions.ex
```diff
@@ -91,6 +91,47 @@ defmodule NervesHub.Devices.BulkActions do
     stream_processing(devices_query, {Devices, :tag_device, [user, tags]})
   end
 
+  @doc """
+  Add tags to many devices, keeping the tags each device already has.
+
+  `tag_devices/3` replaces the tags on every device it touches, which is no use
+  for putting one shared tag on a fleet of individually tagged devices.
+  """
+  @spec add_tags_to_devices([Device.t()] | Ecto.Query.t(), User.t(), list(String.t()) | String.t()) ::
```
nshoes, line 100: This, `remove_tags_to_devices/3`, and `tag_devices/3` are identical expect the atom funciton name. I'd suggest a refactor here, but it's fine if you leave it.

lib/nerves_hub/devices.ex
```diff
@@ -704,24 +705,84 @@ defmodule NervesHub.Devices do
...
+
+  @doc """
+  Remove tags from a device, leaving the rest of its tags alone.
+
+  See `add_tags/3` for why a device that doesn't carry any of the tags is left
+  untouched.
+  """
+  @spec remove_tags(Device.t(), User.t(), list(String.t()) | String.t()) ::
+          {:ok, Device.t()} | {:error, any(), any(), any()}
+  def remove_tags(%Device{} = device, user, tags) do
```
nshoes, line 760: It may be worth to use the `is_list` guard for `tags` here and in `add_tags/3`. There are no code paths where it's possible now, but if `nil` was passed as the `tags` arg in the future, it would remove every tag for a device.

joshk:
Thanks for the feedback @nshoes. I've gone through and implemented your suggested changes. You are a ⭐

lib/nerves_hub/devices/bulk_actions.ex
joshk: I didn't refactor these functions at this time but I will look into a tidy up on this module in another PR

PR #3047, opened 2026-09-14
Add required releases to deployment groups

joshk:
When a deployment group gets a new release, that release is the only one it sends. A device several releases behind goes straight to the newest firmware. That isn't always safe, for example when newer firmware relies on a change an earlier release makes. This lets a release be marked required, either when it's created or at any time afterwards, and devices that haven't reached it take it first.

Adds `deployment_releases.required` (boolean, default false).

## Which release a device gets next

A device's place in the group is the newest release whose firmware it runs. Its next update is the earliest required release after that place, or the current release when there isn't one.

A device running firmware that isn't in any of the group's releases can't be placed by release number. For those devices, a required release is skipped when the device's version is already at or past the release's version. If either version isn't semver, the required release still applies.

`ManagedDeployments.join_target_release/1` implements the rule as two lateral joins, and `target_release/2` runs it for a single device. The code that picks or records the firmware a device is sent now uses it:

- the orchestrator's available-devices query, including the check that a device's delta is ready
- `Updates.resolve_update` and `Updates.check_update`
- `DeviceEvents.schedule_update`, so the inflight update and the audit entries name the firmware the device was sent
- delta generation and the group's delta status, which pair each device's firmware with its own target instead of the current release
- the penalty box audit entry

The per-device lookup only runs for a group that has a required release before its current one (`earlier_required_release?/1`). Other groups join their current release directly, so their queries don't change.

Up-to-date counts, workflow step completion and the orchestrator's device-online check still compare against the current release, because a device that has only reached a required release hasn't finished updating.

## Workflows

Steps before the catch_all roll the current release out in stages, so they now leave out devices that have a required release to take first. A step doesn't claim those devices or send them updates, and it doesn't wait on them, even if it claimed them before the release was marked required. The catch_all picks them up and takes them through the required release and on to the current one.

## UI and API

- The new release form has a "Required release" checkbox.
- Release history shows a "Required" badge and, for users who can update the group, a Mark/Unmark required button. Marking or unmarking a release writes an audit entry, queues any deltas that devices now need, recalculates the group's status and tells the orchestrator to check again.
- The API accepts `required` when a firmware change creates a release, and returns it on `current_release`. There's no API yet for changing it on an existing release.


lawik:
Running an experimental automated review on this. Apologies about any noise. This looked like a good candidate :)

lib/nerves_hub/managed_deployments.ex
```diff
@@ -554,6 +554,159 @@ defmodule NervesHub.ManagedDeployments do
...
+
+      if target_release_id in [nil, current_release.id] do
+        current_release
+      else
+        DeploymentRelease
+        |> Repo.get!(target_release_id)
+        |> Repo.preload([:firmware, :archive])
+      end
+    end
+  end
```
lawik, line 690: **[WARNING] `target_release/2` always runs the per-device lateral-join query, even for a deployment group with no required release.** <sub>Code quality · likely</sub>

Unlike `join_target_release/2`, which checks `earlier_required_release?/1` first and joins the current release directly when there is nothing to work out, `target_release/2` calls `join_target_release/1` unconditionally whenever the device isn't already on the current release's firmware. This function backs `resolve_update`, `check_update` and `schedule_update`, so every deployment group now pays for two lateral subqueries with a semver fragment on essentially every device update check, even though the PR's own stated design is to only do that per-device work when a group actually has an earlier required release.

lawik (commented):
**Pass with warnings**: 1 warning, for `1ab98adc36`.

- **Correctness:** No findings.
- **Security:** No findings.
- **Code quality:** 1 finding, inline at `lib/nerves_hub/managed_deployments.ex:653-678`.
- **Prose, comments and docs:** No findings.

<sub>rummager · prompts c28a7e438902 · reviews are independent, verify before acting.</sub>

lib/nerves_hub/managed_deployments.ex
nshoes: When calling from the context of deltas in `Firmwares`, this will never return any deployment groups when a delta is being built for a required release. This causes the group to be stuck in `:preparing` for forever:
```
status after marking release 2 required:   :preparing
delta fw1->fw2 completed → recalc by fw2.id as arg:  0 groups matched, still :preparing
```
I believe we should be taking into account required releases here, like `recalculate_deployment_group_status/1` has.

nshoes: Since `firmware_pairs` can now span multiple targets with required releases, `Enum.all` is too strict. Claude's note on why this wasn't caught:
```
The existing suite misses both because required_release_test.exs
calls recalculate_deployment_group_status/1 directly rather than
through the by_firmware_id entry point the delta builder actually uses,
and never has two pairs in flight at once.
```

lib/nerves_hub/managed_deployments.ex
```diff
@@ -554,6 +554,159 @@ defmodule NervesHub.ManagedDeployments do
...
+  """
+  @spec set_deployment_release_required(DeploymentRelease.t(), boolean(), User.t()) ::
+          {:ok, DeploymentRelease.t()} | {:error, Changeset.t() | :not_found | :delta_generation_failed}
+  def set_deployment_release_required(%DeploymentRelease{} = release, required, %User{} = user)
+      when is_boolean(required) do
+    Repo.transact(fn ->
+      with {:ok, release} <- Repo.update(DeploymentRelease.required_changeset(release, required)),
+           {:ok, deployment_group} <- get_deployment_group(release.deployment_group_id),
+           :ok <- DeploymentGroupTemplates.audit_release_required_changed(user, deployment_group, release),
+           {:ok, _} <- maybe_trigger_delta_generation(deployment_group, release),
```
nshoes: The audit call on the line above this can be successful, even when delta generation fails, causing an annoying and ambiguous "Please try again" in the UI. Open to solutions here.

nshoes (commented):
This generally LGTM. A few things were missed relating to deltas and the assumption of a single target at callers, but those are easily adjustable.

Other notes:

`lib/nerves_hub/archives.ex:87` - This will always deliver the current releases's archive, instead of the required releases (from `lib/nerves_hub/device_link.ex:779`). I'm not sure we want to be sending a required release's archive because that doesn't really make sense. It's a one-shot and we don't check back on anything regarding the archive, so I'm leaning towards we never send the archive unless the device has updated to the current release.

`lib/nerves_hub/managed_deployments/workflows.ex:154` - Should `claimed_device_count/1` exclude devices needing a required release like `failed_device_count/2` does now?

lib/nerves_hub/managed_deployments.ex
joshk: Good catch, fixed. `target_release/2` now checks `earlier_required_release?/1` too, so a group with no required release before its current one gets the current release back without the per-device query.

lib/nerves_hub/managed_deployments.ex
joshk: Confirmed, and fixed. It now matches groups where the firmware belongs to a required release as well as the current one, so a delta built for a required release updates its group. Covered by a test that goes through this entry point with the delta targeting a required release.

joshk: Agreed. The status is now decided by precedence: any delta still building makes the group :preparing, otherwise any failed or timed out makes it :deltas_failed, otherwise :ready. :unknown_error is no longer produced. A test now runs two deltas in different states. I also decided to give each release its own delta status and drop deployment_groups.status in a follow-up PR, which removes this aggregation entirely.

lib/nerves_hub/managed_deployments.ex
joshk: The audit entry is written inside the same transaction, so it rolls back with everything else and isn't left behind. The UX point stands though: queueing deltas no longer decides whether the release is marked required. attempt_firmware_delta already logs and reports its own failures, a device with no delta gets the whole firmware, and the group's status reports whatever the deltas did. A follow-up PR adds a per-release deltas modal where a failed delta can be retried.

joshk:
> This generally LGTM. A few things were missed relating to deltas and the assumption of a single target at callers, but those are easily adjustable.
> 
> Other notes:
> 
> `lib/nerves_hub/archives.ex:87` - This will always deliver the current releases's archive, instead of the required releases (from `lib/nerves_hub/device_link.ex:779`). I'm not sure we want to be sending a required release's archive because that doesn't really make sense. It's a one-shot and we don't check back on anything regarding the archive, so I'm leaning towards we never send the archive unless the device has updated to the current release.

Agreed, and done. archive_for_deployment_group/2 now takes the device's firmware uuid and only returns the archive when the device is on the current release's firmware. Devices are sent archives again on reconnect, so a device gets it on the connection after it lands on the current release.

> `lib/nerves_hub/managed_deployments/workflows.ex:154` - Should `claimed_device_count/1` exclude devices needing a required release like `failed_device_count/2` does now?

Yes, fixed. It takes the deployment group now and counts through step_devices_query/2 with the same exclusion failed_device_count/2 uses, so a device held back by a required release doesn't take up the step's match limit or count towards its failure tolerance. It also means a step tops up after a device leaves the group.

nshoes (approved):

PR #3048, opened 2026-09-14
Add connecting code to deployment releases

joshk:
Stacked on #3047, so this PR's base is `feature/deployment-required-release`.

A deployment release can now carry connecting code, the Elixir a device runs when it connects. Deployment groups and devices already have connecting code of their own. A release's code runs after the group's code, before it, or instead of it, picked with a run order selector. Device-specific code still runs last.

Adds `deployment_releases.connecting_code` (text) and `deployment_releases.connecting_code_mode` (`first`, `last` or `override`, default `last`).

Screenshot 2026-09-14 at 5 03 07 PM

Screenshot 2026-09-14 at 5 03 12 PM

Screenshot 2026-09-14 at 5 03 23 PM

Screenshot 2026-09-14 at 5 07 36 PM

## Which release's code a device runs

The code comes from the release a device is running, which is the newest release in its group with the firmware the device has. Code written for new firmware only reaches devices that have that firmware, and a device picks up the new code when it reconnects after updating. A device on firmware from no release in its group runs only the group's code and its own.

`ManagedDeployments.join_running_release/1` finds that release. `join_target_release/1` from #3047 now uses it for a device's place in the group instead of its own subquery, so both features agree on which release a device is running.

A release with no code leaves the group's code alone, whatever its mode.

## Editing

Each release in the history shows its run order, with a link to its code. For users who can update the group, the Mark required button from #3047 becomes an Edit button. In the modal it opens, they can mark or unmark the release as required, and add, change or remove its connecting code. Saving connecting code writes an audit entry, and devices get the change the next time they connect.

## API

The deployment group update endpoint doesn't set a release's connecting code. In that request body `connecting_code` is the group's, so it's dropped before a firmware change creates a release. Without that, a request carrying the group's code would give it to the new release instead.


lib/nerves_hub/managed_deployments.ex
```diff
@@ -613,10 +606,39 @@ defmodule NervesHub.ManagedDeployments do
...
+  @spec join_running_release(Ecto.Query.t()) :: Ecto.Query.t()
+  def join_running_release(query) do
+    if has_named_binding?(query, :running_release) do
+      query
+    else
+      running_release =
+        DeploymentRelease
+        |> join(:inner, [r], f in assoc(r, :firmware), as: :firmware)
+        |> where([r], r.deployment_group_id == parent_as(:deployment_group).id)
+        |> where([firmware: f], f.uuid == fragment("? #>> '{\"uuid\"}'", parent_as(:device).firmware_metadata))
```
nshoes, line 636: For devices with firmware belonging to no release in its group, this runs on every connect. I don't think that's necessarily a problem, even at scale, but if we added a product id to this query I believe we could leverage `firmwares_product_id_uuid_index`, which would be easy and we'd never have to think about it.

lib/nerves_hub/devices.ex
```diff
@@ -894,11 +896,31 @@ defmodule NervesHub.Devices do
...
+        }
   def fetch_connecting_code(device_id) do
     Device
-    |> join(:left, [d], dp in assoc(d, :deployment_group))
-    |> select([d, dp], {d.connecting_code, dp.connecting_code})
-    |> where(id: ^device_id)
+    |> from(as: :device)
+    |> join(:left, [device: d], dg in assoc(d, :deployment_group), as: :deployment_group)
+    |> ManagedDeployments.join_running_release()
+    |> join(:left, [running_release: rr], r in DeploymentRelease, on: r.id == rr.id, as: :release)
```
nshoes: If we add to `connecting_code` and `connecting_code_mode` to the select in `join_running_release`, we could nix this join.

lib/nerves_hub/device_link.ex
```diff
@@ -615,18 +615,34 @@ defmodule NervesHub.DeviceLink do
...
+  release the device is running can add code before the group's, after it, or in
+  its place.
+  """
   @spec fetch_connecting_code(DeviceInfo.t()) :: list(binary()) | nil
   def fetch_connecting_code(device_info) do
-    {device_connecting_code, deployment_connecting_code} = Devices.fetch_connecting_code(device_info.device_id)
+    connecting_code = Devices.fetch_connecting_code(device_info.device_id)
 
-    [deployment_connecting_code, device_connecting_code]
+    connecting_code.release
```
nshoes, line 629: I think a simple  `String.trim/1` is worth it here. For every current code path, it's impossible for a release's connecting code to contain only whitespace as Ecto will handle that and set the default value for the `release.connecting_code`, which is `nil`. However, in the future, if a release did have an only whitespace via an `insert_all` or some other changeset that doesn't make the same cast, a device would be sent the connecting code containing only whitespace.

I think it's an easy enough fix that's worth writing a test for to future-proof this, but will defer to whatever you choose to do.

lib/nerves_hub/devices.ex
joshk: What an eye! Fixed!

lib/nerves_hub/device_link.ex
joshk: Great spotting, fixed!

lib/nerves_hub/managed_deployments.ex
joshk: Fixed! I needed to add another index as the one you mentioned is a partial index.

nshoes (approved):

PR #3050, opened 2026-09-14
Refuse a release that matches the current release

joshk:
Stacked on #3048, so this PR's base is `feature/deployment-release-connecting-code`.

A new release with the same firmware and archive as the current release is now refused. It would send devices nothing new. Since #3047 and #3048, a release's required flag and connecting code are changed on the release itself, so neither is a reason to create another one.

Only the current release counts. A release can still go back to firmware from an older release, and the current firmware can be released again with a different archive or with the archive dropped.

The check is a validation in `DeploymentRelease.new_changeset/5`, so the release form shows the error under the firmware select as soon as the current firmware is picked with the current archive. The API gets a 422 with the same message on `firmware`.

`create_deployment_release/6` reads the current release back from the database before checking, locking the deployment group row. A caller's copy of the group can be behind on its current release, and without the lock two identical releases created at once could both get through.


nshoes (approved):
LGTM.

Out of file-change scope notes: 

1. Should workflow definitions that don't change the firmware be blocked? Currently it fails with the "The current release already has this firmware and archive" error. However, if a worflow step stalls or fails, there's no escape hatch to force a rerun, and since you can't re-upload the same step with the same firmware/archive, it's stuck. Claude seems to think there's no way around this. 
2. Updating a deployment group through the API with a payload of `%{firmware: <unchanged>, is_active: true}` will fail now. Definitely an edge-case, may be worth letting other fields be updated even if the firmware field is sent and is unchanged. Will defer to you.

joshk:
> LGTM.
> 
> Out of file-change scope notes:
> 
> 1. Should workflow definitions that don't change the firmware be blocked? Currently it fails with the "The current release already has this firmware and archive" error. However, if a worflow step stalls or fails, there's no escape hatch to force a rerun, and since you can't re-upload the same step with the same firmware/archive, it's stuck. Claude seems to think there's no way around this.

Sorry, I'm a little confused; workflow steps can be retried and skipped. Does this count as an escape hatch?

> 2. Updating a deployment group through the API with a payload of `%{firmware: <unchanged>, is_active: true}` will fail now. Definitely an edge-case, may be worth letting other fields be updated even if the firmware field is sent and is unchanged. Will defer to you.

Good catch. I'd prefer to leave this for now because we really need to move people (and the CLI) to creating a release instead of updating a deployment.

nshoes:
> Sorry, I'm a little confused; workflow steps can be retried and skipped. Does this count as an escape hatch?

Oh, definitely :) I was too lazy to check myself.

> Good catch. I'd prefer to leave this for now because we really need to move people (and the CLI) to creating a release instead of updating a deployment.

Sounds good!

PR #3057, opened 2026-09-15
Give each release its own delta status

joshk:
Stacked on #3050, so this PR's base is `feature/deployment-release-duplicate-check`.

A deployment group has one status, but since #3047 its devices can be headed for more than one release at a time, each with its own deltas. Rolling those into a single value meant one release's deltas held back every device in the group, including the devices going somewhere else. Each release now carries its own delta status, and nothing reads `deployments.status` any more.

Adds `deployment_releases.delta_status` (`ready`, `preparing` or `failed`, default `ready`), backfilled from each group's status for its current release.

Screenshot 2026-09-16 at 11 55 39 AM

## What a release's status counts

The deltas its own devices need: from the firmware each device headed for that release is running, to the release's firmware. While any of them is still being built the release is `preparing`. Once none are, it is `failed` if any of them failed, and `ready` otherwise. `ManagedDeployments.recalculate_release_delta_status/1` records it, and the delta builder recalculates through `recalculate_release_delta_statuses_by_firmware_id/1`, which covers the current release and any required release carrying that firmware.

## What it changes for devices

A device is held back only while the release it is headed for isn't ready. The orchestrator no longer stops for the group as a whole: the available-devices query checks the target release's `delta_status`, so a required release still building deltas leaves the devices already past it alone.

## The group page

The header banner is the current release's own status, read off the release the page already loads. An earlier required release building deltas of its own reports that against itself in the history, rather than speaking for the whole group. The summary tab still lists the current release's deltas. Each release in the history now shows how many deltas its devices need and how many of those failed, with a modal listing them: source firmware, status, size, saving, and retry or delete for users who can update the group.

## Dropping the column

`deployments.status` is no longer read or written, but the column is still there. Dropping it in the same deploy would leave pods still running the old release querying a column that doesn't exist. The `alter table ... remove` belongs in a follow-up migration once this is out.


lib/nerves_hub/managed_deployments.ex
```diff
@@ -378,75 +371,200 @@ defmodule NervesHub.ManagedDeployments do
...
+    |> recalculate_delta_statuses()
+  end
 
-    if Enum.any?(updated_deployment_groups, fn {res, _} -> res == :error end) do
-      {:error, "Failed to recalculate deployment group statuses"}
+  defp recalculate_delta_statuses(releases) do
+    results = Enum.map(releases, &recalculate_release_delta_status/1)
+
+    if Enum.any?(results, &match?({:error, _}, &1)) do
+      {:error, "Failed to recalculate release delta statuses"}
```
nshoes: `{:error, String.t()}` breaks the contract for `set_deployment_release_required/3` and `update_deployment_group/3`. Claude says this will crash LiveView and API paths.

lib/nerves_hub_web/live/deployment_groups/show.ex
```diff
@@ -330,20 +331,24 @@ defmodule NervesHubWeb.Live.DeploymentGroups.Show do
...
     |> follow_release_steps(deployment_group, updated_deployment)
     |> assign(:deployment_group, updated_deployment)
+    |> assign(:delta_status, ManagedDeployments.delta_status(updated_deployment))
     |> assign(:firmware, updated_deployment.current_release.firmware)
     |> noreply()
   end
 
-  def handle_info(%Broadcast{event: "status/updated"}, socket) do
+  # A release's deltas finished, failed, or started building
+  def handle_info(%Broadcast{event: "release/delta_status"}, socket) do
```
nshoes, line 342: This event is only broadcasted when the status of a release changes, so counts can get wrong if multiple fail or succeed in successtion. The Summary tab gets around that by subscribing to the per-firmware `delta/status_update` topic instead (`lib/nerves_hub_web/components/deployment_group_page/summary.ex:813`).

Could be worth fixing, but I'd argue it's low priority.

nshoes (approved):

lib/nerves_hub/managed_deployments.ex
joshk: Thanks for spotting this, I've made some adjustments. recalculate_delta_statuses/1 now logs a release that won't take its new status and returns {:ok, updated} for the rest, so the string can't escape

lib/nerves_hub_web/live/deployment_groups/show.ex
joshk: You star you!

The release history now follows delta/status_update on the target firmware's topic, the way the summary tab already did.

PR #3058, opened 2026-09-16
Capture firmware update history and surface devices failing to update

joshk:
## Why

A device's update lives in `inflight_updates` only while it is running — reaching an outcome deletes the row — so nothing recorded that an update had been attempted, let alone that it failed. The audit log carries a sentence per event, which is readable one device at a time and useless in aggregate: a product whose fleet is failing to update fills pages of it without ever saying how many updates failed this week.

The failure mode that shaped the design is the quiet one. A device can cycle penalty box → retry → penalty box daily for weeks while **never reporting a failure** — no `audit_firmware_upgrade_failed` entry is written at all, because it simply comes back on its old firmware and says nothing. A graph keyed only on device-reported `"failed"` would show an empty chart for exactly the fleet most in trouble.

11-failing-updates

12-tooltip

10-insights-full

### Capture

Six outcomes are recorded wherever an update actually ends:

| Status | Meaning |
|---|---|
| `succeeded` | rejoined running the firmware it was sent |
| `failed` / `ignored` / `rescheduled` | the device reported it |
| `expired` | stopped making progress and timed out |
| `abandoned` | **new** — rejoined still on its old firmware, saying nothing |

`abandoned` is the quiet case above. Previously `maybe_clear_inflight_update/2` in `device_link.ex` dropped it silently; it is now `FirmwareUpdates.abandon_inflight_update/1`, which names what it means.

`rescheduled` is deliberately **not** counted as a failure. A device asking for the update later is negotiating, not failing, and counting it would libel every product whose devices defer updates while they are busy.

### Storage — two halves, on purpose

- **`device_update_history` (ClickHouse)** — one row per finished attempt, carrying the deployment group, the firmware it was moving between, and what the device said. Kept a year, since one row per finished update is orders of magnitude rarer than a health transition.
- **`devices.consecutive_failed_updates` / `first_update_failure_at` / `last_update_failure_at` (PostgreSQL)** — the current run of failures: how many, when it started, when it last moved.

This mirrors `device_health` / `device_health_history` and `device_alarms` / `device_alarm_history`. The counter is not derived from the history because the history is analytics: eventually consistent, only written where `:analytics_enabled` is set, and its buffer sheds rows rather than block. "How many times in a row has this device failed" has to be exact and has to exist everywhere.

`first_update_failure_at` is set via `COALESCE` inside the UPDATE rather than by reading the row back, so concurrent failures cannot race the run's start.

### Insights page

- **Firmware Update History** — successful against failed attempts per day (or hour), with the period tabs the connections graph already has, and a legend carrying both totals and the share that failed. Needs analytics.
- **Failed Firmware Updates** — devices that are both blocked and have a run of failures, worst first, with how long each has been failing. PostgreSQL only, so it is there either way.

The panel excludes blocked devices with no failures, which is what keeps its heading true — a rescheduled device is blocked and has failed at nothing.

### `updates=failed-updates` on the devices list

The panel's "All →" needed a destination showing the same population it counts, and `penalty-box` is only the blocked half of that. `failed-updates` is a new `updates` value — blocked *and* `consecutive_failed_updates > 0` — added as an ordinary filter rather than a special case for the link, so it also appears in the devices list dropdown, in advanced query autocomplete and in the API.

## Worth a reviewer's attention

**`abandoned` can over-count.** A device that reconnects while an update is inflight but before it starts downloading now scores a failure. That is already how the inflight row was being cleared — this only labels it — and arguably such a device *is* having trouble updating. But on a flapping fleet the counter will climb quickly. It is a distinct status in the history, so dropping it from `@failure_statuses` is a one-line change if it proves noisy.

**Existing devices start at zero.** The counter begins at 0 for every device on deploy, so the panel stays empty until devices fail again — about a day for a fleet failing daily.

**The predicate now exists twice** — as an `AdvancedQuery` fragment for the filter and as `Updates.failing_updates/3` for the panel. A test asserts the two return the same devices, since a change to one that misses the other is exactly the mismatch the filter was added to fix.


lib/nerves_hub/device_link.ex
```diff
@@ -878,9 +878,11 @@ defmodule NervesHub.DeviceLink do
   defp maybe_clear_inflight_update(_device, %{"currently_downloading_uuid" => uuid})
        when is_binary(uuid) and byte_size(uuid) > 0, do: :ok
 
+  # Anything left inflight when the device joins without downloading is an
+  # update it did not take: a success has already deleted its own row by this
+  # point, in `firmware_update_successful/2` above.
   defp maybe_clear_inflight_update(device, _) do
-    FirmwareUpdates.clear_inflight_update(device)
-    :ok
+    FirmwareUpdates.abandon_inflight_update(device)
```
nshoes, line 885: Should we only mark the inflights as abondoned for any other inflight status besides `:requested`? Claude made a case for flapping connections and the abundance of `:abandoned` update history rows that could follow.

lib/nerves_hub_web/live/product/insights.html.heex
```diff
@@ -279,6 +360,49 @@
...
+                  <p class="pb-2">The count is attempts in a row that ended without the device running the firmware it was sent, since the last time it did.</p>
+                  <p>Beside it is how long that run has been going: an afternoon is bad luck, a fortnight is a device that will not fix itself.</p>
+                  <div class="bg-surface-muted border-base-700 tooltip-arrow absolute size-2 origin-center rotate-45"></div>
+                </div>
+              </div>
+            </div>
+
+            <.link
+              :if={Enum.any?(@failing_devices)}
+              navigate={~p"/org/#{@current_scope.org}/#{@current_scope.product}/devices?updates=penalty-box"}
```
nshoes: This links to a possibly different list of devices. The count takes into consideration updates blocked and `consecutive_failed_updates > 0`, but the link just displays devices whose updates are blocked.

nshoes (approved):

lib/nerves_hub_web/live/product/insights.html.heex
joshk: Good catch

I've added a new `failed-updates` option for the `updates` filter (blocked *and* `consecutive_failed_updates > 0`) sitting alongside `penalty-box`, and the link just points at it.

lib/nerves_hub/device_link.ex
joshk: This captures all statues, not just `requested`. So if a device starts updating, then reboots (for whatever reason), and comes back online, we mark that update as abandoned.

joshk: The reboot case is narrower than it looks, for two reasons.

The row only reaches this line when `update_firmware_metadata/2` just above saw the **same** firmware UUID. A device that reboots mid-update and comes back running the new firmware has already gone down the `firmware_update_successful/2` path, which deletes the row and records `:succeeded`. So a reboot only lands on `:abandoned` when the device came back on its *old* firmware — the update genuinely didn't take. (`currently_downloading_uuid` doesn't catch it, since a rebooted device isn't downloading.)

And `consecutive_failed_updates` resets to 0 on any success, so a one-off interrupted download vanishes the moment the device next updates. Only a device stuck in a loop reaches the panel.

Where I think @nshoes is right is `:requested` specifically. A device that's been told to update and whose socket flaps before it acts attempted nothing — and if the flapping persists it never gets far enough to succeed, so the counter never resets. That one's a genuine false positive rather than a self-healing one. `maybe_update_update_attempts/1` already singles out `:requested` for much the same reason.

Worth noting either way: the condition here is unchanged from main — these exact joins already deleted the inflight row, just silently. This PR only names and counts it.

Two ways to cut `:requested` if you want it cut: skip the record entirely, or keep it in the history and drop it from `@failure_statuses` so the graph still sees it but the penalty box doesn't. Say which and I'll push it.

---
_Generated by [Claude Code](https://claude.ai/code)_

nshoes (approved):

lib/nerves_hub/device_link.ex
nshoes: Would you like me to decide which option to go with? No changes here are fine with me.

joshk: I'm happy to keep things as they are and evaluate once we see this in action

PR #3069, opened 2026-09-23
Let users drag the device details boxes around

joshk:
Lets each user drag the boxes on the device details tab into their own layout. Stacked on #3068, so review that first. This one takes about 20 minutes: start with `details_columns/3` and the `arrange-device-details` event in `details_tab.ex`, then `DisplayPreferences`.

## How dragging works

Each box has a grip handle at the top left of its header, shown on hover. The handle is the only drag target, so the map, forms and text in a box work as before. Boxes move within a column or across to the other one.

The drag uses [SortableJS](https://github.com/SortableJS/Sortable) (MIT) in a new `DeviceDetailsLayout` hook. After a drop it sends both columns' order to the server, which saves it and renders the boxes in that order. That matches what the drag already did to the DOM.

## Where the layout is stored

Two new `DisplayPreferences` fields, `device_details_left` and `device_details_right`, list box names top to bottom. They stay `nil` until the user first moves a box. Until then the page uses the default layout.

## Which boxes are shown, and where

- Boxes with nothing to show are left out: the map when geo is off or there's no Mapbox token, and Alarms when health reporting is off.
- A hidden box keeps its column and position when the user drags, so turning geo back on puts the map where it was.
- A box added later goes at the bottom of its default column for users with a saved layout.
- Until a user arranges the boxes, the default layout keeps #3068's move of General Info to the top right when the map is hidden. A saved layout is shown as the user left it.

## Refactor and input checks in `details_tab.ex`

The boxes moved out of `render/1` into function components with explicit attrs, so each column can render them in any order. Their markup is unchanged apart from the location box losing its own `:if`.

Box names from the page are only mapped to known atoms. The changeset rejects a layout that lists a box twice.

## Not in this PR

No keyboard way to move a box, and no reset button. Both are left for follow-up PRs.

To try it: open a device's Details tab, hover a box's header, and drag the grip.


lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -865,6 +603,48 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
...
+  # Box names come from the page, so only known ones are turned into atoms.
+  defp to_boxes(names) do
+    known = Map.new(DisplayPreferences.device_details_boxes(), &{Atom.to_string(&1), &1})
+    Enum.flat_map(names, &List.wrap(known[&1]))
+  end
+
+  # A hidden box isn't on the page to be dragged, so it goes back to where it
+  # was in the layout being replaced.
+  defp restore_hidden_boxes({left, right}, {old_left, old_right}, hidden) do
+    Enum.reduce(hidden, {left -- hidden, right -- hidden}, fn box, {left, right} ->
```
nshoes: Claude found an issue with this function (I couldn't summarize any better):

> The reduce walks hidden in canonical box order ([:health, :alarms, :general_info, :deployment, :location, ...]), not in the order the boxes appear in the old layout, and each List.insert_at shifts the indices computed for subsequent boxes. I confirmed this by running the function's logic directly. With health and geo both off (so :alarms and :location are hidden) and a saved left column of [:health, :location, :alarms, :general_info], a drag that changes nothing about the visible boxes rewrites the saved layout to [:health, :location, :general_info, :alarms] — :alarms silently moves past :general_info. The user then re-enables health and finds their layout changed by a drag they made elsewhere. Sorting hidden by its index in the old layout, or inserting from lowest index upward, fixes it.

lib/nerves_hub/accounts/display_preferences.ex
```diff
@@ -38,9 +52,66 @@ defmodule NervesHub.Accounts.User.DisplayPreferences do
       values: @all_deployment_group_list_columns,
       default: nil
     )
+
+    # Where the user has dragged the details tab's boxes to, top to bottom.
+    # Both nil until they first move one.
+    field(:device_details_left, {:array, Ecto.Enum}, values: @all_device_details_boxes, default: nil)
+    field(:device_details_right, {:array, Ecto.Enum}, values: @all_device_details_boxes, default: nil)
```
nshoes: If/when we remove a detail option, things like `Accounts.get_user!/1` in the device's page mount will blow up with `cannot load ["health", "retired_box"] as type {:array, #Ecto.Enum<...>}`. I think it's worth it to note in a comment that specifically _removing_ a detail option requires a data migration as well.

lib/nerves_hub_web/components/device_page/details_tab.ex
```diff
@@ -946,8 +726,402 @@ defmodule NervesHubWeb.Components.DevicePage.DetailsTab do
...
+            <div id="support-script-output" class="hidden" phx-no-format>{@recently_run_support_script_output}</div>
+          </div>
+        </div>
+      </div>
+    </div>
+    """
+  end
+
+  # Sits at the top of the right-hand column when the location map is hidden
+  # and the user hasn't arranged the boxes themselves. See `details_columns/4`.
```
nshoes:
```suggestion
  # and the user hasn't arranged the boxes themselves. See `details_columns/3`.
```

nshoes (approved):

lib/nerves_hub_web/components/device_page/details_tab.ex
joshk: Confirmed with your example: the left column came back as `health, location, general_info, alarms`. Fixed in 17b98055. Each column's hidden boxes now go back from the top down, so an insert can't shift one that's already been placed. The new test "hidden boxes in the same column go back in their old order" covers your case.

lib/nerves_hub/accounts/display_preferences.ex
joshk: Went one step further in 17b98055: the layout fields now use a new `NervesHub.Types.KnownAtoms` type instead of `Ecto.Enum`. Casting is just as strict, but loading skips names it no longer knows, so removing a box can't stop a user loading. The user's next drag saves the layout without the stale name. The comment on the fields says this, and suggests adding an update to the removing PR if the database should be cleaned straight away. `device_list_columns` and `deployment_group_list_columns` have the same problem; I've left them for a separate PR.

lib/nerves_hub_web/components/device_page/details_tab.ex
joshk: Fixed in 17b98055, thanks.

